PR #2, opened 2016-10-20
Introduce Glide for vendor package management

mumoshu:
The purpose for doing this is to:

1. remove a fairly large number of unimported packages in vendor/

  ```
  $ git status | grep "new file" | wc -l
       631
  $ git status | grep "modified" | wc -l
       543
  $ git status | grep "deleted:" | wc -l
      3583
  ```

2. provide developers a unified way to update packages
(3) update vendor with latest packages

I've chosen glide over trash, godep, govendor, etc., because glide seemed well-maintained, used in other coreos products like clair and etcd, prune enough files(e.g. glide only vendors packages imported from non-test go sources, but not too aggressive according to the fast it doesn't flatten nested vendor/ directories or remove *_test.go)


mumoshu:
build passed as can be seen in https://travis-ci.org/mumoshu/kube-aws/builds/169145557

PR #4, opened 2016-10-25
Simple toggle to stop the automatic assignment of public IP's for all the nodes.

pieterlange:
This toggle still requires the operator to bootstrap the VPC with a NAT gateway, but i think it's fair to expect the operator to do this manually/with external tooling.

I might send a subsequent PR to add functionality to create a NAT gateway from kube-aws. Gateway initialization takes several (sometimes 5+) minutes though, so i'm slightly worried cluster bootstrap might be affected.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -65,6 +65,7 @@ func newDefaultCluster() *Cluster {
 		CreateRecordSet:          false,
 		RecordSetTTL:             300,
 		Subnets:                  []*Subnet{},
+		MapPublicIPs:             true,
 		Experimental:             experimental,
 	}
 }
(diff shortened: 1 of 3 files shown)


mumoshu:
LGTM and `mapPublicIps: true` is something I'd like to generally encourage everyone to use for security.

Btw, one gotcha I'm aware of using NAT gateways is that it is limited to 10 Gbps/gateway(http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat-gateway.html#nat-gateway-basics).
You'll be required to create one gateway for each subnet to scale up to `10 Gbps * num of subnets` bandwidth.
If anyone is worrying about this, any suggestion/pull request for documentation or else is welcome.

mumoshu:
Going to merge. Thanks for your contribution @pieterlange !

pieterlange:
There's more than one gotcha for this, which is why i left actual NAT gateway creation up to the operator. :)

The NAT gateway also only lives in one availability zone so now the operator also has to deal with that. In general i think if you have these kinds of problems (including the 10Gbps/gateway problem), the organization should have the resources to help fix this. ;-)

When node pool functionality arrives i think it's worth it to add NAT gateway creation to kube-aws. Thanks for the merge!

PR #6, opened 2016-10-28
Use hosts' DNS resolvers instead of Google

pieterlange:
Also see https://github.com/coreos/coreos-kubernetes/issues/744

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -252,7 +252,7 @@ write_files:
         --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
         --mount=volume=ssl,target=/etc/kubernetes/ssl \
         --uuid-file-save=/var/run/coreos/decrypt-tls-assets.uuid \
-        --dns=8.8.8.8 --dns=8.8.4.4 \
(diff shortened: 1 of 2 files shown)


config/templates/cloud-config-controller
```diff
@@ -252,7 +252,7 @@ write_files:
         --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
         --mount=volume=ssl,target=/etc/kubernetes/ssl \
         --uuid-file-save=/var/run/coreos/decrypt-tls-assets.uuid \
-        --dns=8.8.8.8 --dns=8.8.4.4 \
+        --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
```
mumoshu: nit but should we make `--volume` and `--mount` split into separate lines, to match existing style?

pieterlange: I tend to squash the rkt mount command commands together on the same line because they're _so_ explicit. But i agree we should follow some style.

The squashing is done elsewhere though.

mumoshu: Thanks, got it. So let decide it later in a separate issue (if we remember 😄  )

PR #14, opened 2016-10-30
allow to override the ami id of coreos

qqshfox:
There's no official AMIs for CoreOS in China region. This PR make it possible to launch an instance by a user specified AMI.

### DIFF
--- a/config/templates/cluster.yaml
+++ b/config/templates/cluster.yaml
@@ -13,6 +13,10 @@ externalDNSName: {{.ExternalDNSName}}
 # See coreos.com/releases for more information
 #releaseChannel: stable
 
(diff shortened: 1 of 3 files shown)


pieterlange:
Looks good!

config/templates/cluster.yaml
```diff
@@ -13,6 +13,9 @@ externalDNSName: {{.ExternalDNSName}}
 # See coreos.com/releases for more information
 #releaseChannel: stable
 
+# The AMI ID of CoreOS.
```
qqshfox: Amended and force pushed.

mumoshu:
Thanks for your contribution @qqshfox 🙇

PR #15, opened 2016-10-30
add a new var IsChinaRegion to fix for China Region

qqshfox:
The `Service` in IAM profile is different in Global regions and China region. It's a little bit ugly to expose a var named `IsChinaRegion`, but I don't know whether there is a better way to fix this.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -67,6 +67,7 @@ func newDefaultCluster() *Cluster {
 		Subnets:                  []*Subnet{},
 		MapPublicIPs:             true,
 		Experimental:             experimental,
+		IsChinaRegion:            false,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mumoshu:
Until we see another AWS region like this, this change looks good enough 👍 
The var name states exactly what it is for.
Something like `iamPrincipalService: ec2.amazonaws.com.cn` indeed seem too general today.

mumoshu:
Merging this once we've released 0.9.0.

For anyone interested, here's possibly related upstream issue: https://github.com/kubernetes/kubernetes/issues/21858

mumoshu:
We have to wait for the binary signing process to settle before v0.9.0.
In the meantime, to not stop moving forward, I'd like to make this part of v0.9.0.

PR #17, opened 2016-10-31
Add option to mount existing Amazon EFS volume.

### LINKED ISSUE #10: Option to create and automatically mount EFS volumes
While NFS is kind of a dirty solution for cluster storage, i think it fills a basic (highly available storage across the cluster) need for most operators. Users can use `hostPath` volumes to access their persistent data from any node in any AZ.

Should probably include some caveats about what kinds of data you should (not) store on this.

It's an easy patch, will contribute PR soon.

pieterlange:
This still needs some work but i think it currently qualifies and fills a need.

Creating (and later finding) an EFS volume through cloudformation is more trouble than it's worth so the option currently requires an existing EFS-id. We can later turn this into an option which requires a unique id for creating the volume. We'd need to create the volume before we evaluate the stack-template though and my Go skills are still quite rusty so i think it's better to do that in a separate PR.

(nearly) fixes #10

### DIFF
--- a/config/templates/cluster.yaml
+++ b/config/templates/cluster.yaml
@@ -152,7 +152,7 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #
 # You can create a new EFS volume using the CLI:
 # $ aws efs create-file-system --creation-token $(uuidgen)
-#elasticFileSystem: fs-47a2c22e
+#elasticFileSystemId: fs-47a2c22e
 
 # Determines the container runtime for kubernetes to use. Accepts 'docker' or 'rkt'.
 # containerRuntime: docker
--- a/config/templates/cluster.yaml
+++ b/config/templates/cluster.yaml
@@ -147,6 +147,13 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Use Calico for network policy.
 # useCalico: false
 
+# Create MountTargets for a pre-existing Elastic File System (Amazon EFS). Enter the resource id, eg "fs-47a2c22e"
+# This is a NFS share that will be available across the entire cluster through a hostPath volume on the "/efs" mountpoint
+#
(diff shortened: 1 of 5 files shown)


config/templates/cluster.yaml
```diff
@@ -147,6 +147,13 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Use Calico for network policy.
 # useCalico: false
 
+# Create MountTargets for a pre-existing Elastic File System (Amazon EFS). Enter the resource id, eg "fs-47a2c22e"
+# This is a NFS share that will be available across the entire cluster through a hostPath volume on the "/efs" mountpoint
+#
+# You can create a new EFS volume using the CLI:
+# $ aws efs create-file-system --creation-token $(uuidgen)
+#elasticFileSystem: fs-47a2c22e
```
pieterlange: Agreed. Will fix this.

pieterlange:
I'm investigating some issues regarding service restart and am considering changing this to a full-fledged systemd mount (but that has side-effects too because i need to template the NFS path for the availability zones). Please hold off on merging this. Thanks

mumoshu:
Sure!
It seems things are much more difficult than I had thought.
I believe users (including me) willing to try EFS will thank you a lot to your work 🙏

pieterlange:
Not so difficult, it was just past my bedtime and i wanted to make sure i got it right this time. :nerd_face: 

Verified to be working now! If there is sufficient demand and i have enough time i'll write some code to create the EFS volume as well, but i think the `aws` cli command should be sufficient for now.

mumoshu:
I agree that we should add the EFS volume creation part when we have sufficient demand.
Also, really glad to see this in kube-aws. Thanks as always @pieterlange 👍

PR #31, opened 2016-11-04
[build] only set target os/arch for compile

colhom:
Having `GOOS=darwin` on my fedora box causes `go generate` to fail with a weird error. This fixes that problem (encountered when building release binaries).

\cc @mumoshu


mumoshu:
@colhom LGTM 👍 Thanks as always

Issue #26, opened 2016-11-02
Fix instability while starting kubelet on worker/controller nodes


mumoshu:
i.e. https://github.com/coreos/coreos-kubernetes/pull/697

mumoshu:
I'm going to eventually fix 3 units: docker.service, flanneld.service, kubelet.service.
The reasoning behind this is that I believe

* If `flanneld.service` depends on just local etcd(https://github.com/coreos/coreos-kubernetes/pull/697#issuecomment-254743845), normally it shouldn't fail transiently. But in today's kube-aws, it depends on discrete etcd nodes in controller/worker nods. It may fail when etcd.service's in controller/worker nodes started before the remote etcd's become ready.
* `docker.service` depends on local `flanneld.service` which depends on the remote etcd in controller/worker nodes. It may fail on startup then.
* `kubelet.service` depends on `flanneld.service`. It may also fail then.

fyi @dghubble

mumoshu:
well, some more information regarding this. I'm reconsidering where to fix.

* `flanneld.service` doesn't seem to depend on suspicious services, and [it has `Restart=always`](https://github.com/coreos/coreos-overlay/blob/master/app-admin/flannel-wrapper/files/flanneld.service). It won't fail because of dependency failure then.
* log messages referred in https://github.com/coreos/coreos-kubernetes/issues/710 does show failed `decrypt-tls-assets.service`, which is required by `kubelet.service` and `flanneld.service` in kube-aws

mumoshu:
I've noticed that, in cloud-config-etcd, we should make `decrypt-tls-assets` based rkt instead of docker, as we had done so for workers/controllers.
I'll address that in a separate issue if it matters.

mumoshu:
Let me outline how I'm going to fix the issue based on my initial plan in https://github.com/coreos/coreos-kubernetes/pull/697#issuecomment-254753180

> * Modification to kubelet.service would be useful according to coreos-baremetal#253

Make it not require flannel and/or docker, but want just flannel to up using the `Wants` and `ExecStartPre` pattern. This will prevent kubelet.service from stopped forever in case of startup failures in flanneld.service (and decrypt-tls-assets)

> * Modification to flanneld.service would be useful according to #710

Make it want/require nothing. Just run decrypt-tls-assets in `ExecStartPre` to prevent a single failure of decrypt-tls-assets from breaking flanneld.service forever.

> * Modification to docker.service would be useful according to #682

[...]

PR #34, opened 2016-11-04
Fix instability while starting kubelet on worker/controller nodes

mumoshu:
refs #26


mumoshu:
As this passed e2e as described in https://github.com/coreos/kube-aws/issues/37#issuecomment-258721471, merging.

PR #35, opened 2016-11-05
Allow ECR pull from controller IAM role

c-knowles:
Goes some way to resolve coreos/coreos-kubernetes/issues/620, a replacement of https://github.com/coreos/coreos-kubernetes/pull/731.

### DIFF
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -200,6 +200,19 @@
                   "Action" : "kms:Decrypt",
                   "Effect" : "Allow",
                   "Resource" : "{{.KMSKeyARN}}"
+                },
+                {
+                  "Action": [
+                    "ecr:GetAuthorizationToken",
+                    "ecr:BatchCheckLayerAvailability",
+                    "ecr:GetDownloadUrlForLayer",
+                    "ecr:GetRepositoryPolicy",
+                    "ecr:DescribeRepositories",
+                    "ecr:ListImages",
+                    "ecr:BatchGetImage"
+                  ],
+                  "Resource": "*",
+                  "Effect": "Allow"
                 }
               ],
               "Version": "2012-10-17"


mumoshu:
@c-knowles LGTM. Thanks for your contribution!
FYI, Your PRs are going to be included in v0.9.1-rc.2.

c-knowles:
Great, thanks! I just need to figure out if it's worth re-doing https://github.com/coreos/coreos-kubernetes/pull/716 next. I've been using it already for a while now on a forked build.

mumoshu:
@c-knowles As a kube-aws user, I'd be greatly happy if I could deploy kube-aws created clusters into existing VPCs.

After a deployment, it would be way better if services hosted in both kubernetes and existing infrastructure could start communicating each other (if we demanded so), without manual steps we have to take for now. Those manual steps seem to include (1) maintaining route tables (2) assigning it to kube-aws created subnets, etc?

After reading your great write up in https://github.com/coreos/coreos-kubernetes/pull/716, I now realize that it is basically what your PR addresses, right?

In short: I'm a fan of your PR 👍

c-knowles:
Sure, I'll redo that one soon then. It's exactly what I'm using it for - to access RDS in the same VPC. I'm interested to get some acceptance tests in there too but can discuss in the PR once I've made it.

pieterlange:
/off current ECR topic

The described usecase (using existing services)  is already possible by using `routeTableId`; the PR addressed a need for a separate routetable (and subnet) for the controller only (eg, so one could put the controller into a public subnet).

It might be better to implement this into the nodepool concept we've been talking about..

mumoshu:
@pieterlange Thanks for confirmation. Really helpful 👍 

So, for anyone interested, my comment https://github.com/coreos/kube-aws/pull/35#issuecomment-259014776 was missing the point.
It turns out that it was already possible doing:

1. Add one route for each newly created subnet, to an existing route table (if specified)
2. Associate an existing route table to newly created subnets

...so that we can make services in kubernetes and existing infrastructure communicate to each other.

And what @pieterlange described follows.

(Any correction is welcome!)

c-knowles:
Let's move to https://github.com/coreos/kube-aws/issues/44

PR #36, opened 2016-11-05
Add name tags where applicable

c-knowles:
Not everything is tagged with a name which means the AWS console gets filled with empty named resources, this helps fix it up so they are easier to identify especially with multiple k8s stacks.

A replacement of https://github.com/coreos/coreos-kubernetes/pull/717.


mumoshu:
@c-knowles LGTM. Thanks for your contribution!

PR #41, opened 2016-11-07
Remove "Restart=on-failure" from kube-node-drainer.service

camilb:
Systemd only allows "Restart=no" for  "Type=oneshot" services.

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -152,7 +152,6 @@ coreos:
         [Service]
         Type=oneshot
         RemainAfterExit=true
-        Restart=on-failure
         ExecStartPre=/usr/bin/systemctl is-active kubelet.service
         ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStart=/bin/true


pieterlange:
Thanks! `Restart=` indeed makes no sense on `oneshot` services.

camilb:
Apart from this, maybe it's a good idea to replace `ExecStart: /bin/true`  with:

            ExecStart=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
          /hyperkube kubectl \
          --server=https://{{.ExternalDNSName}}:443 \
          --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
          uncordon $(hostname)'

If the node reboots, it will be marked automatically as `schedulable`.

mumoshu:
@camilb Ah, I know i'm to blame! Thanks for the fix :bow:

mumoshu:
@pieterlange Yes, `$$` is intentional. I remember that I referred to

> To pass a literal dollar sign, use "$$".
> https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines 

Basically I tried my best to make sure `$$(hostname)` to get expanded at runtime. I guess if `$(hostname)` isn't recognized as a systemd variable, it also work as we might expect 👍

PR #47, opened 2016-11-08
build: fix up containerized release build process

colhom:
`rkt run` was doing some weird stuff on fedora machines with SELinux in enforcing mode, so we're back to docker for the containerized build wrapper.

After this, we'll be ready to cut `v0.9.1`.

\cc @mumoshu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .envrc
 coverage.txt
 profile.out
+/.glide
--- a/containerized-build-release-binaries
+++ b/containerized-build-release-binaries
@@ -2,9 +2,15 @@
 set -euo pipefail
 
 COREOS_KUBERNETES_ROOT=$(git rev-parse --show-toplevel)
-sudo rkt run \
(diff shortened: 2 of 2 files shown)


mumoshu:
@colhom Thanks for this! 👍 
It seems that the build failed with missing `glide` command in the Travis env.
Would you mind fixing it or shall I do it after merging?

colhom:
@mumoshu will fix

colhom:
@mumoshu do you thinks it's desirable to have `./build` do `glide install`- or should we simply implicitly trust the checked-in `./vendor` folder? I realize now that may have been the original intent, in which case I will back out the modification to `./build` here.

mumoshu:
@colhom I'm ok with the latter!

// but that's because I have been blindly assuming that's why we've committed `./vendor` into the repo.
// off topic but I began to think that we should document the pre-req/implication of `./build` somewhere regardless of how we fix this build.

mumoshu:
@colhom LGTM 👍 Thanks

Issue #40, opened 2016-11-07
Kube node drainer service fails to start.


camilb:
I think  the line `Restart=on-failure` from `kube-node-drainer.service` should be removed.

Getting these errors:

Failed to restart kubelet.service: Unit kube-node-drainer.service is not loaded properly: Invalid argument.
 kube-node-drainer.service: Service has Restart= setting other than no, which isn't allowed for Type=oneshot services. Refusing.

camilb:
#41

camilb:
#41

pieterlange:
Easy mistake to slip in..

I think i've spotted another one:
```
        ExecStop=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
          /hyperkube kubectl \
          --server=https://{{.ExternalDNSName}}:443 \
          --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
          drain $$(hostname) \
          --ignore-daemonsets \
          --force'
```

Unless i'm missing something here, the double `$$` seems to be an error. Is this some escaping trick? I think we should just be able to run `$(hostname)`, right?

camilb:
@pieterlange I see that it's working with double `$$`as with one `$`.

camilb:
Related to this, but looks like it's affecting other services too, using ExecStartPre=/usr/bin/systemctl is-active kubelet.service will check if the service is active then exit (code=exited, status=3). This being a `oneshot`service, it will not be restarted. Also observed this on `install-calico-system.service` and `install-kube-system.service. Using `ExecStartPre=/usr/bin/systemctl is-active  service.name` it will generate  a lot of errors before kubelet becomes active.

In this case I think is better to use something like:

    After=multi-user.target

    [Install]
    WantedBy=node-drain.target

This way we make sure that all the services are running before we start this one.

Here is a proposal: 

         [Unit]
         Description=drain this k8s node to make running pods time to gracefully shut down before stopping kubelet
         After=multi-user.target
         Wants=decrypt-tls-assets.service kubelet.service docker.service
         
         [Service]
         Type=oneshot
         RemainAfterExit=true
         ExecStart=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
           /hyperkube kubectl \
[...]

PR #48, opened 2016-11-08
Replace docker with rkt in "kube-node-drainer.service".

camilb:
Fix `kube-node-drainer.service` `ExecStart` and `ExecStop` errors . #40

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -144,27 +144,32 @@ coreos:
     {{ if .Experimental.NodeDrainer.Enabled }}
     - name: kube-node-drainer.service
       enable: true
+      command: start
+      runtime: true
       content: |
         [Unit]
         Description=drain this k8s node to make running pods time to gracefully shut down before stopping kubelet
-        Wants=kubelet.service docker.service
+        After=multi-user.target
 
         [Service]
         Type=oneshot
         RemainAfterExit=true
-        ExecStartPre=/usr/bin/systemctl is-active kubelet.service
-        ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStart=/bin/true
-        ExecStop=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
-          /hyperkube kubectl \
+        TimeoutStopSec=30s
+        ExecStop=/bin/sh -c '/usr/bin/rkt run \
+        --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
+        --mount=volume=kube,target=/etc/kubernetes \
+        --net=host \
+        {{.HyperkubeImageRepo}}:{{.K8sVer}} \
+          --exec=/kubectl -- \
           --server=https://{{.ExternalDNSName}}:443 \
           --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
-          drain $$(hostname) \
+          drain $(hostname) \
           --ignore-daemonsets \
           --force'
 
         [Install]
-        RequiredBy=kubelet.service
+        WantedBy=multi-user.target
     {{ end }}
 {{ if $.ElasticFileSystemID }}
     - name: rpc-statd.service


config/templates/cloud-config-worker
```diff
@@ -144,27 +144,32 @@ coreos:
...
 
         [Service]
         Type=oneshot
         RemainAfterExit=true
-        ExecStartPre=/usr/bin/systemctl is-active kubelet.service
-        ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStart=/bin/true
-        ExecStop=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
-          /hyperkube kubectl \
+        TimeoutStopSec=30s
```
mumoshu: Just a question but any specific reason why this is `30s`?

config/templates/cloud-config-worker
```diff
@@ -144,27 +144,32 @@ coreos:
...
+        ExecStop=/bin/sh -c '/usr/bin/rkt run \
+        --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
+        --mount=volume=kube,target=/etc/kubernetes \
+        --net=host \
+        quay.io/coreos/hyperkube:v1.4.5_coreos.0 \
+          --exec=/kubectl -- \
           --server=https://{{.ExternalDNSName}}:443 \
           --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
-          drain $$(hostname) \
+          drain $(hostname) \
```
mumoshu: I now understand that `$$` and `$` to be equivalent in this context because it isn't recognized/expanded as systemd variables anyways. Good!

ref https://github.com/coreos/kube-aws/issues/40#issuecomment-258863442
ref https://github.com/coreos/kube-aws/pull/41#issuecomment-259003781

config/templates/cloud-config-worker
```diff
@@ -144,27 +144,32 @@ coreos:
...
-        ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStart=/bin/true
-        ExecStop=/bin/sh -c '/usr/bin/docker run --rm -v /etc/kubernetes:/etc/kubernetes {{.HyperkubeImageRepo}}:{{.K8sVer}} \
-          /hyperkube kubectl \
+        TimeoutStopSec=30s
+        ExecStop=/bin/sh -c '/usr/bin/rkt run \
+        --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
+        --mount=volume=kube,target=/etc/kubernetes \
+        --net=host \
+        quay.io/coreos/hyperkube:v1.4.5_coreos.0 \
```
mumoshu: Would it be ok to keep using `{{.HyperkubeImageRepo}}:{{.K8sVer}}` instead of `quay.io/coreos/hyperkube:v1.4.5_coreos.0` to avoid hard-coding for future ease in maintenance/more customizability?

config/templates/cloud-config-worker
camilb: Just wanted to make sure this is properly stopped. This configures the amount of time that systemd will wait when stopping  the service. Had several intents with `rkt` without this and the node wasn't marked `SchedulingDisabled`.

config/templates/cloud-config-worker
camilb: Yes, think I was copying this from another place.

config/templates/cloud-config-worker
mumoshu: Got it!

I was just reading https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec= which says that `TimeoutStopSec` defaults to the system-wide value configured via `DefaultTimeoutStopSec`.

Regardless of the default, setting enough timeout like this would be good.
This is what we couldn't have without your effort 👍

config/templates/cloud-config-worker
mumoshu: 👍 

Would it be ok to let me fix this after merging, or would you like to do on your side?

camilb:
Fixed {{.HyperkubeImageRepo}}:{{.K8sVer}} using `git commit --amend. Notice  the comment was erased.

mumoshu:
LGTM 👍
Thanks for your contribution :bow:

Issue #49, opened 2016-11-09
Proposal: Clouformation wait for signal 


gianrubio:
In the current cloudformation template, kube-aws use PauseTime flag to wait for create/update resources.  

```
    UpdatePolicy" : {
        "AutoScalingRollingUpdate" : {
          "MinInstancesInService" : "{{.ControllerCount}}",
          "MaxBatchSize" : "1",
          "PauseTime" : "PT2M"
        }
      }
```

My proposal is to use cfn-signal to control the resources [1][2] and only send the signal after the last service was started. I can implement this proposal and I'm here to listen for suggestions.

```
[Unit]
Description=Cloudformation Signal Ready
After=docker.service
Requires=docker.service
After=etcd2.service
Requires=etcd2.service

[Install]
WantedBy=multi-user.target

[Service]
Type=oneshot
EnvironmentFile=/etc/environment
ExecStart=/bin/bash -c '\
eval $(docker run crewjam/ec2cluster); \
docker run --rm crewjam/awscli cfn-signal \
    --resource MasterAutoscale --stack $TAG_AWS_CLOUDFORMATION_STACK_ID \
    --region $REGION; \
'
```

[1] https://aws.amazon.com/blogs/devops/use-a-creationpolicy-to-wait-for-on-instance-configurations/
[2] http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html

pieterlange:
Big fan of this idea but i'm seeing some issues here:
  - Running containers straight from docker hub outside of the CoreOS management sphere.
  - We obviously would need to do this for both the `AutoScaleController` and `AutoScaleWorker` groups.

We don't actually need `crewjam/ec2cluster` here as all we need to set is `$TAG_AWS_CLOUDFORMATION_STACK_ID` and `$REGION`. We could set both environment variables using the feature described here: https://github.com/coreos/kube-aws/issues/29

There is an image available from CoreOS to replace the `crewjam/awscli` docker command in order to run `cfn-signal`: https://quay.io/repository/coreos/awscli but it could use an update (@colhom).

(Full disclosure: @gianrubio is a colleague :-) )

iwarp:
Also big fan of this, In the UpdatePolicy you will have to include WaitOnResourceSignals=true and set the PauseTime higher in order to give time for the system to come up, otherwise cloudformation will rollback the deploy.

mumoshu:
[...]

PR #58, opened 2016-11-15
Easily manage etcd cluster with env vars

gianrubio:
Exporting  etcdctl environment variables, allow cluster managers to run etcdctl commands without providing arguments.

Currently when you logging in the etcd cluster you can't run etcdctl commands except you provide all the args.  

```
$ etcdctl --cert-file=/etc/etcd2/ssl/etcd-client.pem --ca-file=/etc/etcd2/ssl/ca.pem --key-file=/etc/etcd2/ssl/etcd-client-key.pem  --endpoints=https://ip-10-0-0-5.eu-west-1.compute.internal:2379 cluster-health
member f5654f631de6e415 is healthy: got healthy result from https://ip-10-0-0-5.eu-west-1.compute.internal:2379
cluster is healthy
```

After this PR will be easy for run etcdctl commands

```
$ etcdctl cluster-health
member f5654f631de6e415 is healthy: got healthy result from https://ip-10-0-0-5.eu-west-1.compute.internal:2379
cluster is healthy
```

Also this is important on #49 and signaling cloudformation when the etcd cluster is healthy.

### DIFF
--- a/config/templates/cloud-config-etcd
+++ b/config/templates/cloud-config-etcd
@@ -31,6 +31,7 @@ coreos:
             Environment=ETCD_LISTEN_PEER_URLS=https://%H:2380
             Environment=ETCD_INITIAL_ADVERTISE_PEER_URLS=https://%H:2380
             PermissionsStartOnly=true
+            ExecStartPre=/usr/bin/bash -c "sed -i \"s/^ETCDCTL_ENDPOINT.*$/ETCDCTL_ENDPOINT=https:\/\/$(hostname):2379/\" /etc/environment"
             ExecStartPre=/usr/bin/chown -R etcd:etcd /var/lib/etcd2
       enable: true
       command: start
@@ -86,6 +87,17 @@ coreos:
 
 
 write_files:
+
+  - path: /etc/environment
+    permissions: 0644
+    content: |
+      COREOS_PUBLIC_IPV4=$public_ipv4
+      COREOS_PRIVATE_IPV4=$private_ipv4
+      ETCDCTL_CA_FILE=/etc/etcd2/ssl/ca.pem
+      ETCDCTL_CERT_FILE=/etc/etcd2/ssl/etcd-client.pem
+      ETCDCTL_KEY_FILE=/etc/etcd2/ssl/etcd-client-key.pem
+      ETCDCTL_ENDPOINT=
+
   - path: /opt/bin/ext4-format-volume-once
     permissions: 0700
     owner: root:root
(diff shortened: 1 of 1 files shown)


config/templates/cloud-config-etcd
```diff
@@ -86,6 +87,17 @@ coreos:
...
+
+  - path: /etc/environment
+    permissions: 0644
+    content: |
+      COREOS_PUBLIC_IPV4=$public_ipv4
+      COREOS_PRIVATE_IPV4=$private_ipv4
+      ETCDCTL_CA_FILE=/etc/etcd2/ssl/ca.pem
+      ETCDCTL_CERT_FILE=/etc/etcd2/ssl/etcd-client.pem
+      ETCDCTL_KEY_FILE=/etc/etcd2/ssl/etcd-client-key.pem
+      ETCDCTL_ENDPOINT=
```
mumoshu: I'm just curious but may I ask why you've explicitly emptied `ETCD_ENDPOINT` here?
Just to note that we have to explicitly set it when running `etcdctl`?

config/templates/cloud-config-etcd
```diff
@@ -120,3 +132,11 @@ write_files:
...
     encoding: gzip+base64
     content: {{.TLSConfig.EtcdCert}}
+
+  - path: /etc/etcd2/ssl/etcd-client.pem.enc
+    encoding: gzip+base64
+    content: {{.TLSConfig.EtcdClientCert}}
+
+  - path: /etc/etcd2/ssl/etcd-client-key.pem.enc
+    encoding: gzip+base64
+    content: {{.TLSConfig.EtcdClientKey}}
```
mumoshu: Note to self:

* This would make user-data much bigger than before... Oh, we now have the `--s3-uri` option to workaround that. No problem 😃 
* This does add additional credentials we didn't have in etcd nodes before. Any security concerns for saving more credentials in etcd nodes? No.

config/templates/cloud-config-etcd
mumoshu: I realized that this line is the "marker" for https://github.com/coreos/kube-aws/pull/58/files#diff-c88cb625705bf9711a2f576a1975cb72R34.
Doing this makes it possible to:
* Mention that we have to provided an actual value when running `etcdctl`
* Ensure two things:
  * Appropriate `ETCDCTL_ENDPOINT` value is always configured on starting/restarting `etcd2.service`, automatically.
  * Restarting `etcd2.service` doesn't break `/etc/environment` as we can see in https://github.com/coreos/kube-aws/pull/58/files#diff-c88cb625705bf9711a2f576a1975cb72R34
    * For example, `echo ETCDCTL_ENDPOINT="https://$(hostname):2379/" >> /etc/environment` would break, or needlessly bloat the environment file at least, when `etcd2.service` got restarted.

mumoshu:
@gianrubio I'm impressed with your work 👍

mumoshu:
Thanks for the contribution! Merging.

PR #67, opened 2016-11-17
Add experimental feature "awsEnvironment".

### LINKED ISSUE #29: Proposal: AWS environment variable file
It is currently impossible to directly utilize cloudformation Refs inside of the userdata / `cloud-init` files as those files get passed verbatim to the `UserData` parameters in the `stack-template.json` file.

I think it's useful to provide an `/etc/aws-environment` file in which to pass a list of AWS environment variables by `Ref`, for the end user to utilize using systemd's `EnvironmentFile` [option ](https://coreos.com/os/docs/latest/using-environment-variables-in-systemd-units.html).

We can easily append this right now by using ``Fn::Join`` to the UserData as the last section in `cloud_init` is `write_files`.

Immediate usecase: passing in cloudformation [wait condition urls](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-waitcondition.html).

pieterlange:
This enables users to to pass AWS cloudformation stack references in environment variables to custom scripts.

Use 'EnvironmentFile=/etc/aws-environment` in systemd units to consume the variables.

Fixes #29

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -11,6 +11,31 @@ coreos:
     etcd_keyfile: /etc/kubernetes/ssl/etcd-client-key.pem
 
   units:
+{{if .Experimental.AwsEnvironment.Enabled}}
+    - name: set-aws-environment.service
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Set AWS environment variables in /etc/aws-environment
+        After=network-online.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStartPre=/bin/touch /etc/aws-environment
+        ExecStart=/usr/bin/rkt run \
+          --uuid-file-save=/var/run/coreos/cfn-init.uuid \
+          --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+          --volume=awsenv,kind=host,source=/etc/aws-environment,readOnly=false --mount volume=awsenv,target=/etc/aws-environment \
+          --net=host \
+          --trust-keys-from-https \
+          quay.io/coreos/awscli:edge -- cfn-init -v \
+              --region {{.Region}} \
+              --resource LaunchConfigurationController \
+              --stack {{.ClusterName}}
+{{end}}
     - name: docker.service
       drop-ins:
         - name: 40-flannel.conf
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -141,7 +141,7 @@ coreos:
         WantedBy=multi-user.target
 {{ end }}
 
-    {{ if .Experimental.NodeDrainer.Enabled }}
+{{ if .Experimental.NodeDrainer.Enabled }}
     - name: kube-node-drainer.service
       enable: true
       command: start
@@ -170,7 +170,34 @@ coreos:
 
         [Install]
         WantedBy=multi-user.target
-    {{ end }}
+{{ end }}
+
+{{if .Experimental.AwsEnvironment.Enabled}}
+    - name: set-aws-environment.service
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Set AWS environment variables in /etc/aws-environment
+        After=network-online.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStartPre=/bin/touch /etc/aws-environment
+        ExecStart=/usr/bin/rkt run \
+          --uuid-file-save=/var/run/coreos/cfn-init.uuid \
+          --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+          --volume=awsenv,kind=host,source=/etc/aws-environment,readOnly=false --mount volume=awsenv,target=/etc/aws-environment \
+          --net=host \
+          --trust-keys-from-https \
+          quay.io/coreos/awscli:edge -- cfn-init -v \
+              --region {{.Region}} \
+              --resource LaunchConfigurationWorker \
+              --stack {{.ClusterName}}
+{{end}}
+
 {{ if $.ElasticFileSystemID }}
     - name: rpc-statd.service
       command: start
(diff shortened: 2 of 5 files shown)


config/templates/cloud-config-controller
```diff
@@ -11,6 +11,31 @@ coreos:
...
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStartPre=touch /etc/aws-environment
+        ExecStart=/usr/bin/rkt run \
+          --uuid-file-save=/var/run/coreos/cfn-init.uuid \
+          --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+          --volume=awsenv,kind=host,source=/etc/aws-environment,readOnly=false --mount volume=awsenv,target=/etc/aws-environment \
+          --net=host \
+          --trust-keys-from-https \
+          quay.io/coreos/awscli:edge -- cfn-init -v \
```
mumoshu: fyi cloudformation helper scripts are now included in `quay.io/coreos/awscli`(=`latest`) so you can use that instead 👍

pieterlange: I'm waiting for https://github.com/coreos/awscli/pull/4 to get merged into master (it's only in edge atm)

config/templates/cluster.yaml
```diff
@@ -165,8 +165,12 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # experimental:
 #   nodeDrainer:
 #     enabled: true
+#   awsEnvironment:
+#     enabled: true
+#     environment:
+#       CFNSTACK: '{ "Ref" : "AWS::StackId" }'
```
mumoshu: A very understandable way of configuration, at least for me! Awesome work 👍

config/templates/stack-template.json
```diff
@@ -438,6 +455,23 @@
...
+{{range $variable, $function := .Experimental.AwsEnvironment.Environment}}
+"{{$variable}}=", {{$function}} , "\n",
+{{end}}
+"' > /etc/aws-environment" ] ] }
+              }
+            }
+          }
+        }
+      },
+  {{end}}
```
mumoshu: Just curious but don't we need the same "Metadata" part for the etcd nodes?

config/templates/cloud-config-worker
```diff
@@ -170,7 +170,34 @@ coreos:
...
+
+{{if .Experimental.AwsEnvironment.Enabled}}
+    - name: set-aws-environment.service
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Set AWS environment variables in /etc/aws-environment
+        Before=flannel.service
```
mumoshu: Maybe nit but is there any reason this is `Before=flannel.service`?

mumoshu:
The travis build seems to have failed because of formatting error.
FYI, `make format` is your friend!

config/templates/stack-template.json
pieterlange: Maybe :-) I was actually thinking about these variables in the context of the node pools (so each pool of nodes gets their own set of variables). Also, i'd like to add a etcd autoscaling node group based on @crewjam's work which would need these template variables. TBD

mumoshu:
lgtm overall 👍
This is a kind of elegance I couldn't come up with! Awesome work 👏

I'm ready to merge this once nits and the build failure are cleared.

config/templates/cloud-config-worker
pieterlange: :+1: I just wanted to make it start really early on. What would be better?

config/templates/cluster.yaml
pieterlange: I'm putting it under experimental because it's very easy for users to break their template here. Was debating myself whether to further spec it into function & parameters but i think this would be way too opaque even if we could do some better syntax checking that way.

mumoshu: Could be. I personally like this kind of freedom/power in exchange for clarity but definitely need more thinking. Anyways, something like `CFNSTACK: 'AWS::StackId' won't be fundamentally different from your current implementation, right?

Do we need the new `advanced:` section in cluster.yaml when graduation from experimental? 😈

config/templates/cloud-config-worker
mumoshu: Quick thought but `After=network.service`?

config/templates/stack-template.json
mumoshu: Got it. Thanks for the explanation :bow:

mumoshu:
I've seen `WIP` in the commit message.
Would you like this PR to be open for a while and gather attentions to bring up more discussions/feedbacks?

config/templates/cluster.yaml
pieterlange: There's definitely a trade-off made here. I like to not prevent smart people doing smart/dangerous things :smiling_imp: 

`CFNSTACK: 'AWS::StackId'` crossed my mind as well, but [sometimes](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-eip.html#aws-resource-ec2-eip-ref) you need resource attributes that can only be obtained with [Fn::GetAtt](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html), not by `Ref`.

Better syntax checking could be done with something like:
```
environment:
  CFNSTACK:
    function: Ref
    parameter: AWS::StackId
```

I feel like this would ultimately complicate matters and not help all that much.

As for the `advanced:` section, i think there's some merit.. but i'm not sure yet. Isn't all of this advanced? :)

pieterlange:
I added it in `experimental` so we can change it later... let's get some experience using this so we can fix UX things later (if necessary).

config/templates/cluster.yaml
mumoshu: Thanks for the explanation 😃 
I'm afraid that the syntax checking thing is too much for now.
I'd rather merge this as is and want to see how our users play with this!

> Isn't all of this advanced? :)

Certainly! 😆

mumoshu:
@pieterlange Would you like to address https://github.com/coreos/kube-aws/pull/67#discussion_r88447344 before merging this?

pieterlange:
I did, and just changed it into `After=network-online.target` to it's even better aligned.

mumoshu:
Thanks as always @pieterlange :bow:

PR #69, opened 2016-11-17
EFS systemd unit start ordering.

pieterlange:
EFS can obviously only be mounted if we have a network.


mumoshu:
LGTM 👍 Thanks as always!

PR #72, opened 2016-11-17
Experimental: external etcd

pieterlange:
Refactored to use the `experimental` configuration parameter so we can continue to use this from `master`.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -326,12 +326,14 @@ func (c *Cluster) baseCreateStackInput() *cloudformation.CreateStackInput {
 		tags = append(tags, &cloudformation.Tag{Key: &key, Value: &value})
 	}
 
-	return &cloudformation.CreateStackInput{
-		StackName:    aws.String(c.ClusterName),
-		OnFailure:    aws.String(cloudformation.OnFailureDoNothing),
-		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
-		Tags:         tags,
-		StackPolicyBody: aws.String(`{
+	var creq *cloudformation.CreateStackInput
+	if(c.EtcdLoadBalancer == "") {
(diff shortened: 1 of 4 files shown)


pieterlange:
Will squash after feedback.

cluster/cluster.go
```diff
@@ -348,7 +357,9 @@ func (c *Cluster) baseCreateStackInput() *cloudformation.CreateStackInput {
   ]
 }
 `),
+		}
 	}
+	return creq
```
mumoshu: Almost nit but something like:

```go
var stackPolicyBody string
if c.Experimental.ExternalEtcd.Enabled {
  // We don't manage an etcd cluster in this stack(therefore `ExternalEtcd`), hence no need to lock
  stackPolicyBody = "{}"
} else {
  // Internally managed etcd cluster isn't updatable for now.
  // See https://github.com/coreos/coreos-kubernetes/pull/544#issuecomment-247483015
  stackPolicyBody = `{
    "Statement" : [
      {
        "Effect" : "Deny",
       "Action" : "Update:*",
       "Principal" : "*",
       "Resource" : "LogicalResourceId/InstanceEtcd*"
     },
     {
        "Effect" : "Allow",
        "Principal" : "*",
        "Action" : "Update:*",
        "Resource" : "*"
      }
    ]
  }
`
}

return &cloudformation.CreateStackInput{
    StackName:    aws.String(c.ClusterName),
    OnFailure:    aws.String(cloudformation.OnFailureDoNothing),
    Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
    Tags:         tags,
    StackPolicyBody: aws.String(stackPolicyBody),
}
```

would be better because then we can be explicit about the only difference is `StackPolicyBody`?

pieterlange: Thanks :+1:

PR #75, opened 2016-11-18
cmd/kubeaws: Add S3 location format to help flag

joshrosso:
- kube-aws up, update, and validate all feature the --s3-location flag.
  Users would benefit from knowing the format expected when specifying
  their S3 bucket. This commit adds an example to each instance of the
  flag.


mumoshu:
@joshrosso Thanks, it will certainly improve the user experience 👍

PR #77, opened 2016-11-18
Send sucesss signal to cloudformation when controllers and workers are ready

gianrubio:
This PR make some improvements when creating or updating cluster. #49 
When the resources are creating, cfn-signal will monitor specific services to check if they are healthy. Only after all service are healthy, it send a successful signal to Cloudformation.  The same monitoring will be applied on update command, now we have consistent status for rolling updates.

# Monitoring
On the node controller, cfn-signal will check the status for: calico if enabled, scheduler, api-server, controller-manager and kubelet. 

```
...
/usr/bin/curl -s -m 20 -f  http://127.0.0.1:8080/healthz > /dev/null &&
/usr/bin/curl -s -m 20 -f  http://127.0.0.1:10252/healthz > /dev/null && 
/usr/bin/curl -s -m 20 -f  http://127.0.0.1:10251/healthz > /dev/null &&  
/usr/bin/curl --insecure -s -m 20 -f  https://127.0.0.1:10250/healthz > /dev/null 
....
```

On Worker cfn-signal only monitor kubelet.
```...
/usr/bin/curl  --insecure -s -m 20 -f  https://127.0.0.1:10250/healthz > /dev/null 
```

I'm not totally satisfied with the worker monitoring, only watching kubelet. When upgrading a worker cluster,  there's a little downtime during the pulling images for pods and counting success probe. Suggestions are always welcome.

ps. I'm not yet watching etcd because the actual solution isn't reliable, waiting some improvement on #72

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -49,11 +49,13 @@ func newDefaultCluster() *Cluster {
 		TLSCertDurationDays:      365,
 		ContainerRuntime:         "docker",
 		ControllerCount:          1,
+		ControllerCreateTimeout:  "PT15M",
 		ControllerInstanceType:   "m3.medium",
 		ControllerRootVolumeType: "gp2",
 		ControllerRootVolumeIOPS: 0,
 		ControllerRootVolumeSize: 30,
 		WorkerCount:              1,
+		WorkerCreateTimeout:      "PT15M",
 		WorkerInstanceType:       "m3.medium",
 		WorkerRootVolumeType:     "gp2",
 		WorkerRootVolumeIOPS:     0,
@@ -135,11 +137,13 @@ type Cluster struct {
 	ReleaseChannel           string            `yaml:"releaseChannel,omitempty"`
 	AmiId                    string            `yaml:"amiId,omitempty"`
 	ControllerCount          int               `yaml:"controllerCount,omitempty"`
+	ControllerCreateTimeout  string            `yaml:"controllerCreateTimeout,omitempty"`
 	ControllerInstanceType   string            `yaml:"controllerInstanceType,omitempty"`
 	ControllerRootVolumeType string            `yaml:"controllerRootVolumeType,omitempty"`
 	ControllerRootVolumeIOPS int               `yaml:"controllerRootVolumeIOPS,omitempty"`
 	ControllerRootVolumeSize int               `yaml:"controllerRootVolumeSize,omitempty"`
 	WorkerCount              int               `yaml:"workerCount,omitempty"`
+	WorkerCreateTimeout      string            `yaml:"workerCreateTimeout,omitempty"`
 	WorkerInstanceType       string            `yaml:"workerInstanceType,omitempty"`
 	WorkerRootVolumeType     string            `yaml:"workerRootVolumeType,omitempty"`
 	WorkerRootVolumeIOPS     int               `yaml:"workerRootVolumeIOPS,omitempty"`
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -194,6 +194,29 @@ coreos:
         WantedBy=kubelet.service
 {{ end }}
 
+    - name: cfn-signal.service
+      command: start
+      content: |
+        [Unit]
+        Wants=kubelet.service docker.service
+        After=kubelet.service
+
+        [Service]
+        Type=oneshot
+        ExecStartPre=/usr/bin/bash -c "while sleep 1; do if /usr/bin/curl  --insecure -s -m 20 -f  https://127.0.0.1:10250/healthz > /dev/null ; then break ; fi;  done"
+        {{ if .UseCalico }}
+        ExecStartPre=/usr/bin/systemctl is-active calico-node
+        {{ end }}
+        ExecStart=/usr/bin/rkt run \
+          --uuid-file-save=/var/run/coreos/cfn-signal.uuid \
+          --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+          --net=host \
+          --trust-keys-from-https \
+          quay.io/coreos/awscli:edge -- cfn-signal -e 0 \
+              --region {{.Region}} \
+              --resource AutoScaleWorker \
+              --stack {{.ClusterName}}
+
 write_files:
   - path: /etc/kubernetes/cni/docker_opts_cni.env
     content: |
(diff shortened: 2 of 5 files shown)


pieterlange:
I see no issue at all with the success signal on successful kubelet initialization. Other measures should be in place to guarantee application (pod) availability (like running multiple replicas).

Can confirm functionality:
screenshot from 2016-11-18 17-59-06

config/config.go
```diff
@@ -49,11 +49,13 @@ func newDefaultCluster() *Cluster {
 		TLSCertDurationDays:      365,
 		ContainerRuntime:         "docker",
 		ControllerCount:          1,
+		ControllerCreateTimeout:  "PT15M",
 		ControllerInstanceType:   "m3.medium",
 		ControllerRootVolumeType: "gp2",
 		ControllerRootVolumeIOPS: 0,
 		ControllerRootVolumeSize: 30,
 		WorkerCount:              1,
+		WorkerCreateTimeout:      "PT15M",
```
mumoshu: Really just curious but any specific reason this is being PT15M(=15 min)?(Almost certainly doesn't related to whether this PR is merged sooner or not 😃

mumoshu:
@gianrubio This is really an awesome work 👍 

The little down-time you've mentioned could happen in certain cases(like having `replicas: 1` in a replica set or all the replicas are scheduled in a single node?) but IMHO we could avoid these cases by preparing:

* 2 or more replicas in each replicaset
* 2 or more worker nodes, to allow replicas to spread over nodes
* enough "free resources" in our clusters for scheduling `kubectl drain`ed pods to existing nodes(permanently to an already replaced node, or temporarily to a going-to-be-replaced-soon node)
* pray 🙏 

If necessary, I'd definitely like this to be discussed further in another github issues!

config/templates/cloud-config-worker
```diff
@@ -194,6 +194,29 @@ coreos:
         WantedBy=kubelet.service
 {{ end }}
 
+    - name: cfn-signal.service
+      command: start
+      content: |
+        [Unit]
+        Wants=kubelet.service docker.service
+        After=kubelet.service
```
mumoshu: This way, does `cfn-signal.service` get started only after `kubelet.service` successfully started?

mumoshu:
@gianrubio I've missed the fact that we're going to have `WillReplace: true`!
Edited my comment in https://github.com/coreos/kube-aws/pull/77#issuecomment-261820696 to accommodate that.

mumoshu:
One more idea how we could act to reduce downtime is:

* Run `kubectl drain ${old_node}` after kubelet starts up in a new node so that we can explicitly state that a node is going to replace the `old_node` and only after a successful drain, we signal cloudformation that the new node is ready.
  * One gotcha is that how to correlate a new node with an old node. Maybe suspending node autoscaling before `kube-aws update` and `aws autoscaling describe-autoscaling-instances | <sort  by instance id>` to obtain stably ordered lists of instances for the old asg and the new asg?
  * Sounds really hacky 😆 I wish a simple workaround like `making replicas 2x` would work.

mumoshu:
Am I missing something? Can we just do "rolling update" plus "cloudformation signal" somehow?
If so, avoiding downtime while `kube-aws update` shouldn't be that difficult!

config/templates/cloud-config-worker
pieterlange: That's the idea!

mumoshu: Great!
I guess I can borrow this idea to reduce useless error messages complaining about `systectl is-active flanneld.service` while kubelet.service is getting started.
Thanks 👍

config/config.go
pieterlange: 15 minutes is just a really lenient timeout and should be sufficient of nodes of any type (even t2's). If people want to specify their own timeouts (and fail fast! :+1:) they can.

pieterlange:
Yes, i believe using the experimental `cordon`/`uncordon` feature plus this PR should be sufficient to do mostly downtime-less upgrades. If you have really large images and/or only one replica things might get slightly more hairy. Also: `cordon`ing a node might evict pods to nodes that are about to be replaced, causing more pod movement than strictly necessary.

But what this PR actually does: if the operator makes a mistake in the configuration and new nodes don't show up healthy, the cluster upgrade will first stall and then roll back.

pieterlange:
Sidenote: users that want to prevent downtime should always spec 2 or more replica's. Nodes disappear on their own as well. :skull_and_crossbones:

mumoshu:
@pieterlange 

> Also: cordoning a node might evict pods to nodes that are about to be replaced, causing more pod movement than strictly necessary.

Agreed. That's why I've added the below comment:

> Additionally, I wonder if we could manually/temporarily marking old nodes unschedulable and setting replicasets' replicas values to 2x before running kube-aws update, so that pods are distributed to new nodes earlier

in https://github.com/coreos/kube-aws/pull/77#issuecomment-261820696 👍 

> But what this PR actually does: if the operator makes a mistake in the configuration and new nodes don't show up healthy, the cluster upgrade will first stall and then roll back.

Yes, this behavior itself is really useful!

pieterlange:
Ah.. sorry, i somehow missed that.

I think a lot of these update strategies become moot after we can update the control plane in-place using the self-hosted kubernetes designs. This way the pods stay where they are and no movement is required at all.

config/config.go
mumoshu: Understood. Thanks!

mumoshu:
@pieterlange Btw, I assume that using `WillReplace: true` means that we're not longer doing "rolling update=one by one deployment" but rather "all at once deployment"(=prepare a new ASG filled with instances passed health checks and then terminate all the instances in the old ASG`), right?

Just wanted to confirm because that it is my biggest concern regarding down time and source of my hacky ideans like https://github.com/coreos/kube-aws/pull/77#issuecomment-261845313.

config/config.go
gianrubio: @pieterlange got it, 15 min is a reasonable time for creating workers and controllers but it's easy to customize on clusters.yaml.

gianrubio:
@mumoshu as you suggested, I removed the WillReplace config. Now the cluster takes more time to upgrade, but we're doing in the right way. 

@pieterlange is working in the proposal for cordon/uncordon.

screen shot 2016-11-21 at 16 38 12

mumoshu:
@gianrubio Oh, just removing `WillReplace: true` results in the one-by-one deployment?
Great! 👍 

I'd been somehow misreading the AWS doc/misunderstood that cfn signals only work in combination with `WillReplace: true`, but actually it wasn't. That's good to know.

mumoshu:
Merged to get the ball rolling ⚽️

mumoshu:
Thanks for your efforts on this great improvement @gianrubio @pieterlange!

PR #78, opened 2016-11-21
Various improvements to kube-aws e2e testing script

mumoshu:
* easier ways to connect worker/controller via ssh
* supress verbose logging while waiting for kube-apiserver to be up
* differentiate s3-uri for each cluster so that we can safely run multiple e2e tests with different configurations in parallel
* set KUBE_AWS_USE_CALICO envvar to provision clusters with calico enabled
* print startup time taken since `kube-aws up` until apiserver becomes responsive so that I could notice unexpected effects to startup time more
* Make nodes count smaller because conformance tests seem to pass with 2 nodes now


mumoshu:
Merging as this is already tested multiple times while releasing RCs 😃

PR #79, opened 2016-11-21
Make decrypt-tls-assets faster by consolidating multiple `rkt run`s into one shot

### LINKED ISSUE #76: decrypt-tls-assets needlessly taking time
I've noticed that the current implementation of decrypt-tls-assets running `rkt run` multiple times seems to be resulting in adding up to 20 seconds(or possibly more?) in total run time.

One of the root causes seems to be that `rkt run` runs integrity check on a container image. Possibly related https://github.com/coreos/rkt/issues/1350
Without bypassing the check, we can make total run time shorter via running `rkt run` just once and loop over all the tls assets to decrypt inside the container.

I've benchmarked three implementations from v1 to v3. v1 is the original implementation.
https://gist.github.com/mumoshu/d1bb67ba48b975eccbec4bdf83559ade

v2 looks good as it has a good balance of simplicity and speed.

vanilla v0.9.1-rc.4:
https://gist.github.com/mumoshu/8473065955a4602677fc2698c3720abf

v0.9.1-rc.4 + v2 decrypt-tls-assets:
https://gist.github.com/mumoshu/8bec0b49e7628ce5d929b467b1f70e15

mumoshu:
fix #76

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -292,8 +292,7 @@ write_files:
     content: |
       #!/bin/bash -e
 
-      for encKey in $(find /etc/kubernetes/ssl/*.pem.enc);do
-        sudo rkt run \
+      sudo rkt run \
         --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
         --mount=volume=ssl,target=/etc/kubernetes/ssl \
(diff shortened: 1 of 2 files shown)


mumoshu:
Merging as this is already tested/benchmarked as I've explained in #76 😃

PR #80, opened 2016-11-21
Fix static reference to AWS region

pieterlange:
PR #79 introduced a static reference to `ap-northeast-1`.


mumoshu:
@pieterlange My fault! Thanks for fixing it :bow:

PR #82, opened 2016-11-22
Shield cfn-signal behind experimental flag.

pieterlange:
In preparation for node cordoning/uncordon process 1/2

I think it's safer to have an easy way to disable this while we work on it.

Note that this reverts the PauseTime to 2 minutes and the `WorkerCreateTimeout` / `ControllerCreateTimeout` remain top level variables.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -37,6 +37,10 @@ func newDefaultCluster() *Cluster {
 		AwsEnvironment{
 			Enabled: false,
 		},
+		WaitSignal{
+			Enabled:      false,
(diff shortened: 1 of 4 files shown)


mumoshu:
If you have not yet done so, `make format` seems to be required!

pieterlange:
Will i ever learn..

pieterlange:
Real world check successful. LGTM :+1:

mumoshu:
Thanks as always, @pieterlange!

Issue #46, opened 2016-11-08
Feature: Node pools


mumoshu:
succeed to https://github.com/coreos/coreos-kubernetes/issues/667

---

I'd like to add GKE's "Node Pools" like feature to kube-aws.

With that feature, we can differentiate the following things per pool without maintaining multiple etcd clusters and kubernetes controller planes:

* instance type
* storage size
* instance profile
* security group
* min/max spot price
* auto scaling group or spot fleet
* node labels

It will eventually give us more granular control over:

* security
  * in combination with node selectors, we can
    1. restrict nodes which are able to communicate with databases containing sensitive information and
    2. schedule only necessary pods onto the nodes
* cost effectiveness
  * in combination with node selectors, we can
    1. make pool(s) backed by spot instances and label nodes with e.g. `spot=true`
    2. explicitly schedule pods which can be spawned/killed more aggressively onto those nodes
* scalability
  * creating a separate pool per availability zone allows us utilizing [cluster-autoscaler](https://github.com/kubernetes/contrib/tree/master/cluster-autoscaler) in the multi-AZ setup (see https://github.com/coreos/coreos-kubernetes/issues/668)
* etc
  * If you are aware of one, let us know!

Beware that until the taint-and-toleration feature is implemented into Kubernetes, we'd need to give pods specific node selectors to prevent them from being scheduled onto undesirable nodes.

Edit: Taints and tolerations are partially supported in Kubernetes 1.4.x and introduced in kube-aws since #132 

## TODOs(possibly done in separate pull requests)

* [x] `kube-aws nodepool update` (#130)
  * [x] fix `kube-aws nodepool update` which was always failing 😉 (#140)
* [x] `kube-aws nodepool validate` #161 
* [x] `kube-aws nodepool destroy` (#139)
* [x] Node labels configurable via `node-pools/yourpoolname/cluster.yaml` #149 
* [x] Spot Fleet support (#112 #113)
* [x] confirm https://github.com/coreos/kube-aws/issues/68 doesn't affect node pools
* [x] redo https://github.com/coreos/kube-aws/issues/49 for node pools (#128)

## Non-TODOs

* Multi region deployment a.k.a deploy node pools in another region(s)

mumoshu:
For now, I'm going to build a POC like I've described in https://github.com/coreos/coreos-kubernetes/issues/667#issuecomment-254423979

Any requests, comments, proposals are welcomed!

pieterlange:
[...]

PR #83, opened 2016-11-22
Add additional AWS labels and annotations to node

pieterlange:
Currently adds:

Labels:
- `kube-aws.coreos.com/autoscalinggroup` = autoscalinggroup the node belongs to
- `kube-aws.coreos.com/launchconfiguration` = launchconfiguration of the autoscalinggroup during node intialization

Annotations:
- `kube-aws.coreos.com/securitygroups` = comma separated list of security groups attached to node

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,6 +34,9 @@ func newDefaultCluster() *Cluster {
 		NodeDrainer{
 			Enabled: false,
 		},
+		NodeLabel{
+			Enabled: false,
+		},
 		AwsEnvironment{
 			Enabled: false,
 		},
@@ -201,11 +204,16 @@ type Subnet struct {
 
 type Experimental struct {
 	NodeDrainer    NodeDrainer    `yaml:"nodeDrainer"`
+	NodeLabel      NodeLabel      `yaml:"nodeLabel"`
 	AwsEnvironment AwsEnvironment `yaml:"awsEnvironment"`
 	WaitSignal     WaitSignal     `yaml:"waitSignal"`
 }
 
 type NodeDrainer struct {
+	Enabled bool `yaml:"enabled"`
+}
+
+type NodeLabel struct {
 	Enabled bool `yaml:"enabled"`
 }
 
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -210,7 +210,6 @@ coreos:
         ExecStartPre=/usr/bin/curl http://127.0.0.1:8080/version
         ExecStart=/opt/bin/install-calico-system
 {{ end }}
-
 {{ if $.ElasticFileSystemID }}
     - name: rpc-statd.service
       command: start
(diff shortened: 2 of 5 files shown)


mumoshu:
Thanks @pieterlange! Code lgtm 👍 May I ask what's the use-case for each label?

pieterlange:
The direct usecase here is being able to distinguish nodes from separate pools for the `cfn-signal` cordon/uncordon process in order to prevent excessive pod movement across the cluster.

mumoshu:
@pieterlange Sounds great! but may I ask you how are you going to use those labels in cordon/uncordon process or in its preparation process?

mumoshu:
This PR needs to be rebased as #82 is merged.

pieterlange:
Rebased.

---

> However I'm afraid of encountering race conditions as I've described in #46 (comment) by making a non-zero delay until node is labeled after it became ready.

Pretty tough call. We could start the workers marked as `SchedulingDisabled` and re-enable scheduling after proper labelling took place? We'd have to put the pod in the `/etc/kubernetes/manifests` directory to make sure the label-setter-pod would get started though.

I was originally going to use this route, but as the labels only need to be set on node-init time, it made more sense to use a systemd oneshot instead of a persistent pod.

> Do you think a race condition like that wouldn't happen in practice or wouldn't break your use-case?

In my use case it wouldn't matter: i would use the labels to select nodes to mark `SchedulingDisabled`, and only after they've all been cordoned start the shutdown/drain process of the node. The replace strategy makes sure a fresh node has already been booted, so the pods have only one place to go (the new node), thus preventing pod movement to nodes about to be terminated.

mumoshu:
I've talked with @pieterlange in Slack and now I'd rather like to merge this PR as is and discuss how we sort out various options used to label nodes:

* oneshot systemd service, pod(s) deployed via `/etc/kubernetes/manifests`, `ExecStartPre=curl/docker run kubectl/rkt run kubectl`, etc.
* which to use for calling commands like awscli in containers: docker or rkt
  * `rkt run` formats stdout from containers in its own way.
  * How can we capture only necessary parts of a container's stdout, hopefully without sed/cut/awk etc?
* use `--node-labels` or not to prevent race conditions
  * if not, should we go with `kubelet --register-node=true --register-schedulable=false` to register node while marking it Unschedurable in the beginning, and later uncordon it?
  * or in another way(s)?

...in my upcoming node pools PR or in another issues 😃 

Main motivations for doing so is:
* This change does work for some use-cases
* We don't need to worry about breaking things badly as @pieterlange kindly authored it as an experimental feature 👍 
* There's too much to think about for now. I'd like to just get the ball rolling here

mumoshu:
@pieterlange Would you mind rebasing this again? It seems to conflict with your previous pull request. Also, I guess you'd like this to be merged before I'll make a huge refactoring in preparation to merge the node pools feature 😃

pieterlange:
Rebased again. Thanks for summarizing our talk :+1:

mumoshu:
Thanks for rebasing :bow: Merging.

PR #89, opened 2016-11-23
Add cloudinit sshAuthorizedKeys option to cluster.yaml

pieterlange:
So cluster operators don't have to share the same SSH public key.

Suggested workflow:
- Keep a main organization key (`keyPair`) in a keysafe. Don't use it for normal daily operations.
- Add a select number of cluster operators to the `sshAuthorizedKeys` list


mumoshu:
This would be a nice addition 👍 
No one wants to script `for k in pubkeys/*; do echo "- $(cat $k)" >> cluster.yaml; done` to just automate a duty like this :)

PR #92, opened 2016-11-24
Add experimental feature to enable rbac authorization plugin

gianrubio:
This flag enable the authorization rbac plugin.

http://kubernetes.io/docs/admin/authorization/

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,11 @@ const (
 
 func NewDefaultCluster() *Cluster {
 	experimental := Experimental{
+		AuditLog{
+			Enabled: false,
+			MaxAge:  30,
+			LogPath: "/dev/stdout",
+		},
 		AwsEnvironment{
 			Enabled: false,
 		},
@@ -47,6 +52,11 @@ func NewDefaultCluster() *Cluster {
 		},
 		NodeLabel{
 			Enabled: false,
+		},
+		Plugins{
+			Rbac{
+				Enabled: false,
+			},
 		},
 		WaitSignal{
 			Enabled:      false,
@@ -273,17 +283,25 @@ type Subnet struct {
 }
 
 type Experimental struct {
+	AuditLog              AuditLog              `yaml:"auditLog"`
 	AwsEnvironment        AwsEnvironment        `yaml:"awsEnvironment"`
 	EphemeralImageStorage EphemeralImageStorage `yaml:"ephemeralImageStorage"`
 	LoadBalancer          LoadBalancer          `yaml:"loadBalancer"`
 	NodeDrainer           NodeDrainer           `yaml:"nodeDrainer"`
 	NodeLabel             NodeLabel             `yaml:"nodeLabel"`
+	Plugins               Plugins               `yaml:"plugins"`
 	WaitSignal            WaitSignal            `yaml:"waitSignal"`
 }
 
 type AwsEnvironment struct {
 	Enabled     bool              `yaml:"enabled"`
 	Environment map[string]string `yaml:"environment"`
+}
+
+type AuditLog struct {
+	Enabled bool   `yaml:"enabled"`
+	MaxAge  int    `yaml:"maxage"`
+	LogPath string `yaml:"logpath"`
 }
 
 type EphemeralImageStorage struct {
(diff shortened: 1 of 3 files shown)


pieterlange:
Learn to love `make format` :+1: :laughing:

danielfm:
Not sure if what I'm gonna say next should be considered as part of this issue or not, but I recently updated my cluster configuration to enable RBAC as well, and one thing that helped me a lot along the way is to enable the [API server auditing](http://kubernetes.io/docs/admin/audit/) feature as well.

With the audit logs enabled, I was able to see all denied requests coming from various sources (including internal Kubernetes components, such as the Kubelet) and define the required role bindings without much guess work.

In my configuration, I added these parameters to the API server manifest file:

```
--audit-log-maxage=30
--audit-log-path=/var/log/audit/audit.log
```

In order to be able to ship this logs to our centralized logging solution, I also had to expose the host's`/var/log/audit` as a volume to the API server container, and configured Fluentd to collect logs from this file as well:

```conf
# Somewhere in our td-agent.conf (fluentd)

# API Server audit logs
<source>
  type tail
  path /var/log/audit/audit.log
  pos_file /var/log/audit/audit.log.pos
  format /^(?<time>.*) AUDIT: id=\"(?<id>[^\"]*)\" (response=\"(?<response>[^\"]*)\"$)?(ip=\"(?<ip>[^\"]*)\" method=\"(?<method>[^\"]*)\" user=\"(?<user>[^\"]*)\" as=\"(?<as>[^\"]*)\" namespace=\"(?<namespace>[^\"]*)\" uri=\"(?<uri>[^\"]*)\"$)?/
  tag k8s-api-audit
  read_from_head true
</source>

<match k8s-api-audit>
  format json
  ...
</match>
```

pieterlange:
This is great feedback @danielfm ! With RBAC (as a principle) being a debug nightmare some nice tracing would surely help!

With more plugins (#94) coming, should we simplify this under one unified "Experimental.Plugins" yaml tree?

Ex:
```yaml
experimental:
  plugins:
    rbac:
      enabled: true
      auditlog: /var/log/audit/audit.log
    cronjob:
      enabled: true
```

We're already peppering the userdata and stack-template with plenty of `if`s, this may help for better readability?

danielfm:
@pieterlange the API server audit log functionality is not coupled with RBAC in any way, but the suggested YAML structure might give the impression that it does, so I don't think this is the best way to expose this configuration.

I dig this idea of having all extra/experimental features somewhat decoupled from the basic cluster configuration though. 😄

pieterlange:
I know it's not actually coupled but as you say its highly useful when you enable RBAC so it might make sense to place it under that structure. It was just an example though :-).

mumoshu:
Thanks for this PR @gianrubio! I like this idea.
Could you rebase this?
// Thanks to recent pace of development, we're conflicting badly here and there!

mumoshu:
@danielfm Great feedback! You've saved me/us from scratching our head for multiple days 🤕 

As audit logging being a completely independent feature,

* I'd rather like it to be toggled under an another configuration key like `.experimental.(plugins.)auditlog`
* while encouraging everyone to enable `auditlog` whenever `rbac` is enabled, possibly in comments in `cluster.yaml`

I even guess that we'd better write what you've shared us in https://github.com/coreos/kube-aws/pull/92#issuecomment-262821681 into a doc under `Documentation/`.

gianrubio:
@danielfm good feedback, I created a flag to accept audit log as option on cluster.yaml

```
  auditLog:
    enabled: true
    maxage: 30
    logpath: /var/log/audit/audit.log
```

@pieterlange and @mumoshu  I accepted your suggestion for plugins, now the plugins are a config item on cluster.yaml

```
plugins:
     rbac:
       enabled: true
```

config/config.go
```diff
@@ -244,6 +252,16 @@ type LoadBalancer struct {
 	SecurityGroupIds []string `yaml:"securityGroupIds"`
 }
 
+type Plugins struct {
+	Rbac map[string]string `yaml:"rbac"`
```
pieterlange: Can you make this explicit? This doesn't need free form params and helps with test coverage

EG
```golang
type Plugins struct {
  Rbac Rbac `yaml:"rbac"`
}

type Rbac struct {
  Enabled bool `yaml:"enabled"`
}
```

mumoshu: Yes, making it an another struct would be better 👍
@gianrubio, did you have any chance to address it?

gianrubio:
@mumoshu could you review again?

mumoshu:
@gianrubio LGTM 👍 Thanks for your contribution!

PR #94, opened 2016-11-24
Enable Scheduled Jobs in k8s 1.4  aka Cron Jobs in 1.5

camilb:
`Scheduled Jobs` from 1.4 was renamed to  `Cron Jobs` in 1.5.
This PR works with both versions.

More info here:
http://kubernetes.io/docs/user-guide/cron-jobs/

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -344,7 +344,7 @@ write_files:
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
           - --client-ca-file=/etc/kubernetes/ssl/ca.pem
           - --service-account-key-file=/etc/kubernetes/ssl/apiserver-key.pem
-          - --runtime-config=extensions/v1beta1/networkpolicies=true
+          - --runtime-config=extensions/v1beta1/networkpolicies=true,batch/v2alpha1
           - --cloud-provider=aws
           livenessProbe:
             httpGet:


config/templates/cloud-config-controller
```diff
@@ -344,7 +344,7 @@ write_files:
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
           - --client-ca-file=/etc/kubernetes/ssl/ca.pem
           - --service-account-key-file=/etc/kubernetes/ssl/apiserver-key.pem
-          - --runtime-config=extensions/v1beta1/networkpolicies=true
+          - --runtime-config=extensions/v1beta1/networkpolicies=true, batch/v2alpha1
```
danielfm: @camilb why not `batch/v2alpha1=true`, to be consistent in style with the other flag?

camilb: @danielfm Works both ways and actually I have it like this `batch/v2alpha1=true`. Thought using the recommended method from the documentation `--runtime-config=batch/v2alpha1`

pieterlange: There is a space in the command argument, can you remove that?

I was going to ask to shield this behind an experimental flag like [this PR](https://github.com/coreos/kube-aws/pull/92/files) but i don't think enabling cronjobs is going to be dangerous..

camilb: @pieterlange I'll remove the space, don't remember where I saw an example using the space. Do you think it's also better to use `batch/v2alpha1=true` for consistency?

pieterlange: I don't mind behind explicit about this :+1:

mumoshu:
LGTM 👍 Thanks as always @camilb!

PR #95, opened 2016-11-24
Add option to mount ephemeral disks for image storage

pieterlange:
Intended to be used on instances that come with local storage (eg the `m3`, `c3`,`r3` and`g2` instance types).

When just setting the  `enabled` flag it formats (XFS) & mounts the local `xvdb` disk by default to the appropriate location (depending on container runtime). I chose XFS because of speed and the generous amounts of inodes by default.
Users can choose the filesystem and source disk by overriding the defaults:
```yaml
experimental:
  ephemeralImageStorage:
    enabled: true
    disk: xvdc
    filesystem: ext4
```

:warning: i haven't tested `rkt` as container runtime yet but i guess this will work? :crossed_swords: 

I would like to add the option to RAID together some local disks. The defaults should work all nodes with local storage, but doesn't use additional disks which is a waste if you move up to nodes of reasonable size (`xlarge`). Might add format/mount options later too.

Submitting incomplete work because @mumoshu seems to like rolling balls :laughing: :soccer:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -40,6 +40,11 @@ func newDefaultCluster() *Cluster {
 		AwsEnvironment{
 			Enabled: false,
 		},
+		EphemeralImageStorage{
+			Enabled:    false,
+			Disk:       "xvdb",
+			Filesystem: "xfs",
+		},
 		WaitSignal{
 			Enabled:      false,
 			MaxBatchSize: 1,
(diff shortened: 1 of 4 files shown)


mumoshu:
I like this idea and yes, I like rolling balls! 🍙 💨 
I've left a minor comment. Other than that, I'm ok to merge this.

pieterlange:
Done :+1:

mumoshu:
LGTM 👍 Thanks for your continuous efforts 🙏

tarvip:
Although this PR is already merged, but what about adding `var-lib-docker.mount` as dependency for `docker.service`, to make sure that it gets mounted before docker starts? Something like:
```
    - name: docker.service
      drop-ins:
{{if .Experimental.EphemeralImageStorage.Enabled}}
        - name: 10-docker-mount.conf
          content: |
            [Unit]
            After=var-lib-docker.mount
            Wants=var-lib-docker.mount
{{end}}
```

Using `Wants` allows using this drop-in file even if ephemeral storage is unavailable.

pieterlange:
Sounds like a plan :+1: Will you make the PR? :wink:

PR #98, opened 2016-11-28
fix: Remove problematic states from Config objects

mumoshu:
Config() func had been returning inconsistent result for each call by incrementing private IPs statically assigned to etcd nodes in configuration time.

This has never affected clusters created via kube-aws but does break the upcoming node pools feature. So I'm fixing this now.


mumoshu:
Fixed the commit message to match pull request's subject and description.

PR #102, opened 2016-11-28
refactoring: Extract coreos-cloud-init userdata validation

mumoshu:
Extract it out of the config package into its own package "userdatavalidation" for future use in the upcoming node pools feature.
The former `coreosutil` package is now named `coreos/amiregistry` and the validation is in the `coreos/userdatavalidation` package to reflect the fact that both of them are coreos specific but they have different responsiblity compared to each other.


mumoshu:
Btw, though it should be decided case-by-case, basically [I don't like naming something `util`](https://blog.golang.org/package-names#TOC_5.) so this is natural in that regard 😃

PR #103, opened 2016-11-28
refactoring: Make the Compact func and the KMS ecryption interface

mumoshu:
which is used for compacting TLSassets more open to be reused from other packages

This makes implementing the upcoming node pools feature a bit easier.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -93,7 +93,7 @@ func newDefaultCluster() *Cluster {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


mumoshu:
This does change the signature of the `compact` function but I'm assuming that this is an non-breaking change as the original function was private to the config package.
It might be obvious but I wanted to be clear I'm careful enough 😄

PR #106, opened 2016-11-28
Initial implementation of Node Pools

mumoshu:
The initial implementation for #46.

Several pull requests were already merged to make this easier, smaller:

#97, #98, #99, #100, #101, #102, #103, #104, #105

Our E2E test suite now include `kube-aws node-pools` and Kubernetes conformance tests are passing.

---

Notes on the initial implementation of Node Pools.

## Usage

`kube-aws node-pools (init|render|up) --node-pool-name mypool` commands assume you run `kube-aws` from the same working directory where you've run `kube-aws (init|render|up)` hence the `cluster.yaml` and other assets exist.

```
$ mkdir mycluster
$ cd mycluster
$ kube-aws init --cluster-name mycluster ...
$ kube-aws render credentials
$ kube-aws render stack
$ kube-aws up ...
$ kube-aws node-pools init --node-pool-name mypool ...
$ kube-aws node-pools render stack --node-pool-name mypool ...
$ kube-aws node-pools up --node-pool-name mypool ...
```

## Design

* A node pool is a group of worker nodes.
* Worker nodes in the same node pool have the same instance type, volume type, volume size.
* A cluster is not a node pool.
* A cluster may be called a "main" cluster if needed to be very explicit.
* A cluster has been able to contain 1 or more etcd nodes, 1 or more controller nodes, 1 or more worker nodes.
* A cluster can contain zero or more node pools from now on.
* Vanilla `kube-aws` sub-commands work for a cluster, while `kube-aws node-pools` sub-commands work for a node pool.

## Sources of inspiration

* GKE/the `gcloud` command
  * Both `kube-aws` and `gcloud` uses notion of `<root cmd> node-pools` for subcommands, instead of `nodepool`, `nodepools`, `node-pool`.

## Specifications
- 1 cloudformation stack for each main cluster, 1 stack for each node pool.
- We now have separate `cluster.yaml`, `stack-template.json`, `user-data/cloud-init-worker` for each node pool.
- I tried to make `cluster.yaml` for node pools look familiar to us. You'll see the `worker*` config keys e.g. `workerInstanceType`, `workerRootVolumeSize`, `workerRootVolumeType`, etc. in both main cluster's and node pools' `cluster.yaml`, instead of seeing `instanceType`, `rootVolumeSize`, `rootVolumeType`, etc. without `worker` prefix only in node pools' `cluster.yaml`.
  -  The end result is a bit verbose but consistent, familiar looking `cluster.yaml`s for clusters themselves and their addition - node pools.
- Worker nodes in a main cluster and node pools are assigned to the same security group(the one with the logical name `SecurityGroupWorker` in cloudformation stacks) so that we can allow etcd nodes in the main cluster can communicate with newly added worker nodes from node pools, controller nodes in the main cluster can communicate with the worker nodes.
- A cloudformation stack for a node pool works in two ways
  - If you've omitted `vpcId` and/or `routeTableId` in a node pool's `cluster.yaml`, `kube-aws node-pools up` will generate a cfn stack template containing [`Fn::ImportValue`s](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html) to import required values from the main cluster's cfn stack.
  - If you've specified `vpcId` and/or `routeTableId` in it, `kube-aws nodepool up` doesn't import values from the main stack. It just embed values as is, instead of embedding `Fn::ImportValue`s. This is achieved via computing `RouteTableRef` and `VPCRef` in go code and referencing it from `stack-template.json` for node pools. (refs to [go code](https://github.com/mumoshu/kube-aws/blob/node-pools/nodepool/config/config.go#L150-L160) and [stack-template.json](https://github.com/mumoshu/kube-aws/blob/node-pools/nodepool/config/templates/stack-template.json#L204-L212))

## Implications
- It won't allow us to create kube-aws node pools connected to non-kube-aws main clusters in a clean way. Though allowing it would be out of scope for this issue, I just wanted to note.

## File tree

The whole file tree for a main cluster named `mycluster` with a node pool named `mypool` would look like:

```
$ tree mycluster/
mycluster/
├── cluster.yaml
├── credentials
│   ├── admin-key.pem
│   ├── admin.pem
│   ├── apiserver-key.pem
│   ├── apiserver.pem
│   ├── ca-key.pem
│   ├── ca.pem
│   ├── etcd-client-key.pem
│   ├── etcd-client.pem
│   ├── etcd-key.pem
│   ├── etcd.pem
│   ├── worker-key.pem
│   └── worker.pem
├── kubeconfig
├── node-pools (NEW!)
│   └── mypool
│       ├── cluster.yaml
│       ├── stack-template.json
│       └── userdata
│           └── cloud-config-worker
├── stack-template.json
└── userdata
    ├── cloud-config-controller
    ├── cloud-config-etcd
    └── cloud-config-worker

5 directories, 21 files
```

# Notes for POC users/testers

While in the POC building phase, `nodePoolName` in a node pool's `cluster.yaml` had been `clusterName`, `clusterName` had been `baseClusterName` respectively.

See notes on the POC: https://github.com/coreos/kube-aws/issues/46#issuecomment-261115517

They're now named `nodePoolName` and `clusterName` respectively to be more explicit about the clusterName is clusterName - the same variable in the main cluster's `cluster.yaml`, and the design which differentiates a cluster and a node pool. A cluster is not a node pool, and vice versa.

### DIFF
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -1119,11 +1119,6 @@
   },
 
   "Outputs": {
-    "WorkerSecurityGroup" : {
-      "Description" : "The security group assigned to worker nodes",
-      "Value" :  { "Ref" : "SecurityGroupWorker" },
-      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
-    },
     {{if not .VPCID}}
     "RouteTable" : {
       "Description" : "The route table assigned managed by this stack",
@@ -1134,7 +1129,12 @@
       "Description" : "The VPC managed by this stack",
       "Value" :  { "Ref" : "{{.VPCLogicalName}}" },
       "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-VPC" }}
-    }
+    },
     {{end}}
+    "WorkerSecurityGroup" : {
+      "Description" : "The security group assigned to worker nodes",
+      "Value" :  { "Ref" : "SecurityGroupWorker" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
+    }
   }
 }
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -1116,5 +1116,25 @@
     {{end}}
     {{end}}
 
+  },
+
+  "Outputs": {
+    "WorkerSecurityGroup" : {
+      "Description" : "The security group assigned to worker nodes",
+      "Value" :  { "Ref" : "SecurityGroupWorker" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
+    },
+    {{if not .VPCID}}
+    "RouteTable" : {
+      "Description" : "The route table assigned managed by this stack",
+      "Value" :  { "Ref" : "RouteTable" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-RouteTable" }}
+    },
+    "VPC" : {
+      "Description" : "The VPC managed by this stack",
+      "Value" :  { "Ref" : "{{.VPCLogicalName}}" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-VPC" }}
+    }
+    {{end}}
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /e2e/assets
 *~
 /config/templates.go
+/nodepool/config/templates.go
 .idea/
 .envrc
 coverage.txt
--- a/build
+++ b/build
@@ -19,6 +19,7 @@ fi
 echo Building kube-aws ${VERSION}
 
 go generate ./config
+go generate ./nodepool/config
 
 if [[ ! "${BUILD_GOOS:-}" == "" ]];then
   export GOOS=$BUILD_GOOS
--- a/cmd/nodepool.go
+++ b/cmd/nodepool.go
@@ -1,0 +1,7 @@
+package cmd
+
+import "github.com/coreos/kube-aws/cmd/nodepool"
+
+func init() {
+	RootCmd.AddCommand(nodepool.NodePoolCmd)
+}
--- a/cmd/nodepool/init.go
+++ b/cmd/nodepool/init.go
@@ -1,0 +1,137 @@
+package nodepool
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	cfg "github.com/coreos/kube-aws/config"
+	"github.com/coreos/kube-aws/filegen"
+	"github.com/coreos/kube-aws/nodepool/config"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdInit = &cobra.Command{
+		Use:          "init",
+		Short:        "Initialize default node pool configuration",
+		Long:         ``,
+		RunE:         runCmdInit,
+		SilenceUsage: true,
+	}
+
+	initOpts = config.ComputedConfig{}
+)
+
+func init() {
+	NodePoolCmd.AddCommand(cmdInit)
+	cmdInit.Flags().StringVar(&initOpts.AvailabilityZone, "availability-zone", "", "The AWS availability-zone to deploy to")
+	cmdInit.Flags().StringVar(&initOpts.KeyName, "key-name", "", "The AWS key-pair for ssh access to nodes")
+	cmdInit.Flags().StringVar(&initOpts.KMSKeyARN, "kms-key-arn", "", "The ARN of the AWS KMS key for encrypting TLS assets")
+	cmdInit.Flags().StringVar(&initOpts.AmiId, "ami-id", "", "The AMI ID of CoreOS")
+}
+
+func runCmdInit(cmd *cobra.Command, args []string) error {
+	initOpts.NodePoolName = nodePoolOpts.PoolName
+
+	// Validate flags.
+	required := []struct {
+		name, val string
+	}{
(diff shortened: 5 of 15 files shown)


mumoshu:
Thanks for your interests in Node Pools @pieterlange @c-knowles @camilb @nicr9, I finally made it! 🎉
This will be merged soon to start getting feedback from users.
I'd appreciate it if you could take a look into this regardless of it is before/after merging 😄

c-knowles:
I had a look through today along with latest master while I was trying to figure out how to put shared route tables in there as I'd like to update from my fork of the previous repo. It looks great for the most part, I like that you've started to separate the concerns so they are easier to test. The options have grown significantly in the past couple of weeks, I wonder how to maintain acceptance tests with so many options. Perhaps something like [ginkgo](https://onsi.github.io/ginkgo/)?

config/templates/stack-template.json
```diff
@@ -1116,5 +1116,25 @@
...
     {{end}}
 
+  },
+
+  "Outputs": {
+    "WorkerSecurityGroup" : {
+      "Description" : "The security group assigned to worker nodes",
+      "Value" :  { "Ref" : "SecurityGroupWorker" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
+    },
```
camilb: Using existing `vpcId` will cause an error on `kube-aws up`: 

    Error: Failed to render stack template: invalid character '}' looking for beginning of object key string:
    json syntax error (offset=72048), in this region:
    -------
    }
        },
    
      }
    }

Quickly fixed by moving the comma from L:1126 to L1128

    "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
    }
    {{if not .VPCID}}
    ,
    "RouteTable" : {

camilb:
@mumoshu Congratulations on implementing this feature. Already started some tests on it.

config/templates/stack-template.json
mumoshu: @camilb Good catch! Thanks a lot :bow:

I believe I've made this kind of mistakes several times so far.
To reduce the chance of doing the same again, firstly I've [prepared tests to cover these cases](https://github.com/coreos/kube-aws/pull/106/commits/d68b0350f8c444d42254d745c837afabd92a4a1e) and then [fixed the issue](https://github.com/coreos/kube-aws/pull/106/commits/940ba3f22161eea282d1652694b8a21ef506c439).

I wish I could cover more combinations of settings, options in our e2e test but it had been a bit costly in time and AWS charge 😈
I'd like to just keep covering more cases in unit testing like I did this time for now.

mumoshu:
@c-knowles Thanks for taking a look and your efforts the route table sharing feature 👍 

I've written my feelings about testing shortly in https://github.com/coreos/kube-aws/pull/106#discussion_r89931122

I'd definitely like to go with ginkgo in the future as it is well proven to work in kubernetes's own e2e test suite.

Also, as you've pointed, a challenge is that there're already too many options and settings in kube-aws to test e.g. all the possible combinations.

IMHO what would work is to write tests as much as we can in the following precedence:

* Unit tests under the golang packages in kube-aws
   * Something like [what I've added today for testing node pools](https://github.com/coreos/kube-aws/pull/106/commits/95c35d296bbf6d064dc177e8638121910ce77ecf) would be enough in many cases as it indeed reveal easy mistakes like json syntax errors
* Integration tests under e.g. a directory named like `integration/`
   * Test possible combinations of options and settings as many as we can by validating CloudFormation templates generated by e.g. calling the `ValidateStack` func
   * so that we can ensure these combinations to be "valid" at least in CloudFormation
* E2E tests under the `e2e/` directory (Is it better to call this "acceptance" tests?)
  * Test limited/selected combinations of options and settings as many as we can by running series of kube-aws commands and kubernetes conformance tests and possibly kube-aws specific tests.
  * so that we can ensure these combinations to "work"
  * [They're currently implemented with bash + Docker file + Makefile](https://github.com/coreos/kube-aws/tree/master/e2e) but would be better implemented with ginkgo when it gets more complicated
  * Let me also add that they're occasionally run manually by me 😃 I hope they can be automated somehow in the future but we need at least a dedicated AWS account for testing purpose.

Any thoughts on these and where we should go,
or are there any kind(s) of tests that you could help me implementing?

Thanks!

c-knowles:
Do you think your node pools branch will change significantly before merge? I ask because I will rebase from there for the subnets work if it's reasonably stable.

For testing, I agree with the strategy you mentioned. We should definitely cover the core use cases in E2E but might be prohibitive to try every single option combination even at this point where there are less than there is going to be. Naming wise I don't mind either way, K8S uses [e2e](https://github.com/kubernetes/kubernetes/tree/master/test). My preference is all tests under the same parent directory, K8S seems to do the same at least for non-unit tests. Minimum for a bug should be it goes in with at least a test to cover the issue. For a feature, I think the minimum should be an integration test expressing the feature.

Happy to help out implementing, feel free to send me issues like if we want to improve coverage in a particular package and I will take a look.

mumoshu:
@c-knowles No, it won't change significantly from what you can see now 👍

mumoshu:
@c-knowles Thanks for various feedbacks.

Anyways I'll keep doing what I can do at each point.
Probably something like `test/integration/cfn_validation_test.go` will be coming next time we iterate toward better testing.

Also, for the time being, we need someone to provide us solid infrastructure to continuously or manually run kube-aws's E2E tests (beware that each case takes 1 or more hours to complete because it involves running the whole k8s conformance test suite 😈 )
It is quite out of budget for me to provide it 😉

c-knowles:
It's merged 🎉 For E2E tests I can't promise much so not sure what we can do there but will have a think of options, be good to have a chat at some stage.

Issue #112, opened 2016-11-30
Feature: Spot Fleet support for worker nodes


mumoshu:
Quite self explanatory but I'd like to add this to kube-aws.

Upstream issue: https://github.com/kubernetes/kubernetes/issues/24472

Initial Implementation in this project: #113
Documentation: https://github.com/coreos/kube-aws/blob/master/Documentation/kubernetes-on-aws-node-pool.md#deploying-a-node-pool-powered-by-spot-fleet

Spot fleet backed worker nodes are supported since v0.9.2-rc.3:
```
# Launch a main cluster
kube-aws init ...
kube-aws render
kube-aws up ...

# Launch a node pool powered by Spot Fleet
kube-aws node-pools init --node-pool-name mypoolname ...
echo -e "worker:\n  spotFleet:\n    targetCapacity: 3\n" >> node-pools/mypoolname/cluster.yaml
kube-aws node-pools render --node-pool-name mypoolname
kube-aws node-pools up --node-pool-name mypoolname --s3-uri ...
```

[An experimental feature to automatically taint nodes with user-provided taints is supported](https://github.com/coreos/kube-aws/commit/698a63d75f796b0828ae5b3345736a116a21face) since v0.9.2-rc.4(not yet released) so we can ensure only pods tolerant to frequent node terminations are scheduled to spot instances/spot-fleet-powered nodes:

----

Utilizing Spot Fleet gives us chances to dramatically reduce cost being spent on EC2 instances powering Kubernetes worker nodes
AWS says cost reduction is up to 90%. I can confirm that in my daily used region `ap-northeast-1` it is up to 89% right now, with slightly varying cost for each instance type.

I believe that on top of the recent work on Node Pools #46, it is easier than ever to implement a POC of the Spot Fleet support.
I'll send a pull request to show it shortly.
I'd appreciate your feedback(s)!

Several concerns I've came up with until now:
* cluster-autoscaler doesn't support Spot Fleets
  * If you want to make nodes in a spot fleet auto-scaled, you probably need to tinker resulting CloudFormation templates to include appropriate configuration. See https://aws.amazon.com/jp/blogs/aws/new-auto-scaling-for-ec2-spot-fleets/ for the official announcement of autoscaling for fleets.
  * Upstream issue: https://github.com/kubernetes/contrib/issues/2066
     * We need to teach cluster-autoscaler how it selects which node pool to expand
[...]

PR #113, opened 2016-11-30
Experimental feature: Spot Fleet support for worker nodes

mumoshu:
refs #112


mumoshu:
@c-knowles FYI I've introduced our first test suite which does integration testing with real AWS services including S3, KMS, CloudFormation as we've discussed before 

https://github.com/coreos/kube-aws/pull/113/commits/f46c711f1e50f01998ad2f6611a0ed0a75444db8#diff-17c3b4ff0a8d67faed426a76a03f8430R1

nodepool/config/config.go
```diff
@@ -148,6 +150,22 @@ func ClusterFromBytes(data []byte) (*ProvidedConfig, error) {
...
 	}
 
+	//Computed defaults
+	launchSpecs := []LaunchSpecification{}
+	for _, spec := range c.Worker.SpotFleet.LaunchSpecifications {
+		if spec.RootVolumeType == "" {
+			spec.RootVolumeType = c.WorkerRootVolumeType
+		}
+		if spec.RootVolumeSize == 0 {
+			spec.RootVolumeSize = c.WorkerRootVolumeSize * spec.WeightedCapacity
```
mumoshu: I'd rather like to make this computed via `c.WorkerSpotFleet.UnitRootVolumeSize * spec.WeightedCapacity` while adding `UnitRootVolumeSize` to the SpotFleet struct, so that we can be more explicit that it is not the exact size of EBS volumes but size per "unit".

mumoshu:
E2E tests passed

mumoshu:
Rebased and then squashed.

mumoshu:
All green. Merging 🚥

PR #114, opened 2016-12-02
Ensure ephemeral storage is prepared before docker starts

tarvip:
This was left out in PR https://github.com/coreos/kube-aws/pull/95
I am using `Wants=` instead of `Requires=` so that it works even for instance types that do not have ephemeral storage.


mumoshu:
@tarvip LGTM 👍 Thanks for your contribution!

Issue #90, opened 2016-11-23
Umbrella documentation issue


pieterlange:
There's a bit of documentation already out there, but some of it is hard to find and a lot of it is outdated. The move to the dedicated repo included a bunch of improvements of which nearly all are mostly documented in github issues.

Proposal for documentation checklist:
- [ ] Cluster lifecycle documentation
   -  relevant: #61
- [x] Compiling instructions (in README, #162)
- [x] Feature overview (in README, #162)
- [x] README.md cleanup? (too much info on one page) #162 
- [ ] List of third party blog posts (with dates/versions)
- [x] Add `ROADMAP.md` #162 
- [x] GOALS in README
  - Done in a way other than initially planned
- [x] Fixes
   - [x] [Documentation on external DNS seems out of date](https://github.com/coreos/kube-aws/issues/110) #162 
- [ ] Currently un-documented pre-requisites
   - [x] [Internet Gateway needs to be added to VPC before cluster can be created](https://github.com/coreos/kube-aws/issues/120) #162 
   - [x] [KubernetesCluster tag is not applied to existing objects](https://github.com/coreos/kube-aws/issues/135) #162 
   - [x] [What's the smallest instance size I can use?](https://github.com/coreos/kube-aws/issues/138) #162 
   - [x] [Creating new cluster in brand new VPC requires "DNS Hostnames" setting to be turned on](https://github.com/coreos/kube-aws/issues/119) #162 
   - [ ] [List of IAM permissions](https://github.com/coreos/kube-aws/issues/88)
- [ ] Custom configuration
   - [ ] [Docker container log limit](https://github.com/coreos/kube-aws/issues/134)
- [ ] FAQ/Gotchas (in `Documentation/kubernetes-on-aws-limitations.md`, #162)
   - [x] [does hostPort not work on kube-aws/CoreOS?](https://github.com/coreos/kube-aws/issues/91)

mumoshu:
Thanks for starting this!

* Regarding the compiling instructions part, I guess we'd better document at least the fact we need to `git clone` the repo under `$GOPATH/src/github.com/coreos` like I've explained in https://github.com/coreos/kube-aws/issues/87#issuecomment-262660786
* Should we document about experimental features, too?
* Anyone likes the idea of something like `GOALS.md` which I've mentioned in https://github.com/coreos/kube-aws/issues/59#issuecomment-261824684?

aholbreich:
[...]

PR #115, opened 2016-12-02
Adding initial GOALS.md

aholbreich:
This adds initial GOALS.md as contribution to #90

### DIFF
--- a/GOALS.md
+++ b/GOALS.md
@@ -12,11 +12,11 @@ Kube-aws CLI's goal is to support different experience levels of the users in te
 * CoreOS administration expertise
 * Amazone Web Services expertise
 
-to support this goal it 
-    * maintanes well-documented set of default settings
-    * but allows to configure every aspect of the cluster
-    * enables reuse of (all) custom predefined AWS resources that are precreated in AWS
-    * provides build in! documentation, that is available online as well
+to support this goal it
+* maintanes well-documented set of default settings
+* but allows to configure every aspect of the cluster
+* enables reuse of (all) custom predefined AWS resources that are precreated in AWS
+* provides build in! documentation, that is available online as well
 
 Further more kube-aws allow me to 
 * manages several clusters in convenient way
--- a/GOALS.md
+++ b/GOALS.md
@@ -1,0 +1,29 @@
+# Why Goals
+Stating goals (and non-goals) are intended to sharpen the join understanidng the the kube-aws product vision. Expressiong the vision by 'Goals' seem to be understandable approach.
+
+#Goals
+Kube-aws is Command Line Tool (CLI) that allow me as experienced console-user to:
+
+* to plan ant prepare my kubernetes cluster alongside need AWS resources
+* browse throgh intermediate details of the configuration (of not yet provisioned clusters)
+
+Kube-aws CLI's goal is to support different experience levels of the users in terms of 
+* kubernetes administration expertise
+* CoreOS administration expertise
+* Amazone Web Services expertise
+
+to support this goal it 
+    * maintanes well-documented set of default settings
+    * but allows to configure every aspect of the cluster
+    * enables reuse of (all) custom predefined AWS resources that are precreated in AWS
+    * provides build in! documentation, that is available online as well
+
+Further more kube-aws allow me to 
+* manages several clusters in convenient way
+* is capable of query statuses of maintaned clusters and e.g. list out maintaned AWS resources
+* provides controlled verbosity, warns and guides users 
+* CLI command structure is self-explainable
+
+#Non Goals
+
+* Not intended to be used with other cloud providers than Amazon Web Services at the moment.
--- a/GOALS.md
+++ b/GOALS.md
@@ -4,26 +4,27 @@ Stating goals (and non-goals) are intended to sharpen the join understanidng the
 #Goals
 Kube-aws is Command Line Tool (CLI) that allow me as experienced console-user to:
 
-* to plan ant prepare my kubernetes cluster alongside need AWS resources
-* browse throgh intermediate details of the configuration (of not yet provisioned clusters)
+* to plan and prepare my kubernetes cluster alongside need AWS resources
+* browse through intermediate details of the configuration (of not yet provisioned clusters)
 
-Kube-aws CLI's goal is to support different experience levels of the users in terms of 
+Kube-aws CLI's aims to support different experience levels of the users in terms of 
 * kubernetes administration expertise
 * CoreOS administration expertise
 * Amazone Web Services expertise
 
 to support this goal it
-* maintanes well-documented set of default settings
+* maintains well-documented set of default settings
 * but allows to configure every aspect of the cluster
-* enables reuse of (all) custom predefined AWS resources that are precreated in AWS
-* provides build in! documentation, that is available online as well
+* enables reusing of the existing AWS resources whenever possible (if and only if those resources conforms to the requirements of kube-aws and/or Kubernetes)
+* provides built-in! documentation, that is available online as well
 
 Further more kube-aws allow me to 
-* manages several clusters in convenient way
-* is capable of query statuses of maintaned clusters and e.g. list out maintaned AWS resources
+* manage several clusters in convenient way
+* is capable of query statuses of maintained clusters and e.g. cab list maintained AWS resources
 * provides controlled verbosity, warns and guides users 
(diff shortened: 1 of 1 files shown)


GOALS.md
```diff
@@ -0,0 +1,29 @@
...
+* to plan ant prepare my kubernetes cluster alongside need AWS resources
+* browse throgh intermediate details of the configuration (of not yet provisioned clusters)
+
+Kube-aws CLI's goal is to support different experience levels of the users in terms of 
+* kubernetes administration expertise
+* CoreOS administration expertise
+* Amazone Web Services expertise
+
+to support this goal it
+* maintanes well-documented set of default settings
```
mumoshu: nit: you meant "maintains"?

GOALS.md
```diff
@@ -0,0 +1,29 @@
...
+* but allows to configure every aspect of the cluster
+* enables reuse of (all) custom predefined AWS resources that are precreated in AWS
+* provides build in! documentation, that is available online as well
```
mumoshu: nit: you meant "built in"?

GOALS.md
```diff
@@ -0,0 +1,29 @@
...
+
+Further more kube-aws allow me to 
+* manages several clusters in convenient way
+* is capable of query statuses of maintaned clusters and e.g. list out maintaned AWS resources
```
mumoshu: nit: maintained

GOALS.md
```diff
@@ -0,0 +1,29 @@
...
+* provides controlled verbosity, warns and guides users 
+* CLI command structure is self-explainable
+
+#Non Goals
+
+* Not intended to be used with other cloud providers than Amazon Web Services at the moment.
```
mumoshu: Would you mind adding `No plan to support host OSes other than CoreOS`?

GOALS.md
```diff
@@ -0,0 +1,29 @@
...
+
+Kube-aws CLI's goal is to support different experience levels of the users in terms of 
+* kubernetes administration expertise
+* CoreOS administration expertise
+* Amazone Web Services expertise
+
+to support this goal it
+* maintanes well-documented set of default settings
+* but allows to configure every aspect of the cluster
+* enables reuse of (all) custom predefined AWS resources that are precreated in AWS
```
mumoshu: Would you mind alter this a bit, like "enables reusing of the existing AWS resources whenever possible, if and only if those resources conforms to the requirements of kube-aws and/or Kubernetes" so that we can ensure that kube-aws would e.g. fail-fast when an user tried to reuse invalid/wrongly configured AWS resources which ends up with a non-functional k8s cluster?

mumoshu:
@aholbreich Thanks for the writeup!
I've seen a few nits but I like it overall 😄 

@pieterlange Would you mind taking a look into this, too?
I believe you also have a vision where we should go 😉

GOALS.md
aholbreich: yes.

GOALS.md
iwarp: Any thoughts on support for Open Stack for use for an on-premise deployment?

mumoshu: @iwarp I understand the need(s) but it should be done in another projects ☕️ 

As kube-aws being very specific/highly locked-in to AWS, implementing OpenStack support in kube-aws wouldn't give any merit(e.g. no code reuse at all)
I'm even thinking of integration with [bootkube](https://github.com/kubernetes-incubator/bootkube) for the cloud-agnostic part of Kubernetes deployment so that kube-aws can focus more on the AWS specific part.

I'd rather recommend you to take a look into OpenStack Magnum and/or coreos-baremetal if not yet.

pieterlange:
I like the overall idea but if we're doing this we should be way more specific. Overall this `GOALS.md` does not add any actionable objectives, which is important actually obtaining these goals.

(if you need an obvious example "but allows to configure every aspect of the cluster" is an explicit goal that is unattainable)

@aholbreich i do not want to stomp on your idea, but how do you feel about us refactoring this into a `ROADMAP.md`? 

Every stage
---------------
  * Provide article walking users through:
    * setting up a cluster from scratch
    * using/enabling new features
    * (breaking changes)
  
Stage0
---------
  * Implement node pools
  * Clean cluster upgrades (preventing downtime, make sure they succeed)

Stage1
---------
  * etcd improvements
    * backups
    * recovery
  * yaml cloudformation templates

StageN a few months from now
---------
  * Bootkube switch
    * `kube-aws` can largely go into maintenance mode when k8s upgrades can be safely achieved on self-hosted clusters.

---
Some other notes:

We should clean up `README.md` so these goals are actually explicit as soon as you look at the project. I see this project as the tool to deploy kubernetes on CoreOS on the Amazon Web Services cloud. I think this project should do everything in it's power to make deployments as straightforward (unsurprising) as possible, but this doesn't excuse the user from doing their kubernetes homework.

We should also factor in this is now a community supported project - it can only do well if its users keep contributing PRs. PRs for documentation are just as good as code.

mumoshu:
@pieterlange Thanks for your feedback!

I guess I basically agree to you both.
How about renaming `GOALS.md` to `VISION.md` and adding what you've wrote up to `ROADMAP.md`?

I'd like to have a "foundation" to shape our future actionable goals/concrete roadmap(which would be written to `ROADMAP.md` like you've done).
Honestly, I'm not very confident in its naming yet. But IMHO something named like `VISION.md` containing what @aholbreich wrote up seems to be a good foundation to set our long term direction.

mumoshu:
Oh, and I definitely agree to clean up README so that it contains a subset of GOALS.md 👍

GOALS.md
```diff
@@ -0,0 +1,30 @@
...
+#Goals
+Kube-aws is Command Line Tool (CLI) that allow me as experienced console-user to:
+
+* to plan and prepare my kubernetes cluster alongside need AWS resources
+* browse through intermediate details of the configuration (of not yet provisioned clusters)
+
+Kube-aws CLI's aims to support different experience levels of the users in terms of 
+* kubernetes administration expertise
+* CoreOS administration expertise
+* Amazone Web Services expertise
```
mumoshu: I've talked with @pieterlange and now I'm a bit afraid to state this kind of things in our documentation.
The point is that, we don't know how our users vary in their levels of expertise!

An extreme case but targeting e.g. users don't know AWS at all, in this stage of kube-aws, would make us all unhappy - possibly endless user support and no time to develop features.

So, in the meantime, how about we rewrite goals targeting users with a certain level of expertise?
Maybe after certain period of time with various fixes, improvements, features which would be written in our roadmap introduced to kube-aws, we can gradually re-target our users?

Btw, I don't mean to make kube-aws targeted to advanced users more than today.
It is already advanced henec our current users are certainly advanced 😉 
I just want to make it extra sure that we don't target to "wrong" users at this stage.

GOALS.md
```diff
@@ -0,0 +1,30 @@
...
+
+to support this goal it
+* maintains well-documented set of default settings
+* but allows to configure every aspect of the cluster
+* enables reusing of the existing AWS resources whenever possible (if and only if those resources conforms to the requirements of kube-aws and/or Kubernetes)
+* provides built-in! documentation, that is available online as well
```
mumoshu: @pieterlange Quick thought but should we write these in CONTRIBUTION.md, after rewriting a bit to match the taste of CONTRIBUTION.md?

GOALS.md
```diff
@@ -0,0 +1,30 @@
...
+
+Further more kube-aws allow me to 
+* manage several clusters in convenient way
+* is capable of query statuses of maintained clusters and e.g. cab list maintained AWS resources
+* provides controlled verbosity, warns and guides users 
+* CLI command structure is self-explainable
```
mumoshu: @pieterlange Should these be addressed in `ROADMAP.md`?

mumoshu:
@aholbreich Added several comments. Anyways, I want to be clear that your write-up is very precious because it definitely brought up such a discussion regarding the fundamentals of kube-aws :bow:

GOALS.md
iwarp: Thanks for you comments, I had been wondering how portable the cloudformation is to heat on openstack.

mumoshu:
Sorry for leaving this open too long :bow:
I'm closing this for now as I had updated various documentations and README to include essences since then.
This has not been merged as-is but did help us draw our direction.
Thanks for your contribution @aholbreich!

PR #127, opened 2016-12-06
Allow to attach user defined security groups to worker nodes

### LINKED ISSUE #43: Proposal: enable to attach user defined security groups for worker instances.
Hi community,  thanks for sharing the great project.  I'm very happy if I could contribute this project would be better.

# Motivation 
I think restricting access to ec2 instances by security group is a common pattern.  For example, assume the situation that you  had some database instances and you wanted to restrict access to those db instances by some specific security group.  

# Proposal
In such situations, I believe that `kube-aws` could be more useful if `kube-aws` can attach user defined security groups to its worker nodes.

# Current possible workaround and its disadvantage
We can use the security group which `kube-aws` will generate for worker nodes, let this be `sg-w`.  For above example, that means, we can configure the security group which will be attached to db instances, let this be `sg-db`, so that `sg-db` allows ingress from `sg-w`.

Unfortunately,  in this case, this doesn't allow user to destroy the k8s cluster unless user modified `sg-db` because`sg-w` is referred from `sg-db` and `sg-w` is managed by `kube-aws`.   I believe that this is not preferred by many users because this solution introduces undesirable dependencies.

mumoshu:
We can now specify which existing security groups to be attached to worker nodes created by kube-aws.

`workerSecurityGroupIds` configuration key was added to allow that.

In cluster.yaml, it would look like:

```
workerSecurityGroupIds:
  - sg-87654321
  - sg-98765432
```

Worker nodes in either main clusters and node pools are supported.

fix #43 

TODOs:

- [x] Add unit test cases
- [x] Add integration test cases

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,12 +79,13 @@ func NewDefaultCluster() *Cluster {
 			DNSServiceIP: "10.3.0.10",
 		},
 		WorkerSettings: WorkerSettings{
-			WorkerCount:          1,
-			WorkerCreateTimeout:  "PT15M",
-			WorkerInstanceType:   "m3.medium",
-			WorkerRootVolumeType: "gp2",
-			WorkerRootVolumeIOPS: 0,
-			WorkerRootVolumeSize: 30,
+			WorkerCount:            1,
+			WorkerCreateTimeout:    "PT15M",
+			WorkerInstanceType:     "m3.medium",
+			WorkerRootVolumeType:   "gp2",
+			WorkerRootVolumeIOPS:   0,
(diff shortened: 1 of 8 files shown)


mumoshu:
I'm going to add unit tests for this after #126 is merged because of conflicts.

mumoshu:
We really should be aware of the max number of SGs attached to an EC2 instance is only 5.
I've not yet implemented the validation for that but I'm going to.

mumoshu:
Addressed all the TODOs

PR #129, opened 2016-12-06
Name the node pool subnets after the node pool

c-knowles:
Rather than the cluster name


mumoshu:
@c-knowles Good catch! Thanks 👍

PR #133, opened 2016-12-06
Bumped dashboard to v1.5.0

danielfm:


mumoshu:
LGTM 👍 Thanks for your contribution!

FYI: If you're deploying PetSets, dashboard v1.5.0 won't show them after merging this because PetSets are renamed to StatefulSets in kubernetes v1.5 and dashboard v1.5.0 assumes that.

https://github.com/kubernetes/dashboard/releases

If you'd like to show PetSets in dashboard, I suggest manually modifying cloud-config-controller to ref dashboard v1.4.x.

PR #139, opened 2016-12-08
Add `kube-aws node-pools destroy` command

mumoshu:
which had been missing since the first implementation of Node Pools


mumoshu:
This is already tested manually while running E2E tests

PR #141, opened 2016-12-08
Fix node registration issue for spot-fleet-powered worker nodes

mumoshu:
This change fixes the issue that the second and the subsequent nodes in each spot fleet couldn't be registered thus unable to run any pods.

ref https://github.com/coreos/kube-aws/issues/112#issuecomment-265643237

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -211,6 +211,22 @@ coreos:
               --stack {{.ClusterName}}
 {{end}}
 
+{{if .Worker.SpotFleet.Enabled}}
+    - name: tag-spot-instance.service
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Tag this spot instance with cluster name
+        After=network-online.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/opt/bin/tag-spot-instance
+{{end}}
+
 {{ if $.ElasticFileSystemID }}
     - name: rpc-statd.service
       command: start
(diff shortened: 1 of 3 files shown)


mumoshu:
I've confirmed that node registration just work with the following process:

* Create a main cluster
* Create 2 node pools, each with targetCapacity=3 thus 2 nodes
  * only 1 of 2 nodes is registered
* Apply the change from this PR and create a node pool with targetCapacity=3 thus 2 nodes
  * 2 of 2 nodes is registered, as expected

```
NAME                                              STATUS                     AGE
ip-10-0-0-164.ap-northeast-1.compute.internal     Ready                      5h
ip-10-0-0-197.ap-northeast-1.compute.internal     Ready,SchedulingDisabled   7h
ip-10-0-0-202.ap-northeast-1.compute.internal     Ready                      7h
ip-10-0-101-46.ap-northeast-1.compute.internal    Ready                      1h <---from node pool 1 (without fix)
ip-10-0-102-64.ap-northeast-1.compute.internal    Ready                      1h <---from node pool 2 (without fix)
ip-10-0-103-197.ap-northeast-1.compute.internal   Ready                      1h <---from node pool 3 (with fix)
ip-10-0-103-232.ap-northeast-1.compute.internal   Ready                      1h <---from node pool 3 (with fix)
```

config/templates/cloud-config-worker
```diff
@@ -437,6 +453,36 @@ write_files:
...
+      instance_id=$(curl http://169.254.169.254/latest/meta-data/instance-id)
+
+      sudo rkt run \
+        --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
+        --mount=volume=ssl,target=/etc/kubernetes/ssl \
+        --uuid-file-save=/var/run/coreos/tag-spot-instance.uuid \
+        --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+        --net=host \
+        --trust-keys-from-https \
+        --insecure-options=ondisk \
```
mumoshu: `--insecure-options=ondisk` to skip verification on docker images on the disk.
The verification is IMHO not needed for us as we don't verify when running images using docker.
Skipping verification results in faster container startup.

PR #142, opened 2016-12-08
Allow more control over ASG definition

c-knowles:
Min/max size and min instances in service can now be directly defined.
Defaults taken from the older `controllerCount` and `workerCount`


config/config.go
```diff
@@ -399,8 +394,33 @@ func (c KubeClusterSettings) K8sNetworkPlugin() string {
...
+	}
+	if config.WorkerASG.MinSize == 0 {
+		config.WorkerASG.MinSize = config.WorkerCount
+	}
+	if config.WorkerASG.MaxSize == 0 {
+		config.WorkerASG.MaxSize = util.IntMax(config.WorkerCount, config.WorkerASG.MinSize)
+	}
+	if config.WorkerASG.RollingUpdateMinInstancesInService == 0 {
+		config.WorkerASG.RollingUpdateMinInstancesInService = config.WorkerASG.MaxSize - 1
+	}
```
mumoshu: If we go this way, I guess we also need to set defaults for WorkerASG in `nodepool/config/config.go` or we'll end up with all of MinSize, MaxSize, RollingUPdateMinInstancesInService to be kept as is - 0 for node pools, hence stack creation error.

config/config.go
```diff
@@ -399,8 +394,33 @@ func (c KubeClusterSettings) K8sNetworkPlugin() string {
...
+		config.ControllerASG.RollingUpdateMinInstancesInService = config.ControllerASG.MaxSize - 1
+	}
+
+	// setup worker ASG size, one is the default WorkerCount
+	if config.WorkerCount != 1 && (config.WorkerASG.MinSize != 0 || config.WorkerASG.MaxSize != 0) {
+		return nil, fmt.Errorf("`workerASG.MinSize` and `workerASG.MaxSize` can only be specified without `workerCount`")
+	}
+	if config.WorkerASG.MinSize == 0 {
+		config.WorkerASG.MinSize = config.WorkerCount
+	}
```
mumoshu: Could we just alter `func (c WorkerSettings) MinWorkerCount() int` to achieve the same end result?

e.g.

```
func (c WorkerSettings) MinWorkerCount() int {
	if c.WorkerASG.MinSize == 0 {
		return c.WorkerCount
	}
	return c.WorkerASG.MinSize
	// Instead of:
	// return c.WorkerCount - 1
}
```

So that we don't even need to modify `stack-template.json` and could keep backward-compatibility of what are (exposed to|usable from) stack-template.json

config/config.go
```diff
@@ -399,8 +394,33 @@ func (c KubeClusterSettings) K8sNetworkPlugin() string {
...
+	if config.WorkerASG.MaxSize == 0 {
+		config.WorkerASG.MaxSize = util.IntMax(config.WorkerCount, config.WorkerASG.MinSize)
+	}
```
mumoshu: The same as https://github.com/coreos/kube-aws/pull/142/files#r91637776 for `func (c WorkerSettings) MaxWorkerCount() int`

config/config.go
```diff
@@ -399,8 +394,33 @@ func (c KubeClusterSettings) K8sNetworkPlugin() string {
...
+
+	// setup worker ASG size, one is the default WorkerCount
+	if config.WorkerCount != 1 && (config.WorkerASG.MinSize != 0 || config.WorkerASG.MaxSize != 0) {
+		return nil, fmt.Errorf("`workerASG.MinSize` and `workerASG.MaxSize` can only be specified without `workerCount`")
+	}
+	if config.WorkerASG.MinSize == 0 {
+		config.WorkerASG.MinSize = config.WorkerCount
+	}
+	if config.WorkerASG.MaxSize == 0 {
+		config.WorkerASG.MaxSize = util.IntMax(config.WorkerCount, config.WorkerASG.MinSize)
```
mumoshu: Maybe we'd add a validation like this to `func (c WorkerSettings) Valid() error` rather than silently fixing invalid configuration (i.e. MinSize > WorkerCount).

config/templates/stack-template.json
```diff
@@ -17,19 +17,19 @@
           "{{$subnet.AvailabilityZone}}"
           {{end}}
         ],
-        "DesiredCapacity": "{{.WorkerCount}}",
+        "DesiredCapacity": "{{.WorkerASG.MinSize}}",
```
mumoshu: Probably you've not intended this?
I guess we should leave this `{{.WorkerCount}}` or we become unable set desired capacity

config/templates/stack-template.json
```diff
@@ -99,19 +99,19 @@
           "{{$subnet.AvailabilityZone}}"
           {{end}}
         ],
-        "DesiredCapacity": "{{.ControllerCount}}",
+        "DesiredCapacity": "{{.ControllerASG.MinSize}}",
```
mumoshu: The same as https://github.com/coreos/kube-aws/pull/142/files#r91640304

config/config.go
mumoshu: Also I believe we can do this more immutably/DRY by wrapping RollingUpdateMinInstancesInService field into a func named the same.
Wrote a snippet to show my thoughts: https://play.golang.org/p/c_0Pu_UDMf

nodepool/config/templates/cluster.yaml
```diff
@@ -46,9 +46,15 @@ availabilityZone: {{.AvailabilityZone}}
...
 
-# Number of worker nodes to create
+# Number of worker nodes to create, for more control use `workerASG` and do not use this setting
 #workerCount: 1
 
+# Auto Scaling Group definition for workers. If only `workerCount` is specified, min and max will be the set to that value and `rollingUpdateMinInstancesInService` will be one less.
+# workerASG:
+#   minSize: 1
+#   maxSize: 3
+#   rollingUpdateMinInstancesInService: 2
```
mumoshu: Could we adapt to the new `worker:` key [I've recently added for spot fleets](https://github.com/coreos/kube-aws/blob/46330082520d8307e7062b9e86f34c929008de2b/nodepool/config/templates/cluster.yaml#L79) so that we can have more consistency/distinction between 2 deployment strategy: auto scaling group and spot fleet?

```yaml
#worker:
#  # Auto Scaling Group definition for workers. If only `workerCount` is specified, min and max will be the set to that value and `rollingUpdateMinInstancesInService` will be one less.
#  autoScalingGroup:
#    minSize: 1
#    maxSize: 3
#      rollingUpdateMinInstancesInservice: 2
#  # Spot Fleet definitions for workers
#  # You may specify autoScalingGroup or spotFleet, but not both.
#  # Spot Fleet support is still experimental. It can be changed in backward-incompatible ways
#  spotFleet:
#  *snip*
```

mumoshu:
@c-knowles Thanks for the PR!

Its functionality seems good 👍 however some comments on backward-compatibility, consistency and DRY.
Sorry for not clearly documenting these points anywhere but I believe this is also a good chance for us to share kube-aws' design.
Probably too many comments on a rather simple enhancement functionality-wise but could you take some time to address these?

.gitignore
```diff
@@ -9,3 +9,6 @@
 coverage.txt
 profile.out
 /.glide
+
+# files generated during tests
+config/temp
```
c-knowles: I don't mind removing but one thing is if viewing git diff during a test run, it shows many uncommitted files before this. I think that could be a good reason to leave it in. I'm using some Atom extensions that mean even just saving will run the tests.

config/config.go
c-knowles: As yes you are right, I had thought nodepools config was inherited from the main one. I want to share it somehow as you say so will look into it.

config/config.go
```diff
@@ -399,8 +394,33 @@ func (c KubeClusterSettings) K8sNetworkPlugin() string {
...
+		config.ControllerASG.MaxSize = util.IntMax(config.ControllerCount, config.ControllerASG.MinSize)
+	}
+	if config.ControllerASG.RollingUpdateMinInstancesInService == 0 {
+		config.ControllerASG.RollingUpdateMinInstancesInService = config.ControllerASG.MaxSize - 1
+	}
+
+	// setup worker ASG size, one is the default WorkerCount
+	if config.WorkerCount != 1 && (config.WorkerASG.MinSize != 0 || config.WorkerASG.MaxSize != 0) {
+		return nil, fmt.Errorf("`workerASG.MinSize` and `workerASG.MaxSize` can only be specified without `workerCount`")
+	}
```
c-knowles: Oh I see. I missed that, I'm sure as you are aware it is a little hard to know about the style with old+new code together and such a big file! I will fix it.

config/config.go
c-knowles: I would avoid keeping `MinWorkerCount` for two reasons:
1. I think it's better if the layout of the names in config `cluster.yaml` match those `stack-template.json`
2. Regarding backwards compatibility, `MinWorkerCount` was only introduced in the new kube-aws repo so I think the impact is minimal for anyone with major modifications as they would have to do similar to I have done (switch over and change quite a lot). I don't think we should maintain too many options/functions that do the same thing.

What are you thoughts assuming I do the function/DRY work above?

(Side note, this relates to me wanting to remove the old `workerCount` too as it's trivial for the user to update to min/max compared to keeping all so many different options available)

config/config.go
c-knowles: Can do. What do you want the expected behaviour to be for the setup in `TestASGsMinConfigured`?

config/templates/stack-template.json
c-knowles: My assumption is desired capacity is usually min size for cluster rollout? Afterwards AWS changes the desired capacity when we implement auto scaling. We could allow this to be customised as well? I can't remember the AWS behaviour if desired capacity is outside min/max etc. Right now I just replaced like for like so we could leave the extra config for later, i.e. it was the min count before and it is still the min count.

nodepool/config/templates/cluster.yaml
c-knowles: Yes we can! I saw that and was wondering if it was the way you wanted to go, I can update it. One thing will be we have old settings outside `worker` and new settings inside. My opinion is we need to deprecate some things in order to ensure a consistent and easy to understand tool.

c-knowles:
@mumoshu great, thanks. I'll start on a few of those changes and I put some follow up comments above.

nodepool/config/templates/cluster.yaml
c-knowles: Do you also want the controller ASG in a `controller` YAML property?

.gitignore
cw-kuoka: Rebase to the latest master and `rm -Rf config/temp nodepool/config/temp` and these uncommitted files won't appear again 💮

config/config.go
cw-kuoka: Only static, default values set in `NewDefaultCluster()` are inherited to node pool config.
IMHO currently there's no timing to inherit "computed" values like these.
That is one of reasons I believe it is better to implement "computed" values by wrapping them into functions, instead of mutating values in place. ref https://github.com/coreos/kube-aws/pull/142/files#r91641540

config/templates/stack-template.json
cw-kuoka: Yes. However if that's your intention it should be achieved in an another way.

[`DesiredCapacity` is not a required parameter](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-desiredcapacity).
If you want to leave your initial ASG size as same as MinSize, you could just omit `"DesiredCapacity": "{{.WorkerASG.MinSize}}",` so stack-template would look like:

```json
{{if gt .WorkerCount 0}}"DesiredCapacity": "{{.WorkerCount}}",{{end}}
```

With this, If workerCount is omitted from cluster.yaml, `DesiredCapacity` is automatically set to MinSize by ASG itself and we're even unable to update DesiredCapacity via cfn stack updates. Now, the user is responsible to update DesiredCapacity outside of kube-aws or CloudFormation.

.gitignore
mumoshu: Rebase to the latest master and `rm -Rf config/temp nodepool/config/temp` and these uncommitted files won't appear again 💮 
If it appears, it is a bug. I'd rather like to notice bugs instead of ignoring them.

config/templates/stack-template.json
mumoshu: Yes. However if that's your intention it should be achieved in an another way.

[`DesiredCapacity` is not a required parameter](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-desiredcapacity).
If you want to leave your initial ASG size as same as MinSize, you could just omit `"DesiredCapacity": "{{.WorkerASG.MinSize}}",` so stack-template would look like:

```json
{{if gt .WorkerCount 0}}"DesiredCapacity": "{{.WorkerCount}}",{{end}}
```

With this, If workerCount is omitted from cluster.yaml, `DesiredCapacity` is automatically set to MinSize by ASG itself and we're even unable to update DesiredCapacity via cfn stack updates. Thus, now, the user is responsible to update DesiredCapacity outside of kube-aws or CloudFormation.

config/config.go
mumoshu: Only static, default values set in `NewDefaultCluster()` are inherited to node pool config.
IMHO currently there's no timing to inherit "computed" values like these.
That is one of reasons I believe it is better to implement "computed" values by wrapping them into functions, instead of mutating values in place. ref https://github.com/coreos/kube-aws/pull/142/files#r91641540

mumoshu:
I've mistakenly commented via my another github account several times 😅 Never mind about that.

config/templates/stack-template.json
c-knowles: The intention of this change is that the new `WorkerASG.MinSize` has the same value as the older `WorkerCount`. So if we want to omit it for that case, I would just omit it altogether and that would clarify that we currently do not support configuration of `DesiredCapacity`. What do you think?

It depends whether you want the `DesiredCapacity` to be set to the exact value it is prior to this change which it is `MinSize` is always set, defaulting to `WorkerCount`. In turn, this also keeps `stack-template.json` clean as it only uses the ASG properties rather than having multiple levels of defaulting/switching inside code and templates.

config/config.go
mumoshu: A validation error when MinSize > WorkerCount would be nice

nodepool/config/templates/cluster.yaml
mumoshu: Yes and yes! I'd like to have a new `controller` property and both `worker` and `controller` to be consistently used in both main cluster's and node pools' cluster.yaml

mumoshu:
@c-knowles I've replied to every comment 👍

config/templates/stack-template.json
mumoshu: Quick thought but do you prefer entirely removing `DesiredCapacity` from stack-template.json and later improve `kube-aws update` command to support updating DesiredCapacity e.g. `kube-aws update --num-workers 10`?

c-knowles: Yes, I think so. Since it's currently set to the same size as min and AWS will default to that anyway, it's not currently doing much.

mumoshu: OK so let's remove `DesiredCapacity` then 👍

Btw, I've noticed that omitting `DesiredCapacity` here would make the situation easier for #148 too.

.gitignore
c-knowles: Not sure if anything was changed on latest master that would help this as master hasn't changed since this comment? Anyway, it still happens off latest - my local save runs the tests and if I happen to go to git diff while that is running it shows many files in `config/temp`. Not a big problem, but it is still there.

config/config.go
c-knowles: As part of this, I will need to update `ClusterFromBytes` because currently it sets up `InstanceCIDR` if it's empty with zero subnets but then after adds subnets which is then invalid! i.e. calling `ClusterFromBytes` with basic config and then `.valid()` does not work.

nodepool/config/templates/cluster.yaml
c-knowles: I'm not entire sure how I can get this in without moving `worker.go` since otherwise there will be a duplicate `Worker` inside `Cluster`. Are you ok for me to move it? I saw there was a bunch of placeholder code anyway in `config.go` so it's pretty messy as it is.

c-knowles:
@mumoshu I believe I've done most of what you requested above, I've started to align worker/controller as well as I was forced to combine those to remove duplicate definitions inside config/nodepools. I'd like to move more of the parsing and validation of the model into that package but I kept it to a minimum for now.

config/templates/stack-template.json
```diff
@@ -65,7 +64,7 @@
       {{if .Experimental.WaitSignal.Enabled}}
       "CreationPolicy" : {
         "ResourceSignal" : {
-          "Count" : "{{.WorkerCount}}",
+          "Count" : "{{.MinWorkerCount}}",
```
mumoshu: In `kube-aws update`, this should match the desired capacity at that time rather than min size, or I guess we will end up destroying the former autoscaling group too early the overall cluster capacity temporarily decreased until  the rest of worker nodes(`DesiredCapacity - MinSize`) becomes ready.

So, do you mind that I'll follow up with a pr addressing this after merging?

.gitignore
mumoshu: Thanks for the info!
That's odd 😭
Would you mind reverting this and then raising an another issue specif for that?

config/templates/stack-template.json
c-knowles: Sure, if that's appropriate. Since this is the `CreationPolicy` I thought it didn't apply to stack updates at all?

.gitignore
c-knowles: https://github.com/coreos/kube-aws/issues/156

mumoshu:
@c-knowles Thanks for your continuous efforts :bow:
Before merging this, could you remove `util/math.go` because it is not referenced from anywhere now?

c-knowles:
Done. By the way, I think this resolves https://github.com/coreos/kube-aws/pull/158 as well.

iameli:
Oh nice, that seems true. Thanks!

mumoshu:
@c-knowles Merged. Thanks for the PR and a lot of follow ups 👍

PR #146, opened 2016-12-10
Add configurable instance tenancy

iameli:
Return of [my PR on the coreos-kubernetes repo](https://github.com/coreos/coreos-kubernetes/pull/252), this time with node-pool support added.

Let me know if you'd like tests or anything. The only thing slightly tricky about dedicated tenancy is that it's mutually exclusive with spot instances, I could potentially add a check for that.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -89,6 +89,7 @@ func NewDefaultCluster() *Cluster {
 			WorkerRootVolumeIOPS:   0,
 			WorkerRootVolumeSize:   30,
 			WorkerSecurityGroupIds: []string{},
+			WorkerTenancy:          "default",
 		},
 		ControllerSettings: ControllerSettings{
 			ControllerCount:          1,
@@ -97,6 +98,7 @@ func NewDefaultCluster() *Cluster {
 			ControllerRootVolumeType: "gp2",
 			ControllerRootVolumeIOPS: 0,
 			ControllerRootVolumeSize: 30,
+			ControllerTenancy:        "default",
 		},
 		EtcdSettings: EtcdSettings{
 			EtcdCount:          1,
@@ -107,6 +109,7 @@ func NewDefaultCluster() *Cluster {
 			EtcdDataVolumeSize: 30,
 			EtcdDataVolumeType: "gp2",
 			EtcdDataVolumeIOPS: 0,
+			EtcdTenancy:        "default",
 		},
 		FlannelSettings: FlannelSettings{
 			PodCIDR: "10.2.0.0/16",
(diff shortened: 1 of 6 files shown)


config/templates/cluster.yaml
```diff
@@ -68,6 +68,10 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Number of I/O operations per second (IOPS) that the controller node disk supports. Leave blank if controllerRootVolumeType is not io1
 #controllerRootVolumeIOPS: 0
 
+# Tenancy of the controller nodes. Options are "default" and "dedicated"
+# Documentation: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/dedicated-instance.html
```
mumoshu: I appreciate your documentation here. Linking to the original source of information is always good idea 👍

mumoshu:
Thanks for the PR!

Regarding tests, I'd appreciate it if you could add a test case to https://github.com/csats/kube-aws/blob/42efc6222121df1a4cc81a2c4f2002299ad77dc6/config/config_test.go#L1234, something like:

```go
configYaml: minimalValidConfigYaml + `
workerTenancy: dedicated
controllerTenancy: dedicated
etcdTenancy: dedicated
`
```

so that we can ensure this feature won't break in the future.

mumoshu:
> The only thing slightly tricky about dedicated tenancy is that it's mutually exclusive with spot instances, I could potentially add a check for that.

Thanks for the info!
Then, I'd like `func (c WorkerSettings) Valid() error` to return an error when:

* `if c.Worker.SpotFleet.Enabled() && c.WorkerInstanceTenancy != "default"`
* Or `if c.WorkerSpotPrice != "" && c.WorkerInstanceTenancy != "default"`

iameli:
Tests and configuration validation added. https://github.com/coreos/kube-aws/pull/146/commits/d93ea0ec1f2f3b0dc4abca551ca935a4c74d4554

config/config.go
```diff
@@ -731,6 +737,14 @@ func (c Cluster) valid() error {
 		return err
 	}
 
+	if c.WorkerTenancy != "default" && c.Worker.SpotFleet.Enabled {
```
mumoshu: To follow the recent change in master, `c.Worker.SpotFleet.Enabled` should now be `c.Worker.SpotFleet.Enabled()`

mumoshu:
@iameli I've added the a comment in https://github.com/coreos/kube-aws/pull/146/files#r92318290

Would you mind rebasing on the latest master and address the comment?
Then, I will merge this shortly.

iameli:
@mumoshu Done, rebased, and squashed.

mumoshu:
@iameli LGTM 👍 Thanks for your contribution :bow:

PR #149, opened 2016-12-10
Experimental feature: User-specified node labels for worker nodes

mumoshu:
This complements Node Pools(#46) and Spot Fleet support(#112)

The former `experimental.nodeLabel` configuration key is renamed to `experimental.awsNodeLabels` to avoid collision with newly added `experimental.nodeLabels` and consistency with `experimental.awsEnvironment`.

cc @pieterlange @c-knowles

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -37,6 +37,9 @@ func NewDefaultCluster() *Cluster {
 		AwsEnvironment{
 			Enabled: false,
 		},
+		AwsNodeLabels{
+			Enabled: false,
+		},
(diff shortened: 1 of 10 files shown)


mumoshu:
@pieterlange I'm willing to rename the former `NodeLabel` to `AwsNodeLabels` as I've explained in the description of this PR. Any thoughts on the move and/or naming?

The former `NodeLabel` is added in v0.9.2-rc.1. I'd rather like to merge this before we cut 0.9.2 final so it won't be a breaking change.

PR #153, opened 2016-12-12
Option to pretty print CloudFormation JSON

c-knowles:
Means stack contents are easier to compare out of the box

### DIFF
--- a/cmd/nodepool/update.go
+++ b/cmd/nodepool/update.go
@@ -18,14 +18,15 @@ var (
 	}
 
 	updateOpts = struct {
-		awsDebug bool
-		s3URI    string
+		awsDebug, prettyPrint bool
+		s3URI                 string
 	}{}
 )
 
 func init() {
 	NodePoolCmd.AddCommand(cmdUpdate)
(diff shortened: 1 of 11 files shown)


cmd/nodepool/update.go
```diff
@@ -18,14 +18,15 @@ var (
...
-		s3URI    string
+		awsDebug, prettyPrint bool
+		s3URI                 string
 	}{}
 )
 
 func init() {
 	NodePoolCmd.AddCommand(cmdUpdate)
 	cmdUpdate.Flags().BoolVar(&updateOpts.awsDebug, "aws-debug", false, "Log debug information from aws-sdk-go library")
+	cmdUpdate.Flags().BoolVar(&updateOpts.prettyPrint, "prettyPrint", false, "Pretty print the resulting CloudFormation")
```
mumoshu: @c-knowles I like this more if it were `pretty-print` instead of `prettyPrint` for consistency with other flags like `aws-debug` 😊

c-knowles: Done

mumoshu:
@c-knowles LGTM. Thanks as always 👍

PR #155, opened 2016-12-13
format ephemeral storage only on first boot

dimpavloff:
That way rebooting the agents does not cause a re-download of images
it's likely to have had and now need again nor losing any Pod logs

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -311,6 +311,7 @@ coreos:
       content: |
         [Unit]
         Description=Formats the ephemeral drive
+        ConditionFirstBoot=yes
         After=dev-{{.Experimental.EphemeralImageStorage.Disk}}.device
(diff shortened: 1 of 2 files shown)


mumoshu:
@dimpavloff LGTM. Thanks for contributing!

tarvip:
It seems to me that this is not working (at least with `v0.9.4-rc.2`):

```
● format-ephemeral.service - Formats the ephemeral drive
   Loaded: loaded (/etc/systemd/system/format-ephemeral.service; static; vendor preset: disabled)
   Active: inactive (dead)
Condition: start condition failed at Tue 2017-02-07 09:48:05 UTC; 1min 2s ago
           └─ ConditionFirstBoot=yes was not met
```
This was fresh host. Maybe someone can verify this.

redbaron:
do you have more than one boot in `journalctl --list-boots` output?

tarvip:
I don't have that particular host available any more, but I checked other hosts and they all have one entry.

PR #158, opened 2016-12-14
config: make sure minimum servers is always >= 0

iameli:
I was setting up a new cluster and intended to create all of my worker nodes using node pools, so I  had a `workerCount: 0` line in there. This all works, except it tries to set the ASG's minSize to -1, and CloudFormation complains. This fixes that.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -379,6 +379,9 @@ var supportedReleaseChannels = map[string]bool{
 }
 
 func (c WorkerSettings) MinWorkerCount() int {
+	if c.WorkerCount == 0 {
+		return 0
+	}
 	return c.WorkerCount - 1
 }
 


iameli:
Closed in favor of #142.

mumoshu:
@iameli Thanks for trying kube-aws 👍 
Could I ask more details to follow your use-case?
Which deployment option you've chosen for your node pool, ASG or Spot Fleet?

If you've chosen ASG, workerCount is still required in order to instruct ASG to set its desired capacity.
```yaml
workerCount: 1
```

If you've chosen Spot Fleet, workerCount is not required hence you can keep `workerCount` commented out like `#workerCount: 1` while adding something to define Spot Fleet like:

```yaml
#workerCount: 1

worker:
  spotFleet:
    targetCapacity: 10
    # *snip*
```

Also, regardless of how kube-aws works as of now, Node Pool is still an experimental feature hence there may be a lot of space for improvements and there may even be some friction for users in it.
Please don't hesitate to submit issues filled with your experiences, inconvenience and/or suggestions if necessary!

mumoshu:
Anyways, this and @c-knowles's fix do resolve the problem arises when you try to set workerCount to zero. Thanks for the fix to both of you!

I'm just wanting to know the exact reason why you've tried to do so in the first place so that
I can reveal some hidden variables behind which may contribute to improvements of kube-aws hence better experience for you, in the future.

iameli:
I have one cluster, `test-cluster`, with two node pools, `test-cluster-mem` and `test-cluster-cpu`. As their name implies, one of them has high-memory instances and one has high-CPU instances.

Each node pool has `workerCount: 3` -- each node pool has three workers. But the main `cluster.yaml` file has `workerCount: 0`, so the main cluster itself doesn't have any workers associated with it. I like this because I can add and remove additional node pools and completely change my instance type without changing the settings of the cluster itself.

mumoshu:
@iameli Understood!

That's definitely a valid use-case which I had imagined/intended but not tested myself yet.
Thanks for revealing the clear path to achieve the use-case 👍

c-knowles:
Great, that's definitely the same outcome (worker count of zero in the main cluster def). The decouple of the node pools makes it quite a bit more flexible.

Issue #138, opened 2016-12-07
What's the smallest instance size I can use?


cisnky:
With 0.8.3 I could use t2.micro for the controller and the worker.

On 0.9.1 after setting up the cluster, I can't connect if using t2.micro.

cisnky:
Worked my way up till it worked with t2.medium.

pieterlange:
t2.micro (with 1GB memory) is pushing it for the controller server. On my cluster(s) the kube components peak above 1GB. I would also advise against using the T2 instance type for the controller if you're running anything serious.

Also note that `kube-aws` comes with a bunch of components by default: heapster and dashboard use some memory and CPU as well.

mandrean:
From my company's real-world experiences, I'd say bare minimum for production is 1x m3.medium for both controller and etcd. But preferably at least 3x m3.medium for HA.

Skip the t2s for controller/etcd in production. You're screwed when you run out of CPU credits. Only use them for worker nodes.

jelis:
I got stuff working with on nanos with 0.8.2 .. I really wish smaller instances worked!

mumoshu:
Thanks for bringing up and info everyone!
I'm closing this as it is now documented in #162.

scribnar:
@mandrean What you mean you are screwed if CPU credits expire? The baseline performance is still available so the controller will continue to run albeit with slow performance. I understand thats not a recommended scenario however you comment sounds like that is very dangerous scenario. I may be missing something here. Can you please explain "you are screwed"? Thanks.

Issue #135, opened 2016-12-06
KubernetesCluster tag is not applied to existing objects


flah00:
I opted to use an existing VPC and existing subnets, because I wanted to stick with my existing public and private subnets. Unfortunately, none of the subnets or routing tables were tagged with the `KubernetesCluster` key and value.  As best as I can tell, this prevented me from creating LoadBalancer services in my public subnet. After I tagged the configured subnets, I was able to create LoadBalancer services, with public IP addresses.

`kube-aws`, it seems, should be able to tag existing resources, to avoid headaches like this.

mumoshu:
Hi @flah00,

Thanks for sharing your experience and findings!
I'd definitely like to document the fact you need to manually tag existing AWS resources if you opted to use them like described in [the upstream doc](https://github.com/kubernetes/kubernetes/blob/master/docs/design/aws_under_the_hood.md#tagging).

However, I don't (currently) think that it is a good idea to make `kube-aws` tag them for you.
The main reason is that we can't reliably clean up those tags even if they aren't needed anymore, because we can't let CloudFormation to "tag existing resources".
Tagging existing resources outside of CloudFormation would easily end up with unnecessary tags remained. An example is that you'd initially created a kubernetes cluster via kube-aws and then destroyed it via CloudFormation console. kube-aws could tag existing resources at creation but there's no chance for kube-aws to clean up these tags in such case. Additional headaches.

Does this make sense to you?
Anyways I'd like to document it so that we can better support use-cases like yours.

Oh, and I'd definitely recommend not to reuse subnets but VPCs and route tables.
That's why we'd been not supporting reusing of existing subnets as of today.

Related historical issues i.e. recommended readings:

[Ability to use existing VPC subnets by c-knowles · Pull Request #671 · coreos/coreos-kubernetes](https://github.com/coreos/coreos-kubernetes/pull/671)
[Ability to use existing route tables for controller and workers by c-knowles · Pull Request #716 · coreos/coreos-kubernetes](https://github.com/coreos/coreos-kubernetes/pull/716)
[Allow workers to sit inside private subnets within an existing VPC · Issue #44 · coreos/kube-aws](https://github.com/coreos/kube-aws/issues/44)

flah00:
@mumoshu That definitely makes sense. As for the reuse of existing subnets, I misspoke. We are only reusing the vpc and route table.

BTW, thanks for this utility, it's been a lifesaver.

mumoshu:
[...]

Issue #120, opened 2016-12-05
Internet Gateway needs to be added to VPC before cluster can be created.


jessesanford:
Again I believe that there is a need for mentioning this in the documentation.

mumoshu:
I understand that some users would suffer with it.
I'd appreciate a pull request to include it in documentation 👍

mumoshu:
I'm closing this as it is now documented in #162 :bow:

Issue #119, opened 2016-12-05
Creating new cluster in brand new VPC requires "DNS Hostnames" setting to be turned on


jessesanford:
I believe that this is needed in the documentation/quickstart. It was a small stumble for me. Without it Etcd hostname couldn't be resolved by the controller.

http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html#vpc-dns-hostnames

Dec 05 14:08:55 ip-10-0-0-231.us-west-2.compute.internal etcdctl[8593]: Error: client: etcd cluster is unavailable or misconfigured
Dec 05 14:08:55 ip-10-0-0-231.us-west-2.compute.internal etcdctl[8593]: error #0: dial tcp: lookup ip-10-0-0-4.us-west-2.compute.internal: no such host

jessesanford:
http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html#vpc-dns-hostnames

jessesanford:
Dec 05 14:08:55 ip-10-0-0-231.us-west-2.compute.internal etcdctl[8593]: Error:  client: etcd cluster is unavailable or misconfigured
Dec 05 14:08:55 ip-10-0-0-231.us-west-2.compute.internal etcdctl[8593]: error #0: dial tcp: lookup ip-10-0-0-4.us-west-2.compute.internal: no such host

pieterlange:
You're deploying to an existing VPC? Please copy your cluster.yaml and version used.

jessesanford:
# Unique name of Kubernetes cluster. In order to deploy
# more than one cluster into the same AWS account, this
# name must not conflict with an existing cluster.
clusterName: kubernetes

# DNS name routable to the Kubernetes controller nodes
# from worker nodes and external clients. Configure the options
# below if you'd like kube-aws to create a Route53 record sets/hosted zones
# for you.  Otherwise the deployer is responsible for making this name routable
externalDNSName: <REDACTED>

# CoreOS release channel to use. Currently supported options: alpha, beta, stable
# See coreos.com/releases for more information
#releaseChannel: stable

# The AMI ID of CoreOS.
# If omitted, the latest AMI for the releaseChannel is used.
#amiId: ""

# Set to true if you want kube-aws to create a Route53 A Record for you.
createRecordSet: true

# TTL in seconds for the Route53 RecordSet created if createRecordSet is set to true.
recordSetTTL: 300

[...]

PR #162, opened 2016-12-14
Documentation for v0.9.2

### LINKED ISSUE #110: Documentation on external DNS seems out of date
Hello all,

I just setup a Kubernetes cluster with version `v0.9.1`. I see changes since version `0.8`:
- etcd cluster is now deployed on its own instances
- controller is in an autoscalling group
- controller is behind a load balancer

The doc: https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws-launch.html, says:

> Otherwise, navigate to the DNS registrar hosting the zone for the provided external DNS name. Ensure a single A record exists, routing the value of externalDNSName defined in cluster.yaml to the externally-accessible IP of the master node instance.

> You can invoke kube-aws status to get the cluster API IP address after cluster creation, if necessary. This command can take a while.

I did what I did with version `0.8`, I set an A record in my zone file to the public IP address of my controller node. However it is now behind a load balancer and the security group of the controller prevent HTTPS access from "0.0.0.0" and constraints it to only the load balancer and the worker nodes.

To access my cluster, I had to modify the security group and add a HTTPS inbound rule with "0.0.0.0". I guess I'm not supposed to do that. What are the recommandation here ?

Regards,

mumoshu:
Trys to cover notable features added in v0.9.2 plus fix #110 #120 #135 #138 #119
Not complete but a lot better than nothing!


mumoshu:
I'm going to replace almost all the duplicate chapters and its parts in `README.md` and replace them with appropriate links to documentation under `Documentation/`.

mumoshu:
Also, I'll try to cover every topic raised in issues.

mumoshu:
I've addressed most of the TODOs mentioned in #90

mumoshu:
Added something about customizing node pools.

I'm not satisfied with the amount of information covered by what were added from this PR but I'll leave things for future contribution from the community :bow:

mumoshu:
@pieterlange @camilb @c-knowles FYI, I've added some documentation all over the place.

PR #163, opened 2016-12-14
Fix kube-node-label failure when there is a whitespace in sg name

tarvip:
e.g if you attach external security group to worker nodes and this group name contains whitespace then `systemctl set-environment` fails with `Failed to set environment: Invalid environment assignments`
Also applied same fix for launchconfiguration and autoscaling group labels, though these names are generated, but it is possible to create manually autoscaling group or launchconfiguration that has whitespace in its name


mumoshu:
@tarvip LGTM 👍 Thanks for the contribution!

Issue #152, opened 2016-12-11
Create etcd and workers in private subnets, controllers in public subnet


andrejvanderzee:
My desired HA setup has three AZs each with a public and private subnet. The private subnets have its own route table that routes external, non-VPC traffic to the public NAT gateway in the same AZ. The public subnets route non-VPC traffic to an IGW.

Preferably I want to place etcd and worker nodes in the private subnets, and the controllers in the public subnet (only the controllers require external exposure via ELB).

I have been experimenting with `kube-aws` (really love it!) but it looks like my goal is not supported. Would it make sense to add options for controllers, etcd and workers that allows creation in `existing` subnets (we can create our own VPCs, subnets and routes easily)? Something like this:

```
subnets:
  - type: etcd
    subnetID: <privateID1>
  - type: etcd
    subnetID: <privateID2>
  - type: etcd
    subnetID: <privateID3>
  - type: controller
    subnetID: <publicID1>
  - type: controller
    subnetID: <publicID2>
  - type: controller
    subnetID: <publicID3>
  - type: worker
    subnetID: <privateID1>
  - type: worker
    subnetID: <privateID2>
  - type: worker
    subnetID: <privateID3>
```
 
When someone provides `availabilityZone` and `instanceCIDR` instead of `subnetID` kube-aws could create the subnets like it does now.

pieterlange:
I'd rather have controllers and etcd nodes be deployed through some default nodepools.

You can still deploy your desired situation by manually adding the subnets to the `stack-template.json`.

andrejvanderzee:
It looks like nodepools are under development now, and it looks awesome :-) I hope it supports having a dedicated route table for each nodepool, as we have a clear use case for private subnets with dedicated route table in each AZ. 

If you guys need any testing on nodepools, I would be glad to help.

neoandroid:
It's up to you to choose if you want your controllers in their own private subnet or not, the options is already in the PoC.
The NAT gateways are in the public subnet, and you have one NAT gateway per AZ so if you also put controllers and workers in all AZ you'll get an HA solution.

andrejvanderzee:
OK that exactly covers my use-case :+1: I guess its up to @pieterlange @mumoshu if this is "the right path"?

If I can be of any help in testing and/or code reviewing, I would be happy to make some time.

c-knowles:
[...]

PR #169, opened 2016-12-15
Create etcd and workers in private subnets, controllers in public subnet

neoandroid:
This PR belongs to issue #152 

The issue talks about the following use-cases supported by the PR:
- Place worker nodes in their own private subnet
- Place controller nodes in their own private subnet
- Place etcd nodes in their own private subnet
- Place node-pools on non pre-allocated private subnets
- Create NAT gateways to give Internet access to private subnets
- Use pre-allocated EIP's for the NAT gateways
- Use pre-allocated NAT gateways

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -179,6 +179,11 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 		}
 	}
 
+	// Populate subnets
+	if len(c.Subnets) > 0 && c.WorkerSettings.MinWorkerCount() > 0 && c.WorkerSettings.TopologyPrivate() == false {
+		c.WorkerSettings.PublicSubnets = c.Subnets
+	}
+
 	return c, nil
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -254,14 +254,15 @@ type WorkerSettings struct {
 
 // Part of configuration which is specific to controller nodes
 type ControllerSettings struct {
-	model.Controller         `yaml:"controller,omitempty"`
-	ControllerCount          int    `yaml:"controllerCount,omitempty"`
-	ControllerCreateTimeout  string `yaml:"controllerCreateTimeout,omitempty"`
-	ControllerInstanceType   string `yaml:"controllerInstanceType,omitempty"`
-	ControllerRootVolumeType string `yaml:"controllerRootVolumeType,omitempty"`
-	ControllerRootVolumeIOPS int    `yaml:"controllerRootVolumeIOPS,omitempty"`
-	ControllerRootVolumeSize int    `yaml:"controllerRootVolumeSize,omitempty"`
-	ControllerTenancy        string `yaml:"controllerTenancy,omitempty"`
+	model.Controller               `yaml:"controller,omitempty"`
+	ControllerCount                int    `yaml:"controllerCount,omitempty"`
+	ControllerCreateTimeout        string `yaml:"controllerCreateTimeout,omitempty"`
+	ControllerInstanceType         string `yaml:"controllerInstanceType,omitempty"`
+	ControllerLoadBalancerPrivate  string `yaml:"controllerLoadBalancerPrivate,omitempty"`
+	ControllerRootVolumeType       string `yaml:"controllerRootVolumeType,omitempty"`
+	ControllerRootVolumeIOPS       int    `yaml:"controllerRootVolumeIOPS,omitempty"`
+	ControllerRootVolumeSize       int    `yaml:"controllerRootVolumeSize,omitempty"`
+	ControllerTenancy              string `yaml:"controllerTenancy,omitempty"`
 }
 
 // Part of configuration which is specific to etcd nodes
--- a/config/config.go
+++ b/config/config.go
@@ -249,6 +249,7 @@ type WorkerSettings struct {
 	WorkerSpotPrice        string   `yaml:"workerSpotPrice,omitempty"`
 	WorkerSecurityGroupIds []string `yaml:"workerSecurityGroupIds,omitempty"`
 	WorkerTenancy          string   `yaml:"workerTenancy,omitempty"`
+	WorkerTopologyPrivate  bool     `yaml:"workerTopologyPrivate,omitempty"`
 }
 
 // Part of configuration which is specific to controller nodes
@@ -986,12 +987,16 @@ func (c *Cluster) AvailabilityZones() []string {
 		return []string{c.AvailabilityZone}
 	}
 
-	azs := make([]string, len(c.Subnets))
-	for i := range azs {
-		azs[i] = c.Subnets[i].AvailabilityZone
+	result := []string{}
+	seen := map[string]bool{}
+	for _, s := range c.Subnets{
+		val := s.AvailabilityZone
+		if _, ok := seen[val]; !ok {
+			result = append(result, val)
+			seen[val] = true
+		}
 	}
-
-	return azs
+	return result
 }
 
 /*
--- a/config/config.go
+++ b/config/config.go
@@ -254,15 +254,15 @@ type WorkerSettings struct {
 
 // Part of configuration which is specific to controller nodes
 type ControllerSettings struct {
-	model.Controller               `yaml:"controller,omitempty"`
-	ControllerCount                int    `yaml:"controllerCount,omitempty"`
-	ControllerCreateTimeout        string `yaml:"controllerCreateTimeout,omitempty"`
-	ControllerInstanceType         string `yaml:"controllerInstanceType,omitempty"`
-	ControllerLoadBalancerPrivate  string `yaml:"controllerLoadBalancerPrivate,omitempty"`
-	ControllerRootVolumeType       string `yaml:"controllerRootVolumeType,omitempty"`
-	ControllerRootVolumeIOPS       int    `yaml:"controllerRootVolumeIOPS,omitempty"`
-	ControllerRootVolumeSize       int    `yaml:"controllerRootVolumeSize,omitempty"`
-	ControllerTenancy              string `yaml:"controllerTenancy,omitempty"`
+	model.Controller              `yaml:"controller,omitempty"`
+	ControllerCount               int    `yaml:"controllerCount,omitempty"`
+	ControllerCreateTimeout       string `yaml:"controllerCreateTimeout,omitempty"`
+	ControllerInstanceType        string `yaml:"controllerInstanceType,omitempty"`
+	ControllerLoadBalancerPrivate string `yaml:"controllerLoadBalancerPrivate,omitempty"`
+	ControllerRootVolumeType      string `yaml:"controllerRootVolumeType,omitempty"`
+	ControllerRootVolumeIOPS      int    `yaml:"controllerRootVolumeIOPS,omitempty"`
+	ControllerRootVolumeSize      int    `yaml:"controllerRootVolumeSize,omitempty"`
+	ControllerTenancy             string `yaml:"controllerTenancy,omitempty"`
 }
 
 // Part of configuration which is specific to etcd nodes
@@ -385,7 +385,7 @@ type WaitSignal struct {
 }
 
 const (
-	vpcLogicalName = "VPC"
+	vpcLogicalName             = "VPC"
 	internetGatewayLogicalName = "InternetGateway"
 )
 
@@ -507,8 +507,8 @@ func (c Cluster) Config() (*Config, error) {
 		nextAddr := netutil.IncrementIP(*lastAllocatedAddr[subnet])
 		lastAllocatedAddr[subnet] = &nextAddr
 		instance := etcdInstance{
-			IPAddress:   *lastAllocatedAddr[subnet],
-			Subnet: subnet,
+			IPAddress: *lastAllocatedAddr[subnet],
+			Subnet:    subnet,
 		}
 
 		//TODO(chom): validate we're not overflowing the address space
@@ -659,8 +659,8 @@ func (c Cluster) RenderStackTemplate(opts StackTemplateOptions, prettyPrint bool
 }
 
 type etcdInstance struct {
-	IPAddress   net.IP
-	Subnet      model.Subneter
+	IPAddress net.IP
+	Subnet    model.Subneter
 }
 
 type Config struct {
@@ -698,7 +698,7 @@ func (c Config) InternetGatewayLogicalName() string {
 
 func (c Config) InternetGatewayRef() string {
 	if c.InternetGatewayID != "" {
-		return fmt.Sprintf("%q", c.InternetGatewayID);
+		return fmt.Sprintf("%q", c.InternetGatewayID)
 	} else {
 		return fmt.Sprintf(`{ "Ref" : %q }`, c.InternetGatewayLogicalName())
 	}
@@ -990,7 +990,7 @@ func (c *Cluster) AvailabilityZones() []string {
 
 	result := []string{}
 	seen := map[string]bool{}
-	for _, s := range c.Subnets{
+	for _, s := range c.Subnets {
 		val := s.AvailabilityZone
 		if _, ok := seen[val]; !ok {
 			result = append(result, val)
--- a/config/config.go
+++ b/config/config.go
@@ -264,7 +264,7 @@ type ControllerSettings struct {
 	ControllerCount               int    `yaml:"controllerCount,omitempty"`
 	ControllerCreateTimeout       string `yaml:"controllerCreateTimeout,omitempty"`
 	ControllerInstanceType        string `yaml:"controllerInstanceType,omitempty"`
-	ControllerLoadBalancerPrivate string `yaml:"controllerLoadBalancerPrivate,omitempty"`
+	ControllerLoadBalancerPrivate bool   `yaml:"controllerLoadBalancerPrivate,omitempty"`
 	ControllerRootVolumeType      string `yaml:"controllerRootVolumeType,omitempty"`
 	ControllerRootVolumeIOPS      int    `yaml:"controllerRootVolumeIOPS,omitempty"`
 	ControllerRootVolumeSize      int    `yaml:"controllerRootVolumeSize,omitempty"`
--- a/config/config.go
+++ b/config/config.go
@@ -476,19 +476,21 @@ func (c Cluster) Config() (*Config, error) {
 	config.EtcdInstances = make([]etcdInstance, config.EtcdCount)
 	var etcdEndpoints, etcdInitialCluster bytes.Buffer
 
-	var lastAllocatedAddr = make(map[*model.Subnet]*net.IP)
+	var lastAllocatedAddr = make(map[model.Subneter]*net.IP)
 	for etcdIndex := 0; etcdIndex < config.EtcdCount; etcdIndex++ {
 
 		//Round-robbin etcd instances across all available subnets
 		subnetIndex := etcdIndex % len(config.Subnets)
-		subnet := config.Subnets[subnetIndex]
+		subnet := model.Subneter(config.Subnets[subnetIndex])
+		subnetInstanceCIDR := config.Subnets[subnetIndex].InstanceCIDR
 		if config.Etcd.TopologyPrivate() {
-			subnet = config.Etcd.PrivateSubnets[subnetIndex]
+			subnet = model.Subneter(config.Etcd.PrivateSubnets[subnetIndex])
+			subnetInstanceCIDR = config.Etcd.PrivateSubnets[subnetIndex].InstanceCIDR
 		}
 
-		_, subnetCIDR, err := net.ParseCIDR(subnet.InstanceCIDR)
+		_, subnetCIDR, err := net.ParseCIDR(subnetInstanceCIDR)
 		if err != nil {
-			return nil, fmt.Errorf("error parsing subnet instance cidr %s: %v", subnet.InstanceCIDR, err)
+			return nil, fmt.Errorf("error parsing subnet instance cidr %s: %v", subnetInstanceCIDR, err)
 		}
 
 		if lastAllocatedAddr[subnet] == nil {
@@ -656,7 +658,7 @@ func (c Cluster) RenderStackTemplate(opts StackTemplateOptions, prettyPrint bool
 
 type etcdInstance struct {
 	IPAddress   net.IP
-	Subnet      *model.Subnet
+	Subnet      model.Subneter
 }
 
 type Config struct {
--- a/config/config.go
+++ b/config/config.go
@@ -75,7 +75,8 @@ func NewDefaultCluster() *Cluster {
 			AWSCliImageRepo:    "quay.io/coreos/awscli",
 			AWSCliTag:          "master",
 			ContainerRuntime:   "docker",
-			Subnets:            []*Subnet{},
+			Subnets:            []*model.Subnet{},
+			EIPAllocationIDs:   []string{},
 			MapPublicIPs:       true,
 			Experimental:       experimental,
 		},
@@ -170,7 +171,7 @@ func ClusterFromBytes(data []byte) (*Cluster, error) {
 
 	// For backward-compatibility
 	if len(c.Subnets) == 0 {
-		c.Subnets = []*Subnet{
+		c.Subnets = []*model.Subnet{
 			{
 				AvailabilityZone: c.AvailabilityZone,
 				InstanceCIDR:     c.InstanceCIDR,
@@ -209,14 +210,15 @@ type ComputedDeploymentSettings struct {
 // Though it is highly configurable, it's basically users' responsibility to provide `correct` values if they're going beyond the defaults.
 type DeploymentSettings struct {
 	ComputedDeploymentSettings
-	ClusterName      string `yaml:"clusterName,omitempty"`
-	KeyName          string `yaml:"keyName,omitempty"`
-	Region           string `yaml:"region,omitempty"`
-	AvailabilityZone string `yaml:"availabilityZone,omitempty"`
-	ReleaseChannel   string `yaml:"releaseChannel,omitempty"`
-	AmiId            string `yaml:"amiId,omitempty"`
-	VPCID            string `yaml:"vpcId,omitempty"`
-	RouteTableID     string `yaml:"routeTableId,omitempty"`
+	ClusterName       string `yaml:"clusterName,omitempty"`
+	KeyName           string `yaml:"keyName,omitempty"`
+	Region            string `yaml:"region,omitempty"`
+	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
+	ReleaseChannel    string `yaml:"releaseChannel,omitempty"`
+	AmiId             string `yaml:"amiId,omitempty"`
+	VPCID             string `yaml:"vpcId,omitempty"`
+	InternetGatewayID string `yaml:"internetGatewayId,omitempty"`
+	RouteTableID      string `yaml:"routeTableId,omitempty"`
 	// Required for validations like e.g. if instance cidr is contained in vpc cidr
 	VPCCIDR             string            `yaml:"vpcCIDR,omitempty"`
 	InstanceCIDR        string            `yaml:"instanceCIDR,omitempty"`
@@ -227,7 +229,8 @@ type DeploymentSettings struct {
 	ContainerRuntime    string            `yaml:"containerRuntime,omitempty"`
 	KMSKeyARN           string            `yaml:"kmsKeyArn,omitempty"`
 	StackTags           map[string]string `yaml:"stackTags,omitempty"`
-	Subnets             []*Subnet         `yaml:"subnets,omitempty"`
+	Subnets             []*model.Subnet   `yaml:"subnets,omitempty"`
+	EIPAllocationIDs    []string          `yaml:"eipAllocationIDs,omitempty"`
 	MapPublicIPs        bool              `yaml:"mapPublicIPs,omitempty"`
 	ElasticFileSystemID string            `yaml:"elasticFileSystemId,omitempty"`
 	SSHAuthorizedKeys   []string          `yaml:"sshAuthorizedKeys,omitempty"`
@@ -262,6 +265,7 @@ type ControllerSettings struct {
 
 // Part of configuration which is specific to etcd nodes
 type EtcdSettings struct {
+	model.Etcd              `yaml:"etcd,omitempty"`
 	EtcdCount               int    `yaml:"etcdCount"`
 	EtcdInstanceType        string `yaml:"etcdInstanceType,omitempty"`
 	EtcdRootVolumeSize      int    `yaml:"etcdRootVolumeSize,omitempty"`
@@ -294,12 +298,6 @@ type Cluster struct {
 	HostedZone             string `yaml:"hostedZone,omitempty"`
 	HostedZoneID           string `yaml:"hostedZoneId,omitempty"`
 	providedEncryptService EncryptService
-}
-
-type Subnet struct {
-	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
-	InstanceCIDR      string `yaml:"instanceCIDR,omitempty"`
-	lastAllocatedAddr *net.IP
 }
 
 type Experimental struct {
@@ -386,6 +384,7 @@ type WaitSignal struct {
 
 const (
 	vpcLogicalName = "VPC"
+	internetGatewayLogicalName = "InternetGateway"
 )
 
 var supportedReleaseChannels = map[string]bool{
@@ -476,6 +475,7 @@ func (c Cluster) Config() (*Config, error) {
 
 	//Set logical name constants
 	config.VPCLogicalName = vpcLogicalName
+	config.InternetGatewayLogicalName = internetGatewayLogicalName
 
 	//Set reference strings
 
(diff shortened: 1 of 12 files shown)


mumoshu:
@neoandroid Thanks for the PR! Could I help you with anything towards https://github.com/coreos/kube-aws/issues/152#issuecomment-267197330?

neoandroid:
I'm still working on it. I plan to add the proposed changes before the end of this week.

neoandroid:
I finally merged the work from @icereval , I liked the approach and made a couple of modifications to have the same capabilities that I had with the previous PoC.

- Add a parameter on cluster.yml called "controllerLoadBalancerPrivate" to have the option to put the ELB on the public subnet although I have my controllers on a private subnet. This way I can access them via Internet without the need of setting up a bastion host (a work that could be done later, although it could be considered out of the scope of this project too).
- Add support to create workers on private subnets that are not part of a node pool since it was possible on the PoC and I still need it.

The rest of the changes I've added are small fixes: a typo in cluster.yml and a wrong reference for the Internet Gateway on the cfn template.

Note: The tests  are broken but I'm currently fixing it

andrejvanderzee:
@neoandroid I also prefer to be able to create workers on a private subnet that are not in a node pool.

andrejvanderzee:
@neoandroid Could you please post a working cluster.yaml? I have tried the `origin/masters-private-subnet` branch with this config:

```
clusterName: PRD
externalDNSName: prd.vanderzee-it.eu
createRecordSet: true
recordSetTTL: 320
hostedZoneId: "Z2N238IAN5AZMQ"
keyName: aws-vanderzee-it
region: eu-west-1
kmsKeyArn: "arn:aws:kms:eu-west-1:595627617610:key/de5f05e9-c2c1-4537-a276-5c9d701bf82a"
controllerCount: 3
controller:
  privateSubnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: "10.0.3.0/24"
  - availabilityZone: eu-west-1b
    instanceCIDR: "10.0.4.0/24"
  - availabilityZone: eu-west-1c
    instanceCIDR: "10.0.5.0/24"
workerCount: 3
worker:
  privateSubnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: "10.0.3.0/24"
  - availabilityZone: eu-west-1b
    instanceCIDR: "10.0.4.0/24"
  - availabilityZone: eu-west-1c
    instanceCIDR: "10.0.5.0/24"
etcdCount: 3
etcd:
  privateSubnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: "10.0.3.0/24"
  - availabilityZone: eu-west-1b
    instanceCIDR: "10.0.4.0/24"
  - availabilityZone: eu-west-1c
    instanceCIDR: "10.0.5.0/24"
vpcCIDR: "10.0.0.0/16"
subnets:
- availabilityZone: eu-west-1a
  instanceCIDR: "10.0.0.0/24"
- availabilityZone: eu-west-1b
  instanceCIDR: "10.0.1.0/24"
- availabilityZone: eu-west-1c
  instanceCIDR: "10.0.2.0/24"
serviceCIDR: "10.3.0.0/24"
podCIDR: "10.2.0.0/16"
dnsServiceIP: 10.3.0.10
tlsCADurationDays: 3650
tlsCertDurationDays: 365
kubernetesVersion: v1.5.1_coreos.0
hyperkubeImageRepo: quay.io/coreos/hyperkube
awsCliImageRepo: quay.io/coreos/awscli
useCalico: true
containerRuntime: docker
stackTags:
  Name: "Kubernetes"
  Environment: "Production"
```

But it does not create any private subnets.

Probably I am missing something?

neoandroid:
Here you have a full example:
`clusterName: foo
externalDNSName: foo.example.tld
createRecordSet: true
hostedZoneId: "ABCDEFGHJ12345"
keyName: foobar
region: eu-west-1
kmsKeyArn: "arn:aws:kms:eu-west-1:foobar"
controller:
  # Auto Scaling Group definition for controllers. If only `controllerCount` is specified, min and max will be the set to that value and `rollingUpdateMinInstancesInService` will be one less.
  autoScalingGroup:
    minSize: 3
    maxSize: 3
    rollingUpdateMinInstancesInService: 2
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.6.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.7.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.8.0/24"
worker:
  # Auto Scaling Group definition for workers. If only `workerCount` is specified, min and max will be the set to that value and `rollingUpdateMinInstancesInService` will be one less.
  autoScalingGroup:
    minSize: 3
    maxSize: 6
    rollingUpdateMinInstancesInService: 2
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.9.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.10.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.11.0/24"
etcdCount: 3
etcd:
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.3.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.4.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.5.0/24"
subnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: "10.0.0.0/24"
  - availabilityZone: eu-west-1b
    instanceCIDR: "10.0.1.0/24"
  - availabilityZone: eu-west-1c
    instanceCIDR: "10.0.2.0/24"
useCalico: true
stackTags:
  Name: "Kubernetes"
  Environment: "Production"`

At first glance I see that you have repeated the same CIDRs for worker, controller and etcd subnets. Probably that is the reason why its failing.

andrejvanderzee:
Tried to differ subnets too does not help:

```clusterName: PRD
externalDNSName: xxx
createRecordSet: true
recordSetTTL: 320
hostedZoneId: xxx
keyName: xxx
region: eu-west-1
kmsKeyArn: xxx
controller:
  autoScalingGroup:
    minSize: 3
    maxSize: 3
    rollingUpdateMinInstancesInService: 2
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.6.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.7.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.8.0/24"
worker:
  autoScalingGroup:
    minSize: 3
    maxSize: 6
    rollingUpdateMinInstancesInService: 2
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.9.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.10.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.11.0/24"
etcdCount: 3
etcd:
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: "10.0.3.0/24"
    - availabilityZone: eu-west-1b
      instanceCIDR: "10.0.4.0/24"
    - availabilityZone: eu-west-1c
      instanceCIDR: "10.0.5.0/24"
subnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: "10.0.0.0/24"
  - availabilityZone: eu-west-1b
    instanceCIDR: "10.0.1.0/24"
  - availabilityZone: eu-west-1c
    instanceCIDR: "10.0.2.0/24"
useCalico: true
stackTags:
  Name: "Kubernetes"
  Environment: "Production"
```

It only creates three public subnets. 

No time to dig into sources now, maybe in the new year :-)

icereval:
Can you elaborate on the use case for needing more than three public subnets?

icereval:
Also over at the KOPS project a contributor created an awesome diagram of their desired network (mostly matching) that might make it easier to discuss.

https://github.com/kubernetes/kops/issues/428#issuecomment-252231600

andrejvanderzee:
For us, we require only 3 private and 3 public subnets (1 per AZ). Therore we would prefer to put etcd, workers and controllers spread over 3 private subnets.

andrejvanderzee:
@icereval What I meant is that three public subnets were created as expected, but no private subnets. I have updated the comment to make it more obvious.

icereval:
Hmmm, I just checked out the latest code and I'm seeing the correct results. Let me get my config together and share it.

icereval:
I'm running an environment setup script via cloudformation which generates the base VPC/DNS/Bastion/VPC Peering and so on, but other than that its a typical setup.

With this configuration running `kube-aws up --s3-uri s3://${S3_BUCKET}/` cloudformation proceeds to build the following environment.

#### cluster.yaml

```yaml
clusterName: cluster-1
externalDNSName: cluster-1.dev.example.internal
amiId: ami-69cbda7e  # copy of coreos stable for encrypted root volume
createRecordSet: true
hostedZoneId: ZWB2UKRTYXHY0  # private hosted zone dev.example.internal
region: us-east-1
keyName: ...
kmsKeyArn: ...
vpcId: vpc-76c11e10
internetGatewayId: igw-adce55ca
vpcCIDR: 10.0.0.0/16
tlsCADurationDays: 3650
tlsCertDurationDays: 1825  # modified (5 years)
useCalico: true 

# limited public subnet cidr needed since the bastion will grant
# access into this space via vpc peering
subnets:
  - availabilityZone: us-east-1a
    instanceCIDR: 10.0.0.0/26
  - availabilityZone: us-east-1b
    instanceCIDR: 10.0.0.64/26
  - availabilityZone: us-east-1d
    instanceCIDR: 10.0.0.128/26

controller:
  autoScalingGroup:
    minSize: 3
    maxSize: 4
    rollingUpdateMinInstancesInService: 1
  privateSubnets:
    - availabilityZone: us-east-1a
      instanceCIDR: 10.0.1.0/26
    - availabilityZone: us-east-1b
      instanceCIDR: 10.0.1.64/26
    - availabilityZone: us-east-1d
      instanceCIDR: 10.0.1.128/26

etcdCount: 3
etcd:
  privateSubnets:
    - availabilityZone: us-east-1a
      instanceCIDR: 10.0.2.0/26
    - availabilityZone: us-east-1b
      instanceCIDR: 10.0.2.64/26
    - availabilityZone: us-east-1d
      instanceCIDR: 10.0.2.128/26

worker:
  autoScalingGroup:
    minSize: 3
    maxSize: 4
    rollingUpdateMinInstancesInService: 1
  privateSubnets:
    - availabilityZone: us-east-1a
      instanceCIDR: 10.0.3.0/24
    - availabilityZone: us-east-1b
      instanceCIDR: 10.0.4.0/24
    - availabilityZone: us-east-1d
      instanceCIDR: 10.0.5.0/24
```

#### Subnets

screen shot 2016-12-30 at 3 00 40 pm

#### Route Tables

screen shot 2016-12-30 at 3 09 14 pm

#### Nat Gateways

screen shot 2016-12-30 at 3 03 08 pm

#### EC2 Instances

screen shot 2016-12-30 at 3 02 04 pm

config/templates/stack-template.json
```diff
@@ -1020,139 +1079,305 @@
...
-        "VpcId": {{.VPCRef}}
+        "DestinationCidrBlock": "0.0.0.0/0",
+        "GatewayId": {{$.InternetGatewayRef}},
+        "RouteTableId": {"Ref" : "PublicRouteTable"}
       },
-      "Type": "AWS::EC2::VPCGatewayAttachment"
+      "Type": "AWS::EC2::Route"
     }
+
+    {{if (or .WorkerTopologyPrivate .Etcd.TopologyPrivate .Controller.TopologyPrivate)}}
```
icereval: Should also add `.Worker.TopologyPrivate` for the case where workers are private yet somehow etcd & controller are public and not using node pools.

andrejvanderzee:
Hmmmm not sure what is going on. I more-or-less have copied your cluster.yaml now (expect that I don't create VPC before running `kube-aws`):

```
clusterName: PRD
externalDNSName: xxx
createRecordSet: true
recordSetTTL: 320
hostedZoneId: xxx
keyName: aws-vanderzee-it
region: eu-west-1
kmsKeyArn: xxx
tlsCADurationDays: 3650
tlsCertDurationDays: 1825  # modified (5 years)
vpcCIDR: 10.0.0.0/16

# limited public subnet cidr needed since the bastion will grant
# access into this space via vpc peering
subnets:
  - availabilityZone: eu-west-1a
    instanceCIDR: 10.0.0.0/26
  - availabilityZone: eu-west-1b
    instanceCIDR: 10.0.0.64/26
  - availabilityZone: eu-west-1c
    instanceCIDR: 10.0.0.128/26

controller:
  autoScalingGroup:
    minSize: 3
    maxSize: 4
    rollingUpdateMinInstancesInService: 1
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: 10.0.1.0/26
    - availabilityZone: eu-west-1b
      instanceCIDR: 10.0.1.64/26
    - availabilityZone: eu-west-1c
      instanceCIDR: 10.0.1.128/26

etcdCount: 3
etcd:
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: 10.0.2.0/26
    - availabilityZone: eu-west-1b
      instanceCIDR: 10.0.2.64/26
    - availabilityZone: eu-west-1c
      instanceCIDR: 10.0.2.128/26

worker:
  autoScalingGroup:
    minSize: 3
    maxSize: 4
    rollingUpdateMinInstancesInService: 1
  privateSubnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: 10.0.3.0/24
    - availabilityZone: eu-west-1b
      instanceCIDR: 10.0.4.0/24
    - availabilityZone: eu-west-1c
      instanceCIDR: 10.0.5.0/24
```

No NATs are created, no private subnets.

```
$ git status
On branch masters-private-subnet
Your branch is up-to-date with 'origin/masters-private-subnet'.
$ git remote -v
origin	https://github.com/neoandroid/kube-aws.git (fetch)
origin	https://github.com/neoandroid/kube-aws.git (push)
```

```
./bin/kube-aws render stack
./bin/kube-aws  up --s3-uri XXX
```

icereval:
@andrejvanderzee I've taken your config and ran it in Ireland region and all checks out.

I'm running low on ideas, beyond maybe the wrong binary or permissions which seems unlikely, anyway here is my kube-aws sha256 and git commit sha if you'd like to verify yours.

```bash
> git branch -vv
  master                            dec2402 [origin/master] Update kubernetes-on-aws-limitations.md
* neoandroid-masters-private-subnet 3e8d431 [neoandroid/masters-private-subnet] fix go format
```

```bash
> shasum -a 256 -b ~/go/src/github.com/coreos/kube-aws/bin/kube-aws
e5910f4acc35d46c14c2c6c280913014f34f17c2d9201249d28e09ee25b06dd8 *~/go/src/github.com/coreos/kube-aws/bin/kube-aws
```

config/templates/stack-template.json
neoandroid: Done. Nice catch! :+1:

icereval:
Also its probably worth noting the build I'm generating is on OSX.

andrejvanderzee:
@icereval Thanks, you just made me realize that I was building against the original kube-aws sources in my GOPATH, sorry for the confusion. It is working for me now.

BTW, I have noticed that controllers are now schedulable. Is that intended?

config/templates/cluster.yaml
```diff
@@ -53,15 +53,23 @@ kmsKeyArn: "{{.KMSKeyARN}}"
...
+#    minSize: 1
+#    maxSize: 3
+#    rollingUpdateMinInstancesInService: 2
+#  privateSubnets:
+#    - availabilityZone: us-west-1a
+#      instanceCIDR: "10.0.4.0/24"
+#    - availabilityZone: us-west-1b
+#      instanceCIDR: "10.0.5.0/24"
 
 # Maximum time to wait for controller creation
```
mumoshu: A comment specific to `controllerLoadBalancerPrivate` would be nice!
I'd imagine that something like `Set to true to put the ELB on the public subnet although having controller nodes on private subnets` would be enough but am open to your suggestions.

neoandroid: Fixed :ok_hand:

nodepool/config/templates/stack-template.json
```diff
@@ -8,8 +8,50 @@
         "TargetCapacity": {{$.Worker.SpotFleet.TargetCapacity}},
         "SpotPrice": "{{$.Worker.SpotFleet.SpotPrice}}",
         "LaunchSpecifications": [
+          {{if .Worker.TopologyPrivate}}
+          {{range $subnetIndex, $subnet := $.Worker.PrivateSubnets}}
```
mumoshu: @neoandroid So we now have two conditions:

1. `TopologyPrivate == true`: Worker.PrivateSubnets are used for deployment targets
2. `TopologyPrivate != true`: Worker.Subnets are used

As this introduces some duplicated code into stack-template.json, would you mind me to refactor this to:

0. Rename `Worker.Subnets` to `Worker.PublicSubnets` and remove the `if-else` blocks here.
1. `TopologyPrivate == true`: `Worker.Subnets` are populated with `Worker.PrivateSubnets`
2. `TopologyPrivate != true`: `Worker.Subnets` are populated with `Worker.PrivateSubnets`

?

nodepool/config/templates/stack-template.json
```diff
@@ -60,10 +102,17 @@
     "AutoScaleWorker": {
       "Properties": {
         "AvailabilityZones": [
+          {{if .Worker.TopologyPrivate}}
+          {{range $index, $subnet := .Worker.PrivateSubnets}}
+          {{if gt $index 0}},{{end}}
+          "{{$subnet.AvailabilityZone}}"
+          {{end}}
+          {{else}}
```
mumoshu: If I understand correctly, I'd leave a comment same as https://github.com/coreos/kube-aws/pull/169/files#r94915440

nodepool/config/templates/stack-template.json
```diff
@@ -197,7 +253,7 @@
 {{end}}
 {
   "AWSTemplateFormatVersion": "2010-09-09",
-  "Description": "kube-aws Kubernetes node pool {{.NodePoolName}}",
+  "Description": "kube-aws Kubernetes node pool {{.ClusterName}} {{.NodePoolName}}",
```
mumoshu: LGTM 👍

nodepool/config/templates/stack-template.json
```diff
@@ -317,18 +373,47 @@
...
+      "Type" : "AWS::EFS::MountTarget"
+    }
+    {{end}}
+    {{end}}
+    {{end}}
+
+    {{if $.Worker.TopologyPrivate}}
+    {{range $index, $subnet := .Worker.PrivateSubnets}}
+    ,
+    "{{$subnet.LogicalName}}": {
```
mumoshu: If I understand correctly, I'd leave a comment same as https://github.com/coreos/kube-aws/pull/169/files#r94915440

c-knowles:
@neoandroid I was just having a look through your work, thank you for your effort! I had a question regarding the use case `Place node-pools on private subnets`. Node pools are already place-able in pre-existing private subnets so my understanding of your change is to provide control so `kube-aws` still creates the subnets, it just creates private ones?

neoandroid:
@c-knowles Yes, you're right. I've modified the text to be more accurate.

Btw, I'm working on the changes proposed by @mumoshu and hope to push them tomorrow after testing some more scenarios.

nodepool/config/templates/stack-template.json
neoandroid: Done

nodepool/config/templates/stack-template.json
neoandroid: I'm not sure if I've understood exactly what  you meant but I thought the general idea was to get rid of that many if/else blocks introduced by this private subnet patch.
I hope you like the solution I've applied and I will extend it to the controllers and etcd sections if you approve it

config/templates/stack-template.json
```diff
@@ -89,14 +88,22 @@
       },
       "DependsOn" : ["AutoScaleController"]
     },
+    {{end}}
     "AutoScaleController": {
       "Type": "AWS::AutoScaling::AutoScalingGroup",
       "Properties": {
         "AvailabilityZones": [
```
redbaron: `AvailabilityZones` param of autoscaling group resource can be omitted all together, ASG uses zones from `VPCZoneIdentifier` which we always specify

config/templates/stack-template.json
```diff
@@ -123,11 +130,20 @@
...
+          {{range $index, $controllerSubnet := .Controller.PrivateSubnets}}
+          {{with $controllerSubnetLogicalName := printf "Controller%s" $controllerSubnet.LogicalName}}
+          {{if gt $index 0}},{{end}}
+          {
+            "Ref": "{{$controllerSubnetLogicalName}}"
+          }
+          {{end}}
+          {{end}}
+          {{else}}
           {{range $index, $subnet := .Subnets}}
```
redbaron: can't we implicitly populate `Controller.PrivateSubnets` with "global" subnets if they are not set? Then there will be less branching all over templates

neoandroid: Of course, we can. Look at this [comment](https://github.com/coreos/kube-aws/pull/169#discussion_r95475310)
 I though it was a good idea to wait for @mumoshu approval before doing more work

config/templates/stack-template.json
neoandroid: Good point. If nobody disagrees I will remove the availability zone properties from the template tomorrow.

config/templates/cluster.yaml
```diff
@@ -53,15 +53,23 @@ kmsKeyArn: "{{.KMSKeyARN}}"
...
+#      instanceCIDR: "10.0.4.0/24"
+#    - availabilityZone: us-west-1b
+#      instanceCIDR: "10.0.5.0/24"
+
+# Set to true to put the ELB on the public subnet although having controller nodes on private subnets
+#controllerLoadBalancerPrivate: false
 
 # Maximum time to wait for controller creation
-#controlerCreateTimeout: PT15M
+#controllerCreateTimeout: PT15M
```
mumoshu: Good catch 👍

config/templates/stack-template.json
mumoshu: I agree to the suggestion 👍

c-knowles:
@mumoshu I've been playing around with the great work going on here plus https://github.com/coreos/kube-aws/pull/195, https://github.com/coreos/kube-aws/pull/227 and I can see in general we're all looking for more control over what gets placed where while still leaving kube-aws to do most of the setup (even in shared VPCs). I think overall it'll be good to get some more control and there's a lot of overlap between what everyone wants.

What I'm wondering is how we go about keeping this stable and maintainable while still keeping the ease of use from a user perspective. Ideas I had:
* Should we add some documentation outside of cluster.yaml to show the major use cases we intend to support? I think the self-documenting method for the config switches can go so far but with so many options and flexibility maybe a bit more description would be useful?
* Where possible I think the configuration should minimise multiple layers of defaults (and duplication). For example, for subnets we already have some different layers prior to this PR. Could we simplify and remove the code level defaults and replace `privateSubnets` with a `private` flag inside the subnets? That way we have a single place where subnets are configured for each set of controllers/workers/etcd rather than layers where configuration and defaulting can occur (which at least to me seems harder to use). If we were to break backwards compatibility, we'd be doing so for ultimately better stability and maintainability.
* I'm trying to think of ideas about how we could split up the monolithic templating while still keeping user ease of use.

mumoshu:
@c-knowles Thanks as always for taking a look into this 👍 

> Should we add some documentation outside of cluster.yaml to show the major use cases we intend to support?

Yes, I believe so!

> Where possible I think the configuration should minimize multiple layers of defaults (and duplication)

Yes. However, basically, I'd also like to avoid something like `private` flag which mixes things up different things(private vs public subnet) into a single value (subnet).

My idea on how to tackle it is: fix api visible from `cluster.yaml` and `stack-template.json` anyways in this PR and then after merging this PR we'd discuss/open another PR for more maintainability like you've suggested (while our memories are still fresh 😄)

> I'm trying to think of ideas about how we could split up the monolithic templating while still keeping user ease of use.

It seems to be the way to go for me too but not sure how yet.

c-knowles:
Sure, if you want me to tackle any of those I can take a look. On the `private` point, I would have thought from a user and kube-aws perspective that it is part of the canonical definition of one subnet (and the code it goes through since it only seems to map to `MapPublicIpOnLaunch=false`).

mumoshu:
@c-knowles I'm afraid if I'm following you correctly but why I preferred `privateSubnets` is because we won't allow something like:

```yaml
controller:
  subnets:
    - availabilityZone: eu-west-1a
      instanceCIDR: 10.0.3.0/24
      private: true
    - availabilityZone: eu-west-1b
      instanceCIDR: 10.0.4.0/24
      private: false
```

In other words, I'd wanted to force users to decide not a subnet but a group of subnets to be private or public so that they can't go into trouble.

c-knowles:
@mumoshu IMO I don't think that will make an end user think about it any more or less, a flag seems pretty clear. It's still possible with both ways to mix private/public I think? I think we should allow that flexibility in the case we want mostly private nodes but some public which run NodePort style services (occasionally necessary). Perhaps the code level switching that is needed in either option means what happens is not so transparent for the end user.

mumoshu:
> It's still possible with both ways to mix private/public I think? 

For now, no. I'm not thinking of making it possible for controller and etcd nodes. Anyway for worker nodes, I guess we can allow mixing private/public subnets via node pools without the `private` flag?

c-knowles:
@mumoshu I think we should allow for worker nodes at least, I think we agree there to be a single way to achieving the same thing though regardless of type of node. i.e. align the workings of worker/controller/etcd across main/nodepool.

mumoshu:
@c-knowles Thanks for the reply!

I believe that your req "a single way to achieving the same thing though regardless of type of node" is still achievable even we go ahead with "privateSubnets".
Then, I guess we'd need to discuss about exact use-cases in which we need to mix private/public subnets used for etcd and controller nodes?

mumoshu:
I'm not aware of exact use-cases to mix private/public subnets for etcd and controller; I'd generally recommend putting etcd and controller nodes inside private subnets once this get merged.

c-knowles:
@mumoshu Trying to think of a use case, perhaps using one of the controllers as the bastion host to cut down on instance cost? Not that I especially want to support that case! We could write the minimum cluster instance count in the docs we agreed on above. Do you want me to write up some of that?

I'm good with whatever you decide :) I'd prefer the work to be merged than not. How about we stick with what is here but agree we may later decide to simplify (and we may have to drop some compatibility to keep it simple)?

mumoshu:
@c-knowles I'll decide to go with the simpler one: `subnets:` without `private: true|false` which might be less confusing/conflicting then!

Even with that:

* my fear of trapping users like https://github.com/coreos/kube-aws/pull/169#issuecomment-272757595 (could be aided with more validations and)
* your intention to minimize multiple layers of defaults and duplication

could both be achieved. Am I correct?

c-knowles:
@mumoshu I think the second point is less achievable that way. It's not a huge issue, more a logical one of how we structure in general. Sorry I hijacked this PR a little bit! Happy to chat outside of here too. For example, if sets of both private and public subnets are specified which "wins"? Currently I believe it's the private subnets. We can document it but ultimately we will still have multiple places where one piece of infrastructure can be specified (in this case the worker subnets).

mumoshu:
> if sets of both private and public subnets are specified which "wins"? Currently I believe it's the private subnets.

My assumption is that:
* If you provide both top-level `subnets:` and worker/controller/etcd specific `subnets:`, the latter wins.
* If you provide both private and public subnets under a single `subnets:`, everyone wins i.e. all the nodes distribute over these subnets without useful labels/taints (for now) hence I'm trying to avoid mixing public/private subnets (for now)

Also note that:

* I agree that we should allow defining subnets elsewhere than of today and then reference them from worker|controller|etcd subnets but that's an another issue in IMHO.

Did I cover all the general concerns you might have considered?

@c-knowles Are you in the kubernetes slack? 😃

nodepool/config/templates/stack-template.json
mumoshu: Sorry for being silent for a while but yes, I believe you did a great job implementing my vague idea!
Would you mind going ahead with your work for controller and etcd, too?

mumoshu:
Hi @neoandroid, sorry for a long silence but I'm still wanting this to be merged!

In addition to https://github.com/coreos/kube-aws/pull/169#discussion_r96192429, 

* would you mind changing `privateSubnets` to just `subnets` after the long discussion above?
* would you mind adding support for NAT gateways? (It seems to be missing in the `stack-template.json`)

c-knowles:
@mumoshu great yes, I agree we can simplify slightly later especially if we can leave behind some legacy prior to v1. I am indeed on the k8s slack, username without the hyphen.

config/config.go
```diff
@@ -247,22 +255,25 @@ type WorkerSettings struct {
...
-	ControllerInstanceType   string `yaml:"controllerInstanceType,omitempty"`
-	ControllerRootVolumeType string `yaml:"controllerRootVolumeType,omitempty"`
-	ControllerRootVolumeIOPS int    `yaml:"controllerRootVolumeIOPS,omitempty"`
-	ControllerRootVolumeSize int    `yaml:"controllerRootVolumeSize,omitempty"`
-	ControllerTenancy        string `yaml:"controllerTenancy,omitempty"`
+	model.Controller              `yaml:"controller,omitempty"`
+	ControllerCount               int    `yaml:"controllerCount,omitempty"`
+	ControllerCreateTimeout       string `yaml:"controllerCreateTimeout,omitempty"`
+	ControllerInstanceType        string `yaml:"controllerInstanceType,omitempty"`
+	ControllerLoadBalancerPrivate string `yaml:"controllerLoadBalancerPrivate,omitempty"`
```
acaire: Should this be a `bool`?

neoandroid: You're right. Good catch :+1:

mumoshu:
Hi @neoandroid, I'm going to merge several PRs including this one from now on.
If one of those PRs conflicted with your PR badly, would you mind me to resolve the conflict and merge this?
After that, we can tackle on https://github.com/coreos/kube-aws/pull/169#issuecomment-272817555 if necessary.

neoandroid:
I'm working on the changes commented [here](https://github.com/coreos/kube-aws/pull/169#issuecomment-272817555) and [here](https://github.com/coreos/kube-aws/pull/169#discussion_r96192429) . I expect to push them tomorrow. If you can wait until then to merge it's fine for me. :smiley: 
Of course, if it appears any merge conflict with master branch when you merge any PR I will try to solve it asap.

neoandroid:
Finally I've applied all the changes discussed in the comments, I hope you like them. I've done some e2e tests in my environment and haven't found bugs using the different supported scenarios but I encourage you to make your own verifications.

Since the parameter in `cluster.yml` is now called `subnets` instead of `privateSubnets` I thought it didn't make any sense to have two variables in the models that handle public and private subnets respectively. It was needlessly complicating things so currently there is only one `Subnets` variable in each model and it's populated with top-level defined  subnets if no subnets are defined under etcd//controller/worker.
I've also removed the remaining if/else blocks in cfn template due to private network topology.

mumoshu:
@neoandroid Thanks for all the hard works!
I've merged your work locally and started testing.
FYI, issues I've ever seen so far are only two.

(1) Node pools fail while creating when the topology isn't private, because of the missing import:

```
CREATE_FAILED AWS::CloudFormation::Stack kubeawstest6-nodepool1 No export named PublicRouteTable found
```

(2) and the missing subnets for the worker ASG:

```
CREATE_FAILED AWS::CloudFormation::Stack kubeawstest6-nodepool1 The following resource(s) failed to create: [Workers].
CREATE_FAILED AWS::AutoScaling::AutoScalingGroup Workers Property AvailabilityZones cannot be empty.
```

They were not that hard to fix. Please give me a day or two until it gets merged 👍

mumoshu:
Merged 😄
Thanks for your contribution @neoandroid!
Please feel free to submit another pull request(s) to add further fixes/improvements to this feature.

whereisaaron:
This looks fab @neoandroid! But does it remove the ability to deploy to private subnets without creating a NAT? We already provide non-AWS NAT for our VPCs, and I was using the add-security-group `cluster.yaml` option to add a Security Group that automatically provided access to NAT for instances created by `kube-aws` installs. Does this PR take away that use case? Or is there still a way to specify a private subnet without creating an AWS NAT? I just couldn't spot how to express that in the syntax above?

neoandroid:
No, that scenario is not possible. I didn't expect users where not using AWS NAT gateways. If you need that, you'll probably have to create your own subnets and NAT's and use the ability to reuse preexisting infrastructure #227

whereisaaron:
Well it was possible before, so this is a substantial breaking change for me if I can't work around it 😞. I  would be moving from the fully automated cluster deployment `kube-aws` provides now, back to semi-manual deployment.

I guess the workaround is to patch out the AWS NATs from the `stack.json`? So long as I can still nominate the Route Table and additional Security Group to be applied to private subnets it should be workable. Or maybe if I script/CF the creation of the subnets for each cluster first, just before `kube-aws up` will that avoid creating the unnecessary AWS NAT's?

The other possible problem with mandating people create AWS NAT's is that AWS imposes is a limit of only 5 NAT gateways per AZ. If every cluster deployed with `kube-aws` creates its own NAT instances, you'll quickly hit that limit.

Anyway, I don't want to rain on your parade 🎈 🎉 I'll do my best to make this new approach work for me too. There's always a way!

c-knowles:
I guess it's still possible for kube-aws to create the subnets but link to existing route tables? That's how I'm using kube-aws, at least on an older version. This does remind me a little of @redbaron's [comments](https://github.com/coreos/kube-aws/issues/277#issuecomment-275373813), perhaps we could discuss as a group how we can allow extensions so not every customisation is embedded deeply in kube-aws.

mumoshu:
@c-knowles I believe that I've already implemented a support for an existing route table per subnet in https://github.com/coreos/kube-aws/pull/284#issuecomment-275955785
Would you mind taking a look into my comment there and leave your feedback?

mumoshu:
Let me also add that as it isn't an general purpose aws provisioning tool, `kube-aws` won't automate all the necessary steps required in order to deploy your cluster into an existing vpc but will definitely try to support at least part of those steps.
Also, I'll try to support your existing use-case(s) as much as possible so please don't hesitate to leave your comments in #284!

neoandroid:
@whereisaaron  For your interest I inform you that you can request a limit increase for NAT gateways. 5 is the default limit but it's not a hard limit.

PR #174, opened 2016-12-16
Correct node pool command inconsistencies

c-knowles:
Corrects one of my comments from https://github.com/coreos/kube-aws/issues/44 now we actually have the node pool validation. I'm not sure we even need the `runCmdRender` method as isn't `kube-aws node-pool render` without any subcommand already deprecated? We shouldn't add deprecated things so if we want it to always be `kube-aws node-pool render stack` then I think we remove this part entirely.

### DIFF
--- a/Documentation/kubernetes-on-aws-node-pool.md
+++ b/Documentation/kubernetes-on-aws-node-pool.md
@@ -69,6 +69,14 @@ subnets:
     instanceCIDR: "10.0.2.0/24"
 ```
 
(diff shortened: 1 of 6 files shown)


mumoshu:
Thanks for the PR!

Yes, `kube-aws node-pool render` should not be runnable. I guess removing `RunE: runCmdRender,` results in that?

Also, would you mind fixing README.md and `Documentation/*.md` by replacing `kube-aws node-pools render` with `kube-aws node-pools render stack` if you find any?

c-knowles:
@mumoshu render for node pools was actually missing from the documentation so I added it.

mumoshu:
LGTM 👍 Thanks as always :bow:

Issue #111, opened 2016-11-30
kube-dns-v20 Number of replicas too low (1) for multi-node cluster availability.


jeremyd:
The number of replicas for kube-dns-v20 is set by default to 1 when setting up a new cluster.  This means that during a rolling-upgrade this service will go down multiple times while it gets re-scheduled to different nodes.  Upping the default replica count to 2, or 3 would be a better solution for people un-aware of this setting.  I'm hoping that the next time I spin up a cluster I won't have to remember to do this manually.  (config option or default > 1).

mumoshu:
Hi @jeremyd, thanks for your feedback 👍 

In the meantime, I believe we should set at least 2 replicas for kube-dns by default.
I'd appreciate a pull request for that if you had some time :bow:

In the future, bundling cluster-proportional-autoscaler into kube-aws would be even nicer.
However cluster-proportional-autoscaler does not seem to support AWS yet.
So it will take some time until this nicer solution gets implemented.

Relevant upstream issue: https://github.com/kubernetes/kubernetes/issues/37567

mumoshu:
This is now fixed in #178. Thanks for the feedback @jeremyd!

PR #178, opened 2016-12-19
DNS horizontal autoscaling

### LINKED ISSUE #172: Feature: DNS horizontal autoscaling
The upstream Kubernetes seems to have adapted to utilize cluster-proportional-autoscaler to horizontally auto-scale kube-dns pods since 1.5.
Would it be nice to support it out of box in kube-aws, too?

Relevant upstream issue: https://github.com/kubernetes/kubernetes/pull/33239

Setting min replicas to at least 2 would also result in fixing #111

mumoshu:
fix #172 
depends on #177

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -755,7 +755,7 @@ write_files:
                   - --configmap=kube-dns-autoscaler
                   - --mode=linear
                   - --target=Deployment/kube-dns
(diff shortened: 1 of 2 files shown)


mumoshu:
e2e passed & rebased.

mumoshu:
closes #111 as the min replicas for kube-dns is now 2

Issue #181, opened 2016-12-21
Suggestion: Move away from m3.* for default instance types


p80n:
It looks like AWS is deprecating the m3 instance types? They are no longer available in the new Ohio/us-east-2 region, and "kube-aws up" will fail unless you change out the m3.medium default to, for example, m4.large. They also appear to have removed them from their pricing information page.

Easy to fix - but could create frustration for people new to the tool.

Example output with defaults:
```shell
~>kube-aws up
Creating AWS resources. This should take around 5 minutes.
Error: Error creating cluster: Stack creation failed: CREATE_FAILED : The following resource(s) failed to create: [InstanceController]. 

Printing the most recent failed stack events:
CREATE_FAILED AWS::CloudFormation::Stack petset The following resource(s) failed to create: [InstanceController].
CREATE_FAILED AWS::EC2::Instance InstanceController The requested configuration is currently not supported. Please check the documentation for supported configurations.```

jgmize:
@metadave this looks like the issue you ran into when trying to spin up a cluster in us-east-2.

mumoshu:
@p80n Thanks for the info! I've never realized that there's already an AWS region which is missing `m3`s. I'll fix it shortly.

mumoshu:
@p80n @jgmize Excuse me but I need your additional feedback before fixing this.

[The official Kubernetes doc does explicitly recommend `m3` instance types](http://kubernetes.io/docs/getting-started-guides/aws/#supported-procedure-get-kube) as they provides us:
* Ephemeral Instance Storage
* Consistent CPU performance
* Enough memory to reliably host k8s control plane (on controller nodes)
(With the lowest pricing in the category which provides these points)

What's the minimum instance types of your choice for worker, controller, end etcd nodes when it comes to running a Kubernetes cluster reliably?

mumoshu:
Anyways, I'd like not to make inappropriate instance types to be the defaults regarding reliability of resulting k8s clusters. What is the most viable alternative to `m3.medium`?

FYI:
* [t2's are considered inappropriate in upstream](https://github.com/kubernetes/kubernetes/issues/21151), especially for master(=controller+etcd) nodes.
* t2.medium's are observed to be slightly better than `m3.medium`
  * Recommended reading: http://techblog.bonobos.com/ec2/sysadmin/devops/2014/10/02/the-m3.medium-is-terrible.html

For the time being, walking on the eggshells, I'd suggest `t2.large` as they would be much better than `m3.medium`.

[...]

PR #184, opened 2016-12-27
Move from the old m3.medium instance type to the marginally better t2.medium

mumoshu:
ref #181

### DIFF
--- a/Documentation/kubernetes-on-aws-node-pool.md
+++ b/Documentation/kubernetes-on-aws-node-pool.md
@@ -154,7 +154,7 @@ worker:
     targetCapacity: 5
     launchSpecifications:
     - weightedCapacity: 1
-      instanceType: m3.medium
+      instanceType: t2.medium
     - weightedCapacity: 2
       instanceType: m3.large
     - weightedCapacity: 2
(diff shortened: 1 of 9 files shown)


pieterlange:
-1 Lets not use `t2` for controllers by default - you should really know what you're doing when using this instance class.

Also, 4GB of memory seems a bit on the low side when you start actually using the cluster. My not keep the `m3.medium` and suggest to use `m4.large` or `t2.large`?

mumoshu:
@pieterlange Thanks for your feedback here!

I believe setting good defaults and moving away from the old `m3.medium` instance type is separate thing. So:

* Would you please [read my comment about that](https://github.com/coreos/kube-aws/issues/181#issuecomment-269297176)? To be more clear, It's true that I began to believe `t2.medium` is better than `m3.medium` but I'm not opposed to make `m4.large` or `t2.large` by default.

* Also, if you have any concrete advice on choosing which instance type(s) as default(s), I'd appreciate it if you could [read about my suggestions](https://github.com/coreos/kube-aws/issues/181#issuecomment-268705548) and reply with your suggestions on the instance type for etcd, controller, worker nodes 👍

mumoshu:
Changing the subject according to what we've discussed in https://github.com/coreos/kube-aws/issues/181#issuecomment-269371054

mumoshu:
Things seems to be left up in the air but I'm going to merge this to get the ball rolling :bow:

Let's discuss further about better default(s) in #181

PR #187, opened 2016-12-31
Node Pool Nested Stack Name

icereval:
Nest node pool stack name by cluster name to avoid conflict when running multiple clusters.

e.g.
```
cluster-1
cluster-1-pool-1
cluster-1-pool-2
```

```
cluster-2
cluster-2-pool-1
cluster-2-pool-2
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -693,7 +693,7 @@ type Config struct {
 
 // CloudFormation stack name which is unique in an AWS account.
 // This is intended to be used to reference stack name from cloud-config as the target of awscli or cfn-bootstrap-tools commands e.g. `cfn-init` and `cfn-signal`
-func (c Config) StackName() string {
+func (c Cluster) StackName() string {
 	return c.ClusterName
 }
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -693,7 +693,7 @@ type Config struct {
 
 // CloudFormation stack name which is unique in an AWS account.
 // This is intended to be used to reference stack name from cloud-config as the target of awscli or cfn-bootstrap-tools commands e.g. `cfn-init` and `cfn-signal`
-func (c Config) StackName() string {
+func (c Cluster) StackName() string {
 	return c.ClusterName
```
mumoshu: nit: Any specific reason to move this method from `Config` to `Cluster`? (I guess for now there's no specific reason/it doesn't relate to the main purpose of this PR but you just thought it might be better to do so?

icereval: Only that StackName was not available in the context of Cluster functions.

mumoshu: Got it! I somehow missed from where the `ClusterName` field is coming.

mumoshu:
E2E passed. LGTM 👍 
Thanks for your contribution @icereval

c-knowles:
@mumoshu do we want to also update the [name tags of subresources](https://github.com/coreos/kube-aws/blob/master/nodepool/config/templates/stack-template.json#L331)?

PR #188, opened 2016-12-31
Handle error from ReadOrCreateEncryptedTLSAssets gracefully.

andrejvanderzee:
Causes panic when ReadOrCreateEncryptedTLSAssets, for example when you enter a bad `kmsKeyArn`. 

Ever considered using Dave Cheney's error package? 
https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

### DIFF
--- a/config/tls_config.go
+++ b/config/tls_config.go
@@ -7,14 +7,15 @@ import (
 	"net"
 	"time"
 
+	"io/ioutil"
+	"path/filepath"
+
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


mumoshu:
@andrejvanderzee Thanks as always! LGTM 👍 

Regarding the which `error` package to introduce into kube-aws, I have not yet considered Dave Cheney's one. FYI, reqs for an `error` package for kube-aws in my mind are supports for (1) nested errors and (2) stack traces for each nested error, and my preference so far is `juju/errors`.  I'm open to suggestion(s) though.

PR #200, opened 2017-01-05
Point to the node-pool documentation

ankon:
(aws-experimental-features) is a dead link, but in this case the intended target seems to be the node-pool documentation.


mumoshu:
Good catch! Thanks again :bow:

PR #201, opened 2017-01-05
Fix the SAN for non-us-east-1 AWS regions

ankon:
See http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html#vpc-dns-hostnames


mumoshu:
Good catch! Thanks for your contribution @ankon 👍 

For anyone interested, kube-aws does utilize these SANs in https://github.com/coreos/kube-aws/blob/master/config/config.go#L541-L547
This documentation fix will save users who wants to bring their own certs.

PR #209, opened 2017-01-06
Add customSettings field

redbaron:
This feature is useful to experiment with new setups, have in house customizations etc without resorting to full '--export', so config still remains manageable and flexible.

Example use in stack-template.json:
```
{{ .CfgMap.myCustomOpt }}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -123,6 +123,7 @@ func NewDefaultCluster() *Cluster {
 		TLSCertDurationDays: 365,
 		CreateRecordSet:     false,
 		RecordSetTTL:        300,
+		CustomSettings:      make(map[string]interface{}),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mumoshu:
@redbaron Thanks for the PR!

If I understand correctly, this assumes `cluster.yaml` to allow arbitrary keys in it, right?
In the future, I'd rather like to make `cluster.yaml` be validated more strictly; unknown top-levels keys will be invalid then.
So, would you mind adding `userCustomSettings` top-level key to `cluster.yaml` instead while referring the a custom key in stack-template.json like `{{ .UserCustomSettings.myCustomOpt }}`?

redbaron:
@mumoshu done changes you suggested, would you mind to have a look?

mumoshu:
@redbaron Thanks for your help!
Sorry for the back and forth but would you mind changing it to `customSettings`? It turns out that the former `userCustomSettings` seems a bit non-english-like after googling.
If you are ok, I can also do it on my side after merging your PR.

redbaron:
no problem, updated.

mumoshu:
LGTM & E2E passed.
Thanks for your contribution @redbaron!

PR #210, opened 2017-01-06
document different account may have different supported AZs and solutions

helinwang:
Fixes https://github.com/coreos/kube-aws/issues/206


mumoshu:
Thanks for the PR but this is not correct! Please see my comment in the original issue https://github.com/coreos/kube-aws/issues/206#issuecomment-270986778

helinwang:
@mumoshu Please see updated.

mumoshu:
LGTM 👍 Thanks for your contribution @helinwang!

Issue #215, opened 2017-01-08
taint-and-uncordon task continuously fails on controller nodes due to typo in taint


whereisaaron:
kube-aws version v0.9.3-rc.2

I was surprised that the controllers created by kube-aws 0.9.3-rc.2 were neither cordoned nor tainted. Checking the controller I found the `taint-and-uncordon` task was repeatedly failing due to a typo in the taint.

The attempted taint is `node.alpha.kubernetes.io/role:master:NoSchedule`
but should be `node.alpha.kubernetes.io/role=master:NoSchedule`. This results in the error:

```
Jan 08 05:02:27 ip-1.1.1.1.region.compute.internal taint-and-uncordon[9184]: [  579.594914] hyperkube[5]: error: at least one taint update is required
```

A second problem with this task, is that even though the taint attempt fails and returns an error, the task does not check for errors/success and goes ahead an uncordons the controller anyway. Since the controller is now untainted and uncordoned it starts being scheduled for regular/user pods.

```
run --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true --mount=volume=kube,target=/etc/kubernetes --uuid-file-save=/var/run/coreos/taint-and-uncordon.uuid --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf --net=host --trust-keys-from-https quay.io/coreos/hyperkube:v1.5.1_coreos.0 --exec=/bin/bash -- -vxc echo tainting this node; \
                                                                                    hostname="ip-172-23-146-158.ap-southeast-2.compute.internal"; \
                                                                                    kubectl="/kubectl --server=http://127.0.0.1:8080"; \
                                                                                    taint="$kubectl taint node $hostname"; \
                                                                                    $taint "node.alpha.kubernetes.io/role:master:NoSchedule"; \
                                                                                    echo done. ;\
                                                                                    echo uncordoning this node; \
                                                                                    $kubectl uncordon $hostname;\
                                                                                    echo done.
```

whereisaaron:
I manually tainted the node but that doesn't help because the task doesn't check if the taint is already there or `--overwrite` it.

Over time I get more and more of these 'No such container' messages building up on the controller. About every 60 seconds there is a whole bunch of them dumped in the log.

```
[...]

PR #199, opened 2017-01-04
Controller node not being properly tainted

artushin:
/opt/bin/taint-and-uncordon had a syntax issue with the kubectl taint command. Was failing with `error: at least one taint update is required` and the controller node was still getting pods scheduled to it. PR fixes the syntax as per http://kubernetes.io/docs/user-guide/kubectl/kubectl_taint/

PR #216, opened 2017-01-08
Correct typo in controller/master taint

whereisaaron:
Taint should be "node.alpha.kubernetes.io/role=master:NoSchedule" (with an equals). Ref issue #215.


whereisaaron:
D'uh, this is a duplicate of pull request #199 and doesn't seem to fully resolve the problem anyway. Closing it.

PR #217, opened 2017-01-08
Bash improvements

redbaron:
Set of changes to for cleaner and more correct bash scripts.

Mainly changes are about adding `-e` to make startup scripts exit on errors encountered, which would prevent things like https://github.com/coreos/kube-aws/issues/215#issuecomment-271166798 and also some cosmetics where `; \` noise is removed where possible

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -484,13 +484,13 @@ write_files:
         --trust-keys-from-https \
         --insecure-options=ondisk \
         {{.AWSCliImageRepo}}:{{.AWSCliTag}} --exec=/bin/bash -- \
-          -vxc \
-          'echo tagging this spot instance; \
-           instance_id="'$instance_id'"; \
+          -vxec \
+          'echo tagging this spot instance
+           instance_id="'$instance_id'"
            /usr/bin/aws \
              --region {{.Region}} ec2 create-tags \
              --resource $instance_id \
-             --tags '"'"'Key=KubernetesCluster,Value="{{.ClusterName}}"'"'"' '"'"'Key=Name,Value="{{.StackName}}-kube-aws-worker"'"'"' '"'"'Key="kube-aws:node-pool:name",Value="{{.NodePoolName}}"'"'"'; \
+             --tags '"'"'Key=KubernetesCluster,Value="{{.ClusterName}}"'"'"' '"'"'Key=Name,Value="{{.StackName}}-kube-aws-worker"'"'"' '"'"'Key="kube-aws:node-pool:name",Value="{{.NodePoolName}}"'"'"'
            echo done.'
 
       sudo rkt rm --uuid-file=/var/run/coreos/tag-spot-instance.uuid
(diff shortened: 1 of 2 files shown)


redbaron:
Few things which I didn't change, but looked strange to me:
 - seemingly unnecessary `sudo rkt` calls from script which already run as root as they are called from systemd service files.
 - etcd decrypts assets differently from how it is done for controller and workers
 - shouldn't it be an error if `*.pem.enc` files are not found? current behavior which I kept is to ignore their absence
 - `install-kube-system` script uploads multiple manifests to k8s all in one script,and in case of  an error in one of them systemd will restart `install-kube-system.service`, which executes script again and triggers re-upload of  previously uploaded manifests even those which were successful, how k8s react to it? Is it safe? will it create one more copy of a resource or it will have no effect?

config/templates/cloud-config-worker
```diff
@@ -466,8 +465,6 @@ write_files:
            done;
            echo done.'
 
-      sudo rkt rm --uuid-file=/var/run/coreos/decrypt-tls-assets.uuid
```
mumoshu: Could you share the context behind this? AFAIK, removing explicit `rkt rm`s just means to "defer" removals of rkt pods until `rkt gc` is run. Then, could there be a specific reason to do so?

mumoshu:
> etcd decrypts assets differently from how it is done for controller and workers

@redbaron Thanks for pointing it out. Yes, and as it relates to https://github.com/coreos/kube-aws/issues/218, I'd like `cloud-config-etcd` to be updated similarly to the one for controllers/workers.

mumoshu:
> shouldn't it be an error if *.pem.enc files are not found? current behavior which I kept is to ignore their absence

@redbaron Good catch! Yes, I agree to you there. I'd like decrypt-tls-assets to emit errors when there're missing enc files.

config/templates/cloud-config-worker
redbaron: no particular reason, feels more right that way:
 - there were errors reported regarding `rkt rm` erroring
 - rkt-gc is already there,let it do its job

redbaron:
> Thanks for pointing it out. Yes, and as it relates to #218, I'd like cloud-config-etcd to be updated similarly to the one for controllers/workers.

I have a big change to the way etcd are provisioned, it is updated there. I didn't finish testing it, but can submit WIP if want to have a look

config/templates/cloud-config-worker
mumoshu: @redbaron Sorry missed the context on my side this time. I'd like to leave `rkt rm` as many as possible for system pods like decrypt-tls-assets to provide us chances to immediately notice issues like https://github.com/coreos/kube-aws/pull/199#issuecomment-271000614. Getting rid of `rkt rm`s does defer but doesn't fix an issue like that.

I'm ok with `rkt gc` doing their jobs not on system pods(created via kube-aws) but rather user pods.

mumoshu:
@redbaron Yes, I'd greatly appreciate your PR and am ready to review.

redbaron:
restored `rkt rm` but ignoring errors from it

mumoshu:
@redbaron Thanks! Could you also resolve the conflicts? I guess `taint-and-uncordon` should be conflicted with #199 at least.

mumoshu:
LGTM & E2E passed. Thanks as always @redbaron 👍 :bow:

Issue #214, opened 2017-01-08
Generated ELB Security Group caused API to be inaccessible due to blocking ICMP


whereisaaron:
kube-aws version v0.9.3-rc.2

After bringing up the cluster I could not contact the API server, the error in the controller log was:
```
API server.http: TLS handshake error from 1.1.1.1::12345: EOF
```
After investigating the culprit was the Security Group for ElbAPIServer that `kube-aws` creates. It blocks ICMP and so breaks TCP/IP. The client was sending an ICMP packet to negotiate the MTU but the ELB was blocking it, preventing a successful HTTPS connection.
```
ICMP ip-2.2.2.2.ap-somewhere.compute.internal unreachable - need to frag (mtu 1406), length 556
```
After adding ICMP to the ELB Security Group I could contact the API server no problem.

The problem is here in the stack template, ICMP needs to be added.
```
    "SecurityGroupElbAPIServer" : {
      "Properties": {
        "GroupDescription": {
          "Ref": "AWS::StackName"
        },
        "SecurityGroupIngress": [
          {
            "CidrIp": "0.0.0.0/0",
            "FromPort": 443,
            "IpProtocol": "tcp",
            "ToPort": 443
          }
        ],
...
```

whereisaaron:
I applied this patch to my `stack-template.json` to add ICMP to the API ELB and now the API is accessible as soon as the controller is up and running.

```
*** stack-template.json-orig	2017-01-07 19:52:41.597408095 -0500
--- stack-template.json	2017-01-07 19:53:52.577408095 -0500
***************
*** 628,633 ****
--- 628,639 ----
              "FromPort": 443,
              "IpProtocol": "tcp",
              "ToPort": 443
+           },
+           {
+             "CidrIp": "0.0.0.0/0",
+             "FromPort": -1,
+             "IpProtocol": "icmp",
+             "ToPort": -1
            }
          ],
          "Tags": [
```

redbaron:
Thanks for looking into it! Would you mind to prepare PR? To enable just MTU discovery, this rule will be sufficient:

```
           {
             "CidrIp": "0.0.0.0/0",
             "FromPort": 3,
             "IpProtocol": "icmp",
             "ToPort": 4
}
```

whereisaaron:
Thanks @redbaron I have prepared a PR #220 to unblock ICMP. I unblocked ICMP in general, as I don't see a need to block the other benefits of ICMP like ping or flow control.

[...]

PR #220, opened 2017-01-09
Don't block ICMP for API ELB

whereisaaron:
As discussed in issue #214, stop blocking ICMP for the API ELB. Among other things ICMP is required for MTU discovery to for TCP / HTTPS connections with the API server.


mumoshu:
LGTM & E2E passed.
Thanks as always @whereisaaron!

PR #221, opened 2017-01-09
Setup empty labels in controller

gianrubio:
This labels are important for service auto discovery on [kube-prometheus](https://github.com/coreos/kube-prometheus/tree/master/manifests/k8s/self-hosted). I'm following the exactly same pattern on kops.


mumoshu:
@gianrubio LGTM 👍  Thanks as always!

Issue #222, opened 2017-01-09
kubectl logs not working


mgilbir:
I launched a cluster with `kube-aws v0.9.3-rc.2`, then launched a pod and when I tried to get the logs via kubectl or the dashboard I get:
```
$ kubectl logs nats-d8twb
Error from server (InternalError): an error on the server ("unknown") has prevented the request from succeeding (get pods nats-d8twb)
```

I have another cluster not launched with `kube-aws` that works fine, which makes me think that the local setup is ok. `kubectl` works fine to list nodes and pods in the cluster, so it is not a credentials nor certificates problem either.

I'm a bit at a loss of ideas on where to start looking to fix this. Have you ever seen this? Is `kubectl logs` working for you?

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-22T13:59:22Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1+coreos.0", GitCommit:"cc65f5321f9230bf9a3fa171155c1213d6e3480e", GitTreeState:"clean", BuildDate:"2016-12-14T04:08:28Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}
```

mgilbir:
Just found an open issue about it on the k8s repository: https://github.com/kubernetes/kubernetes/issues/22770

mgilbir:
The fix outlined in the issue worked.

@mumoshu If it doesn't have adverse effects to other users, would you be open to adding `--kubelet-preferred-address-types=InternalIP,Hostname,ExternalIP` in the apiserver setup in `userdata/cloud-config-controller`?

redbaron:
prepare PR, I am sure it will be merged quicker that way

mumoshu:
Thanks a lot for your contribution here :bow:
I'm closing this as the PR adressing this issue is merged but please feel free to reopen if the original issue still persists.

PR #223, opened 2017-01-09
Fix kubectl logs problem due to apiserver config

mgilbir:
Fixes issue #222 with the workaround proposed in https://github.com/kubernetes/kubernetes/issues/22770#issuecomment-266215007


mumoshu:
LGTM & E2E passed.
Thanks for your contribution @mgilbir!

Issue #189, opened 2017-01-01
Existing VPC with custom DHCP Option Set - etcd cluster won't start


rbellamy:
## Overview

Existing VPC in us-west-1 with custom DHCP Option Set - etcd cluster won't start. See https://github.com/coreos/bugs/issues/1272.

DHCP Option Set:
```
domain-name = terradatum.com
domain-name-servers = 10.1.0.2
```

This causes the systemd specifier `%H` (used by the `etcd2.service` unit) to return `ip-10-1-Y-Z.terradatum.com` rather than what kube-aws expects.

## TLDR;
Skip to the bottom to see the currently viable workaround.

Ideally, kube-aws would support this out-of-the-box. I find it hard to believe that we're the only folk that are trying to get kubernetes running in an existing VPC with a custom DHCP Option Set.

Finally, I'm very concerned about the move away from coreos-cloudinit, given that we've only now realized a working environment - what's the likelihood of [Ignition](https://coreos.com/ignition/docs/latest/) and/or [coreos-metadata](https://github.com/coreos/coreos-metadata) addressing what is obviously considered an edge condition by the CoreOS + kube-aws teams?

## Detail

**List of etcd cluster instances**
```
Environment=ETCD_INITIAL_CLUSTER=`.EtcdInitialCluster`
```

Generated by `config.go` lines [542](https://github.com/coreos/kube-aws/blob/639829cc3e0beef04b825f75dbef4ffdceac239a/config/config.go#L542), [549](https://github.com/coreos/kube-aws/blob/639829cc3e0beef04b825f75dbef4ffdceac239a/config/config.go#L549) and finally line [558](https://github.com/coreos/kube-aws/blob/639829cc3e0beef04b825f75dbef4ffdceac239a/config/config.go#L558)

You can see that these values are hard coded for either `ec2.internal` (us-east-1) or `<region>.compute.internal` (everywhere else). So no, it's not a viable option for us to expect to use the `terradatum.com` suffix for our `kube-aws`-controlled cluster.

And of course, the machinery that sets the hostname is tightly coupled with the EC2 launch configuration and is supplied via calls to the EC2 instance metadata at http://169.254.169.254/latest/meta-data/hostname.

**Before altering the hostname**
```
$ hostname
ip-10-1-10-4.terradatum.com
$ hostname -s
ip-10-1-10-4
$ hostname -f
ip-10-1-10-4.terradatum.com
$ hostname -d
terradatum.com
```

[...]

PR #226, opened 2017-01-10
Don't precalculate etcd static IP addresses

redbaron:
As part of effort to add support for provisioning into existing VPCs and subnets, kube-aws should stop trying to calculate static IPs itself: it is likely to cause conflicts for a first cluster in subnet and it is guaranteed not to work when provisioning second one.

This PR works by changing CF to save IP addresses  of InstanceEtcd in CF metadata which is then pulled by `cfn-init` scripts on nodes and placed in the right places.

It also adds explicit `AWS::EC2::NetworkInterface` resource which is then used as eth0 on Etcd instances.  Strictly speaking this change is not needed, but since `kube-aws`  decided not to manage Etcd instances after creation, in the real world they might need changes - having an explicit network interface resource allows instance to be recreated keeping same static IP.

Also few cleanups were done to etcd controller scripts.

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -137,7 +137,7 @@ coreos:
                        --cert-file /etc/kubernetes/ssl/etcd-client.pem \
                        --endpoints "${ETCD_ENDPOINTS}" \
                        cluster-health
-        ExecStartPre=/bin/bas -ec "shopts -s nullglob; sed -i 's|#ETCD_ENDPOINTS#|${ETCD_ENDPOINTS}|' /etc/kubernetes/manifests/* /srv/kubernetes/manifests/*"
+        ExecStartPre=/bin/bash -ec "shopt -s nullglob; sed -i 's|#ETCD_ENDPOINTS#|${ETCD_ENDPOINTS}|' /etc/kubernetes/manifests/* /srv/kubernetes/manifests/*"
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --api-servers=http://localhost:8080 \
         --cni-conf-dir=/etc/kubernetes/cni/net.d \
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -19,9 +19,9 @@ coreos:
         After=network-online.target
 
         [Service]
-        Type=oneshot
+        Restart=on-failure
         RemainAfterExit=true
-        ExecStart=/usr/bin/rkt run \
+        ExecStartPre=/usr/bin/rkt run \
           --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
           --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos \
           --net=host \
@@ -80,13 +80,14 @@ coreos:
         - name: 10-etcd.conf
           content: |
             [Unit]
-            Requires=cfn-etcd-environment.service
+            Wants=cfn-etcd-environment.service
             After=cfn-etcd-environment.service
 
             [Service]
-            EnvironmentFile=/etc/etcd-environment
+            EnvironmentFile=-/etc/etcd-environment
             Environment="ETCD_SSL_DIR=/etc/kubernetes/ssl"
             EnvironmentFile=-/run/flannel/etcd-endpoints.opts
+            ExecStartPre=/usr/bin/systemctl is-active cfn-etcd-environment.service
             ExecStartPre=/bin/sh -ec "echo FLANNELD_ETCD_ENDPOINTS=${ETCD_ENDPOINTS} >/run/flannel/etcd-endpoints.opts"
             ExecStartPre=/opt/bin/decrypt-tls-assets
             ExecStartPre=/usr/bin/etcdctl \
@@ -102,11 +103,10 @@ coreos:
       runtime: true
       content: |
         [Unit]
-        Wants=flanneld.service
-        Requires=cfn-etcd-environment.service
+        Wants=flanneld.service cfn-etcd-environment.service
         After=cfn-etcd-environment.service
         [Service]
-        EnvironmentFile=/etc/etcd-environment
+        EnvironmentFile=-/etc/etcd-environment
         Environment=KUBELET_VERSION={{.K8sVer}}
         Environment=KUBELET_ACI={{.HyperkubeImageRepo}}
         Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf \
@@ -127,6 +127,7 @@ coreos:
         --volume cni-bin,kind=host,source=/opt/cni/bin \
         --mount volume=cni-bin,target=/opt/cni/bin{{ end }}"
         ExecStartPre=/usr/bin/systemctl is-active flanneld.service
+        ExecStartPre=/usr/bin/systemctl is-active cfn-etcd-environment.service
         ExecStartPre=/usr/bin/mkdir -p /var/lib/cni
         ExecStartPre=/usr/bin/mkdir -p /var/log/containers
         ExecStartPre=/usr/bin/mkdir -p /opt/cni/bin
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -4,12 +4,35 @@ coreos:
     reboot-strategy: "off"
   flannel:
     interface: $private_ipv4
-    etcd_endpoints: {{ .EtcdEndpoints }}
     etcd_cafile: /etc/kubernetes/ssl/ca.pem
     etcd_certfile: /etc/kubernetes/ssl/etcd-client.pem
     etcd_keyfile: /etc/kubernetes/ssl/etcd-client-key.pem
 
   units:
+    - name: cfn-etcd-environment.service
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Fetches etcd static IP addresses list from CF
+        After=network-online.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/usr/bin/rkt run \
+          --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
+          --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos \
+          --net=host \
+          --trust-keys-from-https \
+          {{.AWSCliImageRepo}}:{{.AWSCliTag}} -- cfn-init -v \
+              -c "etcd-client" \
+              --region {{.Region}} \
+              --resource LaunchConfigurationController \
+              --stack {{ .ClusterName }}
+        ExecStart=/usr/bin/mv -f /var/run/coreos/etcd-environment /etc/etcd-environment
+
 {{if .Experimental.AwsEnvironment.Enabled}}
     - name: set-aws-environment.service
       enable: true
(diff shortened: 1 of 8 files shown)


redbaron:
Also fixes https://github.com/coreos/kube-aws/issues/218

mumoshu:
@redbaron Thanks again for your contribution!

Explicit `AWS::EC2::NetworkInterface` sounds cool but would mind sharing your thoughts on how an EC2 instance running etcd could be recreated retaining the same private dns name before/after recreation in a cluster created by kube-aws with the change?

redbaron:
So, kube-aws is going to ignore any resources with pattern `InstanceEtcd*` copying them from existing CF fetched from AWS into a new one, which it then uploads back to AWS to apply changes.

This means that any modification to InstanceEtcd is outside of scope of kube-aws and administrator can manually do --export, then make changes needed and then apply them calling CF directly.

explicit `AWS::EC2::NetworkInterface` helps here because as long as it is not destroyed, but simply 
reattached, it keeps it's private IP assigned at creation time, which itself means that `PrivateDNSName` property of `AWS::EC2::Instance` this interface is attached to remains the same even if instance itself (but not the interface!) is recreated.

redbaron:
This has been tested and seems to be working fine. Could you please review this change. I'll squash commits before merge

config/templates/cloud-config-controller
```diff
@@ -4,12 +4,35 @@ coreos:
...
+      enable: true
+      command: start
+      runtime: true
+      content: |
+        [Unit]
+        Description=Fetches etcd static IP addresses list from CF
+        After=network-online.target
+
+        [Service]
+        Type=oneshot
```
mumoshu: You may already know it but according to https://github.com/systemd/systemd/issues/2582, I'd rather like to make this `Type=simple` or implement a loop in bash to retry `cfn-int` on failure for reliability

config/templates/cloud-config-controller
```diff
@@ -56,14 +79,21 @@ coreos:
       drop-ins:
         - name: 10-etcd.conf
           content: |
+            [Unit]
+            Requires=cfn-etcd-environment.service
```
mumoshu: This should be changed according to how you'd approach for https://github.com/coreos/kube-aws/pull/226/files#r96115270.
* Making `cfn-etcd-environment.service` to have `Type=simple` or `Type=oneshot` without implementing retries in bash while having `Requires=cfn-etcd-environment.service` here would permanently keep `flanneld.service` failed once `cfn-etcd-environment.service` failed.
   * So I'd suggest either (1) making `cfn-etcd-environment.service` a `Type=simple` retried by systemd while making `flanneld.service` not to have `Requires=cfn-etcd-environment.service` but to have `Wants=cfn-etcd-environment.service` + `After=cfn-etcd-environment.service` + `ExecStartPre=/usr/bin/systemctl is-active cfn-etcd-environment.service` or (2) making `cfn-etcd-environment.service` a `Type=simple` or `Type=oneshot` retried in bash. AFAIK either (1) or (2) allows `flanneld` to be periodically retried until `cfn-etcd-environment` succeeds. (1) is recommended just because it is used all over the places in `cloud-config-*`.

config/templates/cloud-config-worker
```diff
@@ -31,7 +54,14 @@ coreos:
       drop-ins:
         - name: 10-etcd.conf
           content: |
+            [Unit]
+            Requires=cfn-etcd-environment.service
```
mumoshu: The same as https://github.com/coreos/kube-aws/pull/226/files#r96115418

redbaron:
@mumoshu , all that `cfn-etcd-environment.service` + `ExecStartPre="sed ... "" + `AWS::CloudFormation::Init` might go away if only we could create a private `.kube.local` Route53 zone  where `A` record are created:

```
etcd.clusername.kube.local   7200 IN A <ENI1.IP>
etcd.clusername.kube.local   7200 IN A <ENI2.IP>
etcd.clusername.kube.local   7200 IN A <ENI3.IP>
```

then it will be less crufty to distribute etcd endpoints across the clusters:
  - cloud-config-etcd:   will do something like `ETCD_INITIAL_CLUSTER=$(paste -d, -s <(dig ya.ru +short | xargs -I{} echo '{}=https://{}:PORT'))`
 - everyone else will use a single hardcoded endpoint `https://etcd.clusername.kube.local:PORT`. 

But having a route53 zone is a bigger change from what we have now, but if you think it is OK to go that route, I'd prefer to use it instead of my current approach

mumoshu:
@redbaron Would it be a solution to https://github.com/coreos/kube-aws/issues/189 if you could make the hosted zone used for etcd peer discovery customizable via `cluster.yaml`? (I'm assuming that the hosted zone can be anything if nameservers serving the zone is properly configured for EC2 instances via e.g. DHCP options. Am I correct?)

redbaron:
yes, hosted zone can be configurable, we can possibly reuse same zone which `externalDNSName` is using  by default if `createRecordSet: true`, but there will be no fallback to manually creating entries for etcd.

So if we go that route, then having a Route53 zone becomes a hard requirement, not an optional feature, but what we have instead is simpler and less error-prone etcd bootstrapping.

As for #189, setup described there is broken: if you set `domain-name` and keep using AWS DNS  you are asking for trouble, they even have it in their docs:

>  To change these to custom DNS hostnames, you must set domain-name-servers to a custom DNS server.

Trying to workaround unresolvable hostnames isn't something I am interested in :)

config/templates/cloud-config-controller
```diff
@@ -80,13 +80,14 @@ coreos:
         - name: 10-etcd.conf
           content: |
             [Unit]
-            Requires=cfn-etcd-environment.service
+            Wants=cfn-etcd-environment.service
             After=cfn-etcd-environment.service
 
             [Service]
-            EnvironmentFile=/etc/etcd-environment
+            EnvironmentFile=-/etc/etcd-environment
```
mumoshu: Couldn't find the documentation for it but does the dash(-) is meant to instruct this unit to be started even when the environment file doesn't exist?

mumoshu:
@redbaron Thanks!

> we can possibly reuse same zone which `externalDNSName` is using  by default if `createRecordSet: true`

I'm ok if you meant to automatically create a private hosted zone used to resolve etcd hosts to private ips used among etcd nods and the zone is named after what was specified by `externalDNSName` 👍 
assuming that `hostedZoneId` is currently for specifying "public" hosted zone used to host k8s API endpoint and not for resolving hosts to private ips.

> there will be no fallback to manually creating entries for etcd.

I'm ok with it as long as we document it.

Anyways we can enhance kube-aws later to support accept custom hostnames which results in record-sets in the private hosted zone with something like the below?

```yaml
internalDomainName: terradatum.com
etcd:
  hosts:
  - fqdn: foo.terradatum.com
  - fqdn: bar.terradatum.com
  - fqdn: baz.teradatum.com
```

> As for #189, setup described there is broken: if you set `domain-name` and keep using AWS DNS  you are asking for trouble, they even have it in their docs:

Honestly saying I'm not very familiar with private DNS thing in EC2 but AFAIK a custom `domain-name` in a DHCP options set while pointing to `10.1.0.2`(=The ip address of amazon-provided DNS server) seems valid as long as one correctly created/configured a private hosted zone for the domain for the VPC according to the AWS doc?

> The Amazon DNS server in your VPC is used to resolve the DNS domain names that you specify in a private hosted zone in Amazon Route 53
> http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_DHCP_Options.html#AmazonDNS

Would you let me ask which part of the setup described in #189 looked problematic to you?

redbaron:
> Honestly saying I'm not very familiar with private DNS thing in EC2 but AFAIK a custom domain-name in a DHCP options set while pointing to 10.1.0.2(=The ip address of amazon-provided DNS server) seems valid as long as one correctly created/configured a private hosted zone for the domain for the VPC according to the AWS doc?

My understanding is that only domain names automatically maintained by AWS on instance creation are `compute.internal` and `compute.amazonaws.com`, everything else needs to be managed by AWS user. I just done a quick test to confirm that.

>Would you let me ask which part of the setup described in #189 looked problematic to you?

Unless I misread description, they have situation where hosts receive names which can't be resolved via DNS, ie `dig ip-10-1-0-171.terradatum.com` on a node returns an error. You can't run system reliable where hostnames are unresolvable.

redbaron:
> I'm ok with it as long as we document it.
> Anyways we can enhance kube-aws later to support accept custom hostnames which results in record-sets in the private hosted zone with something like the below?

I'd add just these 2 params to `cluster.yaml`:

```
# DNS zone name to use when creating entries needed for cluster setup
# internalZoneApex: "{{ .clusterName }}.kube-aws.local"

# Uncomment and set to an existing route53 zone id which is capable of serving 'internalZoneApex' domain names. kube-aws will create one itself otherwise. If you use existing 'vpcId', newly created route53 zone will be associated with it.
# internalZoneId: 
```

Whatever in kube-aws is going to use it will figure out itself what to add. 

In case of etcd that would be just one DNS name with  `A` records added: `etcd.{{.InternalZoneApex}}`

mumoshu:
> My understanding is that only domain names automatically maintained by AWS on instance creation are `compute.internal` and `compute.amazonaws.com`, everything else needs to be managed by AWS user. I just done a quick test to confirm that.

Thanks! It matches to my expectation and therefore I'd guess one needs to either
* add a record-set for a host before the host starts up (e.g. via cloudformation like I'm assuming you would do to implement https://github.com/coreos/kube-aws/pull/226#issuecomment-272630759) or
* add a recordset  after the host starts up (e.g. via cloud-config and accordingly override hostname to match the record-set)

ref https://cantina.co/automated-dns-for-aws-instances-using-route-53/

> Unless I misread description, they have situation where hosts receive names which can't be resolved via DNS, ie `dig ip-10-1-0-171.terradatum.com` on a node returns an error. You can't run system reliable where hostnames are unresolvable.

It seems like I had completely missed that part. I agree to you now. Thanks!

mumoshu:
@redbaron Sounds great overall 👍 

Several nits but:
* `internalZoneId: ` should be `internalHostedZoneId` for consistency with the existing `hostedZoneId` (which could be deprecated and renamed to `externalHostedZoneId` in the future)
* AFAIK `"{{ .clusterName }}.kube-aws.local` isn't a zone apex but `kube-aws.local` is and I guess you don't want to restrict it to a zone apex actually. Then, shouldn't the key be `internalDomainName` rather than `internalZoneApex`?

Also, sorry for not being clear enough but my snippet below was meant to support users who wanted to manually name etcd nodes:

```yaml
internalDomainName: terradatum.com
etcd:
  hosts:
  - fqdn: foo.terradatum.com
  - fqdn: bar.terradatum.com
  - fqdn: baz.teradatum.com
```

I believe we don't need to care this now.

mumoshu:
> So if we go that route, then having a Route53 zone becomes a hard requirement, not an optional feature, but what we have instead is simpler and less error-prone etcd bootstrapping.

@redbaron Now I'm glad to make it a hard-req if we can choose to either automatically create one according to `internalDomainName` or provide an existing one via `internalHostedZoneId`. // I'm afraid i might be a time-consuming task though 🍺

mumoshu:
Btw how are you going to enable TLS communication between etcd nodes?

AFAICS, https://github.com/coreos/kube-aws/pull/226#issuecomment-272630759 is meant for etcd peer discovery among etcd nodes + etcd node discovery from client-side; not for etcd peer communication.
I believe you'd need to provide a resolvable hostnames to etcd nodes for etcd peer communication in TLS. Are you going to assume that `domain-name` in a DHCP options are not customized i.e. relying on the default `*.compute.internal` hostnames?
Then, could it be possible to enhance your work to support customized hostnames without many changes? (Just curious. Not meant to scope-creep this PR if it isn't already intended on your side)

redbaron:
@mumoshu , this PR works as of now (nodepools remain untested),  I please consider merging it. DNS etcd endpoint discovery can be a further improvement in  a separate PR. I can't keep rebasing PRs for long, their number grows out of hands :(

mumoshu:
@redbaron ok. I'm merging this anyway!

For anyone interested, AFAICS merging this should make node pools unusable because `etcd-client-env` isn't defined in `nodepool/config/stack-template.json` and private ips of etcd nodes which are required to define `etcd-client-env` isn't provided via cross-stack reference.
I'll fix it after merging several PRs including this but in the meantime if you'd like to use node pools, please consider not to use binaries built from master.
Or even better, submit pull request to fix it! 😄

Issue #52, opened 2016-11-10
Creating cluster with an existing subnet


sdouche:
Hello,
We can specify an existing VPC, but not an existing subnet. At work, I can't create network stuff. I have forked the file `stack-template.json` but It would be cool to have the option built-in.

Thanks.

pieterlange:
Provisioning in existing subnets creates a number of weird edge cases that i do not feel like supporting.

This request keeps showing up though, so out of interest, can you clarify why it's not sufficient to use existing routetables and attaching those to the new subnets? A little bit more background info on what you're trying to do would also help to see why you need this feature.

> At work, I can't create network stuff.
Do you mean in AWS, or at your office?

sdouche:
Hi @pieterlange. Thanks for your response. I don't understand:
> Provisioning in existing subnets creates a number of weird edge cases

What do you talking about? Have you some examples. I use existing subnet w/o issues.

> why it's not sufficient to use existing routetables and attaching those to the new subnets?

I can't create subnets into the AWS account of my company.

mumoshu:
@sdouche Hi, would you mind letting me summarize!

* We're just recommending not to mix up things coming from kube-aws and others, into the same subnets.
* If you meant that your colleague(s) created subnets *dedicated* to kube-aws resources(including ec2, elb etc), it would work.
* AFAIK, the `weird edge cases` @pieterlange mentioned happen only when you *mix up* kube-aws things with things.

c-knowles:
I wanted to share my experience of shared subnets. Having encountered a slight different use case but same end result of wanting to share network resources I did a [pull request here](https://github.com/coreos/coreos-kubernetes/pull/671) for shared subnets. I ended up [replacing that](https://github.com/coreos/coreos-kubernetes/pull/716) with shared route tables.

The main reason I ditched shared subnets is that k8s and kube-aws rely on some aspects of the subnets and to a lesser extent the route tables which meant that spinning up a new cluster was a little fragile. It's definitely related to what @mumoshu says above - dedicated subnets are ok but mixing them tends to case issues. I think it's definitely achievable still though with subnets, I chose the path of slightly lesser resistance. There would have to be details instructions on exactly what subnet and route table requirements there are across k8s and kube-aws.

pieterlange:
One of the side effects of not having the subnets managed by your deployment tool: https://github.com/kubernetes/kubernetes/issues/29298

[...]

PR #195, opened 2017-01-03
[WIP] Direct & Indirect CloudFormation Stack Parameters

icereval:
Based on https://github.com/coreos/kube-aws/pull/169.

@mumoshu, would it be possible to discuss/get some direction on this before going too far down an undesired path :-).

The goal with this PR was to refactor reference stack object to dynamically support both direct and indirect (cross stack) CloudFormation stack integrations.

Examples are always better, so here are a few capabilities of this PR and more generally what it is aiming to support:

Diff from #169 can be viewed here https://github.com/neoandroid/kube-aws/compare/masters-private-subnet...icereval:feature/parameters?expand=1

##### As kube-aws stands today
``` yaml
subnets:
  - availabilityZone: us-east-1a
    instanceCIDR: 10.0.0.0/24
  - availabilityZone: us-east-1b
    instanceCIDR: 10.0.1.0/24
  - availabilityZone: us-east-1d
    instanceCIDR: 10.0.2.0/24
```

##### Add support for Subnet ID, other fields are supported in https://github.com/coreos/kube-aws/pull/169

``` yaml
subnets:
  - availabilityZone: us-east-1a
    id: subnet-5197cb18
    privateRouteTable:
      id: rtb-db36d5a2
    natGateway:
      id: nat-08f4d9afdc670f872
  - availabilityZone: us-east-1b
    id: subnet-35c3a76e
    privateRouteTable:
      id: rtb-2b35d652
    natGateway:
      id: nat-05054cb2d096f2656
  - availabilityZone: us-east-1d
    id: subnet-ae412e83
    privateRouteTable:
      id: rtb-2f35d656
    natGateway:
      id: nat-06b6b84a32a7f4981
```

##### Indirect (cross stack) reference support for Identifier fields (potential to add Value field support)
``` yaml
subnets:
  - availabilityZone: us-east-1a
    idFromStackOutput: Network-PublicSubnet1
    privateRouteTable:
      idFromStackOutput: Network-PrivateRouteTable1
    natGateway:
      idFromStackOutput: Network-NatGateway1
  - availabilityZone: us-east-1b
    idFromStackOutput: Network-PublicSubnet2
    privateRouteTable:
      idFromStackOutput: Network-PrivateRouteTable2 
    natGateway:
      idFromStackOutput: Network-NatGateway1
  - availabilityZone: us-east-1d
    idFromStackOutput: Network-PublicSubnet3
    privateRouteTable:
      idFromStackOutput: Network-PrivateRouteTable3
    natGateway:
[...]

PR #227, opened 2017-01-10
Use existing subnets when creating/updating cluster

sasso:
If you define `subnetId` you can reuse existing subnets.

```
subnets:
  - availabilityZone: us-east-1a
    instanceCIDR: "10.0.1.0/24"
    subnetId: "subnet-xxxxxxxx"
  - availabilityZone: us-east-1b
    instanceCIDR: "10.0.2.0/24"
    subnetId: "subnet-xxxxxxxx"
...
```

there is still some work needed to support node-pools, but I would love to hear opinions or any suggestions on this approach.

It's backward compatible and it uses CF Parameters to overload Subnets with existing subnet id's or references to new resources.

linked to #52

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -302,9 +302,9 @@ type Cluster struct {
 type Subnet struct {
 	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
 	InstanceCIDR      string `yaml:"instanceCIDR,omitempty"`
+	lastAllocatedAddr *net.IP
 	SubnetId          string `yaml:"subnetId,omitempty"`
 	SubnetLogicalName string
-	lastAllocatedAddr *net.IP
 }
 
 type Experimental struct {
--- a/config/config.go
+++ b/config/config.go
@@ -302,6 +302,8 @@ type Cluster struct {
 type Subnet struct {
 	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
 	InstanceCIDR      string `yaml:"instanceCIDR,omitempty"`
+	SubnetId          string `yaml:"subnetId,omitempty"`
+	SubnetLogicalName string
 	lastAllocatedAddr *net.IP
 }
 
@@ -812,7 +814,11 @@ func (c DeploymentSettings) Valid() (*DeploymentValidationResult, error) {
 		}
 
 		var instanceCIDRs = make([]*net.IPNet, 0)
+
 		for i, subnet := range c.Subnets {
+			if subnet.SubnetId != "" {
+				continue
+			}
 			if subnet.AvailabilityZone == "" {
 				return nil, fmt.Errorf("availabilityZone must be set for subnet #%d", i)
 			}
@@ -935,7 +941,6 @@ Validates the an existing VPC and it's existing subnets do not conflict with thi
 cluster configuration
 */
 func (c *Cluster) ValidateExistingVPC(existingVPCCIDR string, existingSubnetCIDRS []string) error {
-
 	_, existingVPC, err := net.ParseCIDR(existingVPCCIDR)
 	if err != nil {
 		return fmt.Errorf("error parsing existing vpc cidr %s : %v", existingVPCCIDR, err)
@@ -970,19 +975,23 @@ func (c *Cluster) ValidateExistingVPC(existingVPCCIDR string, existingSubnetCIDR
 	// Loop through all subnets
 	// Note: legacy instanceCIDR/availabilityZone stuff has already been marshalled into this format
 	for _, subnet := range c.Subnets {
-		_, instanceNet, err := net.ParseCIDR(subnet.InstanceCIDR)
-		if err != nil {
-			return fmt.Errorf("error parsing instances cidr %s : %v", c.InstanceCIDR, err)
-		}
+		if subnet.SubnetId != "" {
+			continue
+		} else {
+			_, instanceNet, err := net.ParseCIDR(subnet.InstanceCIDR)
+			if err != nil {
+				return fmt.Errorf("error parsing instances cidr %s : %v", c.InstanceCIDR, err)
+			}
 
-		//Loop through all existing subnets in the VPC and look for conflicting CIDRS
-		for _, existingSubnet := range existingSubnets {
-			if netutil.CidrOverlap(instanceNet, existingSubnet) {
-				return fmt.Errorf(
-					"instance cidr (%s) conflicts with existing subnet cidr=%s",
-					instanceNet,
-					existingSubnet,
-				)
+			//Loop through all existing subnets in the VPC and look for conflicting CIDRS
+			for _, existingSubnet := range existingSubnets {
+				if netutil.CidrOverlap(instanceNet, existingSubnet) {
+					return fmt.Errorf(
+						"instance cidr (%s) conflicts with existing subnet cidr=%s",
+						instanceNet,
+						existingSubnet,
+					)
+				}
 			}
 		}
 	}
@@ -1074,4 +1083,12 @@ func withHostedZoneIDPrefix(id string) string {
 		return fmt.Sprintf("%s%s", hostedZoneIDPrefix, id)
 	}
 	return id
+}
+
+// Ref returns SubnetId or ref to newly created resource
+func (s Subnet) Ref() string {
+	if s.SubnetId != "" {
+		return fmt.Sprintf("%q", s.SubnetId)
+	}
+	return fmt.Sprintf(`{"Ref" : "%s"}`, s.SubnetLogicalName)
 }
--- a/config/templates/cluster.yaml
+++ b/config/templates/cluster.yaml
@@ -171,8 +171,10 @@ controller:
 # subnets:
 #   - availabilityZone: us-west-1a
 #     instanceCIDR: "10.0.0.0/24"
+#     subnetId: "subnet-xxxxxxxx" #optional
 #   - availabilityZone: us-west-1b
 #     instanceCIDR: "10.0.1.0/24"
+#     subnetId: "subnet-xxxxxxxx" #optional
 
 # CIDR for all service IP addresses
 # serviceCIDR: "10.3.0.0/24"
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -51,12 +51,8 @@
         {{end}}
         "VPCZoneIdentifier": [
           {{range $index, $subnet := .Subnets}}
-          {{with $subnetLogicalName := printf "Subnet%d" $index}}
           {{if gt $index 0}},{{end}}
-          {
-            "Ref": "{{$subnetLogicalName}}"
-          }
-          {{end}}
+          {{$subnet.Ref}}
           {{end}}
         ]
       },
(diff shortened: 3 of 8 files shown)


redbaron:
I am not terribly familiar with CF, but I can't see how would it work:
 subnets must be specified by id `subnet-xxxxx` or via reference to a logical resource name created  within same CF: `{ "Ref": "Subnet0" }`. You solution references using logical resource name but not using `"Ref"` function, therefore it would be treated as subnet id and obviously will be failing

mumoshu:
Thanks for the PR @sasso!

In addition to what @redbaron described, I'd recommend you to take a look into https://github.com/coreos/kube-aws/blob/37c242126fd56db938c3f170e51fddd7a6757207/nodepool/config/config.go#L228 to see how it could be achieved; not tested but maybe what you'd need is something like:

```go
func (s Subnet) LogicalName() string {
  logicalNameOfTheSubnet := // something
  return logicalNameOfTheSubnet
}

func (s Subnet) Ref() string {
	if s.SubnetId != "" {
		return fmt.Sprintf("%q", c.SubnetId)
	}
	return fmt.Sprintf(`{"Ref" : "%s"}}`, s.LogicalName())
}
```

With calling the `Ref()` func from the template like:

```
"Properties": {
          "RouteTableId": "{{$.RouteTableID}}",
          "SubnetId": {{$subnet.Ref}}
        },
```

sasso:
@redbaron, @mumoshu thanks for pointing me to the right direction! I pushed an update and I'm testing it now

config/config.go
```diff
@@ -299,6 +299,9 @@ type Cluster struct {
 type Subnet struct {
 	AvailabilityZone  string `yaml:"availabilityZone,omitempty"`
 	InstanceCIDR      string `yaml:"instanceCIDR,omitempty"`
+	SubnetId          string `yaml:"subnetId,omitempty"`
+	Create            bool
```
redbaron: this looks redundant, it is always `== (subnet.SubnetId != "")` so might as well remove it

config/config.go
```diff
@@ -1032,6 +1044,9 @@ func (c *Cluster) ValidateExistingVPC(existingVPCCIDR string, existingSubnetCIDR
 	// Loop through all subnets
 	// Note: legacy instanceCIDR/availabilityZone stuff has already been marshalled into this format
 	for _, subnet := range c.Subnets {
+		if !subnet.Create {
```
redbaron: validation is still needed, instead of checking for IP ranges, check that subnet actually exist

config/templates/cluster.yaml
```diff
@@ -169,8 +169,10 @@ controller:
 # subnets:
 #   - availabilityZone: us-west-1a
 #     instanceCIDR: "10.0.0.0/24"
+#     subnetId: "subnet-xxxxxxxx" #optional
```
redbaron: ideally it should work with just `subnetId` specified
```
subnets:
 - subnetId: subnet-aaaaa
 - subnetId: subnet-zzzzz
```

does it work like that already?

config/templates/stack-template.json
```diff
@@ -1037,10 +1027,10 @@
     }
     {{if $.ElasticFileSystemID}}
     ,
-    "{{$subnetLogicalName}}MountTarget": {
+    "{{$subnet.SubnetLogicalName}}MountTarget": {
```
redbaron: this makes a confusing resource name if we use existing subnets. Not sure it is worth fixing though :)

config/templates/cluster.yaml
sasso: No I did not pull any data from AWS, it uses {{$subnet.AvailabilityZone}} at some points so I used the static value. I'll look into pulling that from existing subnet

redbaron: why AvailabilityZone is even needed if subnets are specified?

config/templates/stack-template.json
sasso: I also need to move this out of the subnet create block

config/config.go
sasso: is there a reason validation is only done in create https://github.com/coreos/kube-aws/blob/master/cluster/cluster.go#L126

shouldn't we validate on update and render too?

config/templates/cluster.yaml
sasso: So I tried getting data from AWS but I couldn't find a nice place to put that checks as i would also need to set availabilityZone and instanceCIDR to not break other stuff. So i ran into other validation failing when removing this out.

I don't mind setting availabilityZone and instanceCIDR and I will leave it as is for now. It would take me to much time to correctly do this. If anyone has an suggestion, easy fix or want to help whit this that would be awesome.

sasso:
I also added validation for update and export. It seems bad not validating data there.

jeremyd:
This works now.  Tested without subnets (regular mode) and with subnetIDs (pre-existing).  All tests pass!  Anything else we should do to get this accepted?  Thanks!

redbaron:
Tested it too today, works as expected. I didn't test nodepools yet though

mumoshu:
@sasso Thanks for your effort!
I'd greatly appreciate it if you could add some test cases
* for main clusters to https://github.com/coreos/kube-aws/blob/master/config/config_test.go#L980
* for node pools to https://github.com/coreos/kube-aws/blob/master/nodepool/config/config_test.go#L37

jeremyd:
We use node pools but with the static IDs and it is working.  However, I went to test the scenario where users do not have pre-existing subnets and I'm having some problems.  We will keep hacking on this.. Might be a few days..  fyi.  Thanks!

redbaron:
@jeremyd users NOT having pre-existing subnets is the exact scenario kube-aws originally supports,  what kind of problem you encounter? does this PR break it?

jeremyd:
It's not supposed to. Right now it seems to break node pools in that
scenario.  We are working on it.  Unfortunately it's time consuming to test
as the only integration test appears to be hard coded to coreos AWS account.

mumoshu:
@jeremyd If you meant about the E2E test suite, it isn't hard-coded with a CoreOS AWS account(I don't even have access to one 😝) but you bring your own one for to run it. FYI the usage is documented in https://github.com/coreos/kube-aws/tree/master/e2e

jmcarp:
It looks like this PR adds a subset of the functionality in #195. @sasso: could you take a look at #195 and send a sub-PR there if you see anything missing? I think we'll get this done faster if we join forces.

redbaron:
@jmcarp this is much simpler change than  https://github.com/coreos/kube-aws/pull/195 , it is always easier to merge big changes in small pieces. If this one merged, then https://github.com/coreos/kube-aws/pull/195 can make use of it

redbaron:
@sasso , branch with conflicts resolved https://github.com/HotelsDotCom/kube-aws/tree/pull-227-conflicts-1

redbaron:
@sasso , also added commit to remove validation during `cluster.Update()` as it breaks `kube-aws update` command if subnets _were_  created  (== don't have `subnetId`  specified)

redbaron:
@jeremyd, can you check if latest version of this PR still breaks your case?

mumoshu:
@sasso @jeremyd @redbaron It seems like my cluster fails in `kube-aws validate` when there's no `subnetLogicalName: mysubnetname` defined in `cluster.yaml`. Is my understanding correct?

```
$ kube-aws validate --s3-uri s3://mybucket/clusters/kubeawstest4
Validating UserData...
UserData is valid.

Validating stack template...
Error: invalid cloudformation stack: ValidationError: Template format error: Resource name '' is non alphanumeric.
```

The stack template exported via `kube-aws --export` is containing this snippet:

```yaml
"": {
      "Properties": {
        "AvailabilityZone": "ap-northeast-1a",
        "CidrBlock": "10.0.0.0/24",
        "MapPublicIpOnLaunch": true,
        "Tags": [
          {
            "Key": "Name",
            "Value": "kubeawstest4-"
          },
          {
            "Key": "KubernetesCluster",
            "Value": "kubeawstest4"
          }
        ],
        "VpcId": {
          "Ref": "VPC"
        }
      },
      "Type": "AWS::EC2::Subnet"
    },
```

which indeed seem like the source of the error.

mumoshu:
Would you mind if I add some code to default `subnetLogicalName` to something like `subnet${index}` which we had before?

config/templates/stack-template.json
```diff
@@ -1190,14 +1182,12 @@
       "Type": "AWS::EC2::VPCGatewayAttachment"
     }
     {{range $index, $subnet := .Subnets}}
-    {{with $subnetLogicalName := printf "Subnet%d" $index}}
+    {{if not $subnet.SubnetId }}
     ,
-    "{{$subnetLogicalName}}RouteTableAssociation": {
+    "{{$subnet.SubnetLogicalName}}RouteTableAssociation": {
```
mumoshu: nit but we probably need to create a route table association if and only if the one doesn't exist for the subnet or I guess a cluster creation fails because of a duplicated association in the route table?

config/templates/stack-template.json
```diff
@@ -1105,19 +1097,19 @@
       },
       "Type": "AWS::EC2::Subnet"
     }
+    {{end}}
     {{if $.ElasticFileSystemID}}
     ,
-    "{{$subnetLogicalName}}MountTarget": {
+    "{{$subnet.SubnetLogicalName}}MountTarget": {
```
mumoshu: Note: When the `$subnet` is an existing one, we probably need to create a mount target if and only if the mount target doesn't exist in for the pair of the subnet and the EFS.
See https://github.com/coreos/kube-aws/issues/208#issuecomment-272036205

config/templates/cluster.yaml
```diff
@@ -171,8 +171,10 @@ controller:
...
 #   - availabilityZone: us-west-1b
 #     instanceCIDR: "10.0.1.0/24"
+#     subnetId: "subnet-xxxxxxxx" #optional
```
mumoshu: Note: `subnetId` will be renamed to `id` in the future according to https://github.com/coreos/kube-aws/pull/195#issuecomment-272061515
cc @icereval

mumoshu:
Defaulting of subnetLogicalName is addressed in https://github.com/coreos/kube-aws/pull/278/commits/b7adebbb5ae0d2e3fffb7fa82d7c94c081130276

mumoshu:
Thanks to all for your efforts on this :bow:
I've merged this to get the ball rolling.
Please send another pull request(s) for further fixes and improvements if necessary.

jeremyd:
Thanks so much!  We were/are so busy going live in production using kube
from kube-aws and appreciate the maintainers for working with us on this!!


On Mon, Jan 23, 2017, 6:28 PM KUOKA Yusuke <notifications@github.com> wrote:

Thanks to all for your efforts on this 🙇
I've merged this to get the ball rolling.
Please send another pull request(s) for further fixes and improvements if
necessary.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/coreos/kube-aws/pull/227#issuecomment-274683266>, or mute
the thread

Issue #224, opened 2017-01-10
Tainted but idle workers trend to 100% CPU and fail within a few hours


whereisaaron:
So I created a test cluster with 0 workers and in the main pool, a regular pool of 2 workers, and a spot pool of 3 workers. The 3 spot pool workers were tainted and I scheduled nothing on them. When I came back the next day, all the spot instances had flipped to 'NotReady', two can't be pinged any more and the third I can ping but not ssh to (times out). Checking the monitoring, all three have their CPU pegged at 100%.
```
NAME                                        STATUS                     AGE        
ip-172-31-106-46.region.compute.internal    Ready,SchedulingDisabled   1d
ip-172-31-107-195.region.compute.internal   Ready                      1d
ip-172-31-107-39.region.compute.internal    Ready                      1d
ip-172-31-108-131.region.compute.internal   NotReady                   1d
ip-172-31-108-16.region.compute.internal    NotReady                   1d
ip-172-31-108-68.region.compute.internal    NotReady                   1d
```

This is similar to the problem if you run the `untaint-and-cordon` task on the controller (discussed in #199). The task is logged as failed (even if it succeeds) as (`rkt rm` fails), so the task repeats continuously. Some sort of garbage/clean-up accumulates and the controller eventually hits 100% CPU and grinds itself to death. After patching out the `untaint-and-cordon` task on my controllers I no longer have this problem (on the controllers).

It seem like any workers might have the same problem if you use the `experimental.taints` option. This is what I used on my spot instances:
```
experimental:
  taints:
    - key: lifecycle
      value: spot
      effect: NoSchedule
```

Here is the CPU history of the now unpingable tainted spot workers:

![dead-worker](https://cloud.githubusercontent.com/assets/1109585/21794572/7d36078a-d6c9-11e6-8d51-6e4fe262db0e.png)

![dead-m4](https://cloud.githubusercontent.com/assets/1109585/21794708/cff6d1e2-d6ca-11e6-9b0c-20d0f75e5796.png)

mumoshu:
@whereisaaron Thanks again for your continuous feedbacks.

It definitely seems like an issue and I'd like it to be fixed in the same way as we've agreed in https://github.com/coreos/kube-aws/pull/199#issuecomment-271503071.

Would you mind sending a pull request for it or should I?

redbaron:
What about keeping `rkt` but ignoring `rkt rm` errors? I am doing that in https://github.com/coreos/kube-aws/pull/217 

[...]

PR #231, opened 2017-01-10
Change taint-and-uncordon worker task to use docker for now

whereisaaron:
rkt has a gnarly bug (coreos/rkt#3181) that won't be fixed in a hurry (coreos/rkt#3486). It least to continuous task failures that eventually totally wreak worker nodes (#224). In the meantime we can use docker just as easily for this simple task. This work around was discussed in #199.

I tested deploying a node-pool or workers with this patch and the experimental.taint setting enabled and the right thing happened. The taint was applied and the node uncordoned. All looks good in the log and CPU load is nominal.

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -540,33 +540,23 @@ write_files:
 
       hostname=$(hostname)
 
-      sudo rkt run \
-        --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
-        --mount=volume=kube,target=/etc/kubernetes \
-        --uuid-file-save=/var/run/coreos/taint-and-uncordon.uuid \
-        --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf \
-        --net=host \
-        --trust-keys-from-https \
-        {{.HyperkubeImageRepo}}:{{.K8sVer}} --exec=/bin/bash -- \
-          -vxc \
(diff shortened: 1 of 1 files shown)


redbaron:
Shortly after boot pre-patch:
```
cat /proc/mounts | wc -l
203365
```
:)

whereisaaron:
Two hours after boot with patch
```
cat /proc/mounts | wc -l
152
```
You win 😏

config/templates/cloud-config-worker
```diff
@@ -540,33 +540,23 @@ write_files:
...
-           taint="$kubectl taint node $hostname"; \
+           taint="$kubectl taint node --overwrite"; \
            for t in ${taints[@]}; do \
-             $taint "$t"; \
+             $taint "$hostname" "$t"; \
            done; \
-           echo done. ;\
-           echo uncordoning this node; \
+           echo "done."; \
+           echo "uncordoning this node."; \
```
redbaron: how is it live with ocd? :)

redbaron:
don't we need similar change on controllers?

whereisaaron:
The controller has also been patched @redbaron - see #199
I got bitten by that one too 😄

mumoshu:
LGTM & E2E passed.
Thanks again for your contribution @whereisaaron!

config/templates/cloud-config-worker
whereisaaron: "I get by."

PR #242, opened 2017-01-12
Do not require keyName if sshAuthorizedKeys are set

redbaron:
AWS CF reference that it is not mandatory to have one:

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-launchconfig.html#cfn-as-launchconfig-keyname

### DIFF
--- a/nodepool/config/templates/stack-template.json
+++ b/nodepool/config/templates/stack-template.json
@@ -15,7 +15,7 @@
           {
             "ImageId": "{{$.AMI}}",
             "InstanceType": "{{$spec.InstanceType}}",
-            "KeyName": "{{$.KeyName}}",
+            {{if $.KeyName}}"KeyName": "{{$.KeyName}}",{{end}}
             "WeightedCapacity": {{$spec.WeightedCapacity}},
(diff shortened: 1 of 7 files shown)


nodepool/config/templates/stack-template.json
```diff
@@ -161,7 +161,7 @@
         },
         "ImageId": "{{.AMI}}",
         "InstanceType": "{{.WorkerInstanceType}}",
-        "KeyName": "{{.KeyName}}",
+        {{if .KeyName}}"KeyName": "{{.KeyName}}",{{end}}
```
mumoshu: Could you add a test to cover the case in which `KeyName` is missing to our integration test suite for node pools: `aws_test.go`?
The suite runs validations on cfn stack templates calling actual AWS APIs.
That way we can at least save ourselves from breaking things badly in the future.

mumoshu: Would you like to me write the tests instead?

mumoshu:
@redbaron Thanks as always for your PR. I'm merging this anyway to get the ball rolling ⚽️

PR #243, opened 2017-01-12
Allow option to disable certificates management

redbaron:


nodepool/config/templates/cluster.yaml
```diff
@@ -46,6 +46,11 @@ availabilityZone: {{.AvailabilityZone}}
 # ARN of the KMS key used to encrypt TLS assets.
 kmsKeyArn: "{{.KMSKeyARN}}"
 
+# If you do not want kube-aws to manage certificaes, set it to false. If you do that
+# you are responsible for making sure that nodes have correct certificates by the time
+# daemons start up.
```
mumoshu: An example (or a link to a doc explaining it) of how to feed certificates to nodes before daemons start up would be helpful. I'm not really sure how myself and so are others!

Also, a brief explanation of when this feature is intended to be used (a.k.a why "you do not want kube-aws to manage certificates" in the first place) would be good to add, too.

config/config.go
```diff
@@ -621,16 +623,19 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
...
+		compactAssets, err = ReadOrCreateCompactTLSAssets(opts.TLSAssetsDir, KMSConfig{
+			Region:         stackConfig.Config.Region,
+			KMSKeyARN:      c.KMSKeyARN,
+			EncryptService: c.providedEncryptService,
+		})
+		if err != nil {
+			return nil, err
+		}
+		stackConfig.Config.TLSConfig = compactAssets
+	}
```
mumoshu: Could you add a test for this?
I believe `TestReadOrCreateCompactTLSAssets` would be an useful reference.

mumoshu:
@redbaron Thanks as always! I'm impressed to the amount of contribution these days 👍 
LGTM overall but two comments regarding the original intention of adding this feature and testing.

config/config.go
mumoshu: Would you like to me add the tests instead?

redbaron: please adopt / modify any of my PRs if it can speedup merging them to master

PR #245, opened 2017-01-13
BUGFIX for a typo in install-kube-system scripts

redbaron:
It was preventing following deployments to be started:
 - kube-dns
 - kube-dns-autoscaler
 - heapster


mumoshu:
Thanks :+1:

PR #124, opened 2016-12-06
Calico self hosted integration

heschlie:
- Migrated Calico to self hosted install
- Updated Calico versions

PR #248, opened 2017-01-13
Bump calico policy controller to 0.5.2

redbaron:
Fixes

```
2017-01-13 17:56:23,145 6 INFO Syncing 'NetworkPolicy' objects
2017-01-13 17:56:23,148 6 ERROR Unahandled exception killed NetworkPolicy manager
Traceback (most recent call last):
  File "<string>", line 295, in _manage_resource
  File "<string>", line 400, in _sync_resources
TypeError: object of type 'NoneType' has no len()
```

when there is no NetworkPolicy resource on a  new cluster

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -673,7 +673,7 @@ write_files:
         containers:
           # The Calico policy controller.
           - name: kube-policy-controller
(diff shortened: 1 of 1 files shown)


mumoshu:
@redbaron Thanks as always for your contribution! 👍 
This seems to be included in #124 which is merged today!
Could you confirm?

PR #250, opened 2017-01-15
Fix lint.

jmcarp:
Not sure why `go vet` is passing on travis, but it complains about unkeyed fields and fails locally for me.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 
@@ -20,7 +21,6 @@ import (
 	model "github.com/coreos/kube-aws/model"
 	"github.com/coreos/kube-aws/netutil"
 	yaml "gopkg.in/yaml.v2"
-	"strconv"
 )
 
 const (
@@ -652,9 +652,9 @@ func (c Cluster) ValidateUserData(opts StackTemplateOptions) error {
 	}
 
 	err = userdatavalidation.Execute([]userdatavalidation.Entry{
-		{"UserDataWorker", stackConfig.UserDataWorker},
-		{"UserDataController", stackConfig.UserDataController},
-		{"UserDataEtcd", stackConfig.UserDataEtcd},
+		{Name: "UserDataWorker", Content: stackConfig.UserDataWorker},
+		{Name: "UserDataController", Content: stackConfig.UserDataController},
+		{Name: "UserDataEtcd", Content: stackConfig.UserDataEtcd},
 	})
 
 	return err
--- a/nodepool/config/config.go
+++ b/nodepool/config/config.go
@@ -5,6 +5,8 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
+
 	cfg "github.com/coreos/kube-aws/config"
 	"github.com/coreos/kube-aws/coreos/amiregistry"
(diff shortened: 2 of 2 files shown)


mumoshu:
@jmcarp Thanks for the PR!
`go vet` is passing on my machine, too. Excuse me but would you mind telling me how you're running `go vet`?
FYI how it's running on travis can be seen at https://github.com/coreos/kube-aws/blob/master/make/test#L19

jmcarp:
Sure, I'm running it using `make test-with-cover`, like travis. Here's what I get without this patch:

```
config/config.go:655: github.com/coreos/kube-aws/coreos/userdatavalidation.Entry composite literal uses unkeyed fields
config/config.go:656: github.com/coreos/kube-aws/coreos/userdatavalidation.Entry composite literal uses unkeyed fields
config/config.go:657: github.com/coreos/kube-aws/coreos/userdatavalidation.Entry composite literal uses unkeyed fields
exit status 1
nodepool/config/config.go:79: github.com/coreos/kube-aws/coreos/userdatavalidation.Entry composite literal uses unkeyed fields
exit status 1
```

I'm running go1.7.4. I think @icereval ran into the same issue locally.

icereval:
FWIW, the error does not appear when running on the [docker golang container (debian jessie)](https://hub.docker.com/_/golang/), thus Travis' success.

screen shot 2017-01-15 at 1 19 54 pm

I've also tried both the official golang 1.7.4 build and the brew install on OSX with the same result.

mumoshu:
@jmcarp @icereval I could reproduce the issue on my mac after `brew update && brew upgrade golang` with golang 1.7.4 but still having no luck on travis even after upgrading it to use `1.7.x`:
https://github.com/mumoshu/kube-aws/commit/de6560112b3027f573841dbca2b7b4bb24db4824
https://travis-ci.org/mumoshu/kube-aws/builds/192210197

mumoshu:
Ah, never mind. What I was seeing were errors which are not from the current kube-aws itself but from my poc for kube-aws inside the same directory. I'm still unable to reproduce the issue:

```
$ go version
go version go1.7.4 darwin/amd64
$ go vet -v $(go list ./... | grep -v '/vendor/')
$ echo $?
0
```

icereval:
Agreed and its strange, I too built specifically 1.7.4 on travis earlier and narrowed it down just to `go vet` and it results in 0 errors in the test [environment](https://travis-ci.org/icereval/kube-aws/builds).

I can work around this issue for now so its not a big deal.

Out of curiosity whats your 1.7.4 go sha256? mine is:

```bash
$ go version
go version go1.7.4 darwin/amd64
$ shasum -a 256 -b `which go`
e19967912c02675d44806c8bbdd681a6218844fcb7f6977a72dffdaf710c2e71 */usr/local/bin/go
```

mumoshu:
Regarding the sha256, mine is:

```
$ go version
go version go1.7.4 darwin/amd64
$ shasum -a 256 -b `which go`
9118739c7497ffe86a4ba73867937030d011004f1f804ce260823b129241eabd */usr/local/bin/go
```

mumoshu:
@jmcarp @icereval I'd rather like to merge this once the `go vet` errors are confirmed to be valid(which seems true to me).
Honestly I'm not very familiar with `go vet` but the `composites` check of `go vet` which seems to be emitting the errors should be enabled by default in `go vet`. Am I correct?

mumoshu:
I'm merging this anyways! Thanks.

Issue #252, opened 2017-01-16
Userdata limit of 16KB is almost reached 


redbaron:
LaunchController userdata size is almost 16KB and starts to fail if few modifications added. We need to find another way topass userdata, such as s3 bucket.

danielfm:
I just bumped into this when trying to move my customizations from earlier versions of kube-aws to the new ones. The error message when that happens is not pretty as well:

```
$ ~/kube-aws up --s3-uri ...
Creating AWS resources. This should take around 5 minutes.
Error: Error creating cluster: Stack creation failed: CREATE_FAILED : The following resource(s) failed to create: [InstanceEtcd0, LaunchConfigurationController].

Printing the most recent failed stack events:
CREATE_FAILED AWS::CloudFormation::Stack kube-aws-test The following resource(s) failed to create: [InstanceEtcd0, LaunchConfigurationController].
[...]

PR #255, opened 2017-01-17
Bypass 16kb userdata limit by uploading to s3 bucket

redbaron:
Crude fix for #252 .

This is a best-effort rebase+squash of our internal branch on top of master.  It seems to solve this unexpected blocker for us, but code is brittle and moderately ugly, but not much more that already existing code around there :) I have no capacity to refine it further. It _breaks_ nodepools (another reason to get rid of them!).

Probably best thing to do is to make `cluster.yml` to accept `s3Bucket` config option and remove `s3-uri` flag, making presence of a readable bucket for cluster needs a hard requirement.

As I said, I wont be able to make it better, please adopt it, edit it to your needs and fix it,throw away it or whatever, but fix 16KB problem in master one way or another It is a major blocker for us so had to be solved ASAP, I imagine other users feel this pain too.

No testing of this PR is done whatsoever, just hacked it so that it compiles. Internal branch this PR is based on seems to be working fine.


redbaron:
small comment on PR itself: all it is doing is plumbing `stackConfig` (which includes what of *Cluster flavours, I lost track which one)all the way up to a point where CF template upload is happening.

config/config.go
```diff
@@ -577,7 +577,7 @@ func (c Cluster) stackConfig(opts StackTemplateOptions, compressUserData bool) (
 		stackConfig.Config.TLSConfig = compactAssets
 	}
 
-	if stackConfig.UserDataWorker, err = userdatatemplate.GetString(opts.WorkerTmplFile, stackConfig.Config, compressUserData); err != nil {
+	if stackConfig.UserDataWorker, err = userdatatemplate.GetString(opts.WorkerTmplFile, stackConfig.Config, false); err != nil {
```
mumoshu: Thanks! I'd been bitten by this myself 😛

mumoshu:
Thanks for the follow-ups!
I'm testing/refactoring this in my internal branch.
I'll merge this and/or my branch including your commit probably after #227 and/or #169

mumoshu:
Closing this in favor of #268(It is almost ready to be merged!

PR #256, opened 2017-01-17
Bump to k8s 1.5.2

redbaron:


mumoshu:
@redbaron Thanks as always! This is done in #262 by running `contrib/bump-version`.

PR #263, opened 2017-01-18
Fix node pools on master

### LINKED ISSUE #261: Fix node pools on master
I believe the node pools feature is broken since merging #226.
Please consider not to use binaries built from master if you'd like to use node pools.
See https://github.com/coreos/kube-aws/pull/226#issuecomment-273343076 for more information.

mumoshu:
This is a follow up for the PR #226 "Don't precalculate etcd static IP addresses"

To see why this follow up was needed, read the maintainer's comment at #226 (comment)

P.S. As the result of some refactorings done in order to implement this, worker nodes powered by a spot fleet now supports the `awsEnvironment` experimental feature

fixes #261 

cc @redbaron

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -29,8 +29,8 @@ coreos:
           {{.AWSCliImageRepo}}:{{.AWSCliTag}} -- cfn-init -v \
               -c "etcd-client" \
               --region {{.Region}} \
-              --resource LaunchConfigurationController \
-              --stack {{ .ClusterName }}
+              --resource {{.Worker.LogicalName}} \
+              --stack {{ .StackName }}
         ExecStart=/usr/bin/mv -f /var/run/coreos/etcd-environment /etc/etcd-environment
 
     - name: docker.service
@@ -212,8 +212,9 @@ coreos:
           --net=host \
           --trust-keys-from-https \
           {{.AWSCliImageRepo}}:{{.AWSCliTag}} -- cfn-init -v \
+              -c "aws-environment" \
               --region {{.Region}} \
-              --resource LaunchConfigurationWorker \
+              --resource {{.Worker.LogicalName}} \
               --stack {{.StackName}}
 {{end}}
 
@@ -310,7 +311,7 @@ coreos:
           --trust-keys-from-https \
           {{.AWSCliImageRepo}}:{{.AWSCliTag}} -- cfn-signal -e 0 \
               --region {{.Region}} \
-              --resource AutoScaleWorker \
+              --resource Workers \
               --stack {{.StackName}}
 {{end}}
 
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -9,7 +9,7 @@
     }
   },
   "Resources": {
-    "AutoScaleWorker": {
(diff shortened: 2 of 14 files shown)


mumoshu:
@redbaron Thanks for taking your time reviewing this!
> Output json list of IP addresses which pool stack then import and use

Wouldn't it break TLS between etcd and node pools then?

config/templates/cloud-config-worker
```diff
@@ -29,8 +29,8 @@ coreos:
           {{.AWSCliImageRepo}}:{{.AWSCliTag}} -- cfn-init -v \
               -c "etcd-client" \
               --region {{.Region}} \
-              --resource LaunchConfigurationController \
-              --stack {{ .ClusterName }}
```
mumoshu: Sounds great!
However `cfn-init` doesn't seem to have access to a stack other than the one in which the EC2 instance running `cfn-init` is directly/indrectly managed 😢 
`cfn-init` fails like:

```
Jan 18 23:46:32 ip-10-0-1-242.ap-northeast-1.compute.internal rkt[7989]: [ 1044.115954] awscli[5]: ValidationError: Instance i-0507f6146f3c1bd79 in not a member of stack arn:aws:cloudformation:ap-northeast-1:<account-id>:stack/kubeawstest2/4dfa4280-dd8e-11e6-8dbb-500c44f24ce6
```

redbaron:
~private ip addresses~ , etcd DNS names

mumoshu:
Thanks! So your intention is to revive `etcdEndpoints` key in the nodepool's `cluster.yaml` or [what I've done in this work to allow a node pool to import etcd dns names from a main cluster](https://github.com/coreos/kube-aws/pull/263/files#diff-4c2a1b47467158c986623aa3a3d3c323R1244)?

redbaron:
ah, nevermoind, I didn't read it close enough,you already doing export + import :) I saw changes in go code and  thought you were doing something to enable it work in nodepools, but you are  not, so it is all good.

config/templates/stack-template.json
```diff
@@ -87,9 +87,44 @@
           {{end}}
         }
       },
-      "DependsOn" : ["AutoScaleController"]
+      "Metadata" : {
```
mumoshu: Thanks for your comment!
Regarding the mental effort, your original decision sounds very reasonable to me now.
On the other hand, I'd also like only relevant configsets to be included in the metadata of a AWS resource like an ASG for workers and an ASG for controllers, respectively.
So, would you think it is not the best but acceptable/slightly better to extract each configset to a golang subtemplate like:

```yaml
{{define "cfn-init-configset-aws-environment"}}
{
  "commands": {
  "write-environment": {
  "command": { "Fn::Join" : ["", [ "echo '",
  {{range $variable, $function := .Experimental.AwsEnvironment.Environment}}
  "{{$variable}}=", {{$function}} , "\n",
  {{end}}
  "' > /etc/aws-environment" ] ] }
  }
}
{{end}}
{{define "cfn-init-configset-etcd-client"}}
...
{{end}}
{{define "cfn-init-configset-etcd-server"}}
...
{{end}}
```

redbaron: nice and clean,I like it

PR #265, opened 2017-01-18
Add support for Encrypting EBS volumes

sasso:


nodepool/config/templates/stack-template.json
```diff
@@ -73,7 +73,8 @@
                   {{if gt $spec.RootVolumeIOPS 0}}
                   "Iops": "{{$spec.RootVolumeIOPS}}",
                   {{end}}
-                  "VolumeType": "{{$spec.RootVolumeType}}"
+                  "VolumeType": "{{$spec.RootVolumeType}}",
+                  "Encrypted": {{$spec.RootVolumeEncrypted}}
```
mumoshu: Thanks for the PR! :bow:
Any chance you could fix somewhere so that the travis build passes?
I suspect this line as the source of the problem. If I understand correctly, shouldn't it be `WorkerRootVolumeEncrypted` rather than `RootVolumeEncrypted`?

sasso: I'll check and fix the tests

sasso: fixed

sasso:
I need to cherry-pick some fixes from another branch...

sasso:
I'll close this PR and create another one. The only thing that can be encrypted is EtcdDataVolume, everything else needs/can be encrypted by rebuilding CoreOS/ContainerOS with encryption enabled.

PR #268, opened 2017-01-19
Bypass userdata limit with s3 plus follow ups

mumoshu:
@redbaron FYI this is what I have so far.

Unit tests are failing/won't even compile but I've tested this by creating an actual cluster including 1 main cluster with 2 node pools including 1 asg-backed pool and 1 spot-fleet-backed pool.

If interested, please see my latest commit https://github.com/mumoshu/kube-aws/commit/c2358d7b749092a947fd4ebcc576dd729413e992 for what were changed since your work. Basically I've refactored `cfnstack/` so that
* you don't need explicitly render stack templates where it seems unnecessary (e.g. `destroy.go`) anymore
* you don't need to pass s3URI all over the places


mumoshu:
This isn't really a natural order of things I would do to fix something but anyways E2E tests are passing against the cluster.
I'll work on fixing unit tests next.

cluster/cluster.go
```diff
@@ -179,7 +212,19 @@ func (c *Cluster) Create(stackBody string, s3URI string) error {
...
+	stackTemplate, err := c.RenderTemplateAsString()
+	if err != nil {
+		return fmt.Errorf("Error while rendering template : %v", err)
+	}
+
+	uploads := map[string]string{
+		"stack.json":          stackTemplate,
+		"userdata-controller": c.UserDataController,
+		"userdata-worker":     c.UserDataWorker,
+		"userdata-etcd":       c.UserDataEtcd,
```
redbaron: etcd userdata can't be  moved to s3, have to stay in CF template to be excluded from updates

mumoshu: Good catch! I'll fix it. Anyways, it is uploaded but not read from anywhere now.

mumoshu:
> nodepools should write to a different path , otherwise the'll overwrite main cluster userdata. this probably means that `--s3-uri` arg should be moved into cluster.yaml config option

Thanks for the notice 👍 
I'm aware of it and actually it is already addressed in this PR.

mumoshu:
Unit/integration tests are passing.

cfnstack/provisioner.go
```diff
@@ -71,25 +75,25 @@ func (c *Provisioner) UploadTemplate(s3Svc S3ObjectPutterService, s3URI string,
...
 
-func (c *Provisioner) uploadTemplateIfNecessary(s3Svc S3ObjectPutterService, stackBody string, s3URI string) (*string, error) {
-	if len(stackBody) > CFN_TEMPLATE_SIZE_LIMIT {
-		if s3URI == "" {
-			return nil, fmt.Errorf("stack template's size(=%d) exceeds the 51200 bytes limit of cloudformation. `--s3-uri s3://<bucket>/path/to/dir` must be specified to upload it to S3 beforehand", len(stackBody))
-		}
-
-		templateURL, err := c.UploadTemplate(s3Svc, s3URI, stackBody)
-		if err != nil {
+func (c *Provisioner) uploadTemplates(s3Svc S3ObjectPutterService, uploads map[string]string) (*string, error) {
```
mumoshu: @redbaron It turned out to me that you've used `uploadTemplates` to upload not only a template but also files referenced from it(e.g. `cloud-config-worker` and `cloud-config-controller`).
If so, I'd rather like to change its signature to something like `uploadStackAssets(s3Svc S3ObjectPutterService, stackTemplate string, cloudConfigs: map[string]string)` for now and extend/introduce a type like `StackAssets` in the future.
Then, `if filename == "stack.json" {` can be removed i.e. hence less complexity.

mumoshu:
E2E tests passed.

Although I can merge this anytime, I'd like to hold on until other big PRs like #169 #227, which are likely to conflict with this PR, become ready to be merged.

redbaron:
I tested https://github.com/coreos/kube-aws/pull/255/commits/9fce0cb541b3fe103a614d8e0362508387ddda56 change locally,seems to work, consider adding here if you agree

mumoshu:
@redbaron Thanks for wrapping it into a systemd unit 👍 I will test it out later.

Btw, `kube-aws up` seems to take 2x more time than it had taken before while creating ASGs:

image

Any idea how we could improve it?

PR #272, opened 2017-01-19
worker userdata template: availability zone is not pre-pended to the EFS name

vipinsthename:
The worker userdata template does not need to get the availability zone to prepend to the ElasticFileSystemID.

### DIFF
--- a/config/templates/cloud-config-worker
+++ b/config/templates/cloud-config-worker
@@ -262,7 +262,7 @@ coreos:
         [Service]
         Type=oneshot
         ExecStartPre=-/usr/bin/mkdir -p /efs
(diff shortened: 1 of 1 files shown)


mumoshu:
Hi @vipinsthename, thanks for the PR!
AFAIK, the AZ part of the mount target is required according to the [relevant AWS doc](http://docs.aws.amazon.com/efs/latest/ug/mounting-fs.html#mounting-fs-install-nfsclient).
Probably that's not always the case?
I'd appreciate it if you could provide me more context on why you'd decided to make this change.

pieterlange:
You actually do want the AZ there because otherwise you're paying for inter-AZ traffic and you create a more fragile system.

If you don't mind those downsides you could actually just use the native kubernetes nfs volume type.

mumoshu:
I'm closing this according to the @pieterlange comment for now but please feel comment on this if nececssary :bow:

PR #274, opened 2017-01-19
Remove obsolete Calico files

redbaron:


caseydavenport:
CC @heschlie

heschlie:
@redbaron Good catch, these should be taken out after the self-hosted upgrade.  LGTM

mumoshu:
Thanks to all for the improvement :bow: Merging.

PR #282, opened 2017-01-25
Updating Calico images to Calico 2.0.2

heschlie:
Calico 2.0.1 was released with some bug fixes, more details here:

http://docs.projectcalico.org/v2.0/releases/

### DIFF
--- a/config/templates/cloud-config-controller
+++ b/config/templates/cloud-config-controller
@@ -474,7 +474,7 @@ write_files:
             hostNetwork: true
             containers:
               - name: calico-node
-                image: quay.io/calico/node:v1.0.0
+                image: quay.io/calico/node:v1.0.1
                 env:
                   - name: ETCD_ENDPOINTS
                     valueFrom:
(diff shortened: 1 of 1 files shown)


mumoshu:
@heschlie Thanks as always and sorry for the delay but calico 2.0.2 seems to have already been released.
Would you mind updating this pr or send another pr for 2.0.2?

heschlie:
@mumoshu It indeed has, and I will!

redbaron:
@heschlie just for my education, what is relation between felix and calico node? why we use one,but not another here?

heschlie:
@redbaron Felix is part of our backend (the biggest part I believe) that is balled up into calico-node.  I believe it can be run on its own, but calico-node includes Bird (the BGP daemon we use, though not with Flannel) as well as several other utilities to get the ball rolling, see [here](https://github.com/projectcalico/calicoctl/tree/master/calico_node) if you are interested in more details.

@mumoshu Updated to Calico 2.0.2

mumoshu:
I've verified this by manually running E2E tests and therefore merged.

@heschlie Thanks as always!

PR #292, opened 2017-02-01
More robust etcd search and replace step

redbaron:
On workers nodes with calico enabled /etc/kubernetes/net.d/ can contain
directories which breaks `sed  -i '...' /etc/kuberneted/net.d/*`. It
runs smooth on first boot, but subseqent restarts are failing because
of newly created directory.

Change it to use `find -maxdepth 1 -type f | xargs ...` approach
which should be more robust.


mumoshu:
@redbaron Thanks as always for your contribution :bow:

I've verified this by manually running E2E tests.

PR #294, opened 2017-02-02
Add ClusterAutoscaler experimental option to setup iam permissions

pwillie:
Simple change to optionally provision worker nodes with sufficient privileges to run the cluster-autoscaler controller.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,7 +38,7 @@ func NewDefaultCluster() *Cluster {
 		AwsNodeLabels{
 			Enabled: false,
 		},
-		ClusterAutoScaler{
+		ClusterAutoscalerSupport{
 			Enabled: false,
 		},
 		EphemeralImageStorage{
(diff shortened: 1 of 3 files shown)


config/templates/stack-template.json
```diff
@@ -742,7 +754,13 @@
...
           { "Ref" : "SecurityGroupElbAPIServer" }
         ]
-      }
+      },
+      "DependsOn" : [
+        {{range $etcdIndex, $etcdInstance := .EtcdInstances}}
+        {{if gt $etcdIndex 0}},{{end}}
+        "InstanceEtcd{{$etcdIndex}}"
+        {{end}}
+      ]
```
mumoshu: What's your intention of adding this?
This seems to mean that the controller ELB depends on etcd nodes, which isn't true IMHO.

config/config.go
```diff
@@ -404,6 +407,7 @@ type Experimental struct {
 	AuditLog              AuditLog              `yaml:"auditLog"`
 	AwsEnvironment        AwsEnvironment        `yaml:"awsEnvironment"`
 	AwsNodeLabels         AwsNodeLabels         `yaml:"awsNodeLabels"`
+	ClusterAutoScaler     ClusterAutoScaler     `yaml:"clusterAutoScaler"`
```
mumoshu: nit: I prefer `ClusterAutoscaler` and `clusterAutoscaler` because the original CA is spelled `cluster-autoscaler`

mumoshu:
Hi @pwillie, thanks for the PR!
I've left several minor comments but overall LGTM.
Would you mind addressing these comments?

config/config.go
mumoshu: Could you also make these `clusterAutoscalerSupport` in cluster.yaml and `ClusterAutoscalerSupport` in golang code respectively so that we can be consistent among other experimental features to support k8s addons like #297

config/templates/stack-template.json
pwillie: I found that when running within a small private VPC the controller ELB will provision faster than the etcd nodes and more often than not will be provisioned with the same IP that was calculated for one of the etcd nodes resulting in a failed deployment.

config/config.go
pwillie: No problem.  Done

config/templates/stack-template.json
redbaron: > more often than not will be provisioned with the same IP that was calculated for one of the etcd nodes resulting in a failed deployment.

do you use latest master? Etcd servers IP addresses are not calculated by kube-aws anymore

mumoshu: @pwillie As redbaron kindly described, it won't happen with latest master or even with v0.9.4-rc.1.
Could you confirm the fix and if you're ok with that, remove this `DependsOn` part?

pwillie: That explains it.  Was using the latest release at the time of v0.9.3-rc.5.  Will update and re-test.

mumoshu: Thanks for your help :bow: This PR is ready to be merged once this `DependsOn` section is sorted out!

pwillie: Hi @mumoshu.  I have sorted the DependsOn section.  What is the convention with PR.  Should I rebase and squash my commits or ??

mumoshu:
@pwillie LGTM! Thanks for your contribution 👍

Issue #253, opened 2017-01-16
Add a (optional) feature to enable kube2iam


ankon:
(Issue itself is work-in-progress, updating it with the parts I found are needed)

In the kube-aws context it seems reasonable to assume that users have IAM enabled, and have some knowledge about it - and that they want it on pod-level, not on worker-level granularity. It would be awesome if kube-aws would provide some OOBE for setting that up using https://github.com/jtblin/kube2iam.

1. Add an "experimental" feature for kube2iam in cluster.yaml
2. If enabled: add this piece to the worker role:
   ```json
           {
            "Action": "sts:AssumeRole",
            "Resource": "*",
            "Effect": "Allow"
        },
   ```
3. If enabled: Deploy the kube2iam daemonset as part of kube-system (or maybe introduce a kube-aws namespace?)
4. Document how to link the worker role to existing roles (roughly like this: 
   ```sh
   EXISTING_ROLES=....
   ROLE_ID=`aws cloudformation list-stack-resources --stack-name CLUSTER-ID | jq -r '.StackResourceSummaries[]|select(.LogicalResourceId=="IAMRoleWorker")|.PhysicalResourceId'`
   ROLE_ARN=`aws iam get-role --role-name collaborne-20170105-IAMRoleWorker-XX0VV92HCQKG | jq -r .Role.Arn`
   for role in ${EXISTING_ROLES}; do
     aws iam get-role --role-name ${role} | jq ".Role.AssumeRolePolicyDocument|(.Statement |= .+ [{\"Principal\":{\"AWS\":\"${ROLE_ARN}\"},\"Action\":\"sts:AssumeRole\",\"Effect\":\"Allow\"}])" > policy-${role}.json
     aws iam update-assume-role-policy --role-name ${role} --policy-document file://$PWD/policy-${role}.json
   done   
   ```

c-knowles:
There's a helm chart for kube2iam, perhaps supporting helm could be a good idea.

ankon:
Sounds reasonable. Especially since I now found that there is a bug in kube2iam:latest in combination with calico: https://github.com/jtblin/kube2iam/issues/37. The fix proposed there comes via a helm chart, too.

Other than using :0.2.2 everything seems to work though in my setup, so that's promising.

c-knowles:
[...]

PR #297, opened 2017-02-02
adding kube2iam experimental support

Fsero:
If you plan to use kube2iam in existing kubeaws cluster you face two things:

- Default WorkerRole does not have the sts:assumeRole grant blocking kube2iam to assume the role for the container.
- target roles needs to change the trust policy to allow kube-aws worker IAM Role to assume the role. Because normally cloudformation generates the role with the role name and a random ID at the end makes the IAM Role less predictable and therefore more complicated to work with if you want to automate this trust.

This pr introduces kube2iam experimental flag, when kube2iam flag is enabled it will add sts:assumeRole to the worker IAM role to allow kube2iam to work, additionally you can specified a named role to generate a worker IAM Role more human and automation friendly as explained here http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-rolename

If you find the named role useful maybe could be merged in the general configuration not only for the worker

### DIFF
--- a/config/templates/cluster.yaml
+++ b/config/templates/cluster.yaml
@@ -125,6 +125,11 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 # Disk type for worker node (one of standard, io1, or gp2)
 #workerRootVolumeType: gp2
 
+# if you specify workerManagedIamRoleName the role created will not suffix the random id at the end
+# Role will be created with "Ref": {"AWS::StackName"}-{"AWS::Region"}-yourManagedRole
+# to follow the recommendation in AWS documentation  http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html
+#workerManagedIamRoleName: "yourManagedRole"
+
 # Number of I/O operations per second (IOPS) that the worker node disk supports. Leave blank if workerRootVolumeType is not io1
 #workerRootVolumeIOPS: 0
 
@@ -475,6 +480,11 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #     # See https://github.com/coreos/kube-aws/issues/230 for more information.
 #     rbac:
 #       enabled: true
+#    # When enabled it will grant sts:assumeRole permission to everyrole in the IAM Worker Role.
+
+#    kube2IamSupport:
+#      enabled: true
+#
 
 # AWS Tags for cloudformation stack resources
 #stackTags:
--- a/cfnstack/provisioner.go
+++ b/cfnstack/provisioner.go
@@ -166,7 +166,7 @@ func (c *Provisioner) baseCreateStackInput() *cloudformation.CreateStackInput {
 	return &cloudformation.CreateStackInput{
 		StackName:       aws.String(c.stackName),
 		OnFailure:       aws.String(cloudformation.OnFailureDoNothing),
-		Capabilities:    []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
+		Capabilities:    []*string{aws.String(cloudformation.CapabilityCapabilityIam), aws.String(cloudformation.CapabilityCapabilityNamedIam)},
 		Tags:            tags,
 		StackPolicyBody: aws.String(c.stackPolicyBody),
 	}
@@ -180,7 +180,7 @@ func (c *Provisioner) createStackFromTemplateURL(cfSvc CreationService, stackTem
 
 func (c *Provisioner) baseUpdateStackInput() *cloudformation.UpdateStackInput {
 	return &cloudformation.UpdateStackInput{
-		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam)},
+		Capabilities: []*string{aws.String(cloudformation.CapabilityCapabilityIam), aws.String(cloudformation.CapabilityCapabilityNamedIam)},
 		StackName:    aws.String(c.stackName),
 	}
 }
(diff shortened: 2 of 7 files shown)


Fsero:
i think this also helps issue #253

config/templates/cluster.yaml
```diff
@@ -469,6 +469,12 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #     # See https://github.com/coreos/kube-aws/issues/230 for more information.
 #     rbac:
 #       enabled: true
+#    # When enabled it will grant sts:assumeRole permission to everyrole in the IAM Worker Role.
+#    # if you specify workerNamedIAMRole the role created will not suffix the random id at the end
+#    # which is useful to update trust policy in target IAM roles that container wants to use through kube2iam.
+#    kube2IAMSupport:
```
mumoshu: nit but it should be spelled `kube2IamSupport` for consistency among other keys in cluster.yaml

mumoshu:
Thanks for the PR with very informative description, @Fsero!
This is almost merged but could you add an example configuration to the test data for the case named `WithExperimentalFeatures` in `test/integration/maincluster_test.go` so that we can ensure that this feature won't break in the future?

Fsero:
@mumoshu thanks! added the tests in `maincluster_test.go` and `nodepool_test.go` and changed the case of every IAM mention. I hope this is enough, in other case let me know what can i do to get this merged :)

config/templates/cluster.yaml
```diff
@@ -469,6 +469,12 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #     # See https://github.com/coreos/kube-aws/issues/230 for more information.
 #     rbac:
 #       enabled: true
+#    # When enabled it will grant sts:assumeRole permission to everyrole in the IAM Worker Role.
+#    # if you specify workerNamedIAMRole the role created will not suffix the random id at the end
+#    # which is useful to update trust policy in target IAM roles that container wants to use through kube2iam.
+#    kube2IamSupport:
+#      enabled: true
+#      workerNamedIAMRole: "yournamedrole"
```
pieterlange: What is the point of letting people choose the role name here? We don't let them choose resource names elsewhere :)

pieterlange:
After this is merged we should also:
  * add some docs that show how to use kube2iam
  * remove all default controller/worker IAM permissions that are not strictly required (create them as roles and attach them to the kube pods that actually need them)

Great work!

config/templates/cluster.yaml
Fsero: @pieterlange if you have a target role for kube2iam ,let say role A, if role A has a trust policy for role R set in cluster.yaml the cluster recreation does not need to retriever a trust policy update

Fsero: S/retriever/retrigger/g

Fsero: edited in web because it was not clear,  check it if you had read email.

c-knowles: Could we just fix the name based on the stack name postfixed with some static role name? Perhaps even [export it](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html) so it can be referenced from other stacks?

c-knowles:
@pieterlange if we are going to use it for other purposes then perhaps when it's not experimental any more we should consider renaming it from `Kube2IamSupport` to something like `roleName` under the `worker` config?

Maybe we could also consider whether to allow a list of roles that are allowed to be assumed from a k8s cluster or just always allow `*`.

config/templates/cluster.yaml
Fsero: @c-knowles one thing i can do is to change the namedRole to something like {{stack-name}}-IAMRoleWorker-{{suffixRoleName}, and then change  workerNamedIAMRole to workerNamedIAMRoleSuffix and set a sensible default like kube2iam, i would prefer to give the users the option to set the naming it makes most sense in their scenario.

c-knowles: It'll work either way so I don't mind, just the same as @pieterlange I'm wondering what the use case is for allowing it to be changed rather than just fixing it against the stack name.

mumoshu:
Note to self: I'd rather like to move `experimental.kube2IamSupportworkerNamedIAMRole` to `worker.managedIamRoleName` so that we can let users generally use it to trust the role managed for worker nodes by kube-aws from a target role.

Fsero:
@mumoshu is there something else I can do to get this merged? :-)

mumoshu:
@Fsero Sorry for the silence!
I'm still busy working on my own PR #315 but in the meantime would you confirm/apply what I've suggested in https://github.com/coreos/kube-aws/pull/297#issuecomment-278492922 on your side? (Btw, does the suggested name `worker.managedIamRoleName` over the former `kube2IamSupportWorkerNamedIAMRole` make sense to you too?)

Anyways, this PR is going to be merged before my own PR not to delegate the hard work of resolving huge conflicts 👍

Fsero:
sure @mumoshu i can do that :)

mumoshu:
Thanks @Fsero!
Could you also rebase your branch? FYI/AFAICS `ClusterAutoscalerSupport ClusterAutoscalerSupport `yaml:"clusterAutoscalerSupport"` from an another PR is the only line conflicting with yours.

Fsero:
@mumoshu i rewrited my PR with the proposed changes, i kept the experimental feature for kube2iam because you maybe want to have the managedIamRole without using kube2iam, in doesnt make sense we can always drop this feature flag and add this grant in the base IAM Role

mumoshu:
LGTM 👍 Thanks for your contribution @Fsero!

PR #304, opened 2017-02-06
Wait for signals by default

### LINKED ISSUE #259: Wait until apiserver becomes ready to serve API before `kube-aws up` finishes
kube-aws as of today prints the message `You should be able to access the Kubernetes API once the containers finish downloading.` after `kube-aws up` completes.
This message expresses that an user must hold on for a few minutes until the apiserver becomes ready to serve API i.e. `kubectl` succeeds.

I've recently received a question in the k8s slack about why `kube-aws up` didn't create a working cluster and one of the gotcha for the one was `kubectl` doesn't succeed immediately after `kube-aws up` completes which is IMHO obvious if one is very careful. In other words, just printing a message like now isn't very friendly.

One of the ways to make it a bit more friendlier is to make `waitSignal` graduate from experimental so that a controller ASG waits until apiserver becomes responsive.

I'll leave this open for a while to hear from the community for ideas.

gianrubio:
Add skip-wait flag
Enable feature waitSignal as default
Setup initial value for RollingUpdateMinInstancesInService as 1
Require s3-flag on validate command
Fix #259

@mumoshu, could you run your e2e tests and do some reviews? I'm writing some docs explaining how the signal works

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -22,8 +22,8 @@ var (
 	}
 
 	upOpts = struct {
-		awsDebug, export, prettyPrint bool
-		s3URI                         string
+		awsDebug, export, prettyPrint, skipWait bool
+		s3URI                                   string
 	}{}
 )
 
@@ -33,6 +33,7 @@ func init() {
 	cmdUp.Flags().BoolVar(&upOpts.prettyPrint, "pretty-print", false, "Pretty print the resulting CloudFormation")
 	cmdUp.Flags().BoolVar(&upOpts.awsDebug, "aws-debug", false, "Log debug information from aws-sdk-go library")
 	cmdUp.Flags().StringVar(&upOpts.s3URI, "s3-uri", "", "When your template is bigger than the cloudformation limit of 51200 bytes, upload the template to the specified location in S3. S3 location expressed as s3://<bucket>/path/to/dir")
+	cmdUp.Flags().BoolVar(&upOpts.skipWait, "skip-wait", false, "Don't wait for the cluster components be ready")
 }
 
 func runCmdUp(cmd *cobra.Command, args []string) error {
@@ -57,7 +58,7 @@ func runCmdUp(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Failed to read cluster config: %v", err)
 	}
 
-	opts := stackTemplateOptions(upOpts.s3URI, upOpts.prettyPrint)
+	opts := stackTemplateOptions(upOpts.s3URI, upOpts.prettyPrint, upOpts.skipWait)
 
 	cluster, err := cluster.NewCluster(conf, opts, upOpts.awsDebug)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


mumoshu:
Hi @gianrubio, thanks for the PR!
Btw is this still WIP?
I'm willing to merge this once the conflicts are resolved but it's also ok for me to re-implement your work myself after #315 is merged, because #315 greatly affects code related to this work.

gianrubio:
yes @mumoshu, you can merge this

config/config.go
```diff
@@ -518,7 +515,15 @@ func (c WorkerSettings) MaxWorkerCount() int {
...
-		return c.MaxWorkerCount() - 1
+
+		maxWorkerCount := c.MaxWorkerCount()
+
+		//never accept worker size = 0
+		if maxWorkerCount > 0 {
+			return maxWorkerCount - 1
+		}
+
+		return 1
```
mumoshu: Any reason you don't want to return 0 instead of 1 as the minimum value?
I guess returning 1 here would make it impossible to set size of the ASG to zero, which is currently possible and even possible after #315?

config/config.go
```diff
@@ -539,7 +544,15 @@ func (c ControllerSettings) MaxControllerCount() int {
...
-		return c.MaxControllerCount() - 1
+
+		maxControllerCount := c.MaxControllerCount()
+
+		//never accept controller size = 0
+		if maxControllerCount > 0 {
+			return maxControllerCount - 1
+		}
+
+		return 1
```
mumoshu: Just curious but would you also like to make the minimum number of controller nodes returned by `MinControllerCount()` to 1?
If you agree, I'll make it so.

cmd/up.go
```diff
@@ -83,13 +84,13 @@ func runCmdUp(cmd *cobra.Command, args []string) error {
 		if err := ioutil.WriteFile(templatePath, stackTemplate, 0600); err != nil {
 			return fmt.Errorf("Error writing %s : %v", templatePath, err)
 		}
-		if conf.KMSKeyARN == "" {
+		if conf.KMSKeyARN != "" {
 			fmt.Printf("BEWARE: %s contains your TLS secrets!\n", templatePath)
```
mumoshu: Btw do you know what is this message meant for?
AFAIK, stack templates doesn't contain raw TLS secrets but KMS-encrypted versions of them.
If so, why do we need a warning(?) like this?

cmd/up.go
```diff
@@ -83,13 +84,13 @@ func runCmdUp(cmd *cobra.Command, args []string) error {
 		if err := ioutil.WriteFile(templatePath, stackTemplate, 0600); err != nil {
 			return fmt.Errorf("Error writing %s : %v", templatePath, err)
 		}
-		if conf.KMSKeyARN == "" {
+		if conf.KMSKeyARN != "" {
```
mumoshu: Good catch!

mumoshu:
Sorry but I now realize that there're also minor changes irrelevant to the main topic of this PR!
I've left comments on those.
Could you check into those before I merge this?

config/config.go
gianrubio: Fixed the min instance for rolling update in the NewDefaultCluster() to 1. If someone doesn't like this value can easily change on cluster.yaml. 

Why this default value? If you don't specify this option and you have only 1 worker/controller you'll have a downtime when upgrading your cluster.
With this option, UpdateMinInstancesInService, you could setup your minsize and maxsize to 0 without problems.

cmd/up.go
gianrubio: I removed the kms warning (doesn't make sense)

gianrubio:
@mumoshu I'm fixing the testes, wait for merge!

gianrubio:
@mumoshu done!

mumoshu:
LGTM. Thanks for your contribution @gianrubio!

PR #313, opened 2017-02-09
Fix error message when failed to init cluster

cheungpat:
The error message should include the message from the original error, but
the original error is not included in error message.

This will be printed out when an error occurred:

```
Error: Failed to initialize cluster driver : <nil>
```


philidem:
I implemented the same fix and then came here to see if there was already pull request. Looks like you beat me to it.

Change looks good!

gianrubio:
Me either, I did on #304 :)

mumoshu:
@cheungpat @philidem @gianrubio Good catch! Thanks for your supports :bow:

Issue #271, opened 2017-01-19
My findings and suggestions


sasso:
After working on this project I would like to share some of my findings and give some suggestions on how we can improve it. This is my view and I would love to hear maintainers thoughts on this and get a better sense of what your roadmap is and where you see this project in 3 months.

**1. ETCD deployment and management should not be a part of this project.**
This is one of the biggest issues for me and it makes running this project in production a real pain. It would be amazing if this could be managed with a tool like https://github.com/crewjam/etcd-aws or something similar. This would bring stability and security when dealing with ETCD.

**2. Project should be focused on both new and existing infrastructures**
By using resource tags like `Kube-AWS: <cluster>-etcd-node`, `Kube-AWS: <cluster>-worker`, `Kube-AWS: <cluster>-subnet` we would be able to reconstruct state and generate better CF templates. Based on state findings we could decide if we need to create new resources or if we should reuse existing ones.

**3. Node pools should be more tightly integrated**
There is a lot of separation between main cluster and node pools. It would be nice if we would only have controller at the top level and then worker pools that could all reside in one cluster config for easier maintenance. You would inherit default options and if needed override them on a node cluster level. It would also be nice if controller and node clusters would be a separated CF's as there is no big dependency between them

**4. Extending should be easy**
With hooks we could inject custom unit files that would allow us to extend the system and support some custom needs that you don't want in master

**5. No experimental tags**
It makes no sense to me to have experimental tags, in the long run it only makes more work for developers and users as they need to modify their templates. It would also be nice to break CF into smaller pieces and include parts/features when needed. By doing that we would move if's and loops out of templates and make codebase more maintainable.

Overall this is a great project and I'm looking forward to seeing it grow and mature. Great work guys!

redbaron:
Here is my view:

1. don't agree, but **optionally** being able to point to existing etcd cluster and not to create it is a good feature to have.
2.you mean make kube-aws adopt existing setups? generally not possible, as existing resources cannot be "imported" into CF template. achieving this goal would mean ditching CF and move towards kops model. Use of CF is a huge win for **kube-aws**  for multiple reasons, freedom of customization is just one of them.
3. totally agree, see https://github.com/coreos/kube-aws/issues/238#issuecomment-272352977
4. already possibly by utilizing git workflow roughly described in a discussion https://github.com/coreos/kube-aws/issues/238#issuecomment-272047491 and below
5. agreee


I'd like to add another point:
[...]

Issue #238, opened 2017-01-12
Support the use-case to manage multiple kube-aws clusters' configurations optionally inheriting organization-specific customizations with a version control system like Git


redbaron:
We use git branching to track customizations on top of kube-aws generated CF/userdata. Having nodepools as a separate level with arbitrary paths complicates things for us. It also leads to not insignificant code duplication in `nodepool/`.

What was justification for such split? would you consider PR which manages nodepools as part of stack-template.json? Either embedded or as a nested CF stack.

mumoshu:
@redbaron We've split templates to:

* Avoid the size limitation of cfn stack templates and
* To quit hard-coding vars like vpcID, routeTableID, workerSecurityGroupIds into the resulting/exported cfn template (#195 is wip) in the future

Also, would you mind sharing me example(s) of the "insignificant code duplication" you've mentioned? Anyways I guess we could fix that separately without embedded/nested CF stack.

mumoshu:
Could you also share your thoughts on how would the whole files and directories would be structured after the change you've wanted, or does fixing just `stack-template.json` solve your issue?

redbaron:
OK, lets discuss that:
 * AFAIK stack size limitation when using s3 bucket and current main stack already reaches that size.
 * They are already "hardcoded" for main template, right? If nodepool is just another ASG in main CF or a nested CF stack which uses values from parent, how is it worse than what we have now?
 * From the brief look `nodepool/config/config.go` is a subset of `config/config.go`: validation, ClusterFromFile and other funcs are pretty much mirroring what is already coded in config/config.go.

redbaron:
> Could you also share your thoughts on how would the whole files and directories would be structured after the change you've wanted, or does fixing just stack-template.json solve your issue?

Well, having one `cluster.yaml` and one `stack-template.json` seems to be capable of supporting nodepools just fine. Luckily there is worker userdata is already reused

mumoshu:
Thanks for the additional info!

> Well, having one cluster.yaml and one stack-template.json seems to be capable of supporting nodepools just fine. Luckily there is worker userdata is already reused

Sorry for missing the context here but anyways I'd like to inform that user-data/cloud-config-worker is managed separately among a main stack and node pools to allow customization per main/node pool.

mumoshu:
However, for example if duplicated blocks in  `cloud-config-worker` files are a problem, we'd probably introduce e.g. a "cloud-config template helper" to reduce them.

[...]

Issue #168, opened 2016-12-15
Controller autoscale


viktordoit:
is there any issue with this? anybody tried to do so?

mumoshu:
Haven't tried yet but definitely I'm interested.

mumoshu:
Closing this due to inactivity but please feel free to reopen if anyone has a concrete use-case.

PR #315, opened 2017-02-10
Easier to configure, easier to version-control, more tightly integrated node pools

mumoshu:
This is an implementation of #238 from @redbaron especially what I've described in my comment there https://github.com/coreos/kube-aws/issues/238#issuecomment-272128944, and an answer to the request "**3. Node pools should be more tightly integrated**" of #271 from @sasso .
I believe this also achieves what was requested by @andrejvanderzee in https://github.com/coreos/kube-aws/issues/176#issuecomment-268117511.

After applying this change:

1. All the `kube-aws node-pools` sub-commands are dropped
2. You can now bring up a main cluster and one or more node pools at once with `kube-aws up`
3. You can now update all the sub-clusters including a main cluster and node pool(s) by running  `kube-aws update`
4. You can now destroy all the AWS resources spanning main and node pools at once with `kube-aws destroy`
5. You can configure node pools by defining a `worker.nodePools` array in cluster.yaml`
6. `workerCount` is dropped. Please migrate to `worker.nodePools[].count`
7. `node-pools/` and hence `node-pools/<node pool name>` directories, `cluster.yaml`, `stack-template.json`, `user-data/cloud-config-worker` for each node pool are dropped.
8. A typical local file tree would now look like:
  - `cluster.yaml`
  - `stack-templates/` (generated on `kube-aws render`)
     - `root.json.tmpl`
     - `control-plane.json.tmpl`
     - `node-pool.json.tmpl`
  - `userdata/`
     - `cloud-config-worker`
     - `cloud-config-controller`
     - `cloud-config-etcd`
  - `credentials/`
     - *.pem(generated on `kube-aws render`)
     - *.pem.enc(generated on `kube-aws validate` or `kube-aws up`)
  - `exported/` (generated on `kube-aws up --export --s3-uri <s3uri>`)
     - `stacks/`
       - `control-plane/`
         - `stack.json`
         - `user-data-controller`
       - `<node pool name = stack name>/`
         - `stack.json`
         - `user-data-worker`
9. A typical object tree in S3 would now look like:
  - `<bucket and directory from s3URI>`/
    - kube-aws/
      - clusters/
        - `<cluster name>`/
          - `exported`/
            - `stacks`/
              - `control-plane/`
                - `stack.json`
                - `cloud-config-controller`
              - `<node pool name = stack name>`/
                - `stack.json`

Implementation details:

Under the hood, kube-aws utilizes CloudFormation nested stacks to delegate management of multiple stacks as a whole.
kube-aws now creates 1 root stack and nested stacks including 1 main(or currently named "control plane") stack and 0 or more node pool stacks.
kube-aws operates on S3 to upload all the assets required by all the stacks(root, main, node pools) and then on CloudFormation to create/update/destroy a root stack.

An example `cluster.yaml`  I've been used to test this looks like:

```yaml
clusterName: <your cluster name>
externalDNSName: <your external dns name>
hostedZoneId: <your hosted zone id>
keyName: <your key name>
kmsKeyArn: <your kms key arn>
region: ap-northeast-1
createRecordSet: true
experimental:
  waitSignal:
    enabled: true
subnets:
- name: private1
  availabilityZone: ap-northeast-1a
  instanceCIDR: "10.0.1.0/24"
  private: true
- name: private2
  availabilityZone: ap-northeast-1c
  instanceCIDR: "10.0.2.0/24"
  private: true
- name: public1
  availabilityZone: ap-northeast-1a
  instanceCIDR: "10.0.3.0/24"
- name: public2
  availabilityZone: ap-northeast-1c
  instanceCIDR: "10.0.4.0/24"
controller:
  subnets:
  - name: public1
  - name: public2
  loadBalancer:
    private: false
etcd:
  subnets:
  - name: public1
  - name: public2
worker:
  nodePools:
  - name: pool1
    subnets:
    - name: asgPublic1a
  - name: pool2
    subnets: # former `worker.subnets` introduced in v0.9.4-rc.1 via https://github.com/coreos/kube-aws/pull/284
    - name: asgPublic1c
    instanceType: "c4.large" # former `workerInstanceType` in the top-level
    count: 2 # former `workerCount` in the top-level
    rootVolumeSize: ...
    rootVolumeType: ...
    rootVolumeIOPs: ...
    autoScalingGroup:
      minSize: 0
      maxSize: 10
    waitSignal:
      enabled: true
      maxBatchSize: 2
  - name: spotFleetPublic1a
    subnets:
    - name: public1
    spotFleet:
      targetCapacity: 1 
      unitRootVolumeSize: 50
      unitRootvolumeIOPs: 100
      rootVolumeType: gp2
      spotPrice: 0.06
      launchSpecifications:
      - spotPrice: 0.12
         weightedCapacity: 2
         instanceType: m4.xlarge
        rootVolumeType: io1
        rootVolumeIOPs: 200
        rootVolumeSize: 100
```

cc @redbaron @sasso @andrejvanderzee As I remember this is most requested by you, would you mind taking a look into this? Tests are failing but it is already working with the above `cluster.yaml`.

Thanks for your supports!

TODOs
- [x] Drop worker nodes from the main stack
- [x] Create node pool(s) by default
- [x] Update `Documentation/*.md`
- [x] Fix integration tests
- [x] Add integration tests
- [x] Fix E2E tests
- [x] Pass E2E tests

List of planned changes until this PR is finished:

- [x] Rename `nodePools[].nodePoolName` to just `nodePools[].name`
- [x] Rename `nodePools` to `worker.nodePools`
- [x] Remove prefixes of `worker*` e.g. `worker.nodePools[].workerRootVolumeSize` to [ ] `worker.nodePools[].rootVolumeSize`
- [x] Flatten `worker.nodePools[].experimental.*` to `worker.nodePools[]`. For example `worker.nodePools[].experimental.waitSignal` would be just `worker.nodePools[].waitSignal`

Things won't be addressed in this PR:
- Transform `Fn::ImportValue`s into `Default`s of stack parameters and put `Ref`s in place instead (#195 from @icereval, https://github.com/coreos/kube-aws/pull/284#issuecomment-275365832 from @redbaron)
- Transform `Fn::ImportValue`s into `Ref`s of stack parameters while providing the parameters via the root stack (https://github.com/coreos/kube-aws/issues/238#issuecomment-272050366 from @redbaron)
- Large, fundamental refactoring of existing code
- Default settings for `worker.experimental.*` from the top level `experimental.*`
- Default settings for `worker.nodePools[].experimental.*` from the upper level `worker.experimental.*`
- Default settings for `worker.nodePools[].experimental.*` from the top level `experimental.*`
- Selectively deprecate or drop all the former `worker*` keys in the top-level of cluster.yaml


mumoshu:
Update the description several times to include all the topics in my mind.

@redbaron @sasso @andrejvanderzee I wish you like this. I'd appreciate your feedbacks if any!

sasso:
Before I dive into code and testing. Do we need to keep the worker or could we only use the node pools to make things simple?

mumoshu:
@sasso I'm believing/assuming that we don't need workers in a main cluster anymore. Node pools can completely replace workers in a main cluster and are more flexible.

sasso:
just saw the comment in the template 😄

cmd/init.go
```diff
@@ -35,22 +33,13 @@ func init() {
...
-	for _, req := range required {
-		if req.val == "" {
-			missing = append(missing, strconv.Quote(req.name))
-		}
-	}
-	if len(missing) != 0 {
-		return fmt.Errorf("Missing required flag(s): %s", strings.Join(missing, ", "))
+	if err := validateRequired(
+		flag{"--cluster-name", initOpts.ClusterName},
+		flag{"--external-dns-name", initOpts.ExternalDNSName},
```
sasso: Anyone else thinks we should drop `--external-dns-name`,  `--region` and `--availability-zone` and validate later when we know if you are doing HA

sasso:
after a quick test I'm stuck on 

```
(model.Worker) {
 AutoScalingGroup: (model.AutoScalingGroup) {
  MinSize: (int) 0,
  MaxSize: (int) 0,
  RollingUpdateMinInstancesInService: (int) 0
 },
 ClusterAutoscaler: (model.ClusterAutoscaler) {
  MinSize: (int) 0,
  MaxSize: (int) 0
 },
 SpotFleet: (model.SpotFleet) {
  TargetCapacity: (int) 0,
  SpotPrice: (string) "",
  IAMFleetRoleARN: (string) "",
  RootVolumeType: (string) "",
  UnitRootVolumeSize: (int) 0,
  UnitRootVolumeIOPS: (int) 0,
  LaunchSpecifications: ([]model.LaunchSpecification) {
  }
 },
 Subnets: ([]model.Subnet) <nil>
}
Error: Failed to initialize cluster driver : file cluster.yaml: public subnets required by default for node pool missing in cluster.yaml.
define one or more public subnets in cluster.yaml or explicitly reference private subnets from node pool by specifying nodePools[].subnets[].name
exit status 2
```

```
nodePools:
- name: "general"
  subnets:
    - name: default-private.us-east-1a
    - name: default-private.us-east-1c
    - name: default-private.us-east-1d
    - name: default-private.us-east-1e
  workerInstanceType: "c4.large"
```
Seems that subnets are not getting populated (I'm testing with existing subnets). Will debug more tomorrow...

redbaron:
wow, that is a great improvement, thanks for doing this work. 

`cluster.yaml` looks exactly how I imagined it, with one exception: can we allow userdata file to be configurable per nodepool? doing that, not only worker nodes can be removed, controllers can be nodepool'ed too then

mumoshu:
@sasso Thanks for the feedback! I'll investigate it soon :bow:

mumoshu:
@redbaron Thanks for checking in!

> can we allow userdata file to be configurable per nodepool?

Probably via introducing a new setting key e.g. `controlplane.templatePath` so that we can override which template other than the default `stack-templates/controlplane` to be used by kube-aws:

```yaml
controlplane:
  templatePath: my-stack-templates/controlplane.json.tmpl`
```

>  doing that, not only worker nodes can be removed, controllers can be nodepool'ed too then

Sounds interesting!
Could you clarify your use-case(s) a bit?
Are you willing to auto-scale controller nodes with CA by having 1 controller node pool per AZ or anything else?

mumoshu:
List of planned changes in this PR:

- [ ] Rename `nodePools[].nodePoolName` to just `nodePools[].name`
- [ ] Rename `nodePools` to `worker.nodePools`
- [ ] Remove prefixes of `worker*` e.g. `worker.nodePools[].workerRootVolumeSize` to [ ] `worker.nodePools[].rootVolumeSize`
- [ ] Flatten `worker.nodePools[].experimental.*` to `worker.nodePools[]`. For example `worker.nodePools[].experimental.waitSignal` would be just `worker.nodePools[].waitSignal`

c-knowles:
@mumoshu I'm trying out this branch, currently getting this:

```
> kube-aws render stack
panic: runtime error: integer divide by zero

goroutine 1 [running]:
panic(0x6488a0, 0xc4200121d0)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/coreos/kube-aws/core/controlplane/config.Cluster.Config(0xc4200d5540, 0x1a, 0x73366a, 0x9, 0x0, 0x0, 0x0, 0x0, 0xc4201267d0, 0xb, ...)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/core/controlplane/config/config.go:589 +0x9d1
github.com/coreos/kube-aws/core/root.StackAssetsRendererFromFile(0x7356fc, 0xc, 0x5f1bc0, 0xc420126639, 0xc420147c97, 0x4)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/core/root/renderer.go:13 +0x87
github.com/coreos/kube-aws/cmd.runCmdRenderStack(0xb4e260, 0xb6e9b0, 0x0, 0x0, 0x0, 0x0)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/cmd/render.go:71 +0x3d
github.com/coreos/kube-aws/vendor/github.com/spf13/cobra.(*Command).execute(0xb4e260, 0xb6e9b0, 0x0, 0x0, 0xb4e260, 0xb6e9b0)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/vendor/github.com/spf13/cobra/command.go:632 +0x23e
github.com/coreos/kube-aws/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xb4e480, 0x0, 0x0, 0x0)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/vendor/github.com/spf13/cobra/command.go:722 +0x367
github.com/coreos/kube-aws/vendor/github.com/spf13/cobra.(*Command).Execute(0xb4e480, 0xc420147f50, 0xc4200001a0)
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/vendor/github.com/spf13/cobra/command.go:681 +0x2b
main.main()
	/Users/cknowles/Projects/go/src/github.com/coreos/kube-aws/main.go:9 +0x2d
make: *** [kube-aws-render-new] Error 2
```

Culprit seems to be this line:
`subnetIndex := etcdIndex % len(config.Etcd.Subnets)`

I setup a brand new stack and did very little customisation to cluster.yaml. The main part I added was trying to add 3 private subnets against existing NAT Gateways ready for workers to be deployed:
```
subnets:
  - name: ManagedPrivateSubnetWithExistingNGW1
    private: true
    availabilityZone: eu-west-1a
    instanceCIDR: "10.0.18.0/24"
    natGateway:
      idFromStackOutput: vpc-nat-gateway-1-stg
  - name: ManagedPrivateSubnetWithExistingNGW2
    private: true
    availabilityZone: eu-west-1b
    instanceCIDR: "10.0.19.0/24"
    natGateway:
      idFromStackOutput: vpc-nat-gateway-2-stg
  - name: ManagedPrivateSubnetWithExistingNGW3
    private: true
    availabilityZone: eu-west-1c
    instanceCIDR: "10.0.20.0/24"
    natGateway:
      idFromStackOutput: vpc-nat-gateway-3-stg
```

I haven't checked it fully but I guess etcd subnets are not defaulting when some subnets are specified. However, cluster.yaml states this in etcd section - `# If omitted, public subnets are created by kube-aws and used for etcd nodes`.

mumoshu:
@c-knowles Thanks as always for your feedback!
I'm sorry for confusing you but the comment `# If omitted, public subnets are created by kube-aws and used for etcd nodes.` applies only when the `subnets` key is completely missing in cluster.yaml.
If you do define `subnets:` with only the private subnets, you have to explicitly pass which private subnet(s) etcd nodes gets created like:

```
etcd:
  subnets:
  - name: ManagedPrivateSubnetWithExistingNGW1
  - name: ManagedPrivateSubnetWithExistingNGW2
  - name: ManagedPrivateSubnetWithExistingNGW3
```

I guess we can add more validation(s) and/or more kind error messages (of course wihout panicing!) but I don't have a concrete idea about those yet.

c-knowles:
@mumoshu I can confirm as soon as I set `etcd.subnets` to something the stack rendered. I'm not sure how much access the etcd nodes need so I set to the same private subnets with NAT for now.

c-knowles:
@mumoshu I think for now, it's enough for make the cluster.yaml clear as below. For future, could kube-aws do the defaulting where possible? I mean, do we have to specify all subnets if we just want to specify one?

```
# If both `etcd.subnets` and the top level `subnets` are omitted, public subnets are created by kube-aws and used for etcd nodes. If the top level `subnets` is present then `etcd.subnets` must be specified.
```

c-knowles:
I guess the same confusion causes this for API ELB on `up`, this time on the controllers though `render` works but it fails during stack creation:
```
11:33:45 UTC+0800	CREATE_FAILED	AWS::ElasticLoadBalancing::LoadBalancer	ElbAPIServer	Either AvailabilityZones or SubnetIds must be specified
```

mumoshu:
@c-knowles 

> I think for now, it's enough for make the cluster.yaml clear as below

Seems like so! 👍

c-knowles:
@mumoshu I also found that a combination of existing VPC, no IGW specification and some managed public subnets means that an IGW is created by kube-aws as expected but it is not attached to the VPC. As far as I can tell this is supported in theory just the `VPCGatewayAttachment` definition inside `controlplane` is inside the `{{if not .VPCID}}` block.

mumoshu:
@c-knowles I really appreciate your support here!

Sorry for the confusion again but specifying `vpcId` means that an user is responsible (1) to create an appropriate VPC gateway attachment for the existing VPC specified by `vpcId` and (2) to provide an appropriate `internetGatewayId` for the managed subnets or (3) to provide unmanaged, exisiting subnets with routes to the IGW already configured instead.

The general rule behind it is that kube-aws tries its best not to modify existing AWS resources - which is the VPC specified by vpcId in this case.

All kube-aws will do when chosen managed public subnets + existing vpc combination are creating managed subnets and associating the internet gateway specified by `internetGatewayId` to those subnets.
I'd like to add more validations and/or documentation(s) on demand. Please tell me here or more better submit another github issues to express demands!

c-knowles:
@mumoshu sure yes, I thought that's what you were going for it's just not entirely clear in the current cluster.yaml I think. I'll file some issues shortly once I'd gone through a few more parts.

c-knowles:
BTW, this is looking awesome! I'd held off on using node pools until this point but now it's so much easier to manage the various parts.

redbaron:
>> doing that, not only worker nodes can be removed, controllers can be nodepool'ed too then
>Sounds interesting!
>Could you clarify your use-case(s) a bit?
>Are you willing to auto-scale controller nodes with CA by having 1 controller node pool per AZ or anything else?

Not so much for autoscaling, but unifying all but etcd node management configuration under the same config structure and probably unify some code path in kube-aws codebase too.

mumoshu:
Just finished TODOs in https://github.com/coreos/kube-aws/pull/315#issuecomment-279264129 via 	3b65538

@c-knowles @sasso fyi `nodePools` are now moved to `worker.nodePools`. A working example of cluster.yaml for me is now:


```yaml
clusterName: <myclustername>
externalDNSName: <myexternaldnsname>
hostedZoneId: <myhostedzoneid>
keyName: <mykeyname>
kmsKeyArn: <mykmskeyarn>
region: ap-northeast-1
createRecordSet: true
experimental:
  waitSignal:
    enabled: true
subnets:
- name: private1
  availabilityZone: ap-northeast-1a
  instanceCIDR: "10.0.1.0/24"
  private: true
- name: private2
  availabilityZone: ap-northeast-1c
  instanceCIDR: "10.0.2.0/24"
  private: true
- name: public1
  availabilityZone: ap-northeast-1a
  instanceCIDR: "10.0.3.0/24"
- name: public2
  availabilityZone: ap-northeast-1c
  instanceCIDR: "10.0.4.0/24"
controller:
  subnets:
  - name: public1
  - name: public2
  loadBalancer:
    private: false
etcd:
  subnets:
  - name: public1
  - name: public2
worker:
  nodePools:
  - name: "nodepool1"
    availabilityZone: "ap-northeast-1a"
    subnets:
    - name: public1
      availabilityZone: "ap-northeast-1a"
      instanceType: "c4.large"
#- name: "nodepool2"
#  availabilityZone: "ap-northeast-1b"
#  subnets:
#  - name: public2
#    availabilityZone: "ap-northeast-1b"
```

mumoshu:
@sasso The bug which had been preventing subnets from being populated seems to be fixed according to the integration tests. Would you mind trying it again?

sasso:
@mumoshu did a quick run

```
Error: Failed to initialize cluster driver : failed to load node pool #0: failed to render worker cloud config: template: userdata/cloud-config-worker:212:12: executing "userdata/cloud-config-worker" at <.Worker.SpotFleet.En...>: can't evaluate field Worker in type *config.ComputedConfig
exit status 2
```

mumoshu:
@sasso Thanks for testing! It seems like you need to rm `user-data/*` and re-run `kube-aws render`

mumoshu:
@redbaron 

> Not so much for autoscaling, but unifying all but etcd node management configuration under the same config structure and probably unify some code path in kube-aws codebase too.

It turned to be true! Almost everything which was assumed to be specific to worker node pools are now generalized into the same config structure:

```golang
type NodePoolConfig struct {
  AutoScalingGroup  AutoScalingGroup  `yaml:"autoScalingGroup,omitempty"`
  ClusterAutoscaler ClusterAutoscaler `yaml:"clusterAutoscaler"`
  SpotFleet         `yaml:"spotFleet,omitempty"`
  Subnets           []Subnet `yaml:"subnets,omitempty"`
  SpotFleet         SpotFleet         `yaml:"spotFleet,omitempty"`
  Count             int               `yaml:"count,omitempty"`
  CreateTimeout     string            `yaml:"createTimeout,omitempty"`
  InstanceType      string            `yaml:"instanceType,omitempty"`
  RootVolume        `yaml:",inline"`
  SpotPrice         string   `yaml:"spotPrice,omitempty"`
  SecurityGroupIds  []string `yaml:"securityGroupIds,omitempty"`
  Tenancy           string   `yaml:"tenancy,omitempty"`
}
```

Hence each worker node pool is now represented in code as:

```golang
type WorkerNodePoolConfig struct {
  model.NodePoolConfig `yaml:",inline"`
}
```

In the future, we could make another refactoring(s) to make controller node pool'ed by something like:

```golang
type ControllerodePoolConfig struct {
  model.NodePoolConfig `yaml:",inline"`
}
```

mumoshu:
* Integration tests WITHOUT real AWS services are now passing on Travis
* Integration tests WITH real AWS services are now passing on my laptop

c-knowles:
@mumoshu I've tried out the latest branch, so far no major additional issues. Is it possible to set a node pool ASG min to zero? I think it used to be possible when we added the ASGs but I can't seem to get it to be set now. I've tried `worker.nodePools[].autoScalingGroup.minSize` and `worker.nodePools[].count`.

c-knowles:
I think there could be an issue with worker nodepools with wait signal in this branch. If I enable the wait signal, then CF `MaxBatchSize` is always zero which is not allowed on AWS. Setting `nodepools[].waitSignal.maxBatchSize` does not seem to solve it. I think the smallest worker config to reproduce is:

```
worker:
  nodePools:
    - name: worker-pool-a
      availabilityZone: eu-west-1a
      subnets:
        - name: ManagedPrivateSubnetWithExistingNGW1
      waitSignal:
        enabled: true
        # maxBatchSize: 1 # adding/removing this does not work around the issue
```

mumoshu:
@c-knowles Thanks for reporting!

I've fixed the above two issues.

You can choose one of the below to have no worker nodes per node pool:

* set `worker.nodePools[].count` to zero
* set `worker.nodePools[].autoScalingGroup.minSize` to zero

Also, setting `waitSignal.maxBatchSize` would just work now.

c-knowles:
@mumoshu great, that was quick! I will try it out soon.

core/nodepool/config/config.go
```diff
@@ -1,6 +1,6 @@
 package config
 
-//go:generate go run ../../codegen/templates_gen.go DefaultClusterConfig=cluster.yaml StackTemplateTemplate=stack-template.json
+//go:generate go run ../../../codegen/templates_gen.go DefaultClusterConfig=cluster.yaml StackTemplateTemplate=stack-template.json
 //go:generate gofmt -w templates.go
```
c-knowles: Now these files have moved, we should update [.gitignore](https://github.com/coreos/kube-aws/blob/master/.gitignore#L5) to match. I suggest `**/config/templates.go`.

mumoshu: Thanks again and good catch! Updated the .gitignore to reflect your suggestion.

redbaron:
@mumoshu , why did you decide to move `nodepools` under `worker` key? is there going to be anything else but `nodepools`?

cmd/init.go
gianrubio: I vote to move this to the cluster.yaml

cmd/up.go
```diff
@@ -36,55 +31,24 @@ func init() {
...
-
-	opts := stackTemplateOptions(upOpts.s3URI, upOpts.prettyPrint)
-
-	cluster, err := cluster.NewCluster(conf, opts, upOpts.awsDebug)
-	if err != nil {
-		return fmt.Errorf("Failed to initialize cluster driver : %v", cluster)
-	}
-
-	if err := cluster.ValidateUserData(); err != nil {
+	if err := validateRequired(flag{"--s3-uri", upOpts.s3URI}); err != nil {
```
gianrubio: @mumoshu could you disable this flag requirement if upOpts.export is true?

core/controlplane/config/templates/stack-template.json
```diff
@@ -1335,6 +1055,10 @@
       "Value" :  { "Ref" : "SecurityGroupWorker" },
       "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
     },
+    "StackName": {
+      "Description": "The name of this stack which cis used by node pool stacks to import outputs from this stack",
```
gianrubio: typo "cis"

cmd/up.go
mumoshu: Thanks for the question but no!
s3URI is required in order to render stack templates because the URI is parsed and then included in the stack templates as (1) URLs to actual worker/controller cloud-configs in S3 and (2) URLs to nested stack templates referenced from the root stack template.

core/controlplane/config/templates/stack-template.json
mumoshu: Good catch 👍 Fixed.

mumoshu:
@redbaron Thanks for the good question 👍 
I've moved `nodePools` to `worker.nodePools` so that we can be more explicit about that the node pools are just for worker nodes "for now". A node can actually be either a worker or a controller and therefore just saying "node pools` is a bit ambiguous.

And in the future we can also:
   * If necessary, introduce new keys like `worker.defaults.*` or even `worker.*` to configure default settings used among all the worker node pools and
   * If necessary, introduce `controller.nodePools` to make controllers to be managed as node pools while differentiating those from worker node pools to e.g. allow enabling cluster-autoscaler (#168)
      * Node pools can naturally be extended to controllers due to the fact that both workers and controllers are "nodes" in k8s. IMHO only the differences are how they're tainted or not and which pods are scheduled onto them.

mumoshu:
E2E tests passed 😮

mumoshu:
It turned out I'd broken worker node pools powered by spot fleets.
However, they're now fixed and working as before.

mumoshu:
I believe that everything is now implemented as described in the updated description of this PR.
Please tell me if anything is wrong!

core/nodepool/config/config.go
```diff
@@ -0,0 +1,328 @@
...
+func (c ProvidedConfig) StackName() string {
+	return c.NodePoolName
+}
+
+func (c ProvidedConfig) StackNameEnvFileName() string {
+	return "/etc/environment"
+}
+
+func (c ProvidedConfig) StackNameEnvVarName() string {
+	return "KUBE_AWS_STACK_NAME"
```
redbaron: is it *really* need to be configurable? :)

mumoshu: Not really, and it isn't intended to be configurable!
I just wanted to expose it like a constant from the view point of stack templates (hence it is hard-coded

mumoshu: In other words, I thought it would be better than directly embedding the var name "KUBE_AWS_STACK_NAME" multiple places in cloud-config and stack templates ❤️

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -47,17 +38,7 @@ coreos:
         Type=oneshot
         RemainAfterExit=true
         ExecStartPre=/bin/touch /etc/aws-environment
```
redbaron: Instead of having an `/opt/bin/get-stack-name`, you can have
```
Environment={{.StackNameEnvFileName}}
```

in unit files and then use `KUBE_AWS_STACK_NAME` and var inside scripts. adding `set -ue` to these scripts will make them fail loudly if vars are not set/not passed

redbaron:
Just curious, is it worth  to keep using separate CF stack per nodepool? It might be easier just to create ASG per nodepool in a main CF stack, no?

mumoshu:
BTW I'm not satisfied with the packages names like `core`, `config`, `model`, `cluster`, `cfnstack` at all.
However not to keep this open too long, I decide not to make further renames and refactorings now. I'll probably work on these later.
IMHO:
* `model` should be renamed to `config` and
* The `config` packages under `nodepool`, `controlplane` and `root` should be dropped and merged into the top-level `config`(former `model`) respectively
* The `cluster` packages under `core/nodepool` and `core/controlplane` should be dropped and merged into `core/nodepool` and `core/controlplane` respectively. I don't like `cluster.Cluster` and `cluster.Config`!
* `core` should be named `stack`

core/controlplane/config/templates/cloud-config-controller
mumoshu: Thanks. That sounds good! I'll work on it soon

mumoshu:
> Just curious, is it worth to keep using separate CF stack per nodepool? It might be easier just to create ASG per nodepool in a main CF stack, no?

It might be easier but I believe separating stacks would be better in longer term mainly because doing so (1) makes each stack template less complex hence manageable and (2) allows kube-aws to scale to dozens of node pools.

core/controlplane/config/templates/cloud-config-controller
mumoshu: `EnvironmentFile` worked like a charm. Thanks for the info 👍

mumoshu:
Finished rebasing to the latest master and adding various follows up for recently added features.

mumoshu:
Added a minor fix for kube-node-label.service about an issue got worse via some changes in this PR.

redbaron:
Wow, @mumoshu unbelievable piece work done and developed so quickly. thank you!!

sasso:
This PR is amazing, great work @mumoshu. It has been a busy week but I can't wait to spin this up!

PR #316, opened 2017-02-10
Minimal RBAC policies

### LINKED ISSUE #230: RBAC feature should also provide minimally needed roles
I enabled the RBAC feature, and then found that "everything was broken" :)

- the worker nodes couldn't connect to the api server until I provided a ClusterRoleBinding from User 'kube-worker' to ClusterRole 'system:node'
- the kube-dns pod and the dashboard pods failed and restarted because they couldn't connect to the API server, with errors indicating that the api server IP 10.3.0.1 is completely not known (`dial tcp 10.3.0.1:443 i/o timeout`). It turned out that that was true: kube-proxy was unable to talk to the API server and get the endpoints, and therefore failed to setup any iptables rules. It took me a long time to figure out how the networking is supposed to work, and then only by accident did I notice that issue.

This is my "minimal" role bindings so far:
```yaml
apiVersion: rbac.authorization.k8s.io/v1alpha1
kind: ClusterRoleBinding
metadata:
  name: worker-node
roleRef:
  kind: ClusterRole
  name: system:node
subjects:
- kind: User
  name: "kube-worker"
---
apiVersion: rbac.authorization.k8s.io/v1alpha1
kind: ClusterRoleBinding
metadata:
  name: worker-node-proxy
roleRef:
  kind: ClusterRole
  name: system:node-proxier
subjects:
- kind: User
  name: "kube-worker"
```

It would be great if this could be provided by default, and/or at least pointed to in cluster.yaml.

gianrubio:
Allow cluster services to talk with apiserver with minimal rbac policies
Fix #230

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -432,6 +432,12 @@ write_files:
           --data-binary "$1" "$2"
       }
 
+      post_yaml_status_code(){
+          /usr/bin/curl  -H "Content-Type: application/yaml" -XPOST \
+          --write-out %{http_code} --silent --output /dev/null \
+          --data-binary "$1" "$2"
+      }
+
       mfdir=/srv/kubernetes/manifests
       for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de}.yaml; do
           post_yaml "@$mfdir/$manifest" \
@@ -445,6 +451,24 @@ write_files:
           post_yaml "@$mfdir/$manifest" \
           "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
       done
+
+      {{if .Experimental.Plugins.Rbac.Enabled}}
+      # k8s > 1.5 already have this clusterRole, so we accept http.code != 200
+      mfdir=/srv/kubernetes/rbac
+      if [[ $(post_yaml_status_code "@mfdir/cluster-roles/cluster-admin.yaml" \
+          "http://127.0.0.1:8081/apis/rbac.authorization.k8s.io/v1alpha1/clusterroles") -ne "200" ]] ; then
+          echo 'Failed to create the cluster role named "cluster-admin". It is ok if you are using k8s newer than 1.5'
+      fi
+
+      post_yaml "@${mfdir}/cluster-roles/cluster-admin.yaml" \
+      "http://127.0.0.1:8080/apis/rbac.authorization.k8s.io/v1alpha1/clusterroles" |true
+
+      for manifest in {system-worker,kube-worker,kube-admin}; do
+          post_yaml "@${mfdir}/cluster-role-bindings/$manifest.yaml" \
+          "http://127.0.0.1:8080/apis/rbac.authorization.k8s.io/v1alpha1/clusterrolebindings"
+      done
+
+      {{ end }}
 
   - path: /etc/kubernetes/cni/docker_opts_cni.env
     content: |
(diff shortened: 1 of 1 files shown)


config/templates/cloud-config-controller
```diff
@@ -693,6 +705,66 @@ write_files:
...
+        apiVersion: rbac.authorization.k8s.io/v1alpha1
+        metadata:
+            name: full-access
+        rules:
+          - apiGroups: ["*"]
+            resources: ["*"]
+            verbs: ["*"]
+          - nonResourceURLs: ["*"]
+            verbs: ["*"]
+
```
whereisaaron: I've noticed recent kube-aws 1.5.1 clusters come bootstrapped with a 'cluster-admin' cluster role that is the same as the 'full-access' above. Referencing that could reduce this patch. On the other hand, it is nice that the RBAC in those patch is self-contained and without implicit dependencies.

gianrubio: @whereisaaron I renamed the full-access to cluster-admin, so we make this cluster role compatible with k8s < 1.5.1.

whereisaaron: @gianrubio won't that clash with the same-named cluster-admin role created the the RBAC bootstrap? Or will one just replace the other?

Do people use new versions of kube-aws to create old k8s version clusters? Or can we just support from the version that is associated with each kube-aws release? @mumoshu ?

gianrubio: Definitely no, if the same role already exist I'll ignore. 
Take a look on https://github.com/gianrubio/kube-aws/blob/394350c27440e60050fa3f01e21536c744e1bbf8/config/templates/cloud-config-controller#L384

Both versions will use the same role.

config/templates/cloud-config-controller
```diff
@@ -693,6 +706,66 @@ write_files:
            $kubectl uncordon "$hostname"
            echo "done."'
 
+  {{if .Experimental.Plugins.Rbac.Enabled }}
+
+  #https://github.com/coreos/kube-aws/issues/230#issuecomment-274946933
+  - path: /srv/kubernetes/rbac/clusterrole.yaml
```
mumoshu: nit but how about organizing and naming yaml files after their kinds and `metadata.name` fields like:

* /srv/kubernetes/rbac
  * cluster-roles/
     * cluster-admin.yaml
  * cluster-role-bindings/
     * kube-admin.yaml 
     * kube-worker.yaml
     * system-default.yaml

config/templates/cloud-config-controller
```diff
@@ -380,6 +380,19 @@ write_files:
           "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services"
       done
 
+      {{if .Experimental.Plugins.Rbac.Enabled}}
+      # k8s > 1.5 already have this clusterRole, so we accept http.code != 200
+      mfdir=/srv/kubernetes/rbac
+      post_yaml "@${mfdir}/clusterrole.yaml" \
+      "http://127.0.0.1:8080/apis/rbac.authorization.k8s.io/v1alpha1/clusterroles" |true
```
mumoshu: nit but could you print a message when the code isn't 200? probably something like `failed to create the cluster role named "cluster-admin". It's ok if you're using k8s newer than 1.5`

config/templates/cloud-config-controller
gianrubio: Done

config/templates/cloud-config-controller
gianrubio: Done

gianrubio:
@mumoshu I did all the changes.

BTW I bootstrapped my cluster withe the new node pools feature. All the resources are successfully created, congrats!

mumoshu:
Thanks so much for your kind words and quick turnaround 👍

PR #317, opened 2017-02-12
Create EBS volumes for Etcd nodes

redbaron:
Having EBS volumes which survive node termination
makes them more manageable

### DIFF
--- a/config/templates/stack-template.json
+++ b/config/templates/stack-template.json
@@ -518,6 +518,19 @@
       },
       "Type": "AWS::EC2::NetworkInterface"
     },
+    {{if not $.EtcdDataVolumeEphemeral}}
+    "InstanceEtcd{{$etcdIndex}}ebs": {
+      "Properties": {
+          "AvailabilityZone": "{{ $etcdInstance.SubnetAvailabilityZone}}",
+          "Size": "{{$.EtcdDataVolumeSize}}",
+          {{if gt $.EtcdDataVolumeIOPS 0}}
+          "Iops": "{{$.EtcdDataVolumeIOPS}}",
+          {{end}}
+          "VolumeType": "{{$.EtcdDataVolumeType}}"
+      },
+      "Type": "AWS::EC2::Volume"
+    },
+    {{end}}
     "InstanceEtcd{{$etcdIndex}}": {
       "Properties": {
         "BlockDeviceMappings": [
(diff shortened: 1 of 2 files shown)


config/templates/stack-template.json
```diff
@@ -509,7 +509,7 @@
     {{range $etcdIndex, $etcdInstance := .EtcdInstances}}
     "InstanceEtcd{{$etcdIndex}}eni": {
       "Properties": {
-          "SubnetId": {{$etcdInstance.SubnetRef}},
+          "SubnetId": {{$etcdInstance.Subnet.Ref}},
```
mumoshu: Thanks for suggesting it 👍 
However I'd prefer to keep this as before while changing `$etcdInstance.Subnet.AvailabilityZone` to `$etcdInstance.SubnetAvailabilityZone` so that we don't need to expose everything related to a `Subnet` to templates.
`SubnetRef` and `SubnetAvailabilityZone` are the only things required in stack templates, aren't they?
Could I have your thoughts?

mumoshu:
@redbaron Thanks as always for your contribution 👍 
One minor comment but LGTM overall.

Btw are you also going to try enabling EC2 instance recovery on etcd nodes?
With these EBS volumes we can start using the instance recovery feature to auto-recover an etcd node on underlying hardware failure.
However I'm not yet sure if it's useful in practice mainly due to the lack of customization points in EC2 instance status checks.
More concretely, we can't customize status checks to fail when e.g. an etcd process is unhealthy.

mumoshu:
Replying to my own comment:

> More concretely, we can't customize status checks to fail when e.g. an etcd process is unhealthy.

It won't make sense as EC2 instance recovery, by definition, seems to be intended for automating recovery from EC2 instance failure. Unhealthy etcd process should be restarted by e.g. systemd or another monitoring process. It won't be a reason not to enable EC2 instance recovery.

config/templates/stack-template.json
```diff
@@ -572,7 +575,10 @@
           }
         ],
         "Tenancy": "{{$.EtcdTenancy}}",
-        "UserData": { "Fn::FindInMap" : [ "EtcdInstanceParams", "UserData", "cloudconfig"] }
+        "UserData": { "Fn::FindInMap" : [ "EtcdInstanceParams", "UserData", "cloudconfig"] },
+        "Volumes": [
+            { "VolumeId": { "Ref": "InstanceEtcd{{$etcdIndex}}ebs" }, "Device": "/dev/xvdf" }
+        ]
```
mumoshu: I guess this should be wrapped like:

```
{{if not $.EtcdDataVolumeEphemeral}},
        "Volumes": [
            { "VolumeId": { "Ref": "InstanceEtcd{{$etcdIndex}}ebs" }, "Device": "/dev/xvdf" }
        ]
{{end}}

or won't it fail when `EtcdDataVolumeEphemeral` is true?

redbaron: of course! fixed

mumoshu:
@redbaron LGTM 👍 Thanks for your contribution :bow: 🍻

PR #328, opened 2017-02-16
Fix a typo in a image reference

ankon:


mumoshu:
Good catch! Thanks for your contribution 👍

PR #329, opened 2017-02-17
Fix various minor typos

ankon:


mumoshu:
Thanks as always, @ankon 🙏

Issue #260, opened 2017-01-17
why does kube-aws not support custom AWS domain names ?


mariusmitrofan:
I've been battling with these for over 48 hours.

The certificates that you create for `etcd2` to use automatically assume that my hostname belong to either `<region-name>.compute.internal` or `ec2.internal`, but if it doesn't ?

What do I do then?

camilb:
Hi Marius, a quick method is  to edit these lines and compile your own version of kube-aws.

        ./config/config.go:546
        ./config/tls_config.go:135
        ./config/tls_config.go:151

redbaron:
generate your worn certs for the time being

mariusmitrofan:
Thank you both.

Does running this make sense ->
```
openssl genrsa -out ca-key.pem 2048
openssl req -x509 -new -nodes -key ca-key.pem -days 10000 -out ca.pem -subj "/CN=<my_hostname_suffix>"
kube-aws render credentials --ca-cert-path=ca.pem --ca-key-path=ca-key.pem
```

The output of that would be this ->
```
root@pan:/opt/kubernetes/kubernetes-demo-6/credentials# ls -altr
total 56
-rw-r--r-- 1 root root 1679 Jan 17 13:17 ca-key.pem
-rw-r--r-- 1 root root 1107 Jan 17 13:17 ca.pem
-rw------- 1 root root 1168 Jan 17 13:19 worker.pem
-rw------- 1 root root 1675 Jan 17 13:19 worker-key.pem
-rw------- 1 root root 1168 Jan 17 13:19 etcd.pem
-rw------- 1 root root 1679 Jan 17 13:19 etcd-key.pem
-rw------- 1 root root 1099 Jan 17 13:19 etcd-client.pem
-rw------- 1 root root 1679 Jan 17 13:19 etcd-client-key.pem
-rw------- 1 root root 1294 Jan 17 13:19 apiserver.pem
-rw------- 1 root root 1679 Jan 17 13:19 apiserver-key.pem
-rw------- 1 root root 1090 Jan 17 13:19 admin.pem
-rw------- 1 root root 1679 Jan 17 13:19 admin-key.pem
drwxr-xr-x 3 root root 4096 Jan 17 13:20 ..
drwx------ 2 root root 4096 Jan 17 13:20 .
```

mariusmitrofan:
Later update for whoever follows this thread:

[...]

Issue #27, opened 2016-11-02
etcd management


pieterlange:
There are currently some people forking because we're not sure about the current etcd solution. Lets discuss the issues in this topic. A lot of us seem to center around @crewjam's etcd solution but there's also others:
  * https://github.com/MonsantoCo/etcd-aws-cluster
  * https://github.com/sttts/elastic-etcd
  
I have a personal preference for https://crewjam.com/etcd-aws/ (https://github.com/crewjam/etcd-aws) but we should definitely have this conversation as a community (as i tried in the old repo https://github.com/coreos/coreos-kubernetes/pull/629) 
Lets combine our efforts @colhom @camilb @dzavalkinolx

Branches for inspiration:
  * https://github.com/pieterlange/kube-aws/tree/feature/external-etcd (refers to etcd-aws managed by separate cloudformation template)
  * @camilb's work
    * https://github.com/camilb/etcd-aws/tree/ssl (SSL support for etcd-aws)
    * https://github.com/camilb/coreos-kubernetes/tree/etcd-asg

Currently missing features:
- [ ] backup/restore
- [ ] node cycling
- [ ] node discovery from ASG
- [ ] cluster recovery from complete failure

As noted in the overall production readiness issue #9 there's also work being done on etcd being hosted inside of kubernetes itself, which is probably where all of this is going in the end.

pieterlange:
I should add that i've been running a test cluster with my monkeypatch branch and it seems to perform well.

I've been seeing a lot of warnings like these though:
```
Nov 02 10:54:33 ip-10-50-44-174.eu-west-1.compute.internal kubelet-wrapper[1983]: W1102 10:54:33.799124    1983 reflector.go:330] pkg/kubelet/kubelet.go:384: watch of *api.Service ended with: too old resource version: 30825 (30896)
Nov 02 10:54:54 ip-10-50-44-174.eu-west-1.compute.internal kubelet-wrapper[1983]: W1102 10:54:54.711865    1983 reflector.go:330] pkg/kubelet/config/apiserver.go:43: watch of *api.Pod ended with: too old resource version: 30847 (30919)
```
_I think_ it's because the controller is configured to talk to the internal ELB and we might be hitting an etcd node that's lagging on replication but i'm not sure. It doesn't seem to break anything though.

camilb:
@pieterlange made some progress on `etcd-aws` with TLS and just saw this: 
https://github.com/coreos/etcd-operator

[...]

PR #332, opened 2017-02-20
Managed HA etcd cluster

mumoshu:
This is a WIP pull request to achieve "Managed HA etcd cluster" with private IPs resolved via public EC2 hostnames stabilized with a pool of EBS and EIP pairs for etcd nodes.

After this change, EC2 instances backing "virtual" etcd nodes are managed by an ASG.

Supported use-cases:

* Automatic recovery from temporary Etcd node failures
  * Even if all the nodes went down, the cluster recovers eventually as long as the EBS volumes aren't corrupted
* Rolling-update of the instance type for etcd nodes without downtime
  * = Scaling-out of Etcd nodes via NOT modifying the ASG directly BUT indirectly via CloudFormation stack updates
* Other use-cases implied by the fact that the nodes are managed by ASGs
* You can choose "eip" or "eni" for etcd node(=etcd member) identity via the `etcd.memberIdentityProvider` key in cluster.yaml
  * `"eip"`, which is the default setting, is recommended
  * If you want, choose `"eni"`.
  * If you choose `"eni"`, and your region has less than 3 AZs, setting `etcd.internalDomainName` to something other than default is HIGHLY RECOMMENDED to prepare for disaster recovery
  * It is an advanced option but DNS other than Amazon DNS could be used (when `memberIdentityProvider` is `"eni"`, `internalDomainName` is set, `manageRecordSets` is `false`, and every EC2 instance has a custom DNS which is capable of resolving FQDNs under `internalDomainName`)

Unsupported use-cases:

* Automatic recovery from more than `(N-1)/2` permanent Etcd nodes failure.
  * Requires etcd backups and automatic determination of whether the new etcd cluster should be created or not via `ETCD_INITIAL_CLUSTER_STATE`
* Scaling-in of Etcd nodes
  * Just remains untested because it isn't my primary focus in this area. Contributions are welcomed

Relevant issues to be (partly) resolved via this PR:

* Part(s) of #27
* Wait signal for etcd nodes. See #49
* Probably #189 #260 as this relies on stable EC2 public hostnames and AWS DNS for peer communication and discovery regardless of whether an EC2 instance relies on a custom domain/hostname or not

The general idea is to make Etcd nodes "virtual" by retaining the state and the identity of an etcd node in a pair of an EBS volume and an EIP or an ENI, respectively.
This way, we can recover/recreate/rolling-update EC2 instances backing etcd nodes without another moving parts like external apps and/or ASG lifecycle hooks, SQS queues, SNS topics, etc.

Unlike well-known etcd HA solutions like crewjam/etcd-aws and MonsantoCo/etcd-aws-cluster, this is intended to be a less flexible but a simpler alternative or the basis for introducing a similar solutions like those.

### General rules

* If you rely on Route 53 record sets, don't modify ones initially created by CloudFormation
   * Doing so breaks CloudFormation stack deletions because it has no way to know about modified record sets and therefore can't cleanly remove them.
* To prepare for a disaster recovery for a single-AZ etcd cluster(possible when the user relies on an AWS region with 2 or less AZs), use Route 53 record sets or EIPs to retain network identities among AZs
   * ENIs and EBS can't be moved to an another AZ
   * EBS volume can, however, be transferred utilizing a snapshot

### Examples of experimented but not employed strategies

* Static private IPs via a pool of ENIs dynamically assigned to EC2 instances under control of a single ASG
  * ENIs can't move around different AZs. What happens when you have 2 ENIs in and 1 ENI in different AZs and the former AZ goes down? Nothing until the AZ comes up! It isn't the degree of H/A I wish to have at all!
* Dynamic private IPs via stable hostnames using a pool of EIP&EBS pairs, single ASG
  * EBS is required in order to achieve "locking" of a pair associated to an etcd instance
    * First of all, identify the "free" pair by filtering available EBS volumes and try to associate it to the EC2 instance
    * Successful association of an EBS volume means that the paired EIP can also be associated to the instance without race conditions
  * EBS can't move around different AZs. What happens when you have 2 pairs in AZ 1 and 1 pair in AZ 2? Once the AZ 2 goes down, the options you can take are (1) manually altering AZ 2 to have 3 etcd nodes and then manually elect a new leader (2) recreate the etcd cluster within AZ 2 by modifying `etcd.subnets[]` point to AZ 2 in cluster.yaml and running `kube-aws update`, ssh into one of nodes and restore etcd state from a backup. Neither is automatic.

## TODOs

* [x] Move userdata for etcd nodes to S3 (To workaround the cfn limit of 16KB in userdata size, like we've done for worker and controller node
* [x] Make EBS volumes created but not associated on stack creation
* [x] Make EIPs created as part of a control-plane stack but not assign them
* [x] Associate a pair of an EBS volume and an EIP before starting etcd process
* [x] Make each etcd node be managed under a dedicated ASG
  * Just change `EC2:Instance` resources in cfn stack templates to a corresponding pair of a launch configuration and an ASG
* [x] Make each etcd ASG depends on the next etcd ASG
  * So that we can achieve rolling-update of etcd ASGs hence etcd nodes
* [x] Trigger cfn-signal for the etcd ASG managing an etcd EC2 instance
  * Inject the new env var `KUBE_AWS_ETCD_INDEX=<$etcdIndex>` and `KUBE_AWS_STACK_NAME` from the stack template via embedded EC2 userdata
  * `cfn-signal -e 0 --region {{.Region}} --resource {{.Etcd.LogicalName}}${{.EtcdIndexEnvVarName}} --stack ${{.StackNameEnvVarName}}`
* ~~Tweak `ExecStartPre`s of the cfn-signal.service unit if necessary~~
  * `cfn-signal.service` for etcd nodes is set to `Wants=etcd2.service` and `After=etcd2.service`
  * Are these enough to ensure that the etcd2.service is up and running?
    * When a rolling-update is in progress, don't we need to wait until a newly recreated etcd member with outdated data(=persisted in the EBS volume which had been attached to the previously terminated instance replaced by the newly created instance) to catch up with the latest data from running etcd cluster?
  * I'll leave this for further improvement(s)
* ~~Security/Fail-proof: Prevent attaching/associating wrong EBS volumes and EIPs~~
   * ~~Using [resource-level permissions for ec2:AttachVolume](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-supported-iam-actions-resources.html)~~
   * I couldn't make it work
* [x] Various clean-ups
* [x] Fix tests
* [ ] Pass E2e tests

## Non-TODOs(for now)

* Graceful termination of etcd nodes
  * like `kube-node-drainer.service` for worker nodes
  * to elect a new leader when the terminating node was the former leader

### DIFF
--- a/core/controlplane/cluster/cluster.go
+++ b/core/controlplane/cluster/cluster.go
@@ -138,7 +138,7 @@ func (c *Cluster) Assets() (cfnstack.Assets, error) {
 
 	return cfnstack.NewAssetsBuilder(c.StackName(), c.StackConfig.S3URI).
 		Add("userdata-controller", c.UserDataController).
-		Add("userdata-worker", c.UserDataWorker).
+		Add("userdata-etcd", c.UserDataEtcd).
 		Add(STACK_TEMPLATE_FILENAME, stackTemplate).
 		Build(), nil
 }
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -15,6 +15,7 @@ import (
 	"github.com/coreos/kube-aws/coreos/amiregistry"
 	"github.com/coreos/kube-aws/filereader/userdatatemplate"
 	"github.com/coreos/kube-aws/model"
+	"github.com/coreos/kube-aws/model/derived"
 	"github.com/coreos/kube-aws/netutil"
 	yaml "gopkg.in/yaml.v2"
 )
@@ -630,32 +631,10 @@ func (c Cluster) Config() (*Config, error) {
 		config.AMI = c.AmiId
 	}
 
-	config.EtcdInstances = make([]model.EtcdInstance, config.EtcdCount)
-
-	for etcdIndex := 0; etcdIndex < config.EtcdCount; etcdIndex++ {
-
-		//Round-robin etcd instances across all available subnets
-		subnetIndex := etcdIndex % len(config.Etcd.Subnets)
-		subnet := config.Etcd.Subnets[subnetIndex]
-
-		var instance model.EtcdInstance
-
-		if subnet.ManageNATGateway() {
-			ngw, err := c.FindNATGatewayForPrivateSubnet(subnet)
-
-			if err != nil {
-				return nil, fmt.Errorf("failed getting a NAT gateway for the subnet %s in %v: %v", subnet.LogicalName(), c.NATGateways(), err)
-			}
-
-			instance = model.NewEtcdInstanceDependsOnNewlyCreatedNGW(subnet, *ngw)
-		} else {
-			instance = model.NewEtcdInstance(subnet)
-		}
-
-		config.EtcdInstances[etcdIndex] = instance
-
-		//http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-private-addresses
-
+	var err error
+	config.EtcdNodes, err = derived.NewEtcdNodes(c.Etcd.Nodes, c.EtcdCluster())
+	if err != nil {
+		return nil, fmt.Errorf("failed to derived etcd nodes configuration: %v", err)
 	}
 
 	// Populate top-level subnets to model
@@ -668,6 +647,12 @@ func (c Cluster) Config() (*Config, error) {
 	config.IsChinaRegion = strings.HasPrefix(config.Region, "cn")
 
 	return &config, nil
+}
+
+func (c *Cluster) EtcdCluster() derived.EtcdCluster {
+	region := model.RegionForName(c.Region)
+	etcdNetwork := derived.NewNetwork(c.Etcd.Subnets, c.NATGateways())
+	return derived.NewEtcdCluster(c.Etcd.Cluster, region, etcdNetwork, c.EtcdCount)
 }
 
 // releaseVersionIsGreaterThan will return true if the supplied version is greater then
@@ -731,7 +716,7 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 	if stackConfig.UserDataController, err = userdatatemplate.GetString(opts.ControllerTmplFile, stackConfig.Config); err != nil {
 		return nil, fmt.Errorf("failed to render controller cloud config: %v", err)
 	}
-	if stackConfig.userDataEtcd, err = userdatatemplate.GetString(opts.EtcdTmplFile, stackConfig.Config); err != nil {
+	if stackConfig.UserDataEtcd, err = userdatatemplate.GetString(opts.EtcdTmplFile, stackConfig.Config); err != nil {
 		return nil, fmt.Errorf("failed to render etcd cloud config: %v", err)
 	}
 
@@ -750,7 +735,7 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 type Config struct {
 	Cluster
 
-	EtcdInstances []model.EtcdInstance
+	EtcdNodes []derived.EtcdNode
 
 	// Encoded TLS assets
 	TLSConfig *CompactTLSAssets
@@ -762,6 +747,18 @@ type Config struct {
 // This is NOT intended to be used to reference stack name from cloud-config as the target of awscli or cfn-bootstrap-tools commands e.g. `cfn-init` and `cfn-signal`
 func (c Cluster) StackName() string {
 	return "control-plane"
+}
+
+func (c Cluster) StackNameEnvVarName() string {
+	return "KUBE_AWS_STACK_NAME"
+}
+
+func (c Cluster) EtcdNodeEnvFileName() string {
+	return "/var/run/coreos/etcd-node.env"
+}
+
+func (c Cluster) EtcdIndexEnvVarName() string {
+	return "KUBE_AWS_ETCD_INDEX"
 }
 
 func (c Config) VPCLogicalName() string {
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -9,14 +9,20 @@ coreos:
       runtime: true
       content: |
         [Unit]
-        Description=Fetches etcd static IP addresses list from CF
+        Description=Configures EBS volume and R53 record set for this node and derives env vars for etcd bootstrap
         After=network-online.target
+        Before=format-etcd2-volume.service
 
         [Service]
+        EnvironmentFile={{.EtcdNodeEnvFileName}}
         Restart=on-failure
         RemainAfterExit=true
         ExecStartPre=/opt/bin/cfn-etcd-environment
         ExecStart=/usr/bin/mv -f /var/run/coreos/etcd-environment /etc/etcd-environment
+        TimeoutStartSec=120
+
+        [Install]
+        RequiredBy=format-etcd2-volume.service
 
     - name: etcd2.service
       drop-ins:
@@ -27,28 +33,11 @@ coreos:
             After=decrypt-tls-assets.service cfn-etcd-environment.service
 
             [Service]
-            Environment=ETCD_NAME=%H
-
-            Environment=ETCD_PEER_TRUSTED_CA_FILE=/etc/etcd2/ssl/ca.pem
-            Environment=ETCD_PEER_CERT_FILE=/etc/etcd2/ssl/etcd.pem
-            Environment=ETCD_PEER_KEY_FILE=/etc/etcd2/ssl/etcd-key.pem
-
-            Environment=ETCD_CLIENT_CERT_AUTH=true
-            Environment=ETCD_TRUSTED_CA_FILE=/etc/etcd2/ssl/ca.pem
-            Environment=ETCD_CERT_FILE=/etc/etcd2/ssl/etcd.pem
-            Environment=ETCD_KEY_FILE=/etc/etcd2/ssl/etcd-key.pem
-
-            Environment=ETCD_INITIAL_CLUSTER_STATE=new
             EnvironmentFile=-/etc/etcd-environment
-            Environment=ETCD_DATA_DIR=/var/lib/etcd2
-            Environment=ETCD_LISTEN_CLIENT_URLS=https://%H:2379
-            Environment=ETCD_ADVERTISE_CLIENT_URLS=https://%H:2379
-            Environment=ETCD_LISTEN_PEER_URLS=https://%H:2380
-            Environment=ETCD_INITIAL_ADVERTISE_PEER_URLS=https://%H:2380
+
             PermissionsStartOnly=true
             ExecStartPre=/usr/bin/systemctl is-active cfn-etcd-environment.service
             ExecStartPre=/usr/bin/systemctl is-active decrypt-tls-assets.service
-            ExecStartPre=/usr/bin/sed -i 's/^ETCDCTL_ENDPOINT.*$/ETCDCTL_ENDPOINT=https:\/\/%H:2379/' /etc/environment
             ExecStartPre=/usr/bin/chown -R etcd:etcd /var/lib/etcd2
       enable: true
       command: start
@@ -121,6 +110,21 @@ coreos:
         [Install]
         RequiredBy=etcd2.service
 
+{{ if .WaitSignal.Enabled }}
+    - name: cfn-signal.service
+      command: start
+      content: |
+        [Unit]
+        Wants=etcd2.service
+        After=etcd2.service
+
+        [Service]
+        Type=oneshot
+        EnvironmentFile={{.EtcdNodeEnvFileName}}
+        ExecStartPre=/usr/bin/systemctl is-active etcd2.service
+        ExecStart=/opt/bin/cfn-signal
+{{end}}
+
 {{if .SSHAuthorizedKeys}}
 ssh_authorized_keys:
   {{range $sshkey := .SSHAuthorizedKeys}}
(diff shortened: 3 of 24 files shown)


mumoshu:
Just realized that what I'm trying to achieve is similar in concept to [simlodon](https://github.com/UKHomeOffice/smilodon) mentioned in https://github.com/kubernetes/kops/issues/772 😃

mumoshu:
Ah, I had been somehow forgetting the fact that an EBS can't move around AZs, too 😢

mumoshu:
I'm going to create a dedicated ASG for each etcd instance.
Each etcd ASG would depend on the next ASG so that we can hopefully do rolling-updates of ASGs.

mumoshu:
I've verified the current implementation by triggering the rolling-update of instance types for etcd nodes while running k8s conformance test.
The conformance test passed without any failures so I say no visible downtime was observed 😉

mumoshu:
My biggest concern left untouched yet is now:

> When a rolling-update is in progress, don't we need to wait until a newly recreated etcd member with possible outdated data(=persisted in the EBS volume which had been attached to the previously terminated instance replaced by the newly created instance) to catch up with the latest data from running etcd cluster?

mumoshu:
Hi @pieterlange @camilb, could I have your comments/requests regarding the TODOs and Non-TODOs in the description of this PR, if any? 😃

mumoshu:
Hi @redbaron, I've implemented my POC to make etcd cluster a bit more H/A.
Believing you're experienced in this area, woud you mind leaving your comments/requests/etc regarding the TODOs and Non-TODOs in the description of this PR? 😃

mumoshu:
Updated the supported use-cases in the description:

> * Automatic recovery from temporary Etcd node failures
>   * Even if all the nodes went down, the cluster recovers eventually as long as the EBS volumes aren't corrupted

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -166,17 +175,37 @@ write_files:
       # Decide which volume to attach hence hostname to assume
       while [ "$assumed_hostname" = "" ]; do
         sleep 3
-        describe_volume_result=$(aws ec2 describe-volumes --region {{.Region}} --filters Name=tag:aws:cloudformation:stack-name,Values=$stack_name Name=tag:kube-aws:owner:role,Values=etcd Name=status,Values=available --output json)
-        vol_id=$(echo "$describe_volume_result" | jq -r ".Volumes[0].VolumeId")
+
+        aws ec2 describe-volumes \
+          --region {{.Region}} \
+          --filters Name=tag:aws:cloudformation:stack-name,Values=$stack_name Name=tag:kube-aws:owner:role,Values=etcd Name=status,Values=available \
```
redbaron: you can add `Name=availability-zone,Values=$az` here and drop `jq` filter later

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -166,17 +175,37 @@ write_files:
...
+        cat /var/run/coreos/available-volumes.json | \
+          jq -r ".Volumes | map({ VolumeId : .VolumeId, AvailabilityZone: .AvailabilityZone, State: .State, Tags: .Tags, Name: (.Tags[] | select(.Key == \"kube-aws:etcd:hostname\") | .Value) }) | group_by(.Name) | select(all(.[].State == \"available\")) | [.[][]]" \
+          > /var/run/coreos/available-etcd-volumes.json
+
+        vol=$(cat /var/run/coreos/available-etcd-volumes.json | jq -r "[.[] | select(.AvailabilityZone == \"$az\")][0]")
+        if [ "$vol" = ""]; then
+          echo "no etcd volume available in availability zone $az"
+        fi
+
+        vol_id=$(echo "$vol" | jq -r ".VolumeId")
```
redbaron: this  will break if more than 1 volume is found. it shouldn't happen normally, though,but better to guard against it

mumoshu: Thanks but I believe I've guard it properly at line 189 with the final `[0]` in the jq expression?

redbaron:
if you have ASG per AZ,then why don't tag ASG with volume id and ENI id to use, instread of going with lengthy and error prone self-discovery process?

core/controlplane/config/config.go
```diff
@@ -284,6 +284,10 @@ func (c *Cluster) SetDefaults() {
 			c.Etcd.Subnets = c.PublicSubnets()
 		}
 	}
+
+	if c.Etcd.InternalDomainName == "" {
+		c.Etcd.InternalDomainName = fmt.Sprintf("%s.internal", c.ClusterName)
```
redbaron: not `.compute.internal`?

redbaron:
is EIP the only way to go? it will be a no-go for us :(

redbaron:
instead of EIP you can  attach ENI in the same way as you do EBS

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Good catch 👍  Thanks

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Just read your updated comment now 😉

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+          # Assume the hostname corresponds to the data(=the EBS volume identified by the vol_id above)
+          assumed_hostname=$(echo "$vol" | jq -r ".Tags[] | select(.Key == \"kube-aws:etcd:hostname\").Value")
+        fi
+      done
+
+      if [ "$vol" = "" ]; then
+        vol=$(echo "$describe_volume_result" | jq -r ".Volumes[0]")
+      fi
+      vol_id=$(echo "$vol" | jq -r ".VolumeId")
+      eip_alloc_id=$(echo "$vol" | jq -r ".Tags[] | select(.Key == \"kube-aws:etcd:eip-alloc-id\").Value")
```
redbaron: I like how you tied data (EBS) with network identity (EIP) by tagging EBS.

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+      )
+      export $(cat /var/run/coreos/etcd-environment | grep -v ^# | xargs)
+
+      export | grep ETCD
+
+      # TODO: Dynamically fetch a free pair of a network interface and an EBS from the pool and attach it.
+      # Doing so under an EC2 instance under an auto-scaling group would achieve automatic recovery from etcd node failures.
+
+      # TODO: Dynamically attach an EBS volume to /dev/xvdf before var-lib-etcd2.mount happens.
+      # Probably we cant achieve it here but in the "bootstrap" cloud-config embedded in stack-template.json
```
mumoshu: Unnecessary, already addressed TODOs in comments

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+      cfn-init -v -c "etcd-server" --region {{.Region}} --resource {{.Etcd.LogicalName}}${{.EtcdIndexEnvVarName}} --stack ${{.StackNameEnvVarName}}
+
+  - path: /opt/bin/attach-etcd-volume
+    owner: root:root
+    permissions: 0700
+    content: |
+      #!/bin/bash -vxe
+
+      instance_id=$(curl http://169.254.169.254/latest/meta-data/instance-id)
+      private_ip=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
mumoshu: `private_ip` is probably an unused shell variable

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+      until [ "$volume_status" = ok ]; do
+        sleep 3
+        describe_volume_status_result=$(aws ec2 describe-volume-status --region {{.Region}} --volume-id $vol_id --output json)
+        volume_status=$(echo "$describe_volume_status_result" | jq -r ".VolumeStatuses[0].VolumeStatus.Status")
+      done
+
+      echo $assumed_hostname > /var/run/coreos/assumed_hostname
+      echo $eip_alloc_id > /var/run/coreos/eip_alloc_id
+
+  - path: /opt/bin/assume-etcd-hostname-with-private-ip
```
redbaron: if you go with ENI, then IP addresses will be fixed and Route53 records can be created once and for all by CF

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+      instance_id=$(curl http://169.254.169.254/latest/meta-data/instance-id)
+      private_ip=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
+      az=$(curl http://169.254.169.254/latest/meta-data/placement/availability-zone)
+      stack_name=$(
+        aws ec2 describe-tags --region {{.Region}} --filters \
+          "Name=resource-id,Values=$instance_id" \
+          "Name=key,Values=aws:cloudformation:stack-name" \
+          --output json \
+        | jq -r ".Tags[].Value"
+      )
```
mumoshu: nit: stack_name should be taken from the env var `KUBE_AWS_STACK_NAME` injected via the embedded userdata in stack-template.json because it is a more portable way to get the stack name. More concretely, the `aws:cloudformation:stack-name` and other tags won't be populated when the EC2 instance is created via a spot fleet.

redbaron:
> Automatic recovery from temporary Etcd node failures
Even if all the nodes went down, the cluster recovers eventually as long as the EBS volumes aren't corrupted

Is it so? AFAIK if all nodes go down, then quorum is lost and it requires  following disaster-recovery process: https://github.com/coreos/etcd/blob/master/Documentation/v2/admin_guide.md#disaster-recovery

mumoshu:
> instead of EIP you can attach ENI in the same way as you do EBS

I'm almost ok with the ENI instead of EIP but the ENI way isn't more complete than the EIP.
I just wanted to go with a more complete option.

My concern is in the fact an ENI can't move between different AZs.
It implies that, for an user like me who is in an AWS region with only 2 AZs available hence must rely on a single-AZ etcd cluster, we have to recreate all ENIs (with different private IPs) and therefore all the etcd members and a cluster in an another live AZ when the single AZ goes down.

Using EIPs instead allows us to retain and reassign these IPs hence etcd members and etcd cluster while recreating all the backing ASGs and EC2 instances for etcd in an another AZ.

// BTW, in both cases, we need EBS snapshots to recreate EBS volumes with equivalent data in an another AZ

mumoshu:
> Is it so? AFAIK if all nodes go down, then quorum is lost and it requires following disaster-recovery process: https://github.com/coreos/etcd/blob/master/Documentation/v2/admin_guide.md#disaster-recovery

Yes, because we retain data in EBS volumes.
There's a brief downtime(approx. 5min) between all the etcd nodes are terminated and then recreated but the cluster starts functioning again after the downtime.

redbaron:
>Yes, because we retain data in EBS volumes.
There's a brief downtime(approx. 5min) between all the etcd nodes are terminated and then recreated but the cluster starts functioning again after the downtime.

Probably that works only for clean shutfown, etcd2 docs say that if quorum is lost then no auto-recovery possible:
> However, in extreme circumstances, a cluster might permanently lose enough members such that quorum is irrevocably lost. For example, if a three-node cluster suffered two simultaneous and unrecoverable machine failures, it would be normally impossible for the cluster to restore quorum and continue functioning.

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Definitely yes but it makes a disaster recovery a bit difficult(though not possible, just slower) as described in https://github.com/coreos/kube-aws/pull/332#issuecomment-281270147

mumoshu:
@redbaron AFIAK, the `unrecoverable machine failures` in this case is corrupted EBS volumes.
As long as data in EBS volumes aren't corrupted, it doesn't count to the "extreme circumstances"; it is a recoverable failure.

mumoshu:
> is EIP the only way to go? it will be a no-go for us :(

Could you mind sharing me why EIP isn't the way to go for you?
You may already know but using EIPs are used just to stabilize hostnames, eventually resolved to "private IPs" of etcd nodes (by AWS DNS)

redbaron:
> @redbaron AFIAK, the unrecoverable machine failures in this case is corrupted EBS volumes.
As long as data in EBS volumes aren't corrupted, it doesn't count to the "extreme circumstances"; it is a recoverable failure.

I can't see how it can work if AZ with last available leader is lost AND quorum is lost too.So in your 2 AZ case, if AZ with 2 nodes go down and one of them was leadere when it happened, even if you restore EBS into healthy AZ, etcd shouldn't come up, otherwise they allow silent data loss without explicit permission from operator, which I doubt very much.

core/controlplane/cluster/cluster.go
```diff
@@ -157,7 +157,7 @@ func (c *Cluster) Assets() (cfnstack.Assets, error) {
 
 	return cfnstack.NewAssetsBuilder(c.StackName(), c.StackConfig.S3URI).
 		Add("userdata-controller", c.UserDataController).
-		Add("userdata-worker", c.UserDataWorker).
```
gianrubio: Is there a mistake here?

core/controlplane/config/templates/stack-template.json
```diff
@@ -1,13 +1,6 @@
...
-  "Mappings" : {
-    "EtcdInstanceParams" : {
-      "UserData" : {
-       "cloudconfig" : "{{.UserDataEtcd}}"
-      }
-    }
-  },
   "Resources": {
     "{{.Controller.LogicalName}}": {
       "Type": "AWS::AutoScaling::AutoScalingGroup",
```
gianrubio: 👍 for etcd in asg

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -130,34 +134,202 @@ ssh_authorized_keys:
...
+           --set-env={{.StackNameEnvVarName}}=${{.StackNameEnvVarName}} \
+           --set-env={{.EtcdIndexEnvVarName}}=${{.EtcdIndexEnvVarName}} \
+           --net=host \
+           --trust-keys-from-https \
+           {{.AWSCliImageRepo}}:{{.AWSCliTag}} --exec=/opt/bin/$1
+
+           rkt rm --uuid-file=/var/run/coreos/$1.uuid || :
+        }
+
+      run cfn-init-etcd-server
```
gianrubio: This command should run only after the cluster is healthy

core/controlplane/cluster/cluster.go
mumoshu: No!
Although it is required only by a node pool stack, we had been unnecessarily uploading `user-worker` here for a control-plane stack, too.

mumoshu:
> I can't see how it can work if AZ with last available leader is lost AND quorum is lost too.So in your 2 AZ case, if AZ with 2 nodes go down and one of them was leadere when it happened, even if you restore EBS into healthy AZ, etcd shouldn't come up, otherwise they allow silent data loss without explicit permission from operator, which I doubt very much.

Thanks!
That's why I'm no longer going to deploy 2-AZ etcd cluster.
As long as your cluster is distributed among an odd number of AZs, I believe my comment at https://github.com/coreos/kube-aws/pull/332#issuecomment-281273859 applies.

core/controlplane/config/config.go
mumoshu: This is now dead-code which had been used for the ENI+Route53 Record Set way of implementing the network identity for etcd nodes.
I can revive it if you really need it rather than the current, EIP way.

mumoshu:
Thanks as always for various feedbacks, @redbaron 👍 

> if you have ASG per AZ,then why don't tag ASG with volume id and ENI id to use, instread of going with lengthy and error prone self-discovery process?

Your suggested method sounds desirable.
I've tinkered the complicated self-discovery process initially to just make sure that the discovery process of an the pair of an EBS volume and ENI/EIP is general enough to support various possible use-cases e.g. utilizing spot fleets for further cost reduction of an etcd cluster.
However the concrete use-case isn't achievable now because we rely on the `1 etcd : 1 ASG` relationship to achieve rolling-updates of etcd nodes.

mumoshu:
@redbaron Almost as an experiment, I've made it to allow choosing a strategy for providing etcd nodes stable network or member identities in `cluster.yaml` like:

```yaml
etcd:
  # The strategy to provide your etcd nodes stable network identities. Defaults to "EIP".
  # Available options:
  # "EIP": Use EC2 public hostnames stabilized with EIPs and resolved eventually to private IPs
  # "ENI": Use ENIs and Route53 record sets to provide etcd nodes stable hostnames
  #nodeIdentityProvider: EIP
```

However, I'd still like to hear from you about why the EIP method isn't the way to go for you!
https://github.com/coreos/kube-aws/pull/332#issuecomment-281271755

core/controlplane/config/templates/stack-template.json
```diff
@@ -86,40 +78,23 @@
...
-                  { "Fn::GetAtt" : [ "InstanceEtcd{{$index}}", "PrivateDnsName" ] },
-                  "=https://",
-                  { "Fn::GetAtt" : [ "InstanceEtcd{{$index}}", "PrivateDnsName" ] },
-                  ":2380",
-                  {{end}}
-                  "'\n"
-                ]]}
-              }
-            }
-          },
```
mumoshu: The `etc-server-env` config set has been moved to etcd ASGs.

core/controlplane/config/templates/stack-template.json
```diff
@@ -86,40 +78,23 @@
...
+                  {{end}}
                   "'\n"
                 ]]}
               }
             }
           }
         }
       },
-      "DependsOn": ["InstanceEtcd0"]
+      "DependsOn": ["{{$.Etcd.LogicalName}}{{minus $.EtcdCount 1}}"]
```
mumoshu: Currently, the controller ASG depends on the last etcd ASG. However, we can improve it further by making it depends on `(N/2)+1`th etcd ASG for earlier creation of the controller ASG.

core/controlplane/config/templates/stack-template.json
```diff
@@ -328,20 +385,137 @@
...
+        "Tags": [
+          {
+            "Key": "KubernetesCluster",
+            "PropagateAtLaunch": "true",
+            "Value": "{{$.ClusterName}}"
+          },
+          {
+            "Key": "Name",
+            "PropagateAtLaunch": "true",
+            "Value": "{{$.ClusterName}}-{{$.StackName}}-kube-aws-etcd"
```
mumoshu: nit but would anyone like to append the index of an etcd node so that the instance name looks like `<cluster-name>-<stack-name>-kube-aws-etcd-<index>` which provides more visibility?

core/controlplane/config/templates/stack-template.json
```diff
@@ -328,20 +385,137 @@
...
+        ],
+        "VPCZoneIdentifier": [
+          {{$etcdInstance.SubnetRef}}
+        ]
+      },
+      {{if $.WaitSignal.Enabled}}
+      "CreationPolicy" : {
+        "ResourceSignal" : {
+          "Count" : "1",
+          "Timeout" : "{{$.ControllerCreateTimeout}}"
```
mumoshu: I was too lazy to newly introduce `$.Etcd.CreateTimeout` and relied on the existing `ControllerCreateTimeout` for now. Anyone cares?

core/controlplane/config/templates/stack-template.json
```diff
@@ -328,20 +385,137 @@
...
+        }
+      },
+      {{end}}
+      "UpdatePolicy" : {
+        "AutoScalingRollingUpdate" : {
+          "MinInstancesInService" : "0",
+          "MaxBatchSize" : "1",
+          {{if $.WaitSignal.Enabled}}
+          "WaitOnResourceSignals" : "true",
+          "PauseTime": "{{$.ControllerCreateTimeout}}"
```
mumoshu: The same as https://github.com/coreos/kube-aws/pull/332/files#r102399001

core/controlplane/config/templates/stack-template.json
```diff
@@ -328,20 +385,137 @@
...
+                  "'\n"
+                ]]}
+              }
+            }
+          }
+        }
+      },
+      "DependsOn": [
+        {{if $etcdInstance.DependencyExists}}{{$etcdInstance.DependencyRef}},{{end}}
+        {{if $etcdIndex}}"{{$.Etcd.LogicalName}}{{minus $etcdIndex 1}}",{{end}}
```
mumoshu: Here, an etcd ASG(at index N) depends on a previous(at index N-1) etcd ASG so that we can rolling-update etcd ASGs one by one with `kube-aws update`

core/controlplane/config/tls_config.go
```diff
@@ -133,6 +133,7 @@ func (c *Cluster) NewTLSAssets(caKey *rsa.PrivateKey, caCert *x509.Certificate)
 		DNSNames: []string{
 			fmt.Sprintf("*.%s.compute.internal", c.Region),
 			"*.ec2.internal",
+			fmt.Sprintf("*.%s.compute.amazonaws.com", c.Region),
```
mumoshu: Intended to be used only when `etcd.nodeIdentitiyProvider: EIP` because it relies on EC2 public hostnames stabilized with EIPs

mumoshu:
Btw `etcd.nodeIdentityProvider: ENI` doesn't work yet because it is still unable to route packets from/to etcd peers through the secondary NIC(=dynamically assigned ENI).
Anyone can help me tweaking route tables and routes to make it work probably using the `ip`  command? 😃

redbaron:
@mumoshu , sorry for late response, and for making you implement additional etcd addressing :) You are right, EIP can give us a stable hostname without enabling external connectivity. (well, incoming UDP packets will be able to reach etcd, but responses wont).

Another thought: you've seen before how people not using Amazon DNS, therefore this "internal ip for public name" trick wont work for them, so maybe ENI work wasn't for nothing.

mumoshu:
There's still a bug that prevents a control-plane cloudformation stack from being successfully deleted because of EBS volumes kept attached to EC2 instances used to be managed by already terminated etcd ASGs.
It doesn't happen 100% so I believe we can fix it by somehow delaying CloudFormation to start deleting EBS volumes.

```
	The following resource(s) failed to delete: [Controlplane].
16:11:09 UTC+0900	DELETE_FAILED	AWS::CloudFormation::Stack	Controlplane	Embedded stack arn:aws:cloudformation:ap-northeast-1:<myawsaccount>:stack/kubeawstest208-Controlplane-G4JWTEKLIEN9/3b50eba0-f8ca-11e6-81dc-50d5ca9ff482 was not successfully deleted: The following resource(s) failed to delete: [Etcd1EBS].
```

mumoshu:
https://github.com/coreos/kube-aws/pull/332#issuecomment-281842674 seems to be caused by a bug that results in an etcd node to mistakenly attach an EBS volume for another etcd node.
I'll fix it soon.

mumoshu:
`nodeIdentityProvider: "ENI"` is also working now.
I'd generally recommend `"EIP"` because it involves far less moving parts, though.

mumoshu:
Though unit and integration tests must be fixed, I believe this is generally working and testable.
Please try building the kube-aws binary from this branch and creating a cluster to see if it works for you!

mumoshu:
Thanks @redbaron!

> Another thought: you've seen before how people not using Amazon DNS, therefore this "internal ip for public name" trick wont work for them, so maybe ENI work wasn't for nothing.

Unfortunately, The current implementation of `memberIdentityProvider: ENI` involves Route 53 record sets to prepare for disaster recovery in a single-AZ case. The same applies to the `EIP` provider. I've documented it in cluster.yaml.

However, I believe I can add an another provider to allow etcd nodes discover peers directly via private IPs so that it would work without Amazon DNS.

core/controlplane/config/config.go
mumoshu: I revived this to implement `memberIdentityProvider: ENI` 😃

mumoshu:
> However, I believe I can add an another provider to allow etcd nodes discover peers directly via private IPs so that it would work without Amazon DNS.

Ah, that's impossible without knowing private IPs of ENIs beforehand 😢

mumoshu:
* The values for `memberIdentityProvider` changed from `EIP` to `eip`, `ENI` to `eni` respectively
* The `eni` provider doesn't create record sets by default now. Set `etcd.internalDomainName: your-own-domain.internal` to instruct kube-aws to create a hosted zone for the domain and record sets for etcd nodes inside the hosted zone.
* Added advanced, experimental support for a DNS other than Amazon DNS for etcd peer discovery.
   * Set `etcd.internalDomainName: your-own-domain.internal` and also set `etcd.manageRecordSets: false` and (1) manually configure your DNS so that `etcd<index>.your-own-domain.internal` is resolvable among all the nodes including worker, controller, etcd nodes.

See the updated doc in `cluster.yaml`:

```yaml
#etcd:
#  # If omitted, public subnets are created by kube-aws and used for etcd nodes
#  subnets:
#    # References subnets defined under the top-level `subnets` key by their names
#    - name: ManagedPrivateSubnet1
#    - name: ManagedPrivateSubnet2
#
#  # The strategy to provide your etcd nodes, in combination with floating EBS volumes, stable member identities. Defaults to "eip".
#  #
#  # Available options: eip, eni
#  #
#  # With every option, etcd nodes communicate to each other via their private IPs.
#  #
#  # eip: Use EC2 public hostnames stabilized with EIPs and "resolved eventually to private IPs"
#  #      Requires Amazon DNS (at the second IP of your VPC, e.g. 10.0.0.2) to work.
#  #      1st recommendation because less moving parts and relatively easy disaster recovery for a single-AZ etcd cluster.
#  #      If you run a single-AZ etcd cluster and the AZ failed, EBS volumes created from latest snapshots and EIPs can be reused in an another AZ to reproduce your etcd cluster in the AZ.
#  #
#  # eni: [EXPERIMENTAL] Use secondary ENIs and Route53 record sets to provide etcd nodes stable hostnames
#  #      Requires Amazon DNS (at the second IP of your VPC, e.g. 10.0.0.2) or an another DNS which can resolve dns names in the hosted zone managed by kube-aws to work.
#  #      2nd recommendation because relatively easy disaster recovery for a single-AZ etcd cluster but more moving parts than "eip".
#  #      If you run a single-AZ etcd cluster and the AZ failed, EBS volumes created from latest snapshots and record sets can be reused in an another AZ to reproduce your etcd cluster in the AZ.
#  #
#  memberIdentityProvider: eip
#
#  # Domain of the hostname used for etcd peer discovery.
#  # Used only when `memberIdentityProvider: eni` for TLS key/cert generation
#  # If omitted, defaults to "ec2.internal" for us-east-1 region and "<region>.compute.internal" otherwise
#  internalDomainName:
#
#  # Set to `false` to disable creation of record sets.
#  # Used only when `memberIdentityProvider` is set to `eni`
#  # When disabled, it's your responsibility to configure all the etcd nodes so that
#  # they can resolve each other's FQDN(specified via the below `etcd.nods[].fqdn` settings) via your DNS(can be the Amazon DNS or your own DNS. Configure it with e.g. coroes-cloudinit)
#  manageRecordSets:
#
#  # Advanced configuration used only when `memberIdentityProvider: eni`
#  hostedZone:
#    # The hosted zone where record sets for etcd nodes managed by kube-aws are created
#    # If omitted, kube-aws creates a hosted zone for you
#    id:
#
#  # CAUTION: Advanced configuration. This should be omitted unless you have very deep knowledge of etcd and kube-aws
#  nodes:
#  - # The name of this etcd node. Specified only when you want to customize the etcd member's name shown in ETCD_INITIAL_CLUSTER and ETCD_NAME
#    name: etcd0
#    # The FQDN of this etcd node
#    # Usually this should be omitted so that kube-aws can choose a proper value.
#    # Specified only when `memberIdentityProvider: eni` and `manageRecordSets: false` i.e.
#    # it is your responsibility to properly configure EC2 instances to use a DNS which is able to resolve the FQDN.
#    fqdn: etcd0.<internalDomainName>
#  - name: etcd1
#    fqdn: etcd1.<internalDomainName>
#  - name: etcd2
#    fqdn: etcd2.<internalDomainName>
```

mumoshu:
Well, this is getting very complicated but I believe I've addressed every use-case we've seen before.

* You can choose EIP or ENI for node(=etcd member) identity
* DNS other than Amazon DNS could be used

mumoshu:
Testes are fixed and added.
E2E tests are in progress.

mumoshu:
Rebased.

mumoshu:
E2E tests are passed.

redbaron:
@mumoshu , this is incredible, thank you

core/controlplane/config/templates/stack-template.json
billyteves: +1 for etcd in ASG also.

c-knowles:
@mumoshu I've just managed to upgrade etcd instances to a larger type as an uptime upgrade thanks to you! 🎉 

It's seems there was a slight pause on etcd responding in a 3 node cluster when the state was:
- first new node was up, old node terminated
- second new node was running but possible not quite fully linked into cluster yet, old node terminated
- third new node was not up, old node still running

The pause was circa 20 seconds and various processes including kubectl and dashboard became unresponsive momentarily. I just wanted to check if anyone has seen anything similar before trying to diagnose more? Each of the wait signals was passing after around 5 minutes so it looks like this was etcd related somehow.

mumoshu:
@c-knowles I greatly appreciate this kind of feedback! Thanks.
I'm still finding a better way to reduce possible down time like that.

AFAIK, each etcd2 member(=etcd2 process inside a rkt pod) doesn't wait until the member becomes connected and ready to serve requests on its startup, and there's no way to know the member is actually ready.

For example, running `etcdctl --peers <first etcd member's advertised peer url> cluster-health` would block until all the remaining etcd members until the number meets quorum(2 for your cluster). This incomplete solution hits a chicken-and-egg problem like this and break wait signals. That's why it doesn't wait for an etcd2 member to be ready to avoid down time completely.

For me, the down time was less than 1 sec when I first tried but I suspect the result varies from time to time hence your case.

c-knowles:
Yeah, it probably varies a little bit. Is there an issue to track this as a known issue? I had a look but didn't find one unless it's part of etcd v2. If not, we should make one for other users who come across this.

mumoshu:
@c-knowles There's no github issue I'm aware of :cry:
Btw, etcd3 seems to signal systemd for readiness when its systemd unit is set to `Type=notify`.

redbaron:
can't we draw dependencies between ASGs, then CF will roll them one by one. It would allow quorum to be maintained all the time

c-knowles:
I've separated the issue out as above so we can track any progress we make on this.

mumoshu:
Hi @redbaron!
It is already implemented so. ASGs do get (hence etcd nodes managed by them) replaced one-by-one according to dependencies among ASGs combined with wait signals. Therefore my guess is that we're proceeding to next node too early because of insufficient wait signals #411, which would result in a temporary loss of quorum in an extreme case(=rolling-update happened faster than I'd expected).

mumoshu:
@c-knowles Thanks for the good writeup!

trinitronx:
@mumoshu:

Hello, I've just tested out the new `kube-aws` `v0.9.5-rc.6` yesterday and I was unable to get it to work due to an error: `The following resource(s) failed to create: [Controlplane].`

After diving further into this, I found that the reason it was failing was really because we have reached our EIP limit.  The real error was: `The maximum number of addresses has been reached.`

So I have a question about this seemingly new `kube-aws` `memberIdentityProvider` option for Etcd2.   We are already using our limit of EIPs that AWS gives us, and this requirement is new to us.  Why is this now necessary instead of using a [SRV record for Etcd2 nodes][1]?

It seems to me that `SRV` record discovery would be cleaner than having to use EIPs for the Etcd2 nodes, and more dynamic as it would not require IP hardcoding into the Etcd2 nodes in a static config file. This forces stored state of the `ETCD_INITIAL_CLUSTER` variable into files stored on the EBS volume.  Having to manage EIPs and EBS volumes in case of failure or scaling seems a bit backwards & would easily hit the 5 EIP limit that AWS has by default with a larger Etcd2 cluster.

My thought was that `SRV` discovery could allow the CloudFormation template from `kube-aws` to then just manage a single Route53 `SRV` record instead based on the Etcd2 `AutoScalingGroup`.  The current implementation of writing a bunch of hardcoded IPs to `/var/run/coreos/etcd-environment` seems like a bit of an unscalable hack IMHO (no offense intended).

Has this option for Etcd2 been explored yet, or are there any pitfalls / pros / cons involved with this type of configuration?

[1]: https://coreos.com/etcd/docs/latest/v2/clustering.html#dns-discovery

trinitronx:
@mumoshu:  Also perhaps related:  Would waiting for the PR from #417 change the picture any by choosing to avoid Etcd2 for a new cluster, and simply go with Kubernetes `v1.6` and Etcd3?

After reading a bit of the surrounding information regarding disaster recovery, this choice seems to be a pivotal moment that would seal our cluster's fate & ease of future maintenance in the event of a disaster.  Any thoughts or recommendations here?

billyteves:
@mumoshu regarding the ETCD cluster, is the setup for data is in replica? if one server goes down, data are still intact to the other servers? Can you also suggest a recommended etcd storage size?

jsravn:
A bit late to the party, but we've built something similar to the Monsanto solution at https://github.com/sky-uk/etcd-bootstrap. It handles node replacement/cluster expansion&shrinking/new clusters, and we've been using in prod for a while now. It's self contained, just a docker wrapper around etcd that queries the local ASG for the other instances in the group. If you use apiserver on the same node, it's easy to run this and have apiserver hit localhost, with an ELB on the ASG to load balance to the apiservers.

It's not entirely clear to me the benefit of managing EBS/ENI separately - why not just rebuild the node including EBS/ENI? Is that in case the entire cluster dies?

VinceMD:
With the ENI as memberIdentityProvider and 3 private subnets in 3 different AZ, how many ETCD instances (per AZ?) should we run, in order to:

* keep cluster running if one AZ goes down?
* keep cluster running if two AZ go down?

How many instances minimum do we need if we are ok with an ETCD downtime (until ASG re-creates the instances and re-attach EBS and ENI)?

Alternatively if you could point me to doc, so I can do the maths. I'd be happy to read any doc to help me understand how kube-aws solved the problem.

redbaron:
@VinceMD , etcd need to maintain quorum at all times to keep cluster running. Quorum is N/2+1 where N is number of etcd nodes.

Therefore if you run 1 node per AZ, you'll continue to maintain quorum in case one AZ goes down.

To tolerate two AZs failure, you'd need to span your cluster across 5 AZs, 2 of which will probably be in another region. I don't have information how happy etcd cluster will be when it sees significant latency increase for certain member of the cluster.

VinceMD:
Thanks @redbaron.  So I will run 3 nodes. One per AZ we have available in Sydney Region. This will allow 1 AZ failure.

In case 2 nodes are down at the same time from 2 different AZ , would the ASG + ENI + EBS solution from @mumoshu allow the ETCD cluster to recover automatically, with some downtime? Assuming ASG is able to create an EC2 again in the same AZ that were affected (since ENI cannot be moved across from one AZ to another) and re-attach the EBS and ANI fine.
->  I'm just trying to put the case of ENI forward. I somewhat understand the idea, but need to explain it better to my peers.

We had that situation in Sydney recently when 1 AZ was done and another one was affected temporarily. It's also not impossible to see 2 instances fail at the same time from 2 AZ. Rare but not impossible.

If, for some reason ENI+EBS does not help/work, would a manual intervention allow recovering of the cluster by cleaning and allowing to elect a new leader? I think we are fine with downtime in case of 2 Nodes being down, as it's very unlikely. The apps will still run, just Kubernetes won't be able to manage the pods until etcd is fixed, I assume.

redbaron:
There were some bugs in automatic recovery, which hopefully were ironed out, but in theory yes, it recovers once AZ is back.

Why you push for ENI and not for default EIP?  EIP should allow you to restore quorum not waiting for AZ to become available.

VinceMD:
Everything has to be private (private subnets only and no EIP allowed).

redbaron:
Trick is, that even if EIP is used, it is resolved to a private IP address, therefore can be used inside private subents

VinceMD:
I got an AWS error though when I tried it because there was not Internet Gateway in my Subnet. Since the subnet is private and I'm not allowed to use IGW

redbaron:
Are you using amazon DNS?

VinceMD:
yes. Route53 if that is what you mean.

PR #339, opened 2017-02-22
Add experimental support for pod security policies

soellman:
This adds support for enabling pod security policies.

Note - this only enables the admission controller and does not create any pod security policies or bindings themselves. This is a similar issue to #230 in that one might argue kube-aws should also provide some minimal pod security policies so that basic pods can be admitted to the cluster before the administrator affects their own policies and rolebindings.

Unlike #230, however, enabling this doesn't result in a "broken cluster" at like the missing RBAC roles did; the kubernetes components work with PSPs enabled.

I have not done exhaustive testing about the ramifications of having PSPs enabled but having no policies, but I think it's worth having this patch in kube-aws so the administrator can make their own choices about how to initialize the cluster objects. Thanks!

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -26,6 +26,11 @@ const (
 
 func NewDefaultCluster() *Cluster {
 	experimental := Experimental{
+		Admission{
+			PodSecurityPolicy{
+				Enabled: false,
+			},
+		},
 		AuditLog{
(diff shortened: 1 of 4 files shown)


mumoshu:
Thanks @soellman. This looks nice to me 👍 
Let's merge and see how everyone uses this.

PR #342, opened 2017-02-22
Add additional EBS volumes to worker nodes.

jollinshead:
This adds a new feature that allows additional EBS volumes to be deployed and mounted on all specified worker nodes.
The template in cluster.yaml follows:
```yaml
volumeMounts:
- type: "gp2"
  iops: 0
  size: 30
  device: "/dev/xvdf"
  path: "/ebs"
```
Allowing the configuration of various parameters (which are validated).

Systemd .mount and .service resources are used to format the disks and mount them during runtime.

NOTE: The decision to not reformat the RootVolume struct (and create a new "VolumeMount" struct) was made to avoid issues regarding the naming convention of "rootVolumeSize", "rootVolumeType" ... etc. already implemented in the cluster .yaml.

Why have this feature?
The need for additional volumes arose when we realized we didn't want a situation where the root volume would reach it's maximum capacity.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -174,8 +174,19 @@ worker:
 #          # Must be within the range between 100 and 2000
 #          # Defaults to worker.spotFleet.unitRootVolumeIOPS * weightedCapacity if omitted
 #          #rootVolumeIOPS:
+#
 #        - weightedCapacity: 2
 #          instanceType: c4.xlarge
+#
+#    #  # Additional EBS volumes mounted on the worker
+#    #  # No additional EBS volume by default. All parameter values do not default - they must be explicitly defined
+#    #  volumeMounts:
+#    #  - type: "gp2"
+#    #    iops: 0
+#    #    sizeGb: 30
+#    #    # Follow the aws convention of '/dev/xvd*' where '*' is a single letter from 'f' to 'z'
+#    #    device: "/dev/xvdf"
+#    #    mountPath: "/ebs"
 #      # Used to provide `/etc/environment` env vars with values from arbitrary CloudFormation refs
 #      awsEnvironment:
 #        enabled: true
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -91,7 +91,18 @@
                   {{end}}
                   "VolumeType": "{{$spec.RootVolumeType}}"
                 }
+              }{{range $volumeMountSpecIndex, $volumeMountSpec := $.VolumeMounts}},
+              {
+                "DeviceName": "{{$volumeMountSpec.Device}}",
+                "Ebs": {
+                  "VolumeSize": "{{$volumeMountSpec.Size}}",
+                  {{if gt $volumeMountSpec.Iops 0}}
+                  "Iops": "{{$volumeMountSpec.Iops}}",
+                  {{end}}
+                  "VolumeType": "{{$volumeMountSpec.Type}}"
+                }
               }
+              {{- end -}}
             ],
             "SecurityGroups": [
               {{range $sgIndex, $sgRef := $.SecurityGroupRefs}}
@@ -207,7 +218,18 @@
               {{end}}
               "VolumeType": "{{.RootVolumeType}}"
             }
+          }{{range $volumeMountSpecIndex, $volumeMountSpec := .VolumeMounts}},
+          {
+            "DeviceName": "{{$volumeMountSpec.Device}}",
+            "Ebs": {
+              "VolumeSize": "{{$volumeMountSpec.Size}}",
+              {{if gt $volumeMountSpec.Iops 0}}
+              "Iops": "{{$volumeMountSpec.Iops}}",
+              {{end}}
+              "VolumeType": "{{$volumeMountSpec.Type}}"
+            }
           }
+          {{- end -}}
         ],
         "IamInstanceProfile": {
           "Ref": "IAMInstanceProfileWorker"
--- a/model/node_pool_config.go
+++ b/model/node_pool_config.go
@@ -17,6 +17,7 @@ type NodePoolConfig struct {
 	SecurityGroupIds   []string               `yaml:"securityGroupIds,omitempty"`
 	Tenancy            string                 `yaml:"tenancy,omitempty"`
 	CustomSettings     map[string]interface{} `yaml:"customSettings,omitempty"`
+	VolumeMounts       []VolumeMount          `yaml:"volumeMounts,omitempty"`
 }
 
 type ClusterAutoscaler struct {
@@ -106,6 +107,10 @@ func (c NodePoolConfig) Valid() error {
 		return err
 	}
 
+	if err := ValidateVolumeMounts(c.VolumeMounts); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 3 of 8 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -110,7 +110,7 @@ worker:
 #      # Number of worker nodes to create for an autoscaling group based pool
 #      count: 1
 #      # If omitted, public subnets are created by kube-aws and used for worker nodes
-#      subnet:
+#      subnets:
```
mumoshu: Good catch 👍

core/controlplane/config/templates/cluster.yaml
```diff
@@ -174,8 +174,29 @@ worker:
 #          # Must be within the range between 100 and 2000
 #          # Defaults to worker.spotFleet.unitRootVolumeIOPS * weightedCapacity if omitted
 #          #rootVolumeIOPS:
+#
+#        #  # Additional EBS volumes mounted on the worker
+#        #  # No additional EBS volume by default. All parameter values do not default - they must be explicitly defined
+#        #  ebsVolumes:
+#        #  - type: "gp2"
+#        #    iops: 0
+#        #    sizeGb: 30
```
mumoshu: Would you mind changing this to just `size` so that we can be consistent among other keys like `workerRootVolume"Size"` and `controllerRootVolume"Size"`?

core/controlplane/config/templates/cluster.yaml
```diff
@@ -174,8 +174,29 @@ worker:
 #          # Must be within the range between 100 and 2000
 #          # Defaults to worker.spotFleet.unitRootVolumeIOPS * weightedCapacity if omitted
 #          #rootVolumeIOPS:
+#
+#        #  # Additional EBS volumes mounted on the worker
+#        #  # No additional EBS volume by default. All parameter values do not default - they must be explicitly defined
+#        #  ebsVolumes:
```
mumoshu: How about changing this to `volumeMounts:`?

In my mental model:
* `type` and `iops`, `size` are properties belonging to an "EBS volume`
* `device` is a property belonging to an "EBS volume attachment"
* `mountPath` is a property belonging to an "EBS volume mount"

respectively and therefore I took it as a short-hand definition of a set of a volume & an attachment & a mount rather than just a volume.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -174,8 +174,29 @@ worker:
...
+#        #  - type: "gp2"
+#        #    iops: 0
+#        #    sizeGb: 30
+#        #    # Follow the aws convention of '/dev/xvd*' where '*' is a single letter from 'f' to 'z'
+#        #    device: "/dev/xvdf"
+#        #    mountPath: "/ebs"
```
mumoshu: If we go ahead with https://github.com/coreos/kube-aws/pull/342/files#r102595612, `mountPath` should be just `path` because the `mount` part would be redundant in `worker.volumeMounts[].mountPath`

model/ebs_volume.go
```diff
@@ -0,0 +1,51 @@
+package model
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type EbsVolume struct {
```
mumoshu: nit: Can we change `Ebs` to `EBS` so that we can be consistent among it and other symbols like `VPDID`, `RouteTableID`, `NATGateway`, etc?

model/ebs_volume.go
```diff
@@ -0,0 +1,51 @@
...
+	SizeGb    int    `yaml:"sizeGb,omitempty"`
+	Device    string `yaml:"device,omitempty"`
+	MountPath string `yaml:"mountPath,omitempty"`
+}
+
+func (v EbsVolume) SystemdMountName() string {
+	return strings.Join(strings.Split(strings.TrimLeft(v.MountPath, "/"), "/"), "-")
+}
+
+func (v EbsVolume) Validate() error {
```
mumoshu: I like this 👍

model/node_pool_config.go
```diff
@@ -13,9 +13,10 @@ type NodePoolConfig struct {
 	InstanceType       string            `yaml:"instanceType,omitempty"`
 	ManagedIamRoleName string            `yaml:"managedIamRoleName,omitempty"`
 	RootVolume         `yaml:",inline"`
-	SpotPrice          string   `yaml:"spotPrice,omitempty"`
-	SecurityGroupIds   []string `yaml:"securityGroupIds,omitempty"`
-	Tenancy            string   `yaml:"tenancy,omitempty"`
+	SpotPrice          string      `yaml:"spotPrice,omitempty"`
+	SecurityGroupIds   []string    `yaml:"securityGroupIds,omitempty"`
+	Tenancy            string      `yaml:"tenancy,omitempty"`
+	EbsVolumes         []EbsVolume `yaml:"ebsVolumes,omitempty"`
```
mumoshu: Should be changed to `VolumeMounts` if we go ahead with https://github.com/coreos/kube-aws/pull/342/files#r102595612

mumoshu:
Thanks for submitting your great work @jollinshead!
Several minor comments left but LGTM overall 👍

jollinshead:
Thank you for the feedback, I've used it to implement improvements.

I can squash the commits into a single commit once you're happy (LGTM) with the request.

redbaron:
having tests for new validation code would be nice to have

model/ebs_volume.go
```diff
@@ -0,0 +1,70 @@
+package model
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type EBSVolume struct {
```
mumoshu: nit: This should be changed to `VolumeMount` too

core/controlplane/config/templates/cluster.yaml
```diff
@@ -174,8 +183,19 @@ worker:
 #          # Must be within the range between 100 and 2000
 #          # Defaults to worker.spotFleet.unitRootVolumeIOPS * weightedCapacity if omitted
 #          #rootVolumeIOPS:
+#
 #        - weightedCapacity: 2
 #          instanceType: c4.xlarge
+#
+#    #  # Additional EBS volumes mounted on the worker
+#    #  # No additional EBS volume by default. All parameter values do not default - they must be explicitly defined
+#    #  ebsVolumes:
```
mumoshu: nit: This should be changed to `volumeMounts` too

core/nodepool/config/templates/stack-template.json
```diff
@@ -91,7 +91,18 @@
                   {{end}}
                   "VolumeType": "{{$spec.RootVolumeType}}"
                 }
+              }{{range $volumeMountSpecIndex, $volumeMountSpec := $.SpotFleet.VolumeMounts}},
```
mumoshu: How about changing it to `$.Worker.VolumeMounts` instead of `$.SpotFleet.VolumeMounts` and completely removing the `VolumeMounts` field inside the `SpotFleet` struct?
Doing so would reduce some code duplication.

mumoshu:
@jollinshead Just realized there's some duplication in the `VolumeMounts` fields in `NodePoolConfig` and `SpotFleet`. Probably we need only the former one?

Other than that, LGTM!
I especially like your code around unit tests and the validation!

mumoshu:
Thanks @jollinshead!
This PR LGTM now and I've updated the description of this PR accordingly.

mumoshu:
I'm going to merge this once it is rebased.

mumoshu:
Thanks for the great work @jollinshead!
This is LGTM now and going to be merged.

PR #345, opened 2017-02-23
credentials: Add /O=systems:master for kube-admin

### LINKED ISSUE #344: Set /O=system:masters for the admin certificate DN
As per https://kubernetes.io/docs/admin/authorization/#rbac-mode:

> When bootstrapping, superuser credentials should include the system:masters group, for example by creating a client cert with /O=system:masters. This gives those credentials full access to the API and allows an admin to then set up bindings for other users.

With kube-aws v0.9.4-rc.3 I see:
```sh
$ openssl x509 -noout -subject -in credentials/admin.pem
subject= /CN=kube-admin
```

ankon:
Fixes #344 

----
This is tested in so far as that it produces a seemingly correct subject in the admin certificate. I'm still replacing my cluster to confirm that this indeed is enough.

### DIFF
--- a/core/controlplane/config/tls_config.go
+++ b/core/controlplane/config/tls_config.go
@@ -168,8 +168,9 @@ func (c *Cluster) NewTLSAssets(caKey *rsa.PrivateKey, caCert *x509.Certificate) 
 	}
 
 	adminConfig := tlsutil.ClientCertConfig{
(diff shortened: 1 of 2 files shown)


mumoshu:
LGTM. Thanks as always @ankon!

PR #360, opened 2017-02-27
Add apiserver-count parameter in kube-apiserver config

jollinshead:
Issues concerning conflicts between apiservers during leader allocation causes the (kubectl) kubernetes.service endpoint value to change arbitrarily (https://github.com/kubernetes/kubernetes/issues/19989) which we've encountered first-hand. The solution implemented sets the 'apiserver-count' parameter to the number of controllers used in the cluster .

The 'apiserver-count' value is set to the minimum number of controller nodes set in cluster.yaml (if controller auto-scaling is used).

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -898,6 +898,7 @@ write_files:
           command:
           - /hyperkube
           - apiserver
+          - --apiserver-count={{if .MinControllerCount}}{{ .MinControllerCount }}{{else}}{{ .ControllerCount }}{{end}}
           - --bind-address=0.0.0.0
           - --etcd-servers=#ETCD_ENDPOINTS#
           - --etcd-cafile=/etc/kubernetes/ssl/ca.pem


mumoshu:
Thanks @jollinshead, I've never realized about the issue.
Just a quick question but:

> The 'apiserver-count' value is set to the minimum number of controller nodes set in cluster.yaml (if controller auto-scaling is used).

Could you provide me a bit more context here? For me, it seems to fix the issue not completely but partially.

According to [what I've seen in SO](http://stackoverflow.com/a/36339537):

> If X is too small, apiservers will fight over the endpoints for the kubernetes service (as you observe)

Setting apiserver-count to MinControllerCount would still persist the issue when the actual desired capacity of the controller autoscaling group is higher?

pieterlange:
Having MinControllerCount to some value less then the amount of actual apiservers (eg during a rolling update) would result in the apiservers fighting:

> If X is too small, apiservers will fight over the endpoints for the kubernetes service (as you observe).

This is not necessarily bad, as the old controllers (just one in most scenarios) are outnumbered and will certainly lose as the AWS autoscaler delivers the final blow. ;-)

mumoshu:
Thanks @pieterlange!
That makes sense. However, then, how about we set the lower limit of `controller.rollingUpdateMinInstancesInService` to significantly high and make it rely on `Controller.RollingUpdateMinInstancesInService` rather than `MinControllerCount`?
So that we can ensure that `apiserver-count` is always less than the actual num of controller nodes, which means some fights for the endpoint but no down time incurred?

redbaron:
What happens if `apiserver-count` is set to a value larger than number of available controllers?

mumoshu:
According to lavalamp at SO:

> If X is too large, then wrong values will linger in the endpoints for the service and some percentage of client requests to the Kubernetes API from within the cluster will not connect to a running apiserver.

mumoshu:
Let's go ahead 😃

mumoshu:
@jollinshead Thanks again for your contribution 👍

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -898,6 +898,7 @@ write_files:
           command:
           - /hyperkube
           - apiserver
+          - --apiserver-count={{if .MinControllerCount}}{{ .MinControllerCount }}{{else}}{{ .ControllerCount }}{{end}}
```
c-knowles: @mumoshu I believe this may replicate the same behaviour of [`MinControllerCount`](https://github.com/coreos/kube-aws/blob/master/core/controlplane/config/config.go#L568) method? i.e. `.MinControllerCount` is never nil.

c-knowles:
What's the recommended practice here in order to do uptime controller updates? I have one cluster setup as:
```
controller:
 autoScalingGroup:
   minSize: 1
   maxSize: 2
   rollingUpdateMinInstancesInService: 1
```

I just did a `kube-aws update` on this from k8s 1.5.3 to k8s 1.5.4 and all seemed good, no outage of kubectl/dashboard etc as far as I could tell.

Then I have another cluster setup as:
```
controller:
 autoScalingGroup:
   minSize: 2
   maxSize: 3
   rollingUpdateMinInstancesInService: 1
```

However, doing the same update caused at least some outage and lack of control from both kubectl and dashboard started to return various errors (mostly 5xx I believe).

mumoshu:
@c-knowles Thanks for the feedback!
Then, perhaps my assumption at https://github.com/coreos/kube-aws/pull/360#issuecomment-282911573 is correct; could you change that part to `--apiserver-count={{.Controller.AutoScaligGroup.RollingUpdateMinInstancesInService}}` and verify it doesn't incur down time while rolling updating controller nodes?

c-knowles:
@mumoshu I can try at some stage but right now I don't have a suitable cluster to upgrade, I can spin up another later.

mumoshu:
Btw kubernetes-dashboard pods are scheduled to worker nodes.
If you'd like dashboard to stop working while rolling updating worker nodes, you should have at least 2 of them and pod anti-affinity to try its best to distribute among different worker nodes. You also want `worker.nodePools[].nodeDrainer.enabled` set to true so that the dashboard pod gets moved from the terminating node to the other up and running node earlier.

c-knowles:
Yep, understood about that. I can confirm I have the `nodeDrainer` enabled. Dashboard is deployed by kube-aws and in general I'm not touching those containers so I don't have the anti-affinity enabled, just the critical pod annotations that kube-aws adds.

To clarify the dashboard bit, my dashboard was already loaded and I was clicking around the different pages and received the errors logged dynamically as HTML inside the dashboard. When I've seen problems like that before it's usually down to the API server not responding correctly, but I agree it could have been the dashboard pod as well. That's why I went to kubectl to have a poke around but also received persistent errors there even with nodes up and running. Those errors cleared up after several minutes which indicates to me some issue during rolling update, I guess similar to the problem described here.

mumoshu:
Made sense. Thanks for the confirmation!

core/controlplane/config/templates/cloud-config-controller
mumoshu: Good catch! I believe so

mumoshu:
Replying to my own old comment:
> Thanks @pieterlange!
That makes sense. However, then, how about we set controller.rollingUpdateMinInstancesInService to significantly high and make --apiserver-count rely on Controller.RollingUpdateMinInstancesInService rather than MinControllerCount?
So that we can ensure that apiserver-count is always less than the actual num of controller nodes, which means some fights for the endpoint but no down time incurred?

Certainly my suggestion above works ok only when `RollingUpdateMinInstancesInService <= MinControllerCount`.
IMHO We can also think of specifying the minimum value between `RollingUpdateMinInstancesInService` and `MinControllerCount` to `apiserver-count` for maximum availability.

PR #384, opened 2017-03-06
Update file paths in readme

swestcott:
Correct file paths in readme


mumoshu:
@swestcott Good catch 👍 Thanks for the contiribution!

PR #386, opened 2017-03-07
Allow disabling wait signals

### LINKED ISSUE #371: waitSignal cluster.yaml section is not - respected in the template
Assuming that you have the following section configured on your cluster.yaml 
```
# When enabled, autoscaling groups managing controller nodes wait for nodes to be up and running.
# It is enabled by default.
waitSignal:
  enabled: false
  maxBatchSize: 1
```
Unfortunately on the generated cloudformation (after the stack is rendered) - or even in the intermediate template. We always have the section generated! 

For example looking at the sections of `node-pool.json.tmpl`

```
{{if .WaitSignal.Enabled}}
          "WaitOnResourceSignals" : "true",
          "MaxBatchSize" : "{{.WaitSignal.MaxBatchSize}}",
          "PauseTime": "{{.CreateTimeout}}"
          {{else}}
```
it seems that this part is always rendered on the generate stack.json.

Since this, part is causing us problems on our AWS account/setup, we need to manually remove the above part from the `node-pool.json.tmpl` so that is not rendered at all. 

Thanks for your time anyway & and many many thanks for your great tool!

mumoshu:
fixes #371

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -94,7 +94,6 @@ func NewDefaultCluster() *Cluster {
 			MapPublicIPs:       true,
 			Experimental:       experimental,
 			ManageCertificates: true,
-			WaitSignal:         WaitSignal{Enabled: true, MaxBatchSize: 1},
 		},
 		KubeClusterSettings: KubeClusterSettings{
 			DNSServiceIP: "10.3.0.10",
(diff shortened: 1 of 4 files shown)


mumoshu:
cc @javapapo @Camsteack

PR #392, opened 2017-03-07
typo

Vrtak-CZ:


mumoshu:
@Vrtak-CZ Good catch! Thanks for your support.

PR #394, opened 2017-03-08
Introduce the rescheduler

c-knowles:
For https://github.com/coreos/kube-aws/issues/118.

- Currently using raw Pod as per [salt setup](https://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/rescheduler/rescheduler.manifest)
- Logging routed via standard streams
- gcr image used, not sure if hyperkube has the rescheduler

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1045,6 +1045,28 @@ write_files:
             initialDelaySeconds: 15
             timeoutSeconds: 15
 
+  - path: /etc/kubernetes/manifests/kube-rescheduler.yaml
+    content: |
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: rescheduler-v0.2.2
+        namespace: kube-system
+        labels:
+          k8s-app: rescheduler
+          version: v0.2.2
+          kubernetes.io/cluster-service: "true"
+          kubernetes.io/name: "Rescheduler"
+      spec:
+        hostNetwork: true
+        containers:
+        - name: rescheduler
+          image: gcr.io/google-containers/rescheduler:v0.2.2
+          # TODO: Make resource requirements depend on the size of the cluster
+          resources:
+            requests:
+              cpu: 10m
+              memory: 100Mi
 
   - path: /srv/kubernetes/manifests/kube-dns-autoscaler-de.yaml
     content: |


c-knowles:
@mumoshu basic part is there, I understand from https://github.com/coreos/kube-aws/issues/118 we wish for a flag for this. Are we still adding experimental flags or we're agreed to just mark them as experimental in docs?

mumoshu:
Thanks for the PR @c-knowles!
Yes, we can just mark them as experimental in docs for now.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1045,6 +1045,28 @@ write_files:
...
+          k8s-app: rescheduler
+          version: v0.2.2
+          kubernetes.io/cluster-service: "true"
+          kubernetes.io/name: "Rescheduler"
+      spec:
+        hostNetwork: true
+        containers:
+        - name: rescheduler
+          image: gcr.io/google-containers/rescheduler:v0.2.2
+          # TODO: Make resource requirements depend on the size of the cluster
```
mumoshu: So, do we need to manually modify `/etc/kubernetes/manifests/kube-rescheduler.yaml` to update the `resources.requests` after each time worker nodes are scaled out considerably?
Better to document it?

c-knowles: @mumoshu I copied this from https://github.com/kubernetes/kubernetes/commit/c304fa1f06eb23bcdc66b3f110f496681bc80301, it's what the salt setup has in the main repo. I'd be happy to just remove it perhaps as none of the others have it.

c-knowles:
New commits are not showing, perhaps due to org move to Kubernetes Incubator. Going to close and open a new PR.

c-knowles:
Now in https://github.com/kubernetes-incubator/kube-aws/pull/441.

mumoshu:
OK. Thanks for your effort 👍

core/controlplane/config/templates/cloud-config-controller
mumoshu: I see; let it remain here so that we can at least notice users about it

mumoshu: Note: As this pod is "burstable" in cpu/memory(requests set, but no limits set), as long as your cluster has enough capacity, rescheduler would keep working.

PR #412, opened 2017-03-14
Fix typo in ClusterAutoScalerSupport attribute name in template

danielfm:
Changed the attribute name to `enabled` so it matches the [expected name](https://github.com/coreos/kube-aws/blob/master/core/controlplane/config/config.go#L474) for that attribute when parsing the `cluster.yaml` file.


danielfm:
Oh, sorry for the branch name, I used the GitHub interface to send the PR. 😅

mumoshu:
Ah, good catch!
Thanks for the fix @danielfm 👍

PR #414, opened 2017-03-15
[wip] Kubelet TLS bootstrapping

### LINKED ISSUE #406: TLS bootstrapping with the cluster-level CA
Kubernetes 1.4 introduces an [experimental API](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/) for requesting certificates from a cluster-level Certificate Authority (CA) for TLS client certificate bootstrapping for kubelets.

What do you guys think about this? Do you consider adding this to kube-aws at some point?

danielfm:
This PR introduces an experimental feature for provisioning TLS certificates for worker nodes via the also experimental [certificate signing requests](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/) workflow. Since this workflow is used by kube-adm, I think we can rest assured it won't be dropped anytime soon.

If RBAC is enabled in `cluster.yaml`, I also made sure to set up the appropriate cluster role bindings so that the token used for sending the certificate signing requests can only make requests related to certificate provisioning, as instructed by the official documentation.

The work is not complete, though. For now, I hard-coded the access token used in the bootstrap process, so I need some pointers regarding how to implement this properly. I think we have two options here:

1. Generate a token file, i.e. `./credentials/csr-token`, and handle it the same way we do with certificates, which is generating/encrypting when running `kube-aws render credentials` and `kube-aws up`, and handling the decryption in userdata templates
2. Take a more generic approach by allowing the cluster administrator to specify a [CSV token file](https://kubernetes.io/docs/admin/authentication/#static-token-file), i.e. `./credentials/tokens.csv`, (we can even generate a minimal one in `kube-aws render credentials`) in which he/she could add as many entries as needed alongside the one needed by the TLS bootstrap...

```csv
<csr-token>,kubelet-bootstrap,10001,"system:kubelet-bootstrap"
<some-token>,<username>,<id>,<groups>
```

...and have it encrypted/decrypted the usual way.

The first option is more to-the-point, but IMO the second option has the nice advantage of allowing the cluster administrator to manage other tokens for other purposes, perhaps even as the main authentication strategy for the cluster (e.g. by issuing tokens for other users and controlling who does what via RBAC).

Could you guys share your opinions on this so I can proceed?

Closes #406.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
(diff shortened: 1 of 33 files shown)


mumoshu:
@danielfm Thanks for the great work!
I prefer the second option.

However, if a token file is subject to change by e.g. deletion of an user, how should we manage updates to a token file?

AFAIK, we have to restart every apiserver to apply changes in a token file.
If we embed `./credentials/tokens.csv` to `cloud-config-controller`, probably `kube-aws update` replaces controller nodes one by one with new csv file to reflect changes to a token file without down time?

danielfm:
Hey @mumoshu,

> However, if a token file is subject to change by e.g. deletion of an user, how should we manage updates to a token file?
>
> AFAIK, we have to restart every apiserver to apply changes in a token file.
If we embed `./credentials/tokens.csv` to `cloud-config-controller`, probably `kube-aws update` replaces controller nodes one by one with new csv file to reflect changes to a token file without down time?

Yes, that's how I've been doing this so far.

Although I'm sure there are better ways for granting and managing access to a Kubernetes cluster "at scale" (i.e. LDAP via [dex](https://github.com/coreos/dex)), using a token file is simple enough for those with a more static set of tokens.

Since this is not currently supported by kube-aws, what I've been doing so far in my setup is to manually embed the encrypted token file in `cloud-config-controller`, decrypting it along the other encrypted strings when provisioning the instance, and providing the `--token-auth-file` flag to the apiserver.

mumoshu:
Hi @danielfm, thanks for the info.

I agree with you about token files.
I have been dreamed of adding a support for dex into kube-aws but haven't had time to work on it.
In the meantime, the support for token files would be useful to many people.

> Since this is not currently supported by kube-aws, what I've been doing so far in my setup is to manually embed the encrypted token file in `cloud-config-controller`, decrypting it along the other encrypted strings when provisioning the instance, and providing the `--token-auth-file` flag to the apiserver.

Is this something you can introduce to kube-aws as the second option we've talked above?
If so, I'd greatly appreciate it 👍

danielfm:
@mumoshu can I add the token file support as a non-experimental feature? In order to make things easy for the reviewers, I think it's best I send another PR just for that, and resume this one once kube-aws support token files.

mumoshu:
Yes, and the another PR would be indeed preferable :+1:
2017年3月16日(木) 23:06 Daniel Martins <notifications@github.com>:

danielfm:
For some reason, my commits are not showing up here... maybe it has to do with the repository moving from the CoreOS org to the Kubernetes Incubator. 😅

mumoshu:
Hey @danielfm, thanks for the PR added the token file support!
Is your commit https://github.com/danielfm/kube-aws/commit/38e0664ee6df41751f8f3283df4e296a7c67fc32? So am I...
Perhaps you need to recreate this PR, or just rebase to the laster master and force push?

c-knowles:
Same problem for https://github.com/kubernetes-incubator/kube-aws/pull/394, rebase and force push does not seem to solve the problem.

danielfm:
@c-knowles I sent a message to the GitHub staff and they said they'll try to find out what happened, but no ETA was given. 🤷‍♂️

danielfm:
I'll try moving to a new PR to see if it solves the problem.

PR #390, opened 2017-03-07
AWS China region support

camilb:
Hi @mumoshu, my colleague @mpucholblasco and I were working to adapt `kube-aws` for AWS China region and we need community feedback.

There were several problems we encountered and  tried to fix:
* there is no KMS support in `cn-north-1`region. We preferred to disable this entirely for this region only.We think later using https://github.com/mozilla/sops to encrypt the certificates fields.
* China region has a different S3 endpoint. For now we are passing a flag "--s3-region=cn-north-1" when starting the cluster but we are working to get the region automatically from `cluster.yaml`
* there are different ARNs. We automatically detect and change them for `cn-north-1`
* most of the docker or rkt images are not pulled from dockerhub or quay and gcr.io is completely blocked in China. On large images, like hyperkube, you get many timeouts from quay. 
We added an option to specify your own images, hosted on a registry in China. 
We are using Alibaba Cloud registry, which is very fast. We also added an option for rkt to pull images from docker repos.
A special case is with `amd64-pause` image that kubernetes is using it and can not change its name or download it. In this case we added a simple service that pulls the image from our registry in China and tag it as `gcr.io/google_containers/pause-amd64:3.0`

* CoreOS AMIs in `cn-north-1` are outdated, so we're using a simple method to clone a AMI from Singapore region.
#ap-south-1
mkfs.ext4 /dev/xvdb
mkdir /ami
mount /dev/xvdb /ami
cd /ami
dd if=/dev/xvda of=root.img bs=1M
scp -i core.pem /ami/root.img core@cn-north-1_ip_address:/home/core
#cn-north-1
dd if=/home/core/root.img of=/dev/xvdb bs=1M oflag=direct
mkdir -p /ami
mount /dev/xvdb9 /ami
rm /ami/home/core/.ssh/authorized_keys
rm /ami/home/core/.ssh/authorized_keys.d/*
umount /ami
Create a snapshot of the volume.
Create an AMI from the snapshot.

After all of these changes we can successfully deploy kubernetes with `kube-aws` in `cn-north-1` as fast as for other regions and without errors or additional configurations.

We are waiting for suggestions from the community before clean the commits and rebase.

PR #417, opened 2017-03-16
Automatic recovery from permanent failures of etcd3 nodes

### LINKED ISSUE #27: etcd management

### LINKED ISSUE #381: etcd v3 support
ref https://github.com/kubernetes/kubernetes/issues/22448

Notes:

* Etcd v3 is able to serve requests for both the v2 and v3 data
* Running etcd v3 against the v2 data dir doesn't automatically upgrade the data dir to the v3 format
* Keeping to use v2 api against etcd v3 only updates the v2 data stored in etcd
* Though it might be out of scope of kube-aws, [upgrading of etcd to v3 is documented in the etcd doc](https://github.com/coreos/etcd/blob/master/Documentation/upgrades/upgrade_3_0.md)
* The whole [migration process is described in the CoreOS blog](https://coreos.com/blog/migrating-applications-etcd-v3.html)
* `etcdctl` should have an env var `ETCDCTL_API=2` or `ETCDCTL_API=3` provided to talk with an etcd cluster in API v2 or v3 respectively.
  * See https://github.com/coreos/etcd/blob/master/Documentation/upgrades/upgrade_3_0.md#further-considerations
* `kube-apiserver` talks with an etcd cluster in API v3 when `--storage-backend etcd3` is provided.
* Etcd3 unit could be configured like https://github.com/coreos/matchbox/blob/master/examples/ignition/etcd3.yaml

mumoshu:
closes #381, resolves #27

## Features

* Automatic recovery from any number of permanently failed etcd nodes
  * Currently, it is considered to be a "permanent failure" when health checks to an etcd member is failing longer than the threshold(10 seconds by default)

## Notable changes

* Upgraded etcd to v3
* kube-apiserver uses etcd3 API rather than the former etcd2 API for writing v3 keys
  * Although you can stick with v2 api + v3 etcd(then kube-apiserver writes v2 keys in etcd v3!), the current implementation of etcd snapshot/restore doesn't support etcd v2 data. That's why I made this change
  * etcd3 will be the default storage backend since k8s 1.6 anyways
* Wrote a lengthy `etcdadm` script to automate etcd member health checking, saving snapshots, recovering from up to `N/2` permanently failed nodes(=replace failed members one by one), recovering from more than `N/2` permanently failed nodes(=bootstrapping a brand-new cluster with the latest snapshot)

## Remaining issues

- [ ] TLS assets are currently located under `/etc/ssl/certs` rather than the former `/etc/etcd2/ssl`
  - This is due to my misunderstanding that `etcd-member.service` bundled in Container Linux for running etcd3 member doesn't support a convenient way to customize the location. It does expose `ETCD_SSL_DIR` to instruct an etcd3 rkt pod to mount `ETCD_SSL_DIR` to `/etc/ssl/certs` in the pod. So, by setting `ETCD_*_KEY_FILE` and `ETCD_*_CERT_FILE` prefixed with the `/etc/ssl/certs` regardless of where the TLS assets are in the host machine would just work.
- [ ] Etcd snapshots are currently saved in `<bucket from s3-uri>/<path from s3-uri>/kube-aws/clusters/<cluster name>/exported/stacks/etcd-snapshots/*`, which doesn't make sense
- [ ] Etcd data directories are still located at `/var/lib/etcd2` 😃 It should be `/var/lib/etcd` or `/var/lib/etcd3`
- [ ] `etcdadm` can't be modified as it is directly embedded from a kube-aws binary to `cloud-config-etcd`, without rendered at first by `kube-aws render` like other assets.


mumoshu:
Seeing bad certificate errors in the `etcdadm-check.service` on etcd nodes.
Probably I confused etcd peer key/cert with client key/cert :)
Fixing...

mumoshu:
Try triggering a permanent failure by running the below in one of your etcd nodes:

```
sudo bash -c 'find /var/lib/etcd2/member/ -exec bash -c "echo corrupting \$1; echo poison > \$1" -- {} \;; systemctl restart etcd-member'
```

And see how it works by tailing logs:

```
journalctl -f -u etcdadm-reconfigure
```

If you've done the above step to all the nodes at once, a disaster recovery process would be triggered i.e. all the members are recovered from the latest snapshot (if any. snapshots are taken since 120sec after each node started)

If you'd like to replace a permanently failed etcd node at any time you like, run:

```
sudo bash -c 'systemctl stop etcd-member; set -a; source /etc/etcd-environment; source /var/run/coreos/etcadm-environment; set+a; /opt/bin/etcdadm replace; systemctl start etcd-member'
```

mumoshu:
Fixed various bugs

redbaron:
>  Etcd snapshots are currently saved in <bucket from s3-uri>/<path from s3-uri>/kube-aws/clusters/<cluster name>/exported/stacks/etcd-snapshots/*, which doesn't make sense

I kinda like that there is no additional parameter to setup :)

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+    fi
+  done
+  echo "$r"
+}
+
+_current_time() {
+  date +%s
+}
+
+_run_as_root() {
```
redbaron: This might work too:
```bash
_run_as_root=

if [[ "$EUID" > 0 ]]; then
  _run_as_root=sudo
fi
```

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+_error() {
+  echo "$0: error: ${FUNCNAME[1]}: $*" >&2
+}
+
+_panic() {
+  echo "$0: panic: ${FUNCNAME[1]}: $*" >&2
+  exit 1
+}
+
+_array_join() {
```
redbaron:
```
printf '%q ' "${@}" 
```

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+
+config_awscli_docker_image() {
+  echo "${ETCDADM_AWSCLI_DOCKER_IMAGE:-quay.io/coreos/awscli}"
+}
+
+awscli_rkt_image() {
+  echo "docker://$(config_awscli_docker_image)"
+}
+
+config_aws_region() {
```
redbaron: why functions which print values? can't values be just set?

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+  echo "${ETCDADM_STATE_FILES_DIR:-/var/run/coreos/$(member_name)-state}"
+}
+
+cluster_etcd_aci_url() {
+  local v
+  v=v$(config_etcd_version)
+  echo "https://github.com/coreos/etcd/releases/download/$v/etcd-$v-linux-amd64.aci"
+}
+
+cluster_member_indices() {
```
redbaron: every time it is used `seq $(config_member_count)` can be used instead

etcdadm/README.md
```diff
@@ -0,0 +1,104 @@
...
+  1. clearing the contents of the etcd data dir 
+  2. removing and then re-adding the etcd member by running `etcdctl member remove` and then `etcdctl memer add` 
+  
+## Pre-requisites
+
+* etcd3
+* rkt
+* docker
+* systemd
+* bash 3+
```
redbaron: `jq`

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+}
+
+cluster_num_running_nodes() {
+  # TODO aws autoscaling describe-auto-scaling-group
+  if [ -f /sys/hypervisor/uuid ] && [ `head -c 3 /sys/hypervisor/uuid` == ec2 ]; then
+    local describe_instances
+    describe_instances=$(_awscli_command ec2 describe-instances --filter Name=tag:kube-aws:role,Values=etcd Name=tag:KubernetesCluster,Values=$KUBERNETES_CLUSTER)
+    $describe_instances | jq -r '[ .Reservations[].Instances[] ] | length'
+  else
+    local f
```
redbaron: if file cannot be empty, the can be written shorter
```
else
  cat $(tester_num_running_nodes_file) 2>/dev/null || echo 0
fi
```

etcdadm/etcdadm
```diff
@@ -0,0 +1,1559 @@
...
+  _info "quorum=$quorum healthy=$healthy"
+  if (( healthy < quorum )); then
+    _info 'cluster is unhealthy'
+    return 0
+  fi
+  _info 'cluster is healthy'
+  return 1
+}
+
+cluster_majority() {
```
redbaron: can be written shorter:
```
cluster_majority() {
  echo $(( $(config_member_count) / 2 + 1 ))
}
```

but again, given that it is a static value, it can't it be just set and used elsewhere instead of specifying and  calling a function?

mumoshu:
> I kinda like that there is no additional parameter to setup :)

It will be `<bucket from s3-uri>/<path from s3-uri>/kube-aws/clusters/<cluster name>/etcd-snapshots/*` 😄

etcdadm/README.md
mumoshu: Good catch! Thanks

etcdadm/etcdadm
mumoshu: Thanks!
As it is starting from 0, would it be something like `seq 0 (($(config_member_count) -1`, or is there anything better?

etcdadm/etcdadm
mumoshu: I guess these functions especially had been useful only when I had not set `set -e` :)
I'm ok with making them values as long as it can emit an undefined-variable-error where necessary.
I'll try later 👍

etcdadm/etcdadm
mumoshu: Perfect!
Your impl seems to work even for a complex case like this 👍 
```
./test.sh sudo bash -c "bash -c 'echo \"foobar baz\"'" | bash
#=> foobar baz
```

etcdadm/etcdadm
mumoshu: It was written as so to prevent successful `echo` even when `config_member_count` is failed.
However, making it just a variable would be better as I have `set -o nounset` now 👍

etcdadm/etcdadm
redbaron: you are right, `seq 0 $(( $(config_member_count) -1 ))` is good enough probably :)

alternative is `for x in {0..$(( $(config_member_count) -1 ))}`

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -24,15 +24,120 @@ coreos:
...
+        BindsTo=etcd-member.service
+        Before=etcd-member.service
+        Wants=cfn-etcd-environment.service
+        After=cfn-etcd-environment.service
+        Wants=network.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        RestartSec=5
```
redbaron: does systemd restart `oneshot` services now?

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -24,15 +24,120 @@ coreos:
...
+        WantedBy=timers.target
+
+    - name: etcd-member.service
       drop-ins:
-        - name: 20-etcd2-aws-cluster.conf
+        - name: 20-aws-cluster.conf
           content: |
             {{- if .TLSAssetsEncryptionEnabled}}
             [Unit]
             Wants=decrypt-tls-assets.service cfn-etcd-environment.service
```
redbaron: don't we need `cfn-etcd-environment.service` all the time?

core/controlplane/config/templates/cloud-config-etcd
mumoshu: No! Good catch.
Btw, I wanted the reconfiguration to retry until it eventually succeeds. However, setting it to  `Type=simple` breaks `etcd-member.service`. I hope there's any way to easily make it restart-able.

core/controlplane/config/templates/stack-template.json
```diff
@@ -331,6 +331,33 @@
...
+                {{/* Required for `etcdadm reconfigure` to check existence of an etcd snapshot in S3 */}}
+                {
+                  "Effect": "Allow",
+                  "Action": [
+                    "s3:List*"
+                  ],
+                  "Resource": "arn:{{.Region.Partition}}:s3:::{{$.EtcdSnapshotsS3Bucket}}",
+                  "Condition": {
+                    "StringLike": {
+                      "s3:prefix": "{{$.EtcdSnapshotsS3Prefix}}/*"
```
redbaron: nice attention to details!

etcdadm/etcdadm
mumoshu: Thanks.
I tried the alternative but it seems to not work(?)

```
core@core-01 ~ $ num=3; for x in {0..$(( num -1 ))}; do echo $x; done
{0..2}
core@core-01 ~ $ num=3; for x in {0..3}; do echo $x; done
0
1
2
3
```

core/controlplane/config/templates/stack-template.json
```diff
@@ -461,6 +488,11 @@
             "Key": "Name",
             "PropagateAtLaunch": "true",
             "Value": "{{$.ClusterName}}-{{$.StackName}}-kube-aws-etcd-{{$etcdIndex}}"
+          },
+          {
+            "Key": "kube-aws:role",
```
redbaron: can't it use Name tag above? it kind find its own tag and then strip index to find other etcd instances

etcdadm/etcdadm
mumoshu: Elegant 🌹

core/controlplane/config/templates/cloud-config-etcd
mumoshu: It is required to dynamically populate `/etc/etcd-environment` which is then read by etcd?

etcdadm/etcdadm
redbaron: well, it is not really an alternative then :))

core/controlplane/config/templates/stack-template.json
mumoshu: That's possible. I just prefer the dedicated key though because then it won't be affected by changes to `Name` and the etcdadm script becomes a bit simpler

core/controlplane/config/templates/cloud-config-etcd
redbaron: that is what I thought too,but it is inside `if` statement , shouldn't it be outside?

`[Unit]` sees to be inside `if` too, which doesn't look right

core/controlplane/config/templates/stack-template.json
redbaron: reason I asked is because there can be 10 Tags on a resource and company policies might require 4-5 of them to be set, so no much room for new tags :) I'd leave it simple then, until it becomes a problem

mumoshu: argh 😄  certainly it should be considered in (near) future..

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Certainly! I didn't get that until now.
Perhaps it was luckily working when it had changed in #390, by the existence of `Before=format-etcd2-volume.service` of `cfn-etcd-environment.service`. Anyway, it should be fixed.

mumoshu:
A few E2E test cases are failing, probably due to etcd3 storage provider?

```
Summarizing 4 Failures:

[Fail] [k8s.io] Pods Delete Grace Period [It] should be submitted and removed [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/pods.go:176

[Fail] [k8s.io] Kubectl client [k8s.io] Kubectl run --rm job [It] should create a job from an image, then delete the job [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/kubectl.go:1278

[Fail] [k8s.io] Pods [It] should be submitted and removed [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/common/pods.go:250

[Fail] [k8s.io] SchedulerPredicates [Serial] [It] validates resource limits of pods that are allowed to run [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduler_predicates.go:200

Ran 117 of 474 Specs in 4953.109 seconds
FAIL! -- 113 Passed | 4 Failed | 0 Pending | 357 Skipped --- FAIL: TestE2E (4953.12s)
FAIL

Ginkgo ran 1 suite in 1h22m33.44099363s
```

redbaron:
it would mean that not everything in kube works with etcd3 :) maybe tests are just flaky

etcdadm/README.md
mumoshu: Added.

mumoshu:
I hope so 😄 Anyways, these cases seems to be consistently failing so far

mumoshu:
If E2E doesn't pass with the combination of k8s 1.5.4 and etcd3, should we postpone this until k8s v1.6 is released?
Since k8s 1.6, the default storage provider etcd3 and therefore the flakiness should be fixed after the release.

etcdadm/etcdadm
```diff
@@ -6,72 +6,70 @@ unset ETCDCTL_ENDPOINT
 
 export | grep ETCDCTL
 
+set -o nounset
+set -o errexit
+set -o pipefail
```
redbaron: this is dangerous, it makes commands like `cat BIGFILE |  grep -q .` to fail, as `cat` receives `SIGPIPE` when `grep` exits and therefore exits with non 0 code, causing whole expression to fail

mumoshu: Good to know! Thanks.
Are there any better ones than the current combination of options?

etcdadm/etcdadm
```diff
@@ -0,0 +1,1621 @@
...
+      member_set_unit_type notify
+    fi
+
+    if (( running_num < total_num )); then
+      _info "only ${running_num} of ${total_num} nodes for etcd members are running, which means cluster is still in bootstrap process. searching for a etcd snapshot to recover this member"
+      member_bootstrap
+    elif cluster_is_failing_longer_than_limit; then
+      _info "all the nodes for etcd members are running but cluster has been unhealthy for a while, which means cluster is now in disaster recovery process. searching for a etcd snapshot to recover this member"
+      member_bootstrap
+    else
```
mumoshu: Perhaps this block results in the last etcd node to miss restoring itself from an etcd snapshot, if any?

We can probably change this if-else block to just:

```bash
f (( running_num < total_num )); then
  _info "only ${running_num} of ${total_num} nodes for etcd members are running, which means cluster is still in bootstrap process. searching for a etcd snapshot to recover this member"
elif (( running_num == total_num )); then
  _info "all the ${running_num} of ${total_num} nodes for etcd members are running. Almost certainly this node is the last one?"
elif cluster_is_failing_longer_than_limit; then
  _info "all the nodes for etcd members are running but cluster has been unhealthy for a while, which means cluster is now in disaster recovery process. searching for a etcd snapshot to recover this member"
else
  _panic "[bug] invalid state!"
fi
member_bootstrap
```

whereisaaron:
1.6 is out! etcdv3 for all :)

mumoshu:
Yay! 😆
@whereisaaron Btw, are you ok with dropping etcd v2 support form kube-aws after this?
That is a major headache for me as of now 😢

mumoshu:
TODO: A configuration key in cluster.yaml to enable/disable the automatic disaster recovery feature?

whereisaaron:
@mumoshu I don't have any need to support etcd v2 and v3 in the same version of `kube-aws`. Since `kube-aws` versions are often not backwards compatible, I use only a specific version of `kube-aws` to create and update each cluster.

I suggest you tie k8s v1.6 and etd v3 together, and that version of `kube-aws` support only k8s v1.6+ with etcd v3. And stick with etc v2 for k8s 1.5 and earlier.

The automated disaster recovery sounds great to me, but I imagine it might not be everyone's taste? I suggest there are three levels of enabled:
1. Do nothing
2. Snapshots to S3
3. Snapshots to S3 with auto-recovery from snapshot

etcdadm/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+* `ETCDADM_AWSCLI_DOCKER_IMAGE` is the reference to the `awscli` docker image used from `etcdadm`. If omitted, `quay.io/coreos/awscli` is used as the default
+
+## Limitations
+
+Beware that this tool does not support etcd2.
+
+## Other notes
+
+Although it might be a bad idea, `etcadm` is implemented in bash (for now) due to the fact it requires access to rkt, docker, systemd and etcd data dir on a host machine to reconfigure an etcd member.
```
mumoshu: typo: "etcadm" should be "etcdadm"

mumoshu:
@whereisaaron Thanks for the suggestion 👍 

 I've added `etcd.snapshot.automated: true` for taking snapshots periodically, and  `etcd.disasterRecovery.automated: true` for enabling the automated disaster recovery feature.

mumoshu:
I've also added `etcd.version` for choosing either `2` or `3` as a temporary step until we settle to k8s 1.6 and eventually drop the support for etcd2.

mumoshu:
Due to my laziness, I kept the change from `/etc/etcd/ssl/` to `/etc/ssl/certs` as the directory containing enc files and the decrypted files as-is for now. Making it other than `/etc/ssl/certs` makes paths among the host and the etcd3 container inconsistent and complicate generation of several env files.

mumoshu:
Just for the record, conformance test is still failing with k8s 1.5.5 + etcd3 as is well known.

```
Summarizing 5 Failures:

[Fail] [k8s.io] SchedulerPredicates [Serial] [BeforeEach] validates that NodeSelector is respected if matching [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduler_predicates.go:78

[Fail] [k8s.io] Service endpoints latency [It] should not be very high [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/service_latency.go:114

[Fail] [k8s.io] SchedulerPredicates [Serial] [BeforeEach] validates resource limits of pods that are allowed to run [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/scheduler_predicates.go:78

[Fail] [k8s.io] Pods [It] should be submitted and removed [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/common/pods.go:250

[Fail] [k8s.io] Pods Delete Grace Period [It] should be submitted and removed [Conformance]
/go/kubernetes/_output/local/go/src/k8s.io/kubernetes/test/e2e/pods.go:176

Ran 117 of 474 Specs in 4897.388 seconds
FAIL! -- 112 Passed | 5 Failed | 0 Pending | 357 Skipped --- FAIL: TestE2E (4897.42s)
FAIL
```

mumoshu:
As k8s 1.5.5 + etcd2 is still passing conformance test on this branch, I'll merge this shortly and start testing with k8s 1.6 while enabling the optional etcd3 support.

mumoshu:
The integration test suite for etcdadm passed:

```
$ cd etcdadm
$ make test-integration
*snip*
finished testing recovery from permanent member failures

all the tests passed. tailing log...
```

mumoshu:
I'm not satisfied with the overall quality of this work but it seems to be working anyway.
Let's go forward.

Issue #396, opened 2017-03-08
Travis build and kube-aws occasionally fails while fetching Container Linux AMI data


mumoshu:
Producing an error like:
> failed getting AMI for config: error getting ami data for channel stable: failed to get AMI data: stable: Get https://coreos.com/dist/aws/aws-stable.json: dial tcp [2400:cb00:2048:1::6814:2fec]:443: connect: network is unreachable

Perhaps we'd better mock out?

c-knowles:
I've had the same on an actual deploy, perhaps worth adding a retry as well?

mumoshu:
It occurred for me, too. Adding a retry would be better.

mumoshu:
Followed up with #442

PR #418, opened 2017-03-16
Auth token file support

danielfm:
This PR adds support for authentication via static token files, as documented [here](https://kubernetes.io/docs/admin/authentication/#static-token-file).

Authentication via static token files, although not very flexible, is a pretty simple solution for users getting started with Kubernetes and/or want to have some mechanism to give access to different users, but do not want the complexity overhead of a more flexible solution, such as [dex](https://github.com/coreos/dex).

This PR will also be the foundation which #414 will be laid on.

Remaining TODOs:

- [x] Write tests
- [x] Document new feature (including required steps when upgrading an existing cluster, if any)
- [x] Spawn a few test clusters to validate the changes (with and without custom tokens to validate both scenarios)
- [x] Validate auth tokens file according to the current specification[1]

[1] The token file is a csv file with a minimum of 3 columns: token, user name, user uid, followed by optional group names. Note, if you have more than one group the column must be double quoted


danielfm:
@mumoshu et al: I still want to add an extra validation to ensure operators can only create clusters if the provided auth tokens file is valid according to the specs, but overall I think this PR is ready for review.

Please let me know if there's anything I missed.

danielfm:
All done! Waiting for reviews.

danielfm:
The tests are kinda flaky right now, this is the error I've been seeing:

```
failed to create cluster driver : failed to load node pool #0: failed getting AMI for config: error getting ami data for channel stable: failed to get AMI data: stable: invalid status code: 409
```

core/controlplane/config/config.go
```diff
@@ -1178,7 +1203,7 @@ func (c ControllerSettings) Valid() error {
 func (c Experimental) Valid() error {
 	for _, taint := range c.Taints {
 		if taint.Effect != "NoSchedule" && taint.Effect != "PreferNoSchedule" {
-			return fmt.Errorf("Effect must be NoSchdule or PreferNoSchedule, but was %s", taint.Effect)
+			return fmt.Errorf("Effect must be NoSchedule or PreferNoSchedule, but was %s", taint.Effect)
 		}
```
mumoshu: Good catch 👍

core/controlplane/config/config.go
```diff
@@ -1045,7 +1070,7 @@ func (c DeploymentSettings) Valid() (*DeploymentValidationResult, error) {
 	return &DeploymentValidationResult{vpcNet: vpcNet}, nil
 }
 
-func (c DeploymentSettings) TLSAssetsEncryptionEnabled() bool {
+func (c DeploymentSettings) AssetsEncryptionEnabled() bool {
```
mumoshu: Thanks for paying attention to the naming 👍

mumoshu:
@danielfm LGTM 👍 Thanks for your contribution!

Btw sry for the flake. The fix for the it is planned in #396

PR #420, opened 2017-03-18
Upgrade heapster to version 1.3.0

AlmogBaku:
Upgrade heapster to version 1.3.0


AlmogBaku:
@danielfm thanks. fixed.

danielfm:
Also, if I may suggest, instead of keeping the hard-coded version in the template, like this:

```yaml
metadata:
  name: heapster-v1.2.0
```

... why not use the tag from the config, like this:

```yaml
metadata:
  name: heapster-{{.HeapsterImage.Tag}}
```

I did not test this though, but it seems possible to get rid of this duplication here.

mumoshu:
Thanks for the help @AlmogBaku @danielfm!
This change LGTM.

Btw, @danielfm's suggested improvement sounds great. Would you mind submitting an another github issue for that?
(And perhaps we'd better do the same for other images like kube-dns?

PR #425, opened 2017-03-21
Emit errors when kube-aws sees unexpected keys in cluster.yaml

### LINKED ISSUE #404: kube-aws should warn or fail when it sees unexpected keys in cluster.yaml
cc @redbaron I believe you've mentioned about this feature in a comment somewhere but I'm not sure where it was!

mumoshu:
Resolves #404


mumoshu:
This doesn't cover e.g. `subnets`, `worker.nodePools[].subnets`, `worker.nodePools[].spotFleet.launchConfiguration[]`.
However, believing it is much better than nothing, I'd like to set this as the starting point 😃

PR #426, opened 2017-03-21
Automatic invalidations of *.enc files

### LINKED ISSUE #402: `.enc` credential files not recreated if existing on `kube-aws validate`
I was having issues with a bad certificate (was using the wrong domain), so I re-rendered credentials and redeployed.

I was still hitting the issue, and noticed that the timestamps for the `.enc` files in the credentials folder were older than the `.pem` files they corresponded to. Removing the credentials files and redoing the whole process fixed the issue.

mumoshu:
Resolves #402

### DIFF
--- a/cmd/render.go
+++ b/cmd/render.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/coreos/kube-aws/core/controlplane/config"
 	"github.com/coreos/kube-aws/core/root"
-	"github.com/coreos/kube-aws/core/root/render"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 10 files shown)


redbaron:
@mumoshu do I read it correctly, that with this change it is mandatory that unencrypted keys are stored next to encrypted ones? So they have to be commited to Git or be stored elsewhere and then fetched prior to running kube-aws? 

AFAIK old behaviour was using .enc files if they present and not try to recreate raw versions of them, which made it possible to commit them to Git, because on subsequent checkout kube-aws would just use .enc files.

Reason I am asking is that we are migrating away to kube-aws way of managing certs, but we can't commit key files to Git for obvious reasons. How do you solve it?

PR #429, opened 2017-03-22
Allow a custom KMS key with encrypted etcd volume

### LINKED ISSUE #393: Specifying which KMS key to be used for etcd data volume encryption
This feature request is derived from https://github.com/coreos/kube-aws/issues/383#issuecomment-284662241

Can we just reuse the key specified with `kmsKeyArn` for data volume encryption, too?

swestcott:
Specify a custom KMS key ARN to use with an encrypted etcd data volume. If omitted the account default key will be used. Have also fixed up the IAM policy so encrypted volumes can be reattached to future etcd instances

Fixes #393

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -891,6 +891,10 @@ func (c Cluster) valid() error {
 		return err
 	}
 
+	if err := c.EtcdSettings.Valid(); err != nil {
+		return err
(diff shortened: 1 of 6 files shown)


mumoshu:
@swestcott Wow, thanks for the great work!
LGTM 👍

PR #430, opened 2017-03-22
Repair CF rollbacks to work after the S3 userdata improvement

### LINKED ISSUE #352: CF rollback wouldn't work because of S3 userdata
kube-aws should upload userdata to S3 bucket using hash name of the file, that is because if we use `WaitSignal` and we have an error in userdata, then CF tries to rollback to previous CF state, but that state doesn't include data on S3 bucket, therefore instance which is being rolled back continues to come up with broken metadata and CF rollback fail eventually.

mumoshu:
Fixes #352


mumoshu:
After a rollback, the launch configuration refers to the old userdata as expected 👍 

image

image

PR #439, opened 2017-03-23
Automatically creates the token auth file if it isn't present

danielfm:
This PR improves UX after #418, hopefully avoiding issues like [this](https://github.com/kubernetes-incubator/kube-aws/issues/427#issuecomment-288551868).

PR #447, opened 2017-03-23
Check for errors when trying to create the auth token file

danielfm:
@mumoshu addressing your comment in my last PR #439.

### DIFF
--- a/core/controlplane/config/token_config.go
+++ b/core/controlplane/config/token_config.go
@@ -160,7 +160,7 @@ func ReadOrEncryptAuthTokens(dirname string, encryptor CachedEncryptor) (*Encryp
 		if err != nil {
 			return nil, err
 		}
-		defer file.Close()
+		file.Close()
 	}
 
 	if _, err := ReadRawAuthTokens(dirname); err != nil {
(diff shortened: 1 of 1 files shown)


core/controlplane/config/token_config.go
```diff
@@ -156,10 +156,11 @@ func ReadOrEncryptAuthTokens(dirname string, encryptor CachedEncryptor) (*Encryp
 
 	// Auto-creates the auth token file, useful for those coming from previous versions of kube-aws
 	if _, err := os.Stat(authTokenPath); os.IsNotExist(err) {
-		_, err := os.OpenFile(authTokenPath, os.O_RDONLY|os.O_CREATE, 0600)
+		file, err := os.OpenFile(authTokenPath, os.O_RDONLY|os.O_CREATE, 0600)
 		if err != nil {
 			return nil, err
 		}
+		defer file.Close()
```
redbaron: you never use `file`, why `defer`?

danielfm: @redbaron I just wanted to make sure all handles were closed properly. Is this unnecessary in this case?

redbaron: `defer` delays `file.Close()` until function exits. you don't use `file` anywhere, so close it right there, without `defer`

danielfm: @redbaron done! Thanks for your feedback.

mumoshu:
LGTM 👍 
Thanks for contributing @danielfm, and for reviewing @redbaron :bow:

PR #448, opened 2017-03-23
China region improvements and update some deprecated vars.

camilb:
Hi @mumoshu 
Made some small changes, mostly for China region.

1. Replace deprecated vars:
`KUBELET_VERSION` ==>   `KUBELET_IMAGE_TAG`
`KUBELET_ACI`.          ==>   `KUBELET_IMAGE_URL`
`RKT_OPT`                   ==>   `RKT_RUN_ARGS`

2. Fix `pause-amd64.service` to start at first boot.

3. Allow to pull **flannel** image from a docker repo In China region
sometimes it takes ~30 min to pull the image from quay.

Please review.

Thanks!

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -351,10 +351,12 @@ ssh_authorized_keys:
         [Unit]
         Description=Pull and tag a mirror image for pause-amd64
         Wants=docker.service
+        After=docker.service
 
         [Service]
-        Type=oneshot
+        Restart=on-failure
         RemainAfterExit=true
+        ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStartPre=/usr/bin/docker pull {{.PauseImage.RepoWithTag}}
         ExecStart=/usr/bin/docker tag {{.PauseImage.RepoWithTag}} gcr.io/google_containers/pause-amd64:3.0
         ExecStop=/bin/true
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -87,6 +87,18 @@ coreos:
             set /coreos.com/network/config '{"Network" : "{{.PodCIDR}}", "Backend" : {"Type" : "vxlan"}}'
             TimeoutStartSec=120
 
+{{if .FlannelImage.RktPullDocker}}
+        - name: 20-flannel-custom-image.conf
+          content: |
+            [Unit]
+            PartOf=flanneld.service
+            Before=docker.service
+
+            [Service]
+            Environment="FLANNEL_IMAGE={{.FlannelImage.RktRepo}}"
+            Environment="RKT_RUN_ARGS={{.FlannelImage.Options}}"
+{{end}}
+
     - name: kubelet.service
       command: start
       runtime: true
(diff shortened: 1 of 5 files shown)


camilb:
I signed it!

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -416,14 +428,15 @@ ssh_authorized_keys:
       content: |
         [Unit]
         Description=Pull and tag a mirror image for pause-amd64
-        After=docker.service
         Wants=docker.service
 
         [Service]
-        Restart=on-failure
+        Type=oneshot
```
mumoshu: so that `docker.service` (and other dependent services) wait until `pause-amd64.service` finished. Good 👍

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -416,14 +428,15 @@ ssh_authorized_keys:
...
         RemainAfterExit=true
         ExecStartPre=/usr/bin/docker pull {{.PauseImage.RepoWithTag}}
```
mumoshu: You probably want to some retry logic to make this reliable against e.g. a temporary network failure. AFAIK, `oneshot` services don't restart on failures.

mumoshu: Something like `/opt/binn/retry docker pull {{.PauseImage.RepoWithTag}}'`??

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -416,14 +428,15 @@ ssh_authorized_keys:
       content: |
         [Unit]
         Description=Pull and tag a mirror image for pause-amd64
-        After=docker.service
```
mumoshu: What is the intention behind the removal of this line?
You probably want to ensure that `docker.service` is up and running before running `pause-amd64.service`?
Then, `After=docker.service` + `Wants=docker.service` + `retry bash -c '/usr/bin/systemctl is-active docker.service && /usr/bin/docker pull ...'` would be what you want/my recommendation.
If you don't care about temporary `docker.service` failures, i.e. no retrying wanted, `After=docker.service` + `Requires=docker.service` + `/usr/bin/docker pull ...` would be sufficient. However, can we really assume `docker.service` to start reliably?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -100,9 +112,9 @@ coreos:
         After=cfn-etcd-environment.service
         [Service]
         EnvironmentFile=-/etc/etcd-environment
-        Environment=KUBELET_VERSION={{.K8sVer}}
-        Environment=KUBELET_ACI={{.HyperkubeImage.RktRepoWithoutTag}}
-        Environment="RKT_OPTS=--volume dns,kind=host,source=/etc/resolv.conf {{.HyperkubeImage.Options}}\
+        Environment=KUBELET_IMAGE_TAG={{.K8sVer}}
+        Environment=KUBELET_IMAGE_URL={{.HyperkubeImage.RktRepoWithoutTag}}
+        Environment="RKT_RUN_ARGS=--volume dns,kind=host,source=/etc/resolv.conf {{.HyperkubeImage.Options}}\
```
mumoshu: Where can I locate info about those old/new env vars, if any?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -338,14 +350,16 @@ ssh_authorized_keys:
       content: |
         [Unit]
         Description=Pull and tag a mirror image for pause-amd64
-        After=docker.service
```
mumoshu: I guess doing this ends up `pause-amd64.service` to be failed (and not restarted) forever when `pause-amd64.service` "luckily" started earlier than `docker.service`?
Relates: https://github.com/kubernetes-incubator/kube-aws/pull/448/files#r107811729

mumoshu:
Thanks as always @camilb!
I'm happy to review 👍

core/controlplane/config/templates/cloud-config-worker
camilb: https://github.com/coreos/coreos-overlay/blob/master/app-admin/kubelet-wrapper/files/kubelet-wrapper

core/controlplane/config/templates/cloud-config-worker
camilb: This was my first choice but without "retry bash -c" . For unknown reasons, in latest CoreOS - stable version, the service start it's never called. The only way to start the service is to add it as dependency to another service. The "After=docker.service" line probably should be kept, as using only "Wants=docker.service" could fail. I will try again with this setup and make the necessary changes. In the previous version of CoreOs it was working fine.

core/controlplane/config/templates/cloud-config-controller
camilb: The start is never called. In logs I can see "loaded" and "enabled" but not even one intent to start it.

camilb:
Hi @mumoshu just updated "pause-amd64.service" to avoid failures. Tested several times in aws-china.

mumoshu:
LGTM. Thanks for your efforts @camilb 👍

PR #449, opened 2017-03-23
Kubelet TLS bootstrapping

### LINKED ISSUE #406: TLS bootstrapping with the cluster-level CA

danielfm:
This PR introduces an experimental feature for provisioning TLS certificates for worker nodes via the alpha [certificate signing requests](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/) workflow.

If RBAC is enabled in `cluster.yaml`, I also made sure to set up the appropriate cluster role bindings so that the token used for sending the certificate signing requests can only make requests related to certificate provisioning, as instructed by the official documentation.

Checklist:

- [x] Add a `kube-aws render token-file` that creates a token auth file with a cluster TLS bootstrap (if the feature is turned on in `cluster.yaml`, or an empty file otherwise)
- [x] Check whether the auth token file has a proper bootstrap token when the TLS bootstrap feature is turned on; if not, return a self-explaining message so the user knows what to do in order to fix the problem
- [x] Read the bootstrap token from the auth token file and generate an encrypted version of it
- [x] Send the encrypted bootstrap token in worker userdata
- [x] Replace the decrypted token in the kubeconfig used for the TLS bootstrapping workflow
- [x] Fix / add more tests wherever possible
- [x] Manual tests
  - [x] Create clusters with TLS bootstrapping enabled and disabled
  - [x] Make sure it plays well with other experimental features (i.e. self-hosted Calico, AWS node labels, node draining)
  - [x] Try to add worker nodes (and stop/start the kubelet on an existing node) to see if they can join the cluster without the need to manually approve every new certificate signing request
  - [x] Test token rollout methods (**Edit:** `kube-aws update` works as expected, since the generated certificates remain valid even though the token used to request it has changed)

Closes #406.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -53,12 +53,12 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
 	if report != "" {
 		fmt.Fprintf(os.Stderr, "Validation Report: %s\n", report)
 	}
-
 	if err != nil {
 		return err
 	}
-	fmt.Printf("stack template is valid.\n\n")
 
-	fmt.Printf("Validation OK!\n")
+	fmt.Printf("stack template is valid.\n\n")
+	fmt.Println("Validation OK!")
+
 	return nil
 }
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -58,7 +58,7 @@ func NewDefaultCluster() *Cluster {
 		ClusterAutoscalerSupport: ClusterAutoscalerSupport{
 			Enabled: false,
 		},
-		ClusterTLSBootstrap: ClusterTLSBootstrap{
+		TLSBootstrap: TLSBootstrap{
 			Enabled: false,
 		},
 		EphemeralImageStorage: EphemeralImageStorage{
@@ -464,7 +464,7 @@ type Experimental struct {
 	AwsEnvironment           AwsEnvironment           `yaml:"awsEnvironment"`
 	AwsNodeLabels            AwsNodeLabels            `yaml:"awsNodeLabels"`
 	ClusterAutoscalerSupport ClusterAutoscalerSupport `yaml:"clusterAutoscalerSupport"`
-	ClusterTLSBootstrap      ClusterTLSBootstrap      `yaml:"clusterTLSBootstrap"`
+	TLSBootstrap             TLSBootstrap             `yaml:"tlsBootstrap"`
 	EphemeralImageStorage    EphemeralImageStorage    `yaml:"ephemeralImageStorage"`
 	Kube2IamSupport          Kube2IamSupport          `yaml:"kube2IamSupport,omitempty"`
 	LoadBalancer             LoadBalancer             `yaml:"loadBalancer"`
@@ -513,7 +513,7 @@ type ClusterAutoscalerSupport struct {
 	Enabled bool `yaml:"enabled"`
 }
 
-type ClusterTLSBootstrap struct {
+type TLSBootstrap struct {
 	Enabled bool `yaml:"enabled"`
 }
 
@@ -783,7 +783,7 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 	if stackConfig.UserDataEtcd, err = userdatatemplate.GetString(opts.EtcdTmplFile, stackConfig.Config); err != nil {
 		return nil, fmt.Errorf("failed to render etcd cloud config: %v", err)
 	}
-	if len(stackConfig.Config.AuthTokensConfig.KubeletBootstrapToken) == 0 && c.DeploymentSettings.Experimental.ClusterTLSBootstrap.Enabled {
+	if len(stackConfig.Config.AuthTokensConfig.KubeletBootstrapToken) == 0 && c.DeploymentSettings.Experimental.TLSBootstrap.Enabled {
 		bootstrapRecord, err := RandomBootstrapTokenRecord()
 		if err != nil {
 			return nil, err
@@ -969,7 +969,7 @@ func (c Cluster) valid() error {
 		fmt.Println(`WARNING: instance types "t2.nano" and "t2.micro" are not recommended. See https://github.com/kubernetes-incubator/kube-aws/issues/258 for more information`)
 	}
 
-	if c.Experimental.ClusterTLSBootstrap.Enabled && !c.Experimental.Plugins.Rbac.Enabled {
+	if c.Experimental.TLSBootstrap.Enabled && !c.Experimental.Plugins.Rbac.Enabled {
 		fmt.Println(`WARNING: enabling cluster-level TLS bootstrapping without RBAC is not recommended. See https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/ for more information`)
 	}
 
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -58,6 +58,9 @@ func NewDefaultCluster() *Cluster {
 		ClusterAutoscalerSupport: ClusterAutoscalerSupport{
 			Enabled: false,
 		},
+		ClusterTLSBootstrap: ClusterTLSBootstrap{
+			Enabled: false,
+		},
 		EphemeralImageStorage: EphemeralImageStorage{
 			Enabled:    false,
 			Disk:       "xvdb",
@@ -461,6 +464,7 @@ type Experimental struct {
 	AwsEnvironment           AwsEnvironment           `yaml:"awsEnvironment"`
 	AwsNodeLabels            AwsNodeLabels            `yaml:"awsNodeLabels"`
 	ClusterAutoscalerSupport ClusterAutoscalerSupport `yaml:"clusterAutoscalerSupport"`
+	ClusterTLSBootstrap      ClusterTLSBootstrap      `yaml:"clusterTLSBootstrap"`
 	EphemeralImageStorage    EphemeralImageStorage    `yaml:"ephemeralImageStorage"`
 	Kube2IamSupport          Kube2IamSupport          `yaml:"kube2IamSupport,omitempty"`
 	LoadBalancer             LoadBalancer             `yaml:"loadBalancer"`
@@ -506,6 +510,10 @@ type AwsNodeLabels struct {
 }
 
 type ClusterAutoscalerSupport struct {
+	Enabled bool `yaml:"enabled"`
+}
+
+type ClusterTLSBootstrap struct {
 	Enabled bool `yaml:"enabled"`
 }
 
@@ -744,7 +752,7 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 		}
 		stackConfig.Config.AuthTokensConfig = compactAuthTokens
 	} else {
-		rawAuthTokens, err := ReadOrCreateUnecryptedCompactAuthTokens(opts.AssetsDir)
+		rawAuthTokens, err := ReadOrCreateUnencryptedCompactAuthTokens(opts.AssetsDir)
 		if err != nil {
 			return nil, err
 		}
@@ -752,19 +760,15 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 	}
 	if c.ManageCertificates {
 		if c.AssetsEncryptionEnabled() {
-
 			compactAssets, err = ReadOrCreateCompactTLSAssets(opts.AssetsDir, KMSConfig{
 				Region:         stackConfig.Config.Region,
 				KMSKeyARN:      c.KMSKeyARN,
 				EncryptService: c.ProvidedEncryptService,
 			})
-			if err != nil {
-				return nil, err
-			}
 
 			stackConfig.Config.TLSConfig = compactAssets
 		} else {
-			rawAssets, err := ReadOrCreateUnecryptedCompactTLSAssets(opts.AssetsDir)
+			rawAssets, err := ReadOrCreateUnencryptedCompactTLSAssets(opts.AssetsDir)
 			if err != nil {
 				return nil, err
 			}
@@ -778,6 +782,13 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 	}
 	if stackConfig.UserDataEtcd, err = userdatatemplate.GetString(opts.EtcdTmplFile, stackConfig.Config); err != nil {
 		return nil, fmt.Errorf("failed to render etcd cloud config: %v", err)
+	}
+	if len(stackConfig.Config.AuthTokensConfig.KubeletBootstrapToken) == 0 && c.DeploymentSettings.Experimental.ClusterTLSBootstrap.Enabled {
+		bootstrapRecord, err := RandomBootstrapTokenRecord()
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("kubelet bootstrap token not found in ./credentials/tokens.csv\n\nTo fix this, append the following line to ./credentials.tokens.csv:\n%s", bootstrapRecord)
 	}
 
 	stackConfig.StackTemplateOptions = opts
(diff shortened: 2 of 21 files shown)


danielfm:
[Just when I thought I was almost done with this PR](https://kubernetes.io/docs/admin/bootstrap-tokens/). 😅

danielfm:
I believe this PR is ready for review! 🎉 

PS: This changes quite a few things, so please take your time to review this and let me know of _any_ changes you guys think of to get this working as smooth as possible.

/cc @mumoshu

mumoshu:
Thanks for efforts @danielfm!

First of all, [regarding your comment](https://github.com/kubernetes-incubator/kube-aws/pull/449#issuecomment-289944136), should we or do we need to translate this work with the [bootstrap tokens](https://kubernetes.io/docs/admin/bootstrap-tokens/) after k8s 1.6?

core/controlplane/config/config.go
```diff
@@ -63,6 +63,9 @@ func NewDefaultCluster() *Cluster {
 			Disk:       "xvdb",
 			Filesystem: "xfs",
 		},
+		ClusterTLSBootstrap: ClusterTLSBootstrap{
+			Enabled: false,
+		},
```
mumoshu: Very nit picky but moving this block to the line 61 would be better for a consistent ordering between this func and the definition of the Experimental struct.

danielfm:
This stuff just came out, so I'm still reading and learning how it compares to what I did here, but for now I think the bootstrap tokens might not be the ideal solution for kube-aws as it is for a few reasons:

- Bootstrap tokens are new to Kubernetes 1.6, while the implementation in this PR can be used for 1.4.x as well; the only requirements are RBAC and CSR support.
- If I understood correctly, the format used for bootstrap tokens (`[a-z0-9]{6}\.[a-z0-9]{16}`) don't follow the security guidelines outlined [here](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/), which states that a token, to be considered secure, should represent at least 128 bits of entropy. The implementation in this PR, on the other hand, generates tokens that satisfy this requirement.
- Bootstrap tokens have an expiration date, after which the token will be no longer valid. This is nice for security reasons, but it might cause problems for less careful operators in more long-lived clusters. In this PR, the token is valid for as long as it's listed in `./credentials/tokens.csv`, and the operator can rotate the token at any time by changing this file, without risking problems due to expired tokens when trying to scale the cluster up (or adding new node pools) at a later moment.
- Bootstrap tokens' usage scope are defined by `usage-bootstrap-*` attributes (like `authentication` and `signing`), which for me doesn't make much sense since we can use RBAC to restrict (with greater flexibility) which actions can be performed on which resources.

That's my first opinion on this, but I'm eager to hear what you guys think.

core/controlplane/config/config.go
```diff
@@ -724,50 +732,51 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
...
+			return nil, err
+		}
+		stackConfig.Config.AuthTokensConfig = compactAuthTokens
+	} else {
+		rawAuthTokens, err := ReadOrCreateUnencryptedCompactAuthTokens(opts.AssetsDir)
+		if err != nil {
+			return nil, err
+		}
+		stackConfig.Config.AuthTokensConfig = rawAuthTokens
+	}
```
mumoshu: Note to self: Probably this block is from #470

core/nodepool/config/config.go
```diff
@@ -37,7 +40,7 @@ type ProvidedConfig struct {
 	cfg.Experimental        `yaml:",inline"`
 	Private                 bool   `yaml:"private,omitempty"`
 	NodePoolName            string `yaml:"name,omitempty"`
-	providedEncryptService  cfg.EncryptService
+	ProvidedEncryptService  cfg.EncryptService
```
mumoshu: Could you leave `providedEncryptService` being a private member if it isn't referenced from another packages?

core/controlplane/config/config.go
danielfm: @mumoshu Yes, I found and fixed the bug here before #470 was opened. If it gets merged before this PR, I'll remove this.

core/nodepool/config/config.go
danielfm: @mumoshu The `ProvidedEncryptService` from both `.../core/controlplane/config` and `.../core/nodepool/config` are used in function `ConfigFromBytesWithEncryptService ` of `.../core/root/config`.

But I will re-check all my code to see if there are any new member that could be left private 👍

danielfm:
@mumoshu et al: I updated my previous comment regarding the new bootstrap tokens feature in Kubernetes 1.6, let me know what you think.

cmd/validate.go
```diff
@@ -53,12 +53,12 @@ func runCmdValidate(cmd *cobra.Command, args []string) error {
...
 	}
-
 	if err != nil {
 		return err
 	}
+
 	fmt.Printf("stack template is valid.\n\n")
+	fmt.Println("Validation OK!")
 
-	fmt.Printf("Validation OK!\n")
```
mumoshu: Good catch 👍

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -546,15 +557,19 @@ write_files:
       rkt run \
         --volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
         --mount=volume=ssl,target=/etc/kubernetes/ssl \
+        {{- if .Experimental.ClusterTLSBootstrap.Enabled }}
+        --volume=kube,kind=host,source=/etc/kubernetes,readOnly=false \
+        --mount=volume=kube,target=/etc/kubernetes \
+        {{- end }}
```
mumoshu: When the cluster TLS bootstrap feature is enabled, the original volume and the mount for `/etc/kubernetes/ssl` seems not be necessary; then, how about something like:

```
{{- if .Experimental.ClusterTLSBootstrap.Enabled }}
--volume=kube,kind=host,source=/etc/kubernetes,readOnly=false \
--mount=volume=kube,target=/etc/kubernetes \
{{- else}}
--volume=ssl,kind=host,source=/etc/kubernetes/ssl,readOnly=false \
--mount=volume=ssl,target=/etc/kubernetes/ssl \
{{- end }}
```

?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -351,24 +359,25 @@ coreos:
...
+          {{- if .Experimental.ClusterTLSBootstrap.Enabled }}
+          -v /var/run/kubernetes:/var/run/kubernetes \
+          {{- end }}
+          -e INSTANCE_ID=${INSTANCE_ID} \
+          -e SECURITY_GROUPS=${SECURITY_GROUPS} \
+          -e AUTOSCALINGGROUP=${AUTOSCALINGGROUP} \
+          -e LAUNCHCONFIGURATION=${LAUNCHCONFIGURATION} \
+          {{.HyperkubeImage.RepoWithTag}} /bin/bash \
+            -ec 'echo "placing labels and annotations with additional AWS parameters."; \
+             kctl="/kubectl --server=https://{{.ExternalDNSName}}:443 --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml"; \
```
mumoshu: fyi: `ExternalDNSName` should be `APIEndpoint.DNSName` once #468 is merged

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -820,6 +832,37 @@ write_files:
...
+            verbs:
+            - get
+          - apiGroups: ["*"]
+            resources:
+            - pods/eviction
+            verbs:
+            - create
+          - nonResourceURLs: ["*"]
+            verbs: ["*"]
+
```
mumoshu: Are the role and the binding for `bootstrapped-node` necessary for this feature?

core/root/config/config.go
```diff
@@ -112,6 +112,12 @@ func ConfigFromBytesWithEncryptService(data []byte, encryptService controlplane.
 		return nil, err
 	}
 	c.ProvidedEncryptService = encryptService
+
+	// Uses the same encrypt service for node pools
+	for _, p := range c.NodePools {
+		p.ProvidedEncryptService = encryptService
+	}
```
mumoshu: This is just to confirm, but this is required to pass one of newly added tests, right?

test/integration/maincluster_test.go
```diff
@@ -1029,6 +1039,8 @@ worker:
       enabled: true
     clusterAutoscalerSupport:
       enabled: true
+    clusterTLSBootstrap:
+      enabled: true # Must be ignored, value is synced with the one from control plane
```
mumoshu: 👍

core/controlplane/config/templates/cloud-config-worker
danielfm: Yes, great suggestion. 👍

core/controlplane/config/templates/cloud-config-controller
danielfm: Unfortunately, yes.

The certificates generated via the certificate signing workflow are bound to the `system:nodes` group, which do not have those permissions, which are needed by the node drainer and the node labeler (the one that adds AWS metadata as nodes labels and annotations).

Without this, node draining does not work, and the AWS metadata is not added to nodes bootstrapped via this CSR.

mumoshu: It totally make sense 👍 

So, in other words, do we need these role and the binding due to the lack of permissions in the k8s default role and the binding for `system:nodes`?
if that's the case, I guess it would be even nicer to add some comments including a link to the doc or code describing/defining the default role & binding, if any.

core/root/config/config.go
danielfm: Yes, without this, the `TestMainClusterConfig/WithExperimentalFeatures` test fails to complete the `assertConfig`, probably due to the code that's used to encrypt the bootstrap token for inclusion in `cloud-config-worker`.

I'll take a more careful look in this part to see if I missed something here.

core/controlplane/config/templates/cloud-config-controller
danielfm: > So, in other words, do we need these role and the binding due to the lack of permissions in the k8s default role and the binding for system:nodes?

Yes!

> if that's the case, I guess it would be even nicer to add some comments including a link to the doc or code describing/defining the default role & binding, if any.

Added!

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -146,9 +146,13 @@ coreos:
         --cluster_dns={{.DNSServiceIP}} \
         --cluster_domain=cluster.local \
         --cloud-provider=aws \
-        --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml \
+        {{- if .Experimental.ClusterTLSBootstrap.Enabled }}
+        --experimental-bootstrap-kubeconfig=/etc/kubernetes/worker-bootstrap-kubeconfig.yaml \
+        {{- else }}
         --tls-cert-file=/etc/kubernetes/ssl/worker.pem \
-        --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem
+        --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem \
```
mumoshu: I guess the `experimental-bootstrap-kubeconfig` setting result in kubelet to write issued worker key and cert to `/var/run/kubernetes`, which is the well-known location for the cert and the key according to https://kubernetes.io/docs/admin/kubelet/

If that's true, I wonder if we could improve code to assume all the worker key and the cert are located under the well-known location, regardless of we enable the tls bootstrap feature or not. Doing so would require `/opt/bin/decrypt-tls-assets` to write decrypted key and cert to `/var/run/kubernetes/<the well-known name for key or cert, if any>` but result in various if-else statements be simplified; For example, this code block can be just:

```
{{- if .Experimental.ClusterTLSBootstrap.Enabled }}
--experimental-bootstrap-kubeconfig=/etc/kubernetes/worker-bootstrap-kubeconfig.yaml \
{{- end}}
```

i.e. explicit `tls-cert-file` and `tls-private-file` can be omitted when the tls bootstrap is disabled?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -642,6 +664,9 @@ write_files:
       docker run --rm --net=host \
         -v /etc/kubernetes:/etc/kubernetes \
         -v /etc/resolv.conf:/etc/resolv.conf \
+        {{- if .Experimental.ClusterTLSBootstrap.Enabled }}
+        -v /var/run/kubernetes:/var/run/kubernetes \
+        {{- end }}
```
mumoshu: If we go [that](https://github.com/kubernetes-incubator/kube-aws/pull/449/files#r108824744) way, this block can be just:

```
-v /var/run/kubernetes:/var/run/kubernetes
```

without a `if`.

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -565,9 +580,16 @@ write_files:
              | base64 -d > $f
              mv -f $f ${encKey%.enc}
```
mumoshu: If we go [that](https://github.com/kubernetes-incubator/kube-aws/pull/449/files#r108824744) way, the decrypted file must be moved to `/var/run/kubernetes/<a well-known name if any>` rather than `${encKey%.enc}`

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -198,10 +202,14 @@ coreos:
...
         ExecStart=/bin/true
-        TimeoutStopSec=30s
+        TimeoutStopSec=60s
         ExecStop=/bin/sh -c '/usr/bin/rkt run \
         --volume=kube,kind=host,source=/etc/kubernetes,readOnly=true \
         --mount=volume=kube,target=/etc/kubernetes \
+        {{- if .Experimental.ClusterTLSBootstrap.Enabled }}
+        --volume=kuberun,kind=host,source=/var/run/kubernetes,readOnly=true \
+        --mount=volume=kuberun,target=/var/run/kubernetes \
+        {{- end }}
```
mumoshu: If we go [that](https://github.com/kubernetes-incubator/kube-aws/pull/449/files#r108824744) way, this block could be just:

```
--volume=kuberun,kind=host,source=/var/run/kubernetes,readOnly=true \
--mount=volume=kuberun,target=/var/run/kubernetes \
```

without a `if`.

core/controlplane/config/templates/cloud-config-worker
mumoshu: Or is there any flag to alter where the kubelet writes issued keys and certs to `/etc/kubernetes/ssl` rather than `/var/run/kubernetes`? so that we don't need `if-else`s to vary mounts accordingly to whether the TLS bootstrap is enabled or not.

danielfm: Probably `--cert-dir`, but I need to test this.

danielfm: Yes, it worked! 🎉

mumoshu: Great! 👍 :tada:

mumoshu:
LGTM 👍
Could you squash and rebase commits?

danielfm:
Just made a mess trying to squash the whole thing. 😅 

Let me spend a little more time testing things to ensure everything is still working so you can merge.

test/integration/maincluster_test.go
```diff
@@ -893,6 +898,8 @@ experimental:
     enabled: true
   clusterAutoscalerSupport:
     enabled: true
+  clusterTLSBootstrap:
```
mumoshu: Sry but the last nit; Would you mind informing me if there's any specific reason you've prefixed the key with "cluster"? Can we call it just `tlsBootstrap` for consistency with the k8s doc?

danielfm: @mumoshu I actually never gave much thought to the configuration name. I guess `tlsBootstrap` is better. 👍

danielfm:
@mumoshu fixed, and changed a few other minor things as well.

I also took the opportunity to test this PR with Calico enabled, which I forgot to do in my previous tests. I found a bug there, which I'm going to fix ASAP.

danielfm:
Okay, now apparently everything's working. 😄

mumoshu:
LGTM. Thanks a lot for the great work @danielfm 👍

Issue #450, opened 2017-03-23
Repository Renamed but not Code?


jeremyd:
I noticed this repository changed from github.com/coreos/kube-aws to github.com/kubernetes-incubator/kube-aws.  However, the code still imports from the old github.com/coreos/kube-aws.  Should we rename all these imports to kubernetes-incubator?  Otherwise, we can't compile in a fresh checkout.

mumoshu:
Thanks for the report and the PR 👍 
Closing this as #451 is merged.

PR #451, opened 2017-03-23
rename imports from github.com/coreos to github.com/kubernetes-incubator

jeremyd:
Here's a rename of all the imports to use kubernetes-incubator if we want that.  #450


mumoshu:
Good catch 👍 Thanks for the contribution @jeremyd!

Issue #383, opened 2017-03-06
etcdDataVolumeEncrypted not creating encrypted volumes


swestcott:
I've set `etcdDataVolumeEncrypted: true` in my `cluster.yaml` file however the resulting volumes are marked as `Encrypted: Not Encrypted` in the AWS console.

From a initial look, I wonder if 341323e0879f3e30170fe1ebc0ddd6f653245d26 inadvertently removed the conditional logic around `EtcdDataVolumeEncrypted` in `stack-template.json`?

swestcott:
A potential secondary issue, after re-adding `EtcdDataVolumeEncrypted`, it seems that the EBS volumes are encrypted with the default `aws/ebs` key rather than the `kms-key-arn` passed in to `kube-aws init`.

mumoshu:
@swestcott Thanks for the feedbacks 👍 

> From a initial look, I wonder if 341323e0879f3e30170fe1ebc0ddd6f653245d26 inadvertently removed the conditional logic around `EtcdDataVolumeEncrypted` in `stack-template.json`?

It certainly seems so 😢 

> it seems that the EBS volumes are encrypted with the default `aws/ebs` key rather than the `kms-key-arn` passed in to `kube-aws init`.

Good catch. Does it make sense to always use the kms key specified via `kmsKeyArn` for encrypting EBS volumes, or would you like an another configuration key like `etcdDataVolumeEncryptionKmsKeyArn`(very long 😄) for flexibility?

mumoshu:
@swestcott `etcdDataVolumeEncrypted` is fixed in #387 and included in v0.9.5-rc.2.
Could you confirm that it is now working?

mumoshu:
I'll keep this open until the second question in https://github.com/coreos/kube-aws/issues/383#issuecomment-284563722 is answered.

swestcott:
@mumoshu Reusing the existing `kmsKeyArn` value suites me. The core point for me is a kube-aws doesn't share it's key with anything else.

swestcott:
I've confirmed `EtcdDataVolumeEncrypted` is working on `master`

mumoshu:
Thanks for the confirmation @swestcott!
I'm closing this as fixed but please feel free to reopen if necessary.
Btw I've opened an another issue #393 for your feature request.
Let's keep discussing there.

Issue #373, opened 2017-03-01
Allow defining multiple k8s API endpoints with/without managed ELB


mumoshu:
Accordingly to use-cases:

* internal or external
* stable/unversioned one for CI&CD or dynamic/versioned one as we have today

This is intended to eventually resolve all the related issues #281 #285 #343

---

## Configuration examples

`cluster.yaml` after this feature is implemented would look like:

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: true
hostedZoneId: myhostedzone
controller:
  loadBalancer:
    private: false

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: false
    hostedZone:
      id: myhostedzone
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    private: true

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: true
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    private: false

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: false
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    subnets:
    - name: managedPublicSubnet1a

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    subnets:
    - name: managedPublicSubnet1a
```

If you'd like to just add alternate DNS names to a k8s api endpoint:
cc @c-knowles #343 

```yaml
apiEndpoints:
- dnsName: v1.k8s.external.example.com 
  loadBalancer:
    public: true
- dnsName: v1.k8s.external.example.com # this is also included in CNs for TLS cert generation, without creating an ELB
```

If you'd like to choose Route53 round-robin rather than ELB for load-balancing:
cc @spacepluk #281

```yaml
apiEndpoints:
- dnsName: v2.k8s.external.example.com
  dnsRoundRobin: # will enable a bash script in cloud-config-controller to update Route 53 record sets
    hostedZone:
      id: <id for the hosted zone external.example.com>
    securityGroups: # associated to controller nodes
    - id: sg-toallowexternalaccess
```

If you'd like to have both public and private ELBs for a k8s endpoint:
[...]

Issue #281, opened 2017-01-25
Allow to choose between ELB and Route53 round robin for the APIServer


spacepluk:
Hi,
I think this would be a nice option to allow deployments of really tiny (and cheap) clusters.

tarvip:
I would like to use 0.8.3 approach for small clusters where master node HA is not important. Instead of creating autoscaling group and ELB for 1 node, just use ElasticIP for external access and host internal ip for worker nodes.

redbaron:
Whats the point? ELB approach is generic enough to cover both small and HA setups

tarvip:
I guess the biggest difference is that you have to pay a bit more.
ElasticIP is free if it is attached to a host, ELB is not.

Another problem is when you don't want to open API server to the world (basically restrict access in security group to specific IPs). Then you have to create additional internal ELB for worker nodes and attach it to controller autoscaling group otherwise worker nodes cannot access API (this setup also assumes that you have recreated API certificate etc.). I think this is more generic problem, but with small clusters it means 2 ELBs in total.

redbaron:
> Another problem is when you don't want to open API server to the world

use internal ELB,or if you use external ELB add workers SG to ELB SG rules

tarvip:
> if you use external ELB add workers SG to ELB SG rules

I already tried that. It doesn't work.

mumoshu:
Hi @tarvip, I've recently worked on adding a generic support to switch internal/external ELB for serving Kubernetes API in #284.
Would you confirm if it works for your use-case?

To make an external ELB public, set `controller.loadBalancer.private` to false(which is also the default value for the key):

```yaml
controller:
  subnets:
  - name: private1
  - name: private2
  loadBalancer:
    private: false
```

And obviously, set `controller.loadBalancer.private` to true for an internal ELB.

mumoshu:
Hi @spacepluk @tarvip! Thanks for your interest on kube-aws.
Also thanks as always @redbaron for your support.

I occasionally receive feature requests like this.
Personally, I'm not too opposed to add it if it does contribute to more wider acceptance of kube-aws in the world, as long as it doesn't break or affect our primary focus: "a community-driven tool to manage **production-grade** CoreOS Kubernetes clusters on AWS".
Could you share me more details of your use-case? Why and for what do you want a tiny kube-aws cluster?

tarvip:
> Would you confirm if it works for your use-case?

[...]

PR #468, opened 2017-03-27
Support for multiple k8s API endpoints

### LINKED ISSUE #285: DNS names in TLS certificate
Hey guys,

is it possible to make the DNS names of the self signed certificates more dynamic. Right now you're creating the certificates based on

DNSNames: []string{
                        "kubernetes",
                        "kubernetes.default",
                        "kubernetes.default.svc",
                        "kubernetes.default.svc.cluster.local",
                        c.ExternalDNSName,

that DNS names in the tls_config.go. 

Is it possible to make that more flexible so that we are able to add more custom DNS names by using the cluster.config file?

Thank you

### LINKED ISSUE #343: Feature: Allow additional DNS names to be added to API TLS assets
One way to deploy large cluster updates is to deploy an entirely new cluster, export services from the old cluster, import to new cluster and switch DNS (this way works fine for stateless services).

The way I've been doing this is to give the cluster name a version and then set a Route 53 DNS alias to that version. i.e. `k8s-dev.DOMAIN` points to `k8s-dev-v2.DOMAIN`. This means that places referencing or calling the API server don't all need to be updated. It's a bit like a cluster blue/green deployment. To fulfil this, I've been generating our own certs entirely since the TLS bootstrap provided by kube-aws used to be less feature complete. In order for this to work with the TLS the certs need to have a DNS name entry of `k8s-dev.DOMAIN` as well as `k8s-dev-v2.DOMAIN`.

Since implementing self TLS management, kube-aws has added etcd management, HA controllers plus a few other changes meaning it's more difficult to keep up to date with TLS changes. So I'm considering whether to scrap this self-managed way and go with the feature kube-aws now provides. However, there's no way to add other DNS names into the cert and almost everything is coupled together in `render credentials`. The only part decoupled is CA generation.

I think this could be solved in a few ways:
1. Provide up to date guidance on generating all required TLS assets (it's not up to date in [the docs](https://github.com/coreos/kube-aws/blob/master/Documentation/kubernetes-on-aws-render.md#certificates-and-keys)).
1. Allow more DNS names to be passed into kube-aws for the API server cert.
1. Allow some of the TLS assets to be kube-aws generated while others are user generated.

My preference currently sits with 2 since I think it can be quite common to alias and version in this manner - usually for services rather than a whole cluster but the same principles apply.

mumoshu:
Applying this change enables you to use a new, alternative syntax to configure API endpoints.
The new syntax supports one or more API endpoints to be configured while the former syntax allowed only one of them.

There's also an ability to configure various options per API Endpoint. The options includes either reuse an existing ELB or create a managed ELB, whether to create a record set or not for it, and of course what DNS name is being used for the ELB.

The support for DNS round-robin(#281) is planned but not included in this commit

Ref #373
Resolves #285
Resolves #343

## Configuration examples

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: true
hostedZoneId: myhostedzone
controller:
  loadBalancer:
    private: false

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: false
    hostedZone:
      id: myhostedzone
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    private: true

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: true
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    private: false

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    private: false
```

and

```yaml
# before
externalDNSName: k8s.external.example.com
createRecordSet: false
controller:
  loadBalancer:
    subnets:
    - name: managedPublicSubnet1a

# after
apiEndpoints:
- dnsName: k8s.external.example.com 
  loadBalancer:
    subnets:
    - name: managedPublicSubnet1a
```

If you'd like to just add alternate DNS names to a k8s api endpoint:

```yaml
apiEndpoints:
- dnsName: v1.k8s.external.example.com 
  loadBalancer:
    public: true
- dnsName: v1.k8s.external.example.com # this is also included in CNs for TLS cert generation, without creating an ELB
```

---

fyi @c-knowles This is a WIP of #383 without support for DNS round-robin, which will fulfill all of your needs!


core/controlplane/config/templates/cloud-config-worker
```diff
@@ -391,7 +391,7 @@ coreos:
             } \
           } \
           }\"' \
-          https://{{.ExternalDNSName}}:443/api/v1/nodes/$(hostname)"
+          https://{{.APIEndpoint.DNSName}}:443/api/v1/nodes/$(hostname)"
```
mumoshu: Perhaps we'd better provide something like:

```golang
func (c ProvidedConfig) ExternalDNSName() string {
    fmt.Println("WARN: ExternalDNSName is deprecated and will be removed in v0.9.7. Please use APIEndpoint.Name instead")
    return c.APIEndpint.DNSName
}
```

so that customized `cloud-config-worker` files or changes from other PRs still referencing `.ExternalDNSName` won't break.

PR #472, opened 2017-03-28
Update kube-system using kubectl

jollinshead:
Currently, kube-system resources are created using the curl POST command.
If these resources are changed (such as an update), the curl POST command fails to reapply the new configuration - these Kubernetes resources are not updated.

This commit implements another way of updating kube-system resources via "kubectl apply". This method is ready used for calico.


mumoshu:
Thanks @jollinshead 👍 
The change LGTM. Could you sign the CLA?

jollinshead:
@mumoshu Will do - We're in the process of adding our organisation to The Linux Foundation contributors.

jollinshead:
@k8s-ci-robot "I signed it!"

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -468,52 +468,37 @@ write_files:
...
 
       {{if .Experimental.Plugins.Rbac.Enabled}}
-      # k8s > 1.5 already have this clusterRole, so we accept http.code != 200
       mfdir=/srv/kubernetes/rbac
-      if [[ $(post_yaml_status_code "@mfdir/cluster-roles/cluster-admin.yaml" \
-          "http://127.0.0.1:8081/apis/rbac.authorization.k8s.io/v1alpha1/clusterroles") -ne "200" ]] ; then
-          echo 'Failed to create the cluster role named "cluster-admin". It is ok if you are using k8s newer than 1.5'
-      fi
-
-      post_yaml "@${mfdir}/cluster-roles/cluster-admin.yaml" \
```
mumoshu: @jollinshead  Perhaps we'd better not to change the default role in k8s 1.5+?
If so, how about making it something like:

```bash
if ! kubectl_create "${mfdir}/cluster-roles/cluster-admin.yaml"; then
   echo 'Failed to create the cluster role named "cluster-admin". It is ok if you are using k8s newer than 1.5'
fi
```
as per lines from 504 to 407 in original code?

mumoshu:
Thanks for signing 👍 
LGTM overall but a minor comment added(sry

jollinshead:
@mumoshu Done 👍

mumoshu:
LGTM & Resolved conflicts with the recent changes in master. Merging...

mumoshu:
Done. Thanks for the contribution @jollinshead 👍

PR #473, opened 2017-03-28
New settings: nodeMonitorGracePeriod, disableSecurityGroupIngress for controller-manager, nodeStatusUpdateFrequency for worker kubelet

jeremyd:
## disableSecurityGroupIngress

This flag disables automatic security group rule management for ELBs.  
It requires that the user has setup a rule that allows inbound traffic on kubelet ports 
 from the local VPC subnet so that load balancers can access it.
Disable setting a new security group rule for every ELB.

## nodeMonitorGracePeriod

Set NodeMonitorGracePeriod <duration>.  Option on controller-manager (default 40s)
This is the amount of time which we allow running Node to be unresponsive before marking it unhealthy. 
Must be N times more than kubelet's nodeStatusUpdateFrequency.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -165,7 +165,7 @@ coreos:
         {{end}}--register-node=true \
         {{if .Experimental.Taints}}--register-schedulable=false \
         {{end}}--allow-privileged=true \
-        {{if .NodeStatusUpdateFrequency}}--node-status-update-frequency={{.NodeStatusUpdateFrequency}}
+        {{if .NodeStatusUpdateFrequency}}--node-status-update-frequency={{.NodeStatusUpdateFrequency}} \
         {{end}}--pod-manifest-path=/etc/kubernetes/manifests \
         --cluster_dns={{.DNSServiceIP}} \
         --cluster_domain=cluster.local \
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -165,7 +165,7 @@ coreos:
         {{end}}--register-node=true \
         {{if .Experimental.Taints}}--register-schedulable=false \
         {{end}}--allow-privileged=true \
-        {{if .Experimental.NodeStatusUpdateFrequency}}--node-status-update-frequency={{.Experimental.NodeStatusUpdateFrequency}}
+        {{if .NodeStatusUpdateFrequency}}--node-status-update-frequency={{.NodeStatusUpdateFrequency}}
         {{end}}--pod-manifest-path=/etc/kubernetes/manifests \
         --cluster_dns={{.DNSServiceIP}} \
         --cluster_domain=cluster.local \
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -165,7 +165,8 @@ coreos:
         {{end}}--register-node=true \
         {{if .Experimental.Taints}}--register-schedulable=false \
         {{end}}--allow-privileged=true \
-        --pod-manifest-path=/etc/kubernetes/manifests \
+        {{if .Experimental.NodeStatusUpdateFrequency}}--node-status-update-frequency={{.Experimental.NodeStatusUpdateFrequency}}
+        {{end}}--pod-manifest-path=/etc/kubernetes/manifests \
         --cluster_dns={{.DNSServiceIP}} \
         --cluster_domain=cluster.local \
         --cloud-provider=aws \
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -989,6 +989,10 @@ addons:
 #   # This is the amount of time which we allow running Node to be unresponsive before marking it unhealthy. 
 #   # Must be N times more than kubelet's nodeStatusUpdateFrequency (default 10s).
 #   nodeMonitorGracePeriod: "40s"
+#   
+#   # Command line flag passed to kubelet. (default 10s)
+#   # Specifies how often kubelet posts node status to master. Note: be cautious when changing the constant, it must work with nodeMonitorGracePeriod in nodecontroller.
+#   nodeStatusUpdateFrequency: "10s"
 
 # AWS Tags for cloudformation stack resources
 #stackTags:
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -237,6 +237,9 @@ worker:
 #        # Follow the aws convention of '/dev/xvd*' where '*' is a single letter from 'f' to 'z'
 #        device: "/dev/xvdf"
 #        path: "/ebs"
+#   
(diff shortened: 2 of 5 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -836,7 +836,15 @@ worker:
 #     # See https://github.com/kubernetes-incubator/kube-aws/issues/230 for more information.
 #     rbac:
 #       enabled: true
-#
+#   # This flag disables automatic security group rule management for ELBs.
```
mumoshu: I was confused "ELBs" with the one for a k8s api(`controller.loadBalancer`)

Perhaps you'd better express this like "When set to true, configures k8s aws provider so that it doesn't modify every node's security group to include an additional ingress rule per an ELB created for a k8s service whose `type` is `LoadBalancer`"?
Also providing a link to a relevant issue like https://github.com/kubernetes/kubernetes/issues/26670 would help users how to utilize the flag.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -836,7 +836,15 @@ worker:
...
-#
+#   # This flag disables automatic security group rule management for ELBs.  
+#   # It requires that the user has setup a rule that allows inbound traffic on kubelet ports 
+#   # from the local VPC subnet so that load balancers can access it.
+#   # Disable setting a new security group rule for every ELB.
+#   disableSecurityGroupIngress: false
+#   # Set NodeMonitorGracePeriod <duration>.  Option on controller-manager (default 40s)
+#   # This is the amount of time which we allow running Node to be unresponsive before marking it unhealthy. 
+#   # Must be N times more than kubelet's nodeStatusUpdateFrequency.
+#   nodeMonitorGracePeriod: "40s"
```
mumoshu: We we need to also add `nodeStatusUpdateFrequency` as a new setting key then?

core/controlplane/config/templates/cluster.yaml
```diff
@@ -836,7 +836,15 @@ worker:
 #     # See https://github.com/kubernetes-incubator/kube-aws/issues/230 for more information.
 #     rbac:
 #       enabled: true
-#
+#   # This flag disables automatic security group rule management for ELBs.  
+#   # It requires that the user has setup a rule that allows inbound traffic on kubelet ports 
+#   # from the local VPC subnet so that load balancers can access it.
+#   # Disable setting a new security group rule for every ELB.
+#   disableSecurityGroupIngress: false
```
mumoshu: AFAIK, the flag controls whether the k8s aws provider automatically configures every node's security group to include an additional sg ingress rule per k8s managed elb.
Then, how about naming this like `autoConfigureNodeSecurityGroupToAllowTrafficFromELB: true|false` for more clarity?

mumoshu:
Thanks as always @jeremyd!
Reviewed & added several comments to your great work 👍

core/controlplane/config/templates/cluster.yaml
jeremyd: I want to keep the name the same as the command line flag passed to the controller.  Generally I'll use this pattern to prevent obfuscation for any power users who need to dig down to what a flag does.

core/controlplane/config/templates/cluster.yaml
jeremyd: Yes that is more clear, fixed the wording.

core/controlplane/config/templates/cluster.yaml
jeremyd: While I'm not going to change that setting it makes sense to provide it.  Will do.

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -142,7 +142,8 @@ coreos:
         {{end}}--register-node=true \
         {{if .Experimental.Taints}}--register-schedulable=false \
         {{end}}--allow-privileged=true \
-        --pod-manifest-path=/etc/kubernetes/manifests \
+        {{if .Experimental.NodeStatusUpdateFrequency}}--node-status-update-frequency={{.Experimental.NodeStatusUpdateFrequency}}
```
jeremyd: @mumoshu So I attempted to add this setting, however, the setting does not seem to get populated during the rendering of the worker template.  How do I get this setting to be available in rendering the worker's user-data?  I think I'm missing something about how these settings get into the nodepools.  Thanks!  As is, this setting does not get applied when set.

mumoshu: @jeremyd Are you sure that you need to set not `experimental.nodeStatusUpdateFrequency` but `worker.nodePools[].nodeStatusUpdateFrequency` for node pools?

jeremyd: I know I need to get the setting into the node pools... just not quite sure how.  Any example setting that does a similar thing that you could point me at? @mumoshu ?

mumoshu: @jeremyd Perhaps reading comments around the `worker.nodePools[]` key in the `cluster.yaml` would help but to be short:

```yaml
worker:
  nodePools:
  - name: myPool1
    nodeStatusUpdateFrequency: <value as you like>
```

mumoshu:
```yaml
# For worker node pools...
worker:
  nodePools:
  - name: myPool1
    nodeStatusUpdateFrequency: <value as you like>

# For controller nodes...
experimental:
  nodeStatusUpdateFrequency: <value as you like>
```

mumoshu: @jeremyd Any chance you try again with the above info?

jeremyd: Yep, thanks for the pointers.  I think I have it now.  I'm giving it another test this morning to make sure it works.

jeremyd:
Don't merge yet.  Testing revealed some issues, working on it.

jeremyd:
This is ready and was tested both with and without these options specified.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -980,6 +980,19 @@ addons:
...
+#   
+#   # Command line flag passed to the controller-manager. (default 40s)
+#   # This is the amount of time which we allow running Node to be unresponsive before marking it unhealthy. 
+#   # Must be N times more than kubelet's nodeStatusUpdateFrequency (default 10s).
+#   nodeMonitorGracePeriod: "40s"
+#   
+#   # Command line flag passed to kubelet. (default 10s)
+#   # Specifies how often kubelet posts node status to master. Note: be cautious when changing the constant, it must work with nodeMonitorGracePeriod in nodecontroller.
+#   nodeStatusUpdateFrequency: "10s"
```
mumoshu: I guess you intended to add this comment about the `nodeStatusUpdateFrequency` setting, which is for node pools, under the `worker.nodePools[]` key but not under the `experimental` key?

jeremyd: Correct, fixed.

mumoshu:
@jeremyd LGTM 👍 Thanks for the contribution :tada:

PR #486, opened 2017-04-03
Correct file references

c-knowles:
Behaviour before this change is that the rescheduler does not get
deployed, in fact the entire `.Addons.Rescheduler.Enabled` block will
be missing from the controller file even when it is enabled.

Introduced by https://github.com/kubernetes-incubator/kube-aws/pull/472.

Should also fix some the cluster roles and TLS bootstrap, not tested that part but pretty certain it's currently broken.


mumoshu:
My bad 😢 Thanks for the fix!

PR #489, opened 2017-04-03
Improve auth tokens / TLS bootstrapping UX

### LINKED ISSUE #484: Bug: Auth tokens file not found
On latest master, perform the following:
1. `init`
2. `render credentials` with any CA
3. `render stack`

On 3, this receives
```
Error: Failed to initialize cluster driver: failed to read/create auth token assets: auth tokens file not found: credentials/tokens.csv

To fix this, please run the following command to create it: kube-aws render token-file
```

Which is ok apart from the output also says:
```
3. (Optional) Add more authentication tokens in ./credentials/tokens.csv
```

I think it's a little bit confusing. The output stating what can be done next probably needs to be improved. Perhaps on `render credentials` it should be saying `render token-file` is next, or it should be optional.

danielfm:
~~This is mostly done (I still want to add a few more tests to newly added code though), so it's pretty much ready to be reviewed.~~

You guys should be able to review this one commit at a time; I tried to be careful to explain the reasons behind each change. Ask me if something is not clear.

/cc @mumoshu @c-knowles 

Closes #484.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -996,16 +996,7 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
 		if err != nil {
 			return nil, err
 		}
-		return nil, fmt.Errorf("kubelet bootstrap token not found in ./credentials/tokens.csv.\n\nTo fix this, please append the following line to ./credentials.tokens.csv:\n%s", bootstrapRecord)
-
-	}
-
-	if len(stackConfig.Config.AuthTokensConfig.KubeletBootstrapToken) == 0 && c.DeploymentSettings.Experimental.TLSBootstrap.Enabled {
-		bootstrapRecord, err := RandomBootstrapTokenRecord()
-		if err != nil {
-			return nil, err
-		}
-		return nil, fmt.Errorf("kubelet bootstrap token not found in ./credentials/tokens.csv\n\nTo fix this, append the following line to ./credentials.tokens.csv:\n%s", bootstrapRecord)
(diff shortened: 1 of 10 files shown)


mumoshu:
Thanks for the quick-turnaround @danielfm 👍 
Automatically generating tokens.csv like this seems nice. Would you also be ok with this, @c-knowles?

c-knowles:
Cool! The changes look good to me. As I'm not using this feature yet, I don't have much of an opinion on whether it should be a separate command or not which is why the issue I filed was slightly open ended. Based on the sequence of commits I agree with the conclusions of @danielfm and think this change will help a lot of existing users plus reduces complexity of the extra command.

core/controlplane/config/config.go
```diff
@@ -969,12 +981,25 @@ func (c Cluster) StackConfig(opts StackTemplateOptions) (*StackConfig, error) {
...
 	if stackConfig.UserDataEtcd, err = userdatatemplate.GetString(opts.EtcdTmplFile, stackConfig.Config); err != nil {
 		return nil, fmt.Errorf("failed to render etcd cloud config: %v", err)
 	}
+
+	if len(stackConfig.Config.AuthTokensConfig.KubeletBootstrapToken) == 0 && c.DeploymentSettings.Experimental.TLSBootstrap.Enabled {
+		bootstrapRecord, err := RandomBootstrapTokenRecord()
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("kubelet bootstrap token not found in ./credentials/tokens.csv.\n\nTo fix this, please append the following line to ./credentials.tokens.csv:\n%s", bootstrapRecord)
```
mumoshu: > please append the following line to ./credentials.tokens.csv
nit: it should be `./credentials/tokens.csv`

mumoshu: Btw I like the kind error message 👍

mumoshu:
Thanks for your supports! Merging 💨

Issue #494, opened 2017-04-04
Kubernetes v1.6.x: CNI + Self-hosted Calico is broken


mumoshu:
Due to the dead-lock between the installation of cni config for calico to `/etc/kubernetes/cni/net.d` and the node readiness explained in https://github.com/kubernetes/kubernetes/issues/43815

Calico and other pods are pending:

```
kubectl get po -n kube-system
NAME                                                                    READY     STATUS    RESTARTS   AGE
calico-policy-controller-2425378810-7nw53                               0/1       Pending   0          1h
heapster-v1.3.0-170904332-xtdf9                                         0/2       Pending   0          1h
kube-apiserver-ip-10-0-0-140.ap-northeast-1.compute.internal            1/1       Running   0          1h
kube-apiserver-ip-10-0-0-41.ap-northeast-1.compute.internal             1/1       Running   0          1h
kube-controller-manager-ip-10-0-0-140.ap-northeast-1.compute.internal   1/1       Running   0          1h
kube-controller-manager-ip-10-0-0-41.ap-northeast-1.compute.internal    1/1       Running   0          1h
kube-dns-1358247298-l64sl                                               0/4       Pending   0          1h
kube-dns-autoscaler-2700806437-8nkv9                                    0/1       Pending   0          1h
kube-proxy-ip-10-0-0-140.ap-northeast-1.compute.internal                1/1       Running   0          1h
kube-proxy-ip-10-0-0-41.ap-northeast-1.compute.internal                 1/1       Running   0          1h
kube-scheduler-ip-10-0-0-140.ap-northeast-1.compute.internal            1/1       Running   0          1h
kube-scheduler-ip-10-0-0-41.ap-northeast-1.compute.internal             1/1       Running   0          1h
kubernetes-dashboard-v1.5.1-xnwmh                                       0/1       Pending   0          1h
```

 due to the node being unready when the cni implementation(=calico) is not ready:

```
$ journalctl -u kubelet.service
*snip*
cni.go:157] Unable to update cni config: No networks found in /etc/kubernetes/cni/net.d
```

and results in the dead-lock.

Not sure how we could fix it yet.

mumoshu:
So - how Self-hosted Calico can properly be configured after https://github.com/kubernetes/kubernetes/pull/43474?

mumoshu:
[...]

PR #496, opened 2017-04-04
Bump to calico 2.1.1

redbaron:
Works well for us with 1.5.5, might also improve 1.6 compatibility


mumoshu:
Thanks @redbaron!
I'm taking a look into http://docs.projectcalico.org/v2.1/releases/ now.
As `Fix for calico-cni plugin for kubernetes 1.6.` is mentioned in the release note, I hope it fixes https://github.com/kubernetes-incubator/kube-aws/issues/494#issuecomment-291687137

mumoshu:
Note: It turned out that we needed `calico/cni` v1.6.2, which is [not even included in the latest calico v2.1.2](http://docs.projectcalico.org/v2.1/releases/)

mumoshu:
LGTM. Thanks for your support 👍

PR #507, opened 2017-04-06
'Cluster-dump' feature to export Kubernetes Resources to S3

jollinshead:
A new kube-system Deployment has been created which takes and uploads snapshots of all Kubernetes resources within the cluster to S3.

There are two containers:

 **cluster-dump-dumper**:
 - Uses kubectl to snapshot all Kubernetes resources (interval of 24 hours).
 - Each snapshot resides in a timestamped folder
 - The resources have several fields omitted such as _status_ , _uid_, etc ... this is to allow the possibility of restoring resources inside a fresh cluster
 - Resources that reside within namespaces are grouped inside folders labeled with the namespace name
 - Resources outside namespaces are grouped at the same directory level as the namespace folders

 **cluster-dump-pusher**
 - Uses aws to upload snapshots to S3
 - The S3 path is constructed from the _S3URI_ passed into kube-aws binary; a string "/clusterDump/"; and the _timestamped folder_

Notes:
- We decided to use bash as arguments to the containers to avoid any additional docker image dependencies - it makes use of the *aws* and *hyperkube* images already defined in cluster.yaml. Also, as the Deployment definition resides on controller nodes but the pods themselves are on worker nodes, storing the bash as separate files did not fell appropriate.

 - We intend to create a 'restore' script that will deploy resources from snapshot folder into a new cluster. Use Case: We destroy a cluster and recreate it (maybe with new features) but also restore the Kubernetes resources we had originally.

- https://coreos.com/kubernetes/docs/latest/cluster-dump-restore.html

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -150,6 +150,9 @@ func NewDefaultCluster() *Cluster {
 		CreateRecordSet:     false,
 		RecordSetTTL:        300,
 		CustomSettings:      make(map[string]interface{}),
+		KubeResourcesAutosave: KubeResourcesAutosave{
+			Enabled: false,
+		},
 	}
 }
 
@@ -643,6 +646,7 @@ type Cluster struct {
 	HostedZoneID           string `yaml:"hostedZoneId,omitempty"`
 	ProvidedEncryptService EncryptService
 	CustomSettings         map[string]interface{} `yaml:"customSettings,omitempty"`
+	KubeResourcesAutosave  `yaml:"kubeResourcesAutosave,omitempty"`
 }
 
 type Experimental struct {
@@ -715,6 +719,11 @@ type EphemeralImageStorage struct {
 
 type Kube2IamSupport struct {
 	Enabled bool `yaml:"enabled"`
+}
+
+type KubeResourcesAutosave struct {
+	Enabled bool `yaml:"enabled"`
+	S3Path  string
 }
 
 type NodeDrainer struct {
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -497,7 +497,7 @@ write_files:
       {{ end }}
 
       # Deployments
-      for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de}.yaml; do
+      for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de{{ if .KubeResourcesAutosave.Enabled }},kube-resources-autosave{{ end }}}.yaml; do
           kubectl apply -f "${mfdir}/$manifest"
       done
 
@@ -790,6 +790,116 @@ write_files:
       sed -i -e "s#\$ETCDCERT#$etcd_cert#g" /srv/kubernetes/manifests/calico.yaml
       sed -i -e "s#\$ETCDKEY#$etcd_key#g" /srv/kubernetes/manifests/calico.yaml
 
+{{ end }}
+{{ if .KubeResourcesAutosave.Enabled }}
+  - path: /srv/kubernetes/manifests/kube-resources-autosave.yaml
+    content: |
+      ---
+      apiVersion: extensions/v1beta1
+      kind: Deployment
+      metadata:
+        name: kube-resources-autosave
+        namespace: kube-system
+        labels:
+          k8s-app: kube-resources-autosave-policy
+      spec:
+        replicas: 1
+        template:
+          metadata:
+            name: kube-resources-autosave
+            namespace: kube-system
+            labels:
+              k8s-app: kube-resources-autosave-policy
+          spec:
+            containers:
+            - name: kube-resources-autosave-dumper
+              image: {{.HyperkubeImage.RepoWithTag}}
+              command: ["/bin/bash", "-c" ]
+              args:
+                - |
+                    set -x ;
+                    DUMP_DIR_COMPLETE=/kube-resources-autosave/complete ;
+                    mkdir -p ${DUMP_DIR_COMPLETE} ;
+                    while true; do
+                      TIMESTAMP=$(date +%Y-%m-%d_%H-%M-%S)
+                      DUMP_DIR=/kube-resources-autosave/tmp/${TIMESTAMP} ;
+                      mkdir -p ${DUMP_DIR} ;
+                      RESOURCES_OUT_NAMESPACE=( namespaces persistentvolumes nodes storageclasses clusterrolebindings clusterroles ) ;
+                      for r in ${RESOURCES_OUT_NAMESPACE[@]};do
+                        echo " Searching for resources: ${r}" ;
+                        /kubectl get --export -o=json ${r} | \
+                        jq '.items[] |
+                            del(.status,
+                            .metadata.uid,
+                            .metadata.selfLink,
+                            .metadata.resourceVersion,
+                            .metadata.creationTimestamp,
+                            .metadata.generation,
+                            .spec.claimRef
+                          )' > ${DUMP_DIR}/${r}.json ;
+                      done ;
+                      RESOURCES_IN_NAMESPACE=( componentstatuses configmaps daemonsets deployments endpoints events horizontalpodautoscalers
+                      ingresses jobs limitranges networkpolicies  persistentvolumeclaims pods podsecuritypolicies podtemplates replicasets
+                      replicationcontrollers resourcequotas secrets serviceaccounts services statefulsets thirdpartyresources
+                      poddisruptionbudgets roles rolebindings) ;
+                      for ns in $(jq -r '.metadata.name' < ${DUMP_DIR}/namespaces.json);do
+                        echo "Searching in namespace: ${ns}" ;
+                        mkdir -p ${DUMP_DIR}/${ns} ;
+                        for r in ${RESOURCES_IN_NAMESPACE[@]};do
+                          echo " Searching for resources: ${r}" ;
+                          /kubectl --namespace=${ns} get --export -o=json ${r} | \
+                          jq '.items[] |
+                            select(.type!="kubernetes.io/service-account-token") |
+                            del(
+                              .spec.clusterIP,
+                              .metadata.uid,
+                              .metadata.selfLink,
+                              .metadata.resourceVersion,
+                              .metadata.creationTimestamp,
+                              .metadata.generation,
+                              .metadata.annotations."pv.kubernetes.io/bind-completed",
+                              .status
+                            )' > ${DUMP_DIR}/${ns}/${r}.json && touch /probe-token ;
+                        done ;
+                      done ;
+                    mv ${DUMP_DIR} ${DUMP_DIR_COMPLETE}/${TIMESTAMP} ;
+                    rm -r -f ${DUMP_DIR} ;
+                    sleep 24h ;
+                    done
+              livenessProbe:
+                exec:
+                  command: ["/bin/bash", "-c", "(( $(date +%s) - $(stat -c%Y /probe-token) < 25*60*60 ))" ]
+                initialDelaySeconds: 240
+                periodSeconds: 10
+              volumeMounts:
+              - name: dump-dir
+                mountPath: /kube-resources-autosave
+                readOnly: false
+            - name: kube-resources-autosave-pusher
+              image: {{.AWSCliImage.RepoWithTag}}
+              command: ["/bin/bash", "-c" ]
+              args:
+                - |
+                    set -x ;
+                    DUMP_DIR_COMPLETE=/kube-resources-autosave/complete ;
+                    while true; do
+                      for FILE in ${DUMP_DIR_COMPLETE}/* ; do
+                        aws s3 mv ${FILE} s3://{{ .KubeResourcesAutosave.S3Path }}/$(basename ${FILE}) --recursive && rm -r -f ${FILE} && touch /probe-token ;
+                      done ;
+                      sleep 1m ;
+                    done
+              livenessProbe:
+                exec:
+                  command: ["/bin/bash", "-c", "(( $(date +%s) - $(stat -c%Y /probe-token) < 25*60*60 ))" ]
+                initialDelaySeconds: 240
+                periodSeconds: 10
+              volumeMounts:
+              - name: dump-dir
+                mountPath: /kube-resources-autosave
+                readOnly: false
+            volumes:
+            - name: dump-dir
+              emptyDir: {}
 {{ end }}
 
 {{if .AssetsEncryptionEnabled }}
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -933,6 +933,11 @@ worker:
 #  enabled: true
 #   maxBatchSize: 1
 
+# Exports all Kubernetes resources (in .json format) to a bucket 's3://S3URI/clusterBackup/*'.
+# The export process executes on start-up and repeats every 24 hours.
+#kubeResourcesAutosave:
+#  enabled: false
+
 # Addon features
 addons:
   # When enabled, Kubernetes rescheduler is deployed to the cluster controller(s)
--- a/core/nodepool/config/config.go
+++ b/core/nodepool/config/config.go
@@ -50,7 +50,8 @@ type DeploymentSettings struct {
 }
 
 type MainClusterSettings struct {
-	EtcdNodes []derived.EtcdNode
+	EtcdNodes             []derived.EtcdNode
+	KubeResourcesAutosave cfg.KubeResourcesAutosave
 }
 
 type StackTemplateOptions struct {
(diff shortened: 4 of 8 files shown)


c-knowles:
Oh, interesting. We're running roughly the same sort of thing but external to the cluster. It's in a pipeline to do blue-green cluster switchovers. It seems like everyone has their own solution to this problem. One relevant issue is at https://github.com/kubernetes/kubernetes/issues/24229.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,119 @@ write_files:
...
+                    DUMP_DIR_COMPLETE=/cluster-dump/complete ;
+                    while true; do
+                      for FILE in ${DUMP_DIR_COMPLETE}/* ; do
+                        aws s3 mv ${FILE} s3://{{ .KubeResourcesS3Path }}/$(basename ${FILE}) --recursive && rm -r -f ${FILE} && touch /probe-token ;
+                      done ;
+                      sleep 1m ;
+                    done
+              livenessProbe:
+                exec:
+                  command: ["/bin/bash", "-c", "(( $(date +%s) - $(stat -c%Y /probe-token) < 25*60*60 ))" ]
```
mumoshu: Can we utilize k8s' cronjobs instead?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,119 @@ write_files:
...
+                              .metadata.selfLink,
+                              .metadata.resourceVersion,
+                              .metadata.creationTimestamp,
+                              .metadata.generation,
+                              .metadata.annotations."pv.kubernetes.io/bind-completed",
+                              .status,
+                              .spec.template.spec.securityContext,
+                              .spec.template.spec.dnsPolicy,
+                              .spec.template.spec.terminationGracePeriodSeconds,
+                              .spec.template.spec.restartPolicy
```
mumoshu: May I ask why you prefer to delete these keys under the pod template?
I guess these keys contains the values originally defined in manifests and therefore
losing these keys from backups results in incomplete backups❓

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,119 @@ write_files:
...
+                            .metadata.selfLink,
+                            .metadata.resourceVersion,
+                            .metadata.creationTimestamp,
+                            .metadata.generation,
+                            .spec.claimRef
+                          )' > ${DUMP_DIR}/${r}.json ;
+                      done ;
+                      RESOURCES_IN_NAMESPACE=( componentstatuses configmaps daemonsets deployments endpoints events horizontalpodautoscalers
+                      ingresses jobs limitranges networkpolicies  persistentvolumeclaims pods podsecuritypolicies podtemplates replicasets
+                      replicationcontrollers resourcequotas secrets serviceaccounts services statefulsets thirdpartyresources ) ;
```
mumoshu: Did you miss `clusterrolebindings`, `clusterroles`, `namespaces`, `poddisruptionbudgets`, `rolebindings`, `roles`, `storageclasses`?
How does it work for `persistentvolumes`? Are `persistentvolumes` able to be backed up and restored?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,119 @@ write_files:
...
+              args:
+                - |
+                    set -x ;
+                    DUMP_DIR_COMPLETE=/cluster-dump/complete ;
+                    mkdir -p ${DUMP_DIR_COMPLETE} ;
+                    while true; do
+                      TIMESTAMP=$(date +%Y-%m-%d_%H-%M-%S)
+                      DUMP_DIR=/cluster-dump/tmp/${TIMESTAMP} ;
+                      mkdir -p ${DUMP_DIR} ;
+                      RESOURCES_OUT_NAMESPACE=( namespaces persistentvolumes nodes storageclasses ) ;
```
mumoshu: How does it work for `persistentvolumes`? Can you retain the data persisted in a persistent volume after restored from a backup?

core/controlplane/config/templates/cloud-config-controller
mumoshu: AFAIK `endpoints` and `nodes` are different among clusters and therefore shouldn't be backed up?

core/controlplane/config/templates/cloud-config-controller
mumoshu: Just realized that perhaps your work is based on https://coreos.com/kubernetes/docs/latest/cluster-dump-restore.html(?)

mumoshu:
I'm sorry to go out of topic but,
@c-knowles Btw, how do you treat background-job workers in your blue-green cluster switch-over process?

If your background-job workers are tied to a shared work queue backed by e.g. kafka or sqs, running desired num of background-job workers on BOTH a blue and a green cluster would temporarily increase loads by read requests from 2x background-job workers.
Do you have enough extra capacity on your work queue, or you don't have such services, or else?

mumoshu:
So I have never assumed a simple dump&restore would work when you're running long-running jobs involving work queues.

Perhaps, you're willing to use this feature to dump&restore k8s clusters for running periodic(e.g. hourly, daily, weekly, etc) batch jobs so that you can cut your infrastructure cost?

core/controlplane/config/templates/cloud-config-controller
jollinshead: Correct - keeping certain fields caused issues when it came to restoring.
I'll also include that link in the original post.

core/controlplane/config/templates/cloud-config-controller
jollinshead: True. We can remove them from the list since they are not necessarily needed for a restore

core/controlplane/config/templates/cloud-config-controller
jollinshead: This was our exact use case - to restore a cluster's state which made use of stateful resources. Our tests so far have been successful regarding restoring persistentvolumes/statefulsets. Although the ordering of restoring resources is important.

core/controlplane/config/templates/cloud-config-controller
jollinshead: Same goes for including any missing resources.

core/controlplane/config/templates/cloud-config-controller
jollinshead: That could be a possibility. 
We went down the Deployment route as it seemed appropriate to make use of Kubernetes where possible.

core/controlplane/config/templates/cloud-config-controller
jollinshead: We assume the original volume is not deleted from AWS though.

core/controlplane/config/templates/cloud-config-controller
mumoshu: Thanks. I'm interested! Would you mind sharing me why it seemed appropriate?

core/controlplane/config/templates/cloud-config-controller
mumoshu: Looking forward to it - I'm not yet sure why a field like `.spec.template.spec.restartPolicy` causes issues in a restore.

mumoshu:
FYI @pieterlange Does this relate to [kube-backup](https://github.com/pieterlange/kube-backup)?
Would you guys like to work together on this, if your use-cases are similar enough?

core/controlplane/config/templates/cloud-config-controller
mumoshu: Good to know. Thanks :bow:

mumoshu: Would you mind documenting somewhere about a possibly supported/expected use-case for this feature?

mumoshu:
@jollinshead Finished reviewing. Perhaps are you willing to add some comments/documentations  according to my review?
If so, please feel free to notify me when you're done and want this to be reviewed again!

c-knowles:
@mumoshu You are correct that switchover can increase load on some backend systems such as databases and queues. We're not fully up to speed on all that yet since the systems we use scale well so it's not too big of an issue. The main parts we are actively working on are around "draining" background workers on the older cluster. In cases where the jobs do not need to be complex right away, we can drain the old cluster before we bring up the new cluster. In cases where the jobs need to complete as soon as possible, we generally leave the workers fully running in the old cluster and only turn them off once the new cluster is fully operational. Important to this is that we do not change worker software versions at the same time as cluster switchover (or any software versions actually).

pieterlange:
@mumoshu i'd say these efforts are highly related indeed. :+1:

Some general nits:
- This is using a deployment that just sleeps for 24h, so you have no real control over when the backups actually happens (as you do with a `CronJob`, but those have other problems.)
- Lots of hardcoding, it's not very configurable which resource types to configure
- Exports & stores secrets in plaintext
- Strips a lot of metadata & podspec. The metadata is OK to strip as it's mostly cluster generated, but stripping the podspec will affect the application behavior.

In general, if this is how @jollinshead likes to run his backups that's fine but i don't see why this is specific to kube-aws or why it should be included as a default cluster backup option.

core/controlplane/config/templates/cloud-config-controller
redbaron: deployment + liveness probe allows "more natural" monitoring of backup status. logic is simple - if pod is up and  "live" -> backed were done. moment it wen't "not healthy" -> backups are outdate for whatever reason.  CronJob allows periodic execution, but IMHO makes it a bit harder to reason about status of a backup for not much gain

core/controlplane/config/templates/cloud-config-controller
redbaron: agree, `spec.template` fields doesn't seem to be autopopulated like lets say `status`, can they be kept or maybe deleted only for certain type of resource.

core/controlplane/config/templates/cloud-config-controller
jollinshead: We felt containing the process within Kubernetes adhered to Kubernetes' concept: Managing containerized applications across a cluster of nodes homogeneously.

Containing the process outside Kubernetes could introduce additional work (monitoring/maintaining) and would introduce a disconnect that defeats the purpose of Kubernetes.

Also, triggering a manual snapshot is simpler inside Kuberenetes: ```kubectl delete pod cluster-dump```.

redbaron:
> This is using a deployment that just sleeps for 24h, so you have no real control over when the backups actually happens (as you do with a CronJob, but those have other problems.)
just delete the pod :) backup happens on first start

> Lots of hardcoding, it's not very configurable which resource types to configure
IMHO it is better approach to backup ALL and then have flexibility on a restore side. Thats where you can configure individual namespaces/resources to be restored

> i don't see why this is specific to kube-aws or why it should be included as a default cluster backup option.
it makes use of S3 bucket +paths already "dedicated" to cluster so it is just natural to add it to kube-aws. Besides with recent changes to `kube-aws` clusters often need to be recreated and making a tool to support "recreate cluster, but bring it back as it used to be" makes sense

core/controlplane/config/templates/cloud-config-controller
jollinshead: Agreed - I'll keep spec.template fields. 

The restore script can always be modified to include/exclude field anyway.

pieterlange:
Can we at least make sure to use KMS to encrypt the secrets in this case?

Looks OK otherwise.

jollinshead:
@mumoshu Done

core/controlplane/config/config.go
```diff
@@ -643,6 +644,8 @@ type Cluster struct {
 	HostedZoneID           string `yaml:"hostedZoneId,omitempty"`
 	ProvidedEncryptService EncryptService
 	CustomSettings         map[string]interface{} `yaml:"customSettings,omitempty"`
+	ExportKubeResources    bool                   `yaml:"exportKubeResources,omitempty"`
+	KubeResourcesS3Path    string                 `yaml:"-"`
```
mumoshu: Shouldn't the annotation be `yaml:"kubeResourcesS3Path,omitempty"` or is there an expression like "-" which I'm not familiar with?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -497,7 +497,7 @@ write_files:
       {{ end }}
 
       # Deployments
-      for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de}.yaml; do
+      for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de{{ if .ExportKubeResources }},cluster-dump{{ end }}}.yaml; do
```
mumoshu: After thinking about this feature for several days, I'd suggest it to name `KubeResourcesAutosaveEnabled` rather than `ExportKubeResources` so that it can tell users that a save(export) happens automatically.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,116 @@ write_files:
       sed -i -e "s#\$ETCDKEY#$etcd_key#g" /srv/kubernetes/manifests/calico.yaml
 
 {{ end }}
+{{ if .ExportKubeResources }}
+  - path: /srv/kubernetes/manifests/cluster-dump.yaml
```
mumoshu: If we go [that way](https://github.com/kubernetes-incubator/kube-aws/pull/507/commits/20a1f392a5925ccdf728e7373000b508d0c7b364#r111071194), I'd expect this to be named like `kube-resources-autosave.yaml` so that it will be consistent across configuration key names, func names, and manifest file names.

mumoshu: And I'd also expect the `cluster-dump`s to be changed to `kube-resources-autosave` accordingly.

core/nodepool/config/config.go
```diff
@@ -115,6 +117,7 @@ func (c ProvidedConfig) StackConfig(opts StackTemplateOptions) (*StackConfig, er
 
 	baseS3URI := strings.TrimSuffix(opts.S3URI, "/")
 	stackConfig.S3URI = fmt.Sprintf("%s/kube-aws/clusters/%s/exported/stacks", baseS3URI, c.ClusterName)
+	stackConfig.KubeResourcesS3Path = fmt.Sprintf("%sclusterBackup", strings.TrimPrefix(opts.S3URI, "s3://"))
```
mumoshu: Are you willing to restore cluster(s) named differently from the original one or not?
If it was latter, it would be better to make KubeResourcesS3Path to be scoped like:
```
fmt.Sprintf("%s/kube-aws/clusters/%s/backup", strings.TrimPrefix(baseS3URI "s3://"), c.ClusterName)
```

core/controlplane/config/templates/cluster.yaml
```diff
@@ -933,6 +933,10 @@ worker:
 #  enabled: true
 #   maxBatchSize: 1
 
+# Exports all Kubernetes resources (in .json format) to a bucket 's3://S3URI/clusterBackup/*'.
+# The export process executes on start-up and repeats every 24 hours.
+#exportKubeResources: false
```
mumoshu: I'd prefer:
```yaml
kubeResourcesAutosave:
  enabled: false
```
per https://github.com/kubernetes-incubator/kube-aws/pull/507/commits/20a1f392a5925ccdf728e7373000b508d0c7b364#r111071194

mumoshu:
@jollinshead Sorry for the delay in reviewing!
I've read through your work. Though it looks good overall, I've added some nits around the naming - I believe the naming helps us maintain this feature in longer term.
Your thoughts?

jollinshead:
Your proposed changes look good to me - I'll notify you once they're implemented.

jollinshead:
@mumoshu Done 👍

mumoshu:
@pieterlange I believe it will be good to do! As I think it would affect how the "restore" side of this feature works, maybe we can revisit it once the restore side is implemented?

mumoshu:
@jollinshead Thanks for the quick turnaround. LGTM 👍

core/controlplane/config/templates/cloud-config-controller
c-knowles: I thought above @mumoshu was referring to [Cron Jobs](https://kubernetes.io/docs/user-guide/cron-jobs/) built into k8s? That is managed inside Kubernetes and failed jobs are kept (plus it's possible to set limits on how many failed jobs we keep etc).

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -791,6 +791,116 @@ write_files:
...
+                      ingresses jobs limitranges networkpolicies  persistentvolumeclaims pods podsecuritypolicies podtemplates replicasets
+                      replicationcontrollers resourcequotas secrets serviceaccounts services statefulsets thirdpartyresources
+                      poddisruptionbudgets roles rolebindings) ;
+                      for ns in $(jq -r '.metadata.name' < ${DUMP_DIR}/namespaces.json);do
+                        echo "Searching in namespace: ${ns}" ;
+                        mkdir -p ${DUMP_DIR}/${ns} ;
+                        for r in ${RESOURCES_IN_NAMESPACE[@]};do
+                          echo " Searching for resources: ${r}" ;
+                          /kubectl --namespace=${ns} get --export -o=json ${r} | \
+                          jq '.items[] |
```
c-knowles: Not quite using this yet but took a bit of inspiration to give our similar scripts a quick dust. I think the `json` file output from this command is separate JSON objects written one after another to the same file? It works nicely for YAML but harder to validate as JSON. Not sure if it matters for your restore script but if you wanted these as valid JSON files using Kubernetes kind of `List`, I believe you can use the form `jq '.items[] |= (select() | del())'`.

jollinshead: Correct. Exporting as Kubernetes' 'Lists' would be a nice enhancement.

c-knowles: Sorry I tested this a bit more and the `select` returns nil if no match so this should be more like this:
```
jq '.items |= ([ .[] | select() | del() ])'
```

jollinshead: Nice! I'll make a pull request

PR #510, opened 2017-04-07
New options: customFiles and customSystemdUnits

jeremyd:
This adds the capability to write user defined files and systemd units to the 3 main types of servers: Etcd, Controller, NodePool Worker as specified by the cluster.yaml configuration for the cluster.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -9,6 +9,20 @@ coreos:
     etcd_keyfile: /etc/kubernetes/ssl/etcd-client-key.pem
 
   units:
+{{- range $u := .Controller.CustomSystemdUnits}}
+    - name: {{$u.Name}}
+      command: {{ $u.Command }}
+      {{- if $u.Enable }}
+      enable: {{ $u.Enable }}
+      {{- end }}
+      {{- if $u.Runtime }}
+      runtime: {{ $u.Runtime }}
+      {{- end }}
+      content: |
+        {{- range $l := $u.ContentArray}}
+        {{ $l }}
+        {{- end }}
+{{- end}}
     - name: cfn-etcd-environment.service
       enable: true
       command: start
@@ -394,6 +408,16 @@ ssh_authorized_keys:
         WantedBy=install-kube-system.service
 {{end}}
 write_files:
+{{- if .Controller.CustomFiles}}
+  {{- range $w := .Controller.CustomFiles}}
+  - path: {{$w.Path}}
+    permissions: {{$w.PermissionsString}}
+    content: |
+      {{- range $l := $w.ContentArray}}
+      {{$l}}
+      {{- end}}
+  {{- end }}
+{{- end }}
 {{if .Experimental.DisableSecurityGroupIngress}}
   - path: /etc/kubernetes/additional-configs/cloud.config
     owner: root:root
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -412,10 +412,8 @@ write_files:
   {{- range $w := .Controller.CustomFiles}}
   - path: {{$w.Path}}
     permissions: {{$w.PermissionsString}}
-    content: |
-      {{- range $l := $w.ContentArray}}
-      {{$l}}
-      {{- end}}
+    encoding: gzip+base64
+    content: {{$w.GzippedBase64Content}}
   {{- end }}
 {{- end }}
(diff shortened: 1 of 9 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -9,6 +9,20 @@ coreos:
     etcd_keyfile: /etc/kubernetes/ssl/etcd-client-key.pem
 
   units:
+{{- range $thisunit := .Controller.CustomUnits}}
```
mumoshu: nit: I prefer just `$u` rather than `$thisunit` as per the golang convention of using a single character for a loop variable.

mumoshu: I prefer `CustomSystemdUnits` rather than `CustomUnits` - Just saying it "unit" makes things a bit ambiguous between this and ones in spot fleets.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -394,6 +408,16 @@ ssh_authorized_keys:
         WantedBy=install-kube-system.service
 {{end}}
 write_files:
+{{- if .Controller.CustomWriteFiles}}
```
mumoshu: I prefer `customFiles`/`CustomFiles` rather than `CustomWriteFiles`. I believe it is named after `write_files` in cloud-config but I guess that's an implementation detail. Although it isn't followed everywhere but I'd like to hide implementation details whenever possible.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -163,6 +163,24 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #    # If you like specific private subnets to be used by the internal ELB:
 #    # subnets:
 #    #   - name: ManagedPrivateSubnet0
+#  
+#  # User defined files that will be added to the Controller cluster cloud-init configuration in the "write_files:" section.
```
mumoshu: Good documentation 👍

model/custom_unit.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (c CustomUnit) RuntimeString() string {
+	return strconv.FormatBool(c.Runtime)
+}
+
+func (c CustomUnit) SillyName() string {
+	return c.Name
+}
+
+func (c CustomUnit) SillyCommand() string {
```
mumoshu: Maybe an unused func?

model/custom_unit.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (c CustomUnit) EnableString() string {
+	return strconv.FormatBool(c.Enable)
+}
+
+func (c CustomUnit) RuntimeString() string {
+	return strconv.FormatBool(c.Runtime)
+}
+
+func (c CustomUnit) SillyName() string {
```
mumoshu: Maybe an unused func?

core/controlplane/config/templates/cloud-config-controller
jeremyd: agree, fixed changed the vars to golang convention in all the places you mentioned.

jeremyd: renamed to customSystemdUnits

core/controlplane/config/templates/cloud-config-controller
jeremyd: Renamed to customFiles

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -476,6 +490,16 @@ ssh_authorized_keys:
...
 {{end}}
 write_files:
+{{- if .CustomFiles}}
+  {{- range $w := .CustomFiles}}
+  - path: {{$w.Path}}
+    permissions: {{$w.PermissionsString}}
+    content: |
+      {{- range $l := $w.ContentArray}}
+      {{$l}}
+      {{- end}}
```
mumoshu: @jeremyd  Sorry to add a review comment like this after the delay but
How about making this:

```yaml
encoding: gzip+base64
content: {{$w.GzippedBase64Content}}
```

like how other relatively large files(`.pem` and `.pem.enc` for example) are currently written via write_files, so that we won't bloat the userdata size unnecessarily?
FYI, if we go that way, the `gzipcompressor.CompressString()` func would be useful.

jeremyd: Sure no prob, added the gzip/base64 encoding to these.

mumoshu:
@jeremyd Thanks for the quick turnaround! LGTM 👍

Issue #512, opened 2017-04-08
Expected way to access API server from pods on HA controllers


c-knowles:
### Summary

I finally managed to work on my comment from https://github.com/kubernetes-incubator/kube-aws/issues/409#issuecomment-286950126 to figure out what is going on. I tried quite a few combinations of connectivity from Daemon Set pods to API server in a HA controller environment and there seems to be some problems. I'm not entirely sure of the root cause. I have not tested with a single controller.

My current tests show:
1. `dnsPolicy: ClusterFirst` + `KUBERNETES_SERVICE_HOST` env variable has issues on only some controllers
1. `dnsPolicy: Default` + `KUBERNETES_SERVICE_HOST` env variable has issues on all controllers (expected I think)
1. `dnsPolicy: ClusterFirst` + `externalDNSName` from `cluster.yaml` seems ok
1. `dnsPolicy: ClusterFirst` + `kubernetes.default.svc.cluster.local` has issues on only some controllers even in containers that will resolve that domain correctly (I'm aware that some images don't play nicely with kube DNS)

So 3 seems to work ok but I'm still not sure 1/4 are the expected behaviour. The slightly confusing bit for me is that when I say "some" above it is because when I was initially diagnosing this problem I went about terminating controller nodes via AWS to see if brand new nodes had the same problems. In the case of 4, sometimes both controllers would work fine and sometimes just one would work but usually neither! With 1, the situation seems to be similar but most of the time neither of the controllers work.

It could be me misconfiguring, but the intermittent behaviour makes me think something else is wrong. The environment variable `KUBERNETES_SERVICE_HOST` is used by some apps/addons to access the API server via service accounts so I think it should work ok. `kubernetes.default.svc.cluster.local` should definitely certainly work with `ClusterFirst` DNS.

### Steps to reproduce
1. Spin up a cluster with HA controller, I used 2 controllers
2. Deploy the Daemon Set below, it just periodically tries to connect to the API server. There are a few combinations in there which can be tried
3. Follow the logs on the containers running on the controllers, you should see it times out while trying to connect:
```
✗ stern --since 1m --namespace=kube-system service-account-test
+ service-account-test-9q0wk › test
service-account-test-9q0wk test connect: Operation timed out
service-account-test-9q0wk test connect:errno=110
service-account-test-9q0wk test verify depth is 1
```
4. If you don't see timeouts, try terminating a controller node and check the behaviour on the new node that replaces it.

### Details
[...]

PR #514, opened 2017-04-10
Fix API endpoint from HA controllers

c-knowles:
Pod environment variable `KUBERNETES_SERVICE_HOST` and DNS entry
`kubernetes.default.svc.cluster.local` should now function correctly
and be queryable from controllers.

For https://github.com/kubernetes-incubator/kube-aws/issues/512


mumoshu:
Thanks a lot :tada: LGTM 👍

Issue #520, opened 2017-04-10
Configuration/docs for apiEndpoints


c-knowles:
I just got around to using the multiple API endpoints from https://github.com/kubernetes-incubator/kube-aws/pull/468. I have some issues configuring it to achieve the addition of a DNS name to the cert without another ELB.

The relevant details of my cluster.yaml are:

```
#externalDNSName: k8s-dev-v1.REDACTED
#createRecordSet: true
#recordSetTTL: 60
#hostedZone: ""
#hostedZoneId: REDACTED

apiEndpoints:
  - name: versioned
    createRecordSet: true
    dnsName: k8s-dev-v1.REDACTED
    hostedZone:
      id: REDACTED
    recordSetTTL: 60
  - name: unversioned
    dnsName: k8s-dev.REDACTED

worker:
  apiEndpointName: versioned
  nodePools:
    - name: worker-pool-a
    - name: worker-pool-b
    - name: worker-pool-c
```

The output of `kube-aws validate` is:

```
WARN: the worker node pool "worker-pool-a" is associated to a k8s API endpoint behind the DNS name "k8s-dev-v1.REDACTED" managed by YOU!
Please never point the DNS record for it to a different k8s cluster, especially when the name is a "stable" one which is shared among multiple k8s clusters for achieving blue-green deployments of k8s clusters!
kube-aws can't save users from mistakes like that
WARN: the worker node pool "worker-pool-b" is associated to a k8s API endpoint behind the DNS name "k8s-dev-v1.REDACTED" managed by YOU!
Please never point the DNS record for it to a different k8s cluster, especially when the name is a "stable" one which is shared among multiple k8s clusters for achieving blue-green deployments of k8s clusters!
kube-aws can't save users from mistakes like that
WARN: the worker node pool "worker-pool-c" is associated to a k8s API endpoint behind the DNS name "k8s-dev-v1.REDACTED" managed by YOU!
Please never point the DNS record for it to a different k8s cluster, especially when the name is a "stable" one which is shared among multiple k8s clusters for achieving blue-green deployments of k8s clusters!
kube-aws can't save users from mistakes like that
Error: Failed to initialize cluster driver: file cluster.yaml: unknown keys found in worker.nodePools[0].apiEndpoints[0]: createRecordSet, hostedZone, recordSetTTL
```

Side note - it seems the default cluster.yaml does not have all the details yet. For example `worker.apiEndpointName` which I found out about by accident:

```
[...]

PR #526, opened 2017-04-11
Fix up API endpoints config

c-knowles:
For https://github.com/kubernetes-incubator/kube-aws/issues/520. `loadBalancer` was missing entirely.

Not sure where to put `worker.apiEndpointName`, does it default to anything? I'd like to resolve this documentation issue in the same PR as `worker.apiEndpointName` is not listed in the default cluster.yaml:

```
Error: Failed to initialize cluster driver: file cluster.yaml: worker.apiEndpointName must not be empty when there're 2 or more API endpoints under the key `apiEndpoints` and one of worker.nodePools[] are missing apiEndpointName
```

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -41,32 +41,36 @@ externalDNSName: {{.ExternalDNSName}}
 #  # to be referenced from other parts of cluster.yaml
 #- name: template
 #
-#  # Specifies an existing load-balancer used for load-balancing controller nodes and serving this endpoint
-#  # Setting id requires all the other settings excluding `name` to be omitted because reusing an ELB implies that configuring other resources
-#  # like a Route 53 record set for the endpoint is now your responsibility!
-#  id: existing-elb
+#  # DNS name for this endpoint, added to the kube-apiserver TLS cert
+#  dnsName: dns-name.tld
 #
-#  # Set to false when you want to disable creation of the record set for this api load balancer
-#  # Must be omitted when `id` is specified
-#  createRecordSet: true
+#  loadBalancer:
+#    # Specifies an existing load-balancer used for load-balancing controller nodes and serving this endpoint
+#    # Setting id requires all the other settings excluding `name` to be omitted because reusing an ELB implies that configuring other resources
+#    # like a Route 53 record set for the endpoint is now your responsibility!
+#    id: existing-elb
 #
-#  # All the subnets assigned to this load-balancer. Specified only when this load balancer is not reused but managed one
-#  # Must be omitted when `id` is specified
-#  subnets:
-#  - name: managedPublic1
+#    # Set to false when you want to disable creation of the record set for this api load balancer
+#    # Must be omitted when `id` is specified
+#    createRecordSet: true
 #
-#  # Set to true so that the managed ELB becomes an `internal` one rather than `internet-facing` one
-#  # When set to true while subnets are omitted, one or more private subnets in the top-level `subnets` must exist
-#  # Must be omitted when `id` is specified
-#  private: true
+#    # All the subnets assigned to this load-balancer. Specified only when this load balancer is not reused but managed one
+#    # Must be omitted when `id` is specified
+#    subnets:
+#    - name: managedPublic1
 #
-#  # TTL in seconds for the Route53 RecordSet created if createRecordSet is set to true.
-#  recordSetTTL: 300
+#    # Set to true so that the managed ELB becomes an `internal` one rather than `internet-facing` one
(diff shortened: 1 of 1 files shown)


mumoshu:
@c-knowles Thanks for the fix!
Regarding `worker.apiEndpointName`, it defaults to the only API endpoint if omitted. If it was omitted when `apiEndpoints` has 2 or more items, a validation error occurs.

c-knowles:
@mumoshu ok I've added those details.

I was trying to come up with idea to better document the configuration, I feel the current method is becoming unwieldy and difficult to read. My last commit is one idea, feel free to remove it before merge or I can. The idea is we would remove a lot of the comments from `cluster.yaml` and place them in structured markdown instead for an easier to digest form. `cluster.yaml` could still contain a skeleton plus some defaults. Another idea is I could setup some better documentation using a github static site generator or whatever they use for the new kube API reference.

c-knowles:
p.s. if you like any of the ideas, I can pick up a task to migrate everything across including any PRs in flight.

mumoshu:
@c-knowles Thanks for the suggestion!

I like the idea of writing a documentation dedicated for explaining cluster.yaml but I'd basically like to keep up to, for example, 3 lines per the comment for a setting key in cluster.yaml.

However, I do believe migrating the following long comments to dedicated documentations would help:

* https://github.com/kubernetes-incubator/kube-aws/blob/aeccbb266/core/controlplane/config/templates/cluster.yaml#L609-L622
   * Probably into a doc named like "Bringing your own route table"?
* https://github.com/kubernetes-incubator/kube-aws/blob/aeccbb266/core/controlplane/config/templates/cluster.yaml#L652-L783
   * Probably into a doc named like "Customizing a network topology"?

c-knowles:
@mumoshu ok sure, can I ask the reasons why you wish to keep cluster.yaml verbose? I'm just curious about overall plans to improve this area. Ideally I'm still thinking a full schema would be good.

I will take a task about separating those ones you mention. My initial thoughts are adding both of them as a section in a simple github pages website (plus the bit I've already done). I'm not sure about the publishing to the coreos website, not even sure why that link is still maintained?

mumoshu:
@c-knowles Thanks for your efforts!

> ok sure, can I ask the reasons why you wish to keep cluster.yaml verbose?

I'd just like to keep cluster.yaml self-contained enough for basic setups.
Relatively long comments might be for more advanced setups and therefore I thought they should go to dedicated docs.

> My initial thoughts are adding both of them as a section in a simple github pages website (plus the bit I've already done).

It sounds good but the only concern I have is that, in longer term, aggregating too many topics into a single doc would make it harder to be read?
I just wanted separate docs dedicated to respective topics so that we don't bloat each doc size like we (unfortunately) do for cluster.yaml today.

Your thoughts?

c-knowles:
@mumoshu ok makes sense, at first we can just separate the particularly verbose parts then and see where we get to. A github site using simple Jekyll theme would allow separate markdown files for the different sections/topics so I think it fits with your idea of having separate docs just they are on the same site. If you are happy for me to proceed with that I'll create an issue now and start on it soon.

mumoshu:
@c-knowles I like the idea of having a github pages site for the doc but I'm not really sure if I have an enough permission to enable it on this repo

image

So, would it work if we add the `doc/` dir to the master branch?
I'd been using the `gh-pages` branch for a long time in my other projects which worked great, but there's no option in the web ui to choose to use `gh-pages` branch.

mumoshu:
@c-knowles Would you mind starting from just adding new docs to `Documentation/` (and links to them from README.md) which already exists as of today, and then migrate to a gh pages site (if possible) or an another website (if any) in the future?

c-knowles:
@mumoshu the `docs` folder [has to exist](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch) for it to be selectable.

c-knowles:
@mumoshu I've extracted your two current top items from https://github.com/kubernetes-incubator/kube-aws/pull/526#issuecomment-293773312 into separate tasks. For this PR do you want me to remove the idea markdown for now or rename it?

mumoshu:
@c-knowles Assuming we're going to add a completely different docs for the separate tasks, I guess we'd better remove the doc added in this PR for now.
I'm all for ways you'd feel easier though!

c-knowles:
@mumoshu I agree, let's assume that for now - all ready here. I've kept a copy of the idea locally for when we do the new docs.

mumoshu:
@c-knowles LGTM. Thanks for your continuous efforts 👍

PR #528, opened 2017-04-11
Minor fixup for etcd unit files

redbaron:
nothing major, just small nits found while moving to new master

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -83,10 +83,9 @@ coreos:
         Description=etcd health check
 
         [Service]
-        Type=simple
+        Type=oneshot
         EnvironmentFile=-/etc/etcd-environment
         EnvironmentFile=-/var/run/coreos/etcdadm-environment
-        Restart=no
         ExecStartPre=/usr/bin/systemctl is-active {{.Etcd.SystemdUnitName}}
         ExecStart=/opt/bin/etcdadm check
         TimeoutStartSec=120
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -251,13 +249,12 @@ coreos:
       command: start
       content: |
         [Unit]
-        Wants={{.Etcd.SystemdUnitName}}
+        Requires={{.Etcd.SystemdUnitName}}
```
mumoshu: Wouldn't modifying this to `Requires` result in cfn-signal.service to be kept failed permanently when `etcd-member.service` failed at first but then restarted via `Restart=on-failure`(according to /usr/lib64/systemd/system/etcd-member.service` and the restart was successful?

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -115,10 +114,9 @@ coreos:
         Description=etcd snapshot
 
         [Service]
-        Type=simple
+        Type=oneshot
```
mumoshu: Thanks! This change seems to work.

core/controlplane/config/templates/cloud-config-etcd
mumoshu: @redbaron Any chance you could confirm my comment above?

redbaron: it seems that email responses don't show up here :(

It is `oneshot` service therefore it is never restarted, yet it all worked fine

mumoshu: @redbaron Argh, maybe I had forgotten about that multiple times until now!
Thanks for the confirmation 👍 
LGTM for now.

In the future, perhaps we'd better transform all the oneshot services into simple services with all the useful commands in "ExecStartPre"s, while having `ExecStart=/bin/true`?

mumoshu:
Thanks for your contribution as always, @redbaron 🎉

PR #535, opened 2017-04-13
'Restore' feature to restore Kubernetes Resources from S3 backup

jollinshead:
Restoring a Kubernetes environment (from a previous backup) may be executed through the use of a bash script

The script was designed to be used by the cluster provisioner and assumes that their local machine has:
- 'kubectl' installed locally
- 'aws' installed locally
- 'jq' installed locally
- 'kubeconfig' configured correctly
- 'jq' configured correctly
- cluster is reachable

The script will **(kubectl) create** Kubernetes resources per namespace. All namespaces (and associated resources) are restored by default with the exception of namespaces: 'kube-system' and 'default'.
To whitelist namespaces, include the desired namespace names as additional arguments after the initial BUCKET_URI argument.
The BUCKET_URI argument is the URL of the timestamped folder.

## Implementation
1) The script will initially validate that the namespaces it is to restore are not already existing.
2) Kubernetes resources common to all namespaces (such as *StorageClass*) are created.
3) For each namespace to restore, the namespace is created and Kubernetes resources within the namespace are created.

### Considerations
- Restoring *PersistentVolumes* does not provision new volumes from AWS - it assumes such volumes already exist in AWS.
- You may have to tailor the script for your specific needs.


mumoshu:
@jollinshead Thanks for the PR!
I especially like the well-written documentation 👍 

One nit for now: Would you mind moving your script under `contrib` like e.g. `contrib/cluster-backup/restore.sh`?
Also, having a seperate OWNERS file at `contrib/cluster-backup/OWNERS` like [cluster-autoscaler does](https://github.com/kubernetes/contrib/blob/master/cluster-autoscaler/OWNERS) would also help us maintain the script.
If you're ok with that, please fill your name in the approvers and reviewers section of the file.

jollinshead:
@mumoshu Done 👍

mumoshu:
@jollinshead Thanks for your efforts 👍

contrib/cluster-backup/README.md
```diff
@@ -0,0 +1,37 @@
+# Restore
+Restoring a Kubernetes environment (from a previous backup) may be executed through the use of the bash script: [kubernetes-on-aws-restore.sh](/Documentation/scripts/kubernetes-on-aws-backup.sh)
```
c-knowles: I think this link should be to `contrib/cluster-backup/restore.sh` rather than the docs backup.sh script?

jollinshead: Well spotted - I'll correct this.

PR #538, opened 2017-04-13
Bugfix: Add missing '/' when constructing the Autosave S3 put path

jollinshead:
This issue prevented the autosave to push to s3 since it was pushing to the wrong path.


mumoshu:
@jollinshead LGTM. Thanks for the fix 👍

PR #555, opened 2017-04-19
Don't mount /var/lib/rkt into kubelet

redbaron:
Due to https://github.com/rkt/rkt/issues/3465 no rkt containers can be GCed on
a node once kubelet is up and running.

Symptoms are that if you run following command, container can't be GCed:

```
rkt run  --insecure-options=image docker://busybox --exec date; rkt gc --grace-period=0;
```

So if any service periodically runs rkt containers, then node will eventually fill
up with many mounts and slow down operation.

As an extra-cleanup, also removing mounts of /tmp and rkt binary as they are
not used by kubelet

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -154,14 +154,16 @@ coreos:
         --set-env=ETCD_CERT_FILE=/etc/kubernetes/ssl/etcd-client.pem \
         --set-env=ETCD_KEY_FILE=/etc/kubernetes/ssl/etcd-client-key.pem \
         --mount volume=dns,target=/etc/resolv.conf \
+        {{ if eq .ContainerRuntime "rkt" -}}
         --volume rkt,kind=host,source=/opt/bin/host-rkt \
         --mount volume=rkt,target=/usr/bin/rkt \
         --volume var-lib-rkt,kind=host,source=/var/lib/rkt \
         --mount volume=var-lib-rkt,target=/var/lib/rkt \
-        --volume var-lib-cni,kind=host,source=/var/lib/cni \
-        --mount volume=var-lib-cni,target=/var/lib/cni \
         --volume stage,kind=host,source=/tmp \
         --mount volume=stage,target=/tmp \
+        {{ end -}}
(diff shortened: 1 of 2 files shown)


mumoshu:
@redbaron Thanks for the PR!
I wonder if this change implies that rktnetes becomes unusable due to the missing rkt binary and the rkt pod states(?) dir `/var/lib/rkt` from kubelets?

mumoshu:
If so, should/can we just transform all the periodic rkt pods to equivalent docker containers?

redbaron:
does anybody use rktnetes? do we test it at all? I see that container runtime is optional and not always set to docker.

are you OK with me changing this to be conditional on `{{ .ContainerRuntime }}` ?

redbaron:
> If so, should/can we just transform all the periodic rkt pods to equivalent docker containers instead?
rkt is already there and very convenient to use, we use it in our custom units files a  lot and have nor problems.I see no reason to convert it to docker apart from broken GC, but this thing is fixable. kubelet should have minimal number of mounts, only those which are needed for its operation anyway

mumoshu:
@redbaron 

> are you OK with me changing this to be conditional on {{ .ContainerRuntime }} ?

Yes, I guess it is the way to go for now!
Including a link to rkt/rkt#3465 and a short description of the problem in a comment of `{{if eq .ContainerRuntime "rkt"}}` would be also helpful.

We can discuss about deprecating and removing the `containerRuntime` option if it is really desirable, but in an another issue.

redbaron:
I made it conditional. I am not sure that linking rkt issue is needed, that is because even if it wasn't a problem, not doing mounts which are not required by a given runtime is still a good thing.

redbaron:
I am not sure that `{{ if ` will play niceley with surrounding line breaks, do i need to make it `{{- if` and {{-
 end` instead?

mumoshu:
@redbaron Haven't tried that case my self but it would be better to make them `{{if .... -}}` and `{{end -}` at the same indent level with other unconditioned flags, so that all the empty lines would go way

redbaron:
thanks, made  it as suggested

mumoshu:
LGTM & Manually verified that a cluster successfully starts after this change.

mumoshu:
Thanks for the fix 👍

Issue #500, opened 2017-04-05
Worker node pool takes much time to create IAM role and instance profile


mumoshu:
![image](https://cloud.githubusercontent.com/assets/22009/24688239/a9321392-19f9-11e7-9ef0-27c1ccf3f3bc.png)

Can we just create a default worker role/instance profile in a control-plane stack and then import them to worker node pool stacks so that we can cut 2 or 3 min down per an additional worker node pool?

mumoshu:
@redbaron @c-knowles @danielfm Do you have any thoughts on this?

c-knowles:
Would `IAMRoleWorker` then need to be shared also? `IAMInstanceProfileWorker` refers to it and since the worker CF refers to controlplane CF but not vice versa, I think this would be the case.

If we did this, these are the switches we have in `IAMRoleWorker` which we'd have to consider the original use case and whether those need overrides or not:
1. `.ManagedIamRoleName`, probably ok to share between pools
1. `.Kube2IamSupport.Enabled`, share between pools
1. `.AssetsEncryptionEnabled`
1. `.WaitSignal.Enabled`, probably ok to share between pools
1. `.AwsNodeLabels.Enabled`
1. `.ClusterAutoscalerSupport.Enabled`
1. `.SpotFleet.Enabled`, do not think this can be shared between pools
1. `.LoadBalancer.Enabled` / `.TargetGroup.Enabled`

mumoshu:
@c-knowles Ah, good catch! It seems to be harder than I had initially thought then.

Fsero:
would not be better to give nodepools the ability to attach a preexisting IAM role and we optionally could create one worker role in control plane to attach?

mumoshu:
@Fsero sounds cool but I'm wondering how we could implement an intuitive
configuration syntax for that.
Also, how would you like to ensure that an iam role provided by you has
enough policies for what features are enabled for the node pool?
2017年4月6日(木) 22:08 Fabián Sellés Rosa <notifications@github.com>:

Fsero:
@mumoshu i'm not sure either about how implement this in a sane / easy way but i'm eager to write a PoC for that if it's interesting. One naive approach will be create a new key under worker block named like WorkerUseExistingIamRoleArn and attach the arn.

[...]

PR #556, opened 2017-04-19
Export worker stack names and worker IAM role ARNs

jpb:
This allows node pool IAM roles to be referenced in external cloudformation
stacks that provide roles for kube2iam to use. The worker IAM role is required
to grant permissions to the worker instances in an AssumeRolePolicyDocument.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1488,6 +1488,11 @@
     "StackName": {
       "Description": "The name of this stack which is used by node pool stacks to import outputs from this stack",
       "Value": { "Ref": "AWS::StackName" }
+    },
+    "ControllerIAMRoleArn": {
+      "Description": "The ARN of the IAM role for this Node Pool",
+      "Value": { "Fn::GetAtt": ["IAMRoleController", "Arn"] },
+      "Export": { "Name": { "Fn::Sub": "${AWS::StackName}-ControllerIAMRoleArn" } }
     }
   }
 }
--- a/core/root/config/templates/stack-template.json
+++ b/core/root/config/templates/stack-template.json
@@ -45,6 +45,21 @@
     "ControlPlaneStackName": {
       "Description": "The name of the control plane stack",
       "Value": {"Fn::GetAtt" : [ "{{$.ControlPlane.Name}}" , "Outputs.StackName" ]}
-    }
+    },
+    "ControllerIAMRoleArn": {
+      "Description": "The IAM Role ARN for controller nodes",
+      "Value": {"Fn::GetAtt" : [ "{{$.ControlPlane.Name}}" , "Outputs.ControllerIAMRoleArn" ]},
+      "Export": { "Name": { "Fn::Sub": "${AWS::StackName}-ControllerIAMRoleArn" } }
+    }{{range $_, $p := .NodePools}},
+    "NodePool{{ $p.Name }}StackName": {
+      "Description": "The name of the {{$p.Name}} node pool stack",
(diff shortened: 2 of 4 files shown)


core/root/config/templates/stack-template.json
```diff
@@ -45,6 +45,11 @@
     "ControlPlaneStackName": {
       "Description": "The name of the control plane stack",
       "Value": {"Fn::GetAtt" : [ "{{$.ControlPlane.Name}}" , "Outputs.StackName" ]}
-    }
+    }{{range $_, $p := .NodePools}},
+    "{{ $p.Name }}StackName": {
```
mumoshu: nit: I prefer `"NodePool{{$p.Name}}StackName"` for more clarify ✨

mumoshu:
@jpb Thanks for the PR!
SGTM but I'd greatly appreciate it if you could also add a short documentation on how the external cloudformation stack template would look like!

mumoshu:
Just for confirmation but, in this way, we don't need to utilize `worker.managedIamRoleName` added via #297 to make a role name predictable?

mumoshu:
@jpb Could you sign the CLA? Thanks!

Fsero:
@mumoshu this will not help the predictability of the IAM role, it's just to be able to retrieve easily from other cloudformation stacks, but if cluster is recreated existing roles that assume the role exported role will break.

jpb:
I have now signed the CLA 👍 

When using `worker.managedIamRoleName`, the worker IAM roles are prefixed with the worker stack name (`StackName-Region-MangedIamRoleName`) which contains a random value. It may still be desirable to use `worker.managedIamRoleName` with this change, and only use the worker stack name that is exported from the parent stack to construct the worker IAM ARN.

I haven't tested yet if you can combine CloudFormation imports in a useful way (that is, first import the worker stack name and then use that imported value to import the worker's IAM role).

Either way, I thought it would be best to export both the stack name in the parent stack and the IAM role in the node pool stack.

I'll test both cases out and write documentation for the method that seems most reasonable.

jpb:
CloudFormation's [`Fn::ImportValue` can not depend on the return value of another `Fn::ImportValue`](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html), so I've updated this PR to export the controller and worker IAM roles from the "root" stack. This way you can reference a node pool's role from another stack using `{ "Fn::ImportValue" : {"Fn::Sub": "${KubeAWSStackName}-NodePoolDefaultIAMRoleWorkerArn" } }` (if you had a node pool called "default" and provide the root stack name as a parameter called `KubeAWSStackName`). This could be used with or without `worker.managedIamRoleName`.

Example "root" stack outputs (where `data-k8s-amazing-test` is the stack name and `default` is the node pool name):

screen shot 2017-04-20 at 5 52 53 pm

core/controlplane/config/templates/stack-template.json
```diff
@@ -1488,6 +1488,11 @@
     "StackName": {
       "Description": "The name of this stack which is used by node pool stacks to import outputs from this stack",
       "Value": { "Ref": "AWS::StackName" }
+    },
+    "ControllerIAMRoleArn": {
+      "Description": "The ARN of the IAM role for this Node Pool",
```
Fsero: there is a typo, The ARN of the IAM role for this Node Pool should be The ARN of the IAM role for the Controllers

Fsero:
im working on #500 and i will include this PR, could be merged first or we can try to merge both.

mumoshu:
@jpb LGTM. Thanks for your contribution 👍

PR #557, opened 2017-04-19
Fix hyperlink to restore script in Readme.md.

jollinshead:
Minor fixes:
- Fix hyperlink to restore script in Readme.md.
- Reference 'autosave' rather than 'export' in comments of cluster.yaml.


mumoshu:
@jollinshead LGTM. Thanks for the fix 👍

PR #558, opened 2017-04-19
Fix to calico configuration file etcd endpoints

kevtaylor:
ETCD_ENDPOINTS tokens are not being replaced in the 10-calico.conf file


mumoshu:
@kevtaylor What a mistake I made 😱 Thanks for the fix.
For my education, would you mind informing me the actual issues you've seen due to the mistake?

kevtaylor:
@mumoshu This becomes a problem when you try to load a container on a master by setting a taint toleration (in our case daemon set). The container fails to start and shows cni errors when you describe the pod. If you then try to terminate the pod, it goes into a spin-cycle and cannot delete the pod whatsoever. I had to restart kubelet without the cni options in order for the container to be removed. Once the ETCD endpoints have been added, this condition goes away. I was able to reproduce this just by attempting to delete/re-create the calico-node pod that you will see on the master(s)

PR #559, opened 2017-04-19
Add sprig templating functions

tyrannasaurusbanks:
The sprig library will give us some more functionality in the templates, and the next step would be to refactor the usage of the 2 custom template functions (minus and sha1) to make use of sprig implementations


glide.lock
```diff
@@ -49,15 +50,26 @@ imports:
...
 - name: github.com/spf13/pflag
   version: dabebe21bf790f782ea4c7bbd2efc430de182afd
+- name: golang.org/x/crypto
+  version: 1f22c0103821b9390939b6776727195525381532
+  subpackages:
+  - pbkdf2
+  - scrypt
 - name: gopkg.in/yaml.v2
-  version: a5b47d31c556af34a302ce5d659e6fea44d90de0
+  version: a83829b6f1293c91addabc89d0571c246397bbf4
```
mumoshu: go-yaml seems to be reverted to an older version here according to https://github.com/go-yaml/yaml/commits/v2
Did you intend to do so?

mumoshu:
@tyrannasaurusbanks Thanks for the PR!
The eventual goal of refactoring templates sounds good to me.
One minor comment for now.
Would you mind confirming it?

glide.lock
redbaron: how do you add new dependency without re-resolving everything?

what I do:
 1. add new dependency to "import" block in a source code where I need it
 2. `glide get <dependency>`
 3. `git add vendor; git commit -a -m "added <depepndency>"`

but this process still updates glide.lock and vendor even of unrelated packages.

tyrannasaurusbanks: @redbaron Yes this was the approach I used.
@mumoshu No i did not intend to pull in an older version - I'm quite surprised and a bit confused by this. Would you like me to revert the go-yaml version change?

tyrannasaurusbanks:
No worries, great project, I thought it was about time I got on board and contributed, I'll refactor the templates after this one!

glide.lock
mumoshu: Ah! If I remember correctly, although the gopkg.in is just a proxy for the repo in GitHub, the go-yaml v2 package served by gopkg.in has been a bit out-dated.
I believe I had manually modified the `version` after I ran `glide update` last time.

I guess it will be safe to rewrite the version manually back to `a5b47d31c556af34a302ce5d659e6fea44d90de0` after you run `glide get`.

mumoshu: Your approach seems good to me.

mumoshu: Hm, sorry for the back and forth but even go-ini seems to be reverted to an older version
https://github.com/kubernetes-incubator/kube-aws/pull/559/files#diff-f16a80eae23d5b298c2652448ec420cfR53

mumoshu: I can't reproduce your result by just running `glide get github.com/Masterminds/sprig`.
Mine does change the version of go-yaml as expected but doesn't change the version of go-ini for example.

mumoshu:
@redbaron @tyrannasaurusbanks Which versions of glide are you using?

For me:

```
$ glide --version
glide version 0.13.0-dev
```

tyrannasaurusbanks:
`$ glide --version
glide version 0.12.3`
Where did you get 0.13.0-dev from, did you build it from source?

glide.lock
tyrannasaurusbanks: I've manually rewritten the version's back locally and it builds, so if you're happy with that then i suggest we stick with those:
    - name: gopkg.in/yaml.v2
      version: a5b47d31c556af34a302ce5d659e6fea44d90de0
    - name: github.com/go-ini/ini
      version: 6e4869b434bd001f6983749881c7ead3545887d8

mumoshu:
> Where did you get 0.13.0-dev from, did you build it from source?

@tyrannasaurusbanks Thanks for the confirmation!
Yes - However I don't remeber the exact reason why I had done so.
According to the [full changelog](https://github.com/Masterminds/glide/compare/v0.12.3...master), I believe I don't need to stick with the dev version. Please let me confirm how it works for me with v0.12.3

mumoshu:
@tyrannasaurusbanks Even with v0.12.3, It seems like I can't reproduce your result
https://gist.github.com/mumoshu/5e5ccec4e9c3b88637680411dd9d127a

Could you provide me your output of `glide get`?

tyrannasaurusbanks:
Yeah sure, I've just reverted the changes on my local branch and re-added the dependency: 
https://gist.github.com/tyrannasaurusbanks/b139abcc787aa2a5e20cc8c4926dc022

mumoshu:
@tyrannasaurusbanks Thanks for the info!
Could you run `cd ~/.glide/cache/src/https-gopkg.in-yaml.v2 && git branch` and share me the commit id for the HEAD?

For me, it printed the commit id [which equals to the one seen in the updated glide.yaml](https://gist.github.com/mumoshu/5e5ccec4e9c3b88637680411dd9d127a#file-git-diff-log-L46).

```
$ cd ~/.glide/cache/src/https-gopkg.in-yaml.v2
$ git branch
* (HEAD detached at a3f3340)
  v2
```

I guess that t prints `a83829b` for you?

The issue in glide which seemed relevant to me is: https://github.com/Masterminds/glide/issues/592

mumoshu:
@tyrannasaurusbanks I guess running `glide cc & glide get github.com/Masterminds/sprig` allows us to produce the equivalent results.

Would you mind running it in your machine and confirming your result is equivalent to [the one shown in my gist](https://gist.github.com/mumoshu/0522dc9d76c7f34e3f87a6db9192bc38)?

tyrannasaurusbanks:
@mumoshu yes, `glide cc & glide get github.com/Masterminds/sprig` [has done the trick](https://gist.github.com/tyrannasaurusbanks/09c30b23e1ee32581e2d9525bf069947).

I've got some newer commit id's for github.com/aws/aws-sdk-go, but that makes sense given the updated timestamp in my log/run.
Thanks for taking the time to get to the bottom of this! I'll update my PR now

filereader/texttemplate/texttemplate.go
```diff
@@ -13,12 +12,8 @@ func GetBytesBuffer(filename string, data interface{}) (*bytes.Buffer, error) {
 	if err != nil {
 		return nil, err
 	}
-	funcMap := template.FuncMap{
-		"sha1":  func(v string) string { return fmt.Sprintf("%x", sha1.Sum([]byte(v))) },
```
mumoshu: Seems like we used the `sha1` func nowhere in templates. Good catch 👍

mumoshu:
@tyrannasaurusbanks Great! Thanks a lot for taking your time to finish this 👍

tyrannasaurusbanks:
@mumoshu no worries, thanks for taking the time to help me get my first commit merged - hopefully I can start looking at some more meatier/impactful ones now!

PR #561, opened 2017-04-20
Fix unwanted AWS resource creation/Add extra validation on internetGatewayID + vpcID

### LINKED ISSUE #318: Extra validation required for existing VPC plus no IGW spec
A combination of existing VPC but no IGW specification means that an IGW is created by kube-aws as expected but it is not attached to the VPC. kube-aws will currently fail on this at the point of stack creation. The fix is to specify an existing IGW in the same VPC within the `cluster.yaml`.

In theory this could be supported by kube-aws, we'd just need to move the `VPCGatewayAttachment` definition inside `controlplane` that is currently inside the `{{if not .VPCID}}` block.

I'm happy to support/not support this scenario if we have a single way we do existing infrastructure (we already support subnet/NAT/EIP management or pre-existing). IGWs do not add additional cost so it could make sense for kube-aws to always add it to keep the setup more uniform? On the other hand, if we choose to not support this in kube-aws then ideally we should support some validation error before stack creation and more explicit details in `cluster.yaml`.

Extracted from https://github.com/coreos/kube-aws/pull/315.

### LINKED ISSUE #553: Unwanted, orphaned IGW is created when existing route table is specified
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/318

mumoshu:
Fixes #318
Fixes #553

### DIFF
--- a/core/controlplane/cluster/cluster_test.go
+++ b/core/controlplane/cluster/cluster_test.go
@@ -175,10 +175,12 @@ instanceCIDR: 10.5.11.0/24
 `, `
 vpcCIDR: 192.168.1.0/24
 vpcId: vpc-xxx2
+internetGatewayId: igw-xxx1
 instanceCIDR: 192.168.1.50/28
 `, `
 vpcCIDR: 192.168.1.0/24
 vpcId: vpc-xxx2
+internetGatewayId: igw-xxx1
 subnets:
   - instanceCIDR: 192.168.1.0/28
   - instanceCIDR: 192.168.1.32/28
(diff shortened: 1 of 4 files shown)


redbaron:
we also see that IGW is created when existing subnets are specified, but no routing table (that is because existing subnets already have routing table and kube-aws shouldn't be concerned about it ;) )

mumoshu:
@redbaron Thanks for the info!
I guess this PR also resolves your case, as kube-aws now create an IGW only when vpcId is not given.

PR #565, opened 2017-04-20
Add reclaim policy

kevtaylor:
A small amendment to the Persistent Volume created as EFS.

We found that the default policy for the pv is Reclaim. When you delete the VolumeClaim and try to recreate it, it enters a pending state unless the volume itself is deleted and recreated.

Using Recycle as a reclaim policy allows reruns of the volume claim without having to recreate the volume

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -105,7 +105,8 @@
                   "    storage: 500Gi\n",
                   "  nfs:\n",
                   "    path: /\n",
-                  "    server: ", {"Ref": "FileSystemCustom"}, ".efs.{{ $.Region }}.amazonaws.com", "\n"
+                  "    server: ", {"Ref": "FileSystemCustom"}, ".efs.{{ $.Region }}.amazonaws.com", "\n",
+                  "  persistentVolumeReclaimPolicy: Recycle\n"
                 ]]}
               }
             }


mumoshu:
@kevtaylor Thanks for the fix!
Do you mean `Retain` rather than `Reclaim`, right?
If so, can I confirm if you're OK with k8s removing the contents of a directory in the EFS like [described in the upstream doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaim-policy)?

kevtaylor:
@mumoshu Sorry, yes meant to be Retain. And we are ok with the behaviour. A nice enhancement to this for the future would be to have a template of the volume definitions, so that it isn't so fixed in its approach. This route works for us in the meantime

kevtaylor:
@c-knowles Agreed. I think that adding some configuration options to the cluster.yaml will make this future proof. Do you want me to add a comment somewhere? Is there a link to the release notes?

mumoshu:
Please inform me your desired part of release note so that I can add it in
the release process :+1:
2017年4月21日(金) 18:00 Kevin Taylor <notifications@github.com>:

kevtaylor:
@mumoshu I think for the moment we just need to add something like:
#565 EFS PersistentVolume introduces Recycle policy which is a breaking change. Be aware that if all the persistent volume claims are removed, the EFS will recycle and remove all data.

mumoshu:
@kevtaylor Thanks. I'm ok with the description 👍 
@c-knowles Perhaps (we'd better|you'd like) to have a dedicated `Breaking changes` section like I've made in https://github.com/kubernetes-incubator/kube-aws/releases/tag/v0.9.5 for notes like this?

kevtaylor:
@mumoshu @c-knowles I think that looks a good idea - makes those issues stand out well

c-knowles:
@mumoshu perfect! I think this is good for our users and ease of adoption.

mumoshu:
Thanks for your confimations 👍

Issue #506, opened 2017-04-06
Using dex with kube-aws


mumoshu:
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/502#issuecomment-292064360
cc @qpaulson

camilb:
Hi @mumoshu, we plan to start working on this next week, if there is no one already working on it.

mumoshu:
AFAIK, no one is working on this yet. I'd greatly appreciate it if you could 👍

c-knowles:
Dex is already support with issues https://github.com/kubernetes-incubator/kube-aws/issues/601 and https://github.com/kubernetes-incubator/kube-aws/issues/668 still outstanding.

Issue #509, opened 2017-04-06
Plugins


mumoshu:
Would it be useful/possible to introduce "kube-aws plugins" to define sets of configurations and resources(settings, k8s manifests, required iam policies, etc) used to extend kube-aws, without complicating the core of kube-aws?

For example, I guess with the feature, #507 can be implemented as a plugin composed of:

## Design 1

* `plugins/enabled/`
  * `<unique name of an instance of plugin e.g. cluster-dump>/`
    * `controller`
      * `manifests/`
         * `cluster-dump.yaml.tmpl` with the content of [cluster-dump.yaml](https://github.com/kubernetes-incubator/kube-aws/pull/507/files#diff-ef25536c536667a40b993d4d24ab7567R795)
      * `iam-policies/`
        * `allow-puts-to-s3-path.json.tmpl` with the content of [the iam policy required by cluster-dump](https://github.com/kubernetes-incubator/kube-aws/pull/507/files#diff-7824eb2dc187893e9c52944adc99cebcR331)

Intentions:

* The directory is named `plugins/enabled/<unique name of an instance of plugin>` so that
  * we can later introduce `plugins/available/<unique name of a plugin>` for defining a plugin=a template of instances of plugin
* The `controller` directory is there to place controller node specific files and settings

## Design 2 

* `plugins/available/`
  * `<unique name of a plugin e.g. cluster-dump>/`
    * `controller`
      * `manifests/`
         * `cluster-dump.yaml.tmpl`
      * `iam-policies/`
        * `allow-puts-to-s3-path.json.tmpl`
* `cluster.yaml`
  * With the new configuration key to allow enabling and customizing plugins
     ```yaml
     clusterName: ***
     externalDNSName: ***
     kmsKeyARN: ***
     *snip*
     plugins:
       clusterDump:
       - pluginName: <unique name of a plugin e.g. cluster-dump>
           s3Prefix: <customs3prefix_to_where_cluster_is_dumped>
       rescheduler:
       - pluginName: rescheduler
    ```

and we bundle default plugins into kube-aws binaries, but putting files to `plugins/available` would allow customizing a default plugin.

---

@c-knowles I can't locate the exact link to it but if I remember correctly, we've discussed before about introducing "plugins" to make kube-aws extendable by users?

jeremyd:
[...]

PR #568, opened 2017-04-20
Dex integration.

camilb:
Hi @mumoshu I don't know if this is the best approach to integrate `dex` into `kube-aws` 
as I'm trying to make it as simple as possible but in the same time customisable and my skills with Go are limited. 

The current setup: 
* a Kubernetes "Deployment" + a minimal config in a "ConfigMap"
* all the parameters are set inside the "ConfigMap" which is generated from `cluster.yaml`
* after the cluster is up, the user can customize `dex` "ConfigMap" and "Deployment"
* upgrading/scaling the masters will not modify the "ConfigMap" or the "Deployment"
*`dex` is running on http on port 5556
* the user decide how to expose the service using a internal/public LoadBalancer, a Ingress Controller or a NodePort
* the required SSL certificates are managed by the user. In our case we prefer using a LoadBalancer with AWS Certificate Manager that will also remove the need to configure the `--oidc-ca-file` parameter 

Issues:
* some steps are still manual, like creating the service and configuring the Loadbalancer or Ingress but these are  pretty easy to configure and we can provide examples 
* advanced configurations require modifying the "ConfigMap"/"Deployment". For example, if you plan to use the LDAP connector. Adding the desired configuration in `cluster.yaml` is not parsed well in this case. It's fine for the other connectors.
* `--oidc-ca-file` parameter is working but there is no method yet to load the CA certificate on masters

Possible improvements I'm thinking at:
* generate self signed certificates using `kube-aws`
* find a method to put sensitive data (for example the credentials for the connectors) in secrets (this can be done now manually after the cluster is up)
* configure a Ingress/LoadBalancer/NodePort automatically (in this case Tectonic is using a ingress that terminates the TLS for all the services, including `dex`, and put a TCP LoadBalancer in front)

I'm open for suggestions on how we can improve this.
Related to #506

Thanks!

### DIFF
--- a/contrib/dex/README.md
+++ b/contrib/dex/README.md
@@ -1,4 +1,4 @@
-# Exposing DEX service
+## Exposing DEX service
 After the cluster is up, you have to manually expose dex service using a ELB or Ingress.
 Faster is to use the `expose-service.sh` script or you can manually configure the services using the examples from `contrib/dex` directory.
 
@@ -17,3 +17,49 @@ You pay only for the AWS resources you create to run your application. This is t
 After deploying the Ingress you have to configure the workers security group to allow access on port 443 and optionally on port 80.
 Please note that if you plan to restrict access of these ports to your IP, you also have to allow access from controller nodes, as the API server will access the dex endpoint.
 
+##Configure `kubectl` for token authentication
+
+* `kubectl` config using command line example:
+
+
+    kubectl config set-credentials admin@example.com  \
+    --auth-provider=oidc \   
+    --auth-provider-arg=idp-issuer-url=https://dex.example.com \
+    --auth-provider-arg=client-id=example-app \
+    --auth-provider-arg=client-secret=ZXhhbXBsZS1hcHAtc2VjcmV0 \   
+    --auth-provider-arg=refresh-token=refresh_token \   
+    --auth-provider-arg=idp-certificate-authority=/etc/kubernetes/ssl/ca.pem \   
+    --auth-provider-arg=id-token=id_token \
+    --auth-provider-arg=extra-scopes=groups
+
+* `kubectl` config file example:
+
+
+    apiVersion: v1
+    clusters:
+    - cluster:
+        certificate-authority-data: ca.pem_base64_encoded
+        server: https://kubeapi.example.com
+      name: your_cluster_name
+    contexts:
+    - context:
+        cluster: your_cluster_name
+        user: admin@example.com
+      name: your_cluster_name
+    current-context: your_cluster_name
+    kind: Config
+    preferences: {}
+    users:
+    - name: admin@example.com
+      user:
+        auth-provider:
+          config:
+            access-token: id_token
+            client-id: example-app 
+            client-secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+            extra-scopes: groups
+            id-token: id_token
+            idp-issuer-url: https://dex.example.com
+            refresh-token: refresh_token
+          name: oidc
+          
--- a/contrib/dex/README.md
+++ b/contrib/dex/README.md
@@ -1,0 +1,19 @@
+# Exposing DEX service
+After the cluster is up, you have to manually expose dex service using a ELB or Ingress.
+Faster is to use the `expose-service.sh` script or you can manually configure the services using the examples from `contrib/dex` directory.
+
+1. ELB
+
+First option is to use a Public or Internal ELB.
+
+In this case you have to edit one of the files from `contrib/dex/elb` directory and set your certificate `arn` and `domainName`.
+
+Note: 
+* SSL/TLS certificates provisioned through AWS Certificate Manager are free. 
+You pay only for the AWS resources you create to run your application. This is the recommended method.
+
+2. Ingress
+
+After deploying the Ingress you have to configure the workers security group to allow access on port 443 and optionally on port 80.
+Please note that if you plan to restrict access of these ports to your IP, you also have to allow access from controller nodes, as the API server will access the dex endpoint.
+
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -621,6 +621,16 @@ write_files:
       {{ end }}
       {{ end }}
 
+      {{ if .Experimental.Dex.Enabled }}
+      if kubectl get cm dex -n kube-system;
+        then echo "Configmap for dex already exists and could have been customized. Skipping."; 
+      else 
+        mfdir=/srv/kubernetes/manifests
+        kubectl apply -f "${mfdir}/dex-cm.yaml"
+        kubectl apply -f "${mfdir}/dex-deployment.yaml";
+      fi
+      {{ end }}
+
   - path: /etc/kubernetes/cni/docker_opts_cni.env
     content: |
       DOCKER_OPT_BIP=""
@@ -1263,6 +1273,19 @@ write_files:
           - --advertise-address=$private_ipv4
           - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass{{if .Experimental.Admission.PodSecurityPolicy.Enabled}},PodSecurityPolicy{{ end }},ResourceQuota
           - --anonymous-auth=false
+          {{if .Experimental.Dex.Enabled}}
+          - --oidc-issuer-url={{.Experimental.Dex.Url}}
+          - --oidc-client-id={{.Experimental.Dex.ClientId}}
+          {{if .Experimental.Dex.Username}}
+          - --oidc-username-claim={{.Experimental.Dex.Username}}
+          {{ end -}}
+          {{if .Experimental.Dex.Groups}}
+          - --oidc-groups-claim={{.Experimental.Dex.Groups}}
+          {{ end -}}
+          {{if .Experimental.Dex.CaFile}}
+          - --oidc-ca-file={{.Experimental.Dex.CaFile}}
+          {{ end -}}
+          {{ end -}}
           - --cert-dir=/etc/kubernetes/ssl
           - --tls-cert-file=/etc/kubernetes/ssl/apiserver.pem
           - --tls-private-key-file=/etc/kubernetes/ssl/apiserver-key.pem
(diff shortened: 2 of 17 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1071,6 +1071,43 @@ addons:
 #   # When enabled, `kubectl drain` is run on the node going to shut-down
 #   nodeDrainer:
 #     enabled: true
+#   # Enable dex  integration - https://github.com/coreos/dex
+#   # Configure OpenID Connect token authenticator plugin in Kubernetes API server. Notice: always use "https" for the "url",
+#   # otherwise the Kubernetes API server will not start correctly. The "caFile" can be skipped if you use a certificate signed
+#   # by a trusted CA.
+#   dex:
+#     enabled: true
+#     url: "https://dex.example.com"
```
mumoshu: Should we ideally provision an ELB and a record set for serving the url?
// However, it might be a separate task. I'd just like to discuss about this for now

camilb: As I already added the functionality to generate the certificates from `kube-aws` I think maybe It will be  better to configure `dex`with https on port 5556 and create a NodePort service. Then add a ELB and a record set.

mumoshu: Great 👍 Thanks for the addition.
Yes - I agree to the ideas.
Are those something you can address in this PR or in another PR(s)?

model/dex.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/url"
+)
+
+type Dex struct {
+	Enabled         bool             `yaml:"enabled"`
+	Url             string           `yaml:"url"`
+	ClientId        string           `yaml:"clientId"`
+	Username        string           `yaml:"username"`
+	Groups          string           `yaml:"groups,omitempty"`
+	CaFile          string           `yaml:"caFile,omitempty"`
```
mumoshu: Perhaps we'd better set CaFile to `/etc/kubernetes/ssl/ca.pem` by default?

mumoshu: FYI you can implement `UnmarshalYAML` like [this](https://github.com/kubernetes-incubator/kube-aws/blob/master/core/nodepool/config/config.go#L161-L167) to automatically fill default values into a Dex struct.

core/controlplane/config/templates/cluster.yaml
camilb: After reading @c-knowles's comment from #509, I think using helm in the future will be a better option to keep `kube-aws`code clean. So, i think is better to provide some examples and a script to easily expose dex service using ELB or Ingress after the cluster is up.  Already made the commit. As for the Route53 record, I'm always use a kops's dns-controller to configure all the records automatically for Ingress and ELBs, but there are many solutions for this.

model/dex.go
camilb: If we use signed certificates and leaving CaFile empty, will not create the flag for ca. Maybe using a bool instead will be better.

camilb:
@mumoshu After integrating the latest changes from `master`branch, can not pass the tests anymore. It's strange because even the latest `master` branch doesn't pass the tests locally. There is something you are aware of, or I should look on my side?

camilb:
Thanks @mumoshu, didn't see it. Locally looks fine now, I think my firewall was blocking some connections to coreos.com.
screen shot 2017-04-26 at 07 14 17

camilb:
Hi @mumoshu, at this moment it works pretty good on my tests. Added some examples on how to expose the dex service and even if is not completely automated, is really fast to do it and more customisable. I think it will be better to use helm in the future to configure it and leave only the API flags part in `kube-aws`. Please let me know if other changes are necessary. Thanks!

contrib/dex/README.md
```diff
@@ -0,0 +1,65 @@
+## Exposing DEX service
+After the cluster is up, you have to manually expose dex service using a ELB or Ingress.
```
mumoshu: @camilb Not necessary in this PR but just my two cents.

I guess it will be even better if we could enable users to create an elb and a record set configured appropriately by writing cluster.yaml like we do for apiEndpoints[]:

```yaml
dex:
  exposedBy:
    loadBalancer:
      hostedZone:
        id: hostedzone-abcdefg
```

so that we can ensure people set up dex in a more automated, fail-proof way.

camilb: I agree. I'll take a look on it later today.

c-knowles: Could `dex` go under the `addons` top level config I added during the rescheduler work?

mumoshu: I'm ok with either it gets done in this PR or not but anyway, I think so.

camilb: @c-knowles Yes, I think it makes more sense to put it under `addons`. I'll make the changes.
@mumoshu I prefer doing this part in a separate PR. Right now the TLS termination is handled by ELB or Ingress and we need to configure the certificates for the ELB. Kubernetes + Amazon Certificate Manager makes this easy. Of course I can enable `https` on `dex`but then lose the possibility to use the certificates provided by Amazon. Have to think more on this part.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1817,6 +1840,15 @@ write_files:
...
     encoding: gzip+base64
     content: {{.TLSConfig.EtcdClientKey}}
+
+  - path: /etc/kubernetes/ssl/dex.pem{{if .AssetsEncryptionEnabled}}.enc{{end}}
+    encoding: gzip+base64
+    content: {{.TLSConfig.DexCert}}
+
+  - path: /etc/kubernetes/ssl/dex-key.pem{{if .AssetsEncryptionEnabled}}.enc{{end}}
+    encoding: gzip+base64
+    content: {{.TLSConfig.DexKey}}
```
mumoshu: nit: dex credentials should be written only when `{{if .Experimental.Dex.Enabled}}`

contrib/dex/README.md
mumoshu: @camilb ok 👍 Then, this PR is LGTM. Please remove the `WIP` in the title of this PR when you've done/this PR is ready to be merged.

I also prefer using ACM for managing the cert for dex ELB.
I have not yet tried it myself but [the ACM certificate resource](http://docs.aws.amazon.com/ja_jp/AWSCloudFormation/latest/UserGuide/aws-resource-certificatemanager-certificate.html) seems to allow us to create an ACM cert managed by a cfn stack. AFAICS, a manual certificate approval seems to be required by a human though(=it isn't fully automated)

mumoshu:
@camilb Squashed and merged. Thanks a lot for your contribution 👍

mumoshu:
In cluster.yaml:

> userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"

@camilb How would you generate an userID like this for example? It looks like an UUID anyway.

camilb:
@mumoshu Im using  [uuidgen](http://man7.org/linux/man-pages/man1/uuidgen.1.html)

mumoshu:
@camilb Thanks for the info 👍

contrib/dex/README.md
c-knowles: Maybe for now we could allow the user to specify the Service annotation `service.beta.kubernetes.io/aws-load-balancer-ssl-cert` which needs the ARN of the certificate?

c-knowles:
I have a question about [these docs](https://github.com/coreos/dex/blob/master/Documentation/storage.md#kubernetes-third-party-resources) which say Dex requires to the non-namespaced ThirdPartyResource type. This doesn't seem to be present in kube-aws, is it not needed as we have more global access for apps running the kube-system namespace? I've been trying to get this running, definitely we need to investigate the Helm chart along with adding the Service definition to expose Dex.

camilb:
Hi @c-knowles, Dex is creating these ThirdPartyResource

    $ kubectl get thirdpartyresource
    NAME                               DESCRIPTION                                                                VERSION(S)
    auth-code.oidc.coreos.com          A code which can be claimed for an access token.                           v1
    auth-request.oidc.coreos.com       A request for an end user to authorize a client.                           v1
    connector.oidc.coreos.com          Connectors available for login                                             v1
    o-auth2-client.oidc.coreos.com     An OpenID Connect client.                                                  v1
    offline-sessions.oidc.coreos.com   User sessions with an active refresh token.                                v1
    password.oidc.coreos.com           Passwords managed by the OIDC server.                                      v1
    refresh-token.oidc.coreos.com      Refresh tokens for clients to continuously act on behalf of an end user.   v1
    signing-key.oidc.coreos.com        Keys used to sign and verify OpenID Connect tokens.                        v1

Are you asking about the RBAC permissions to create ThirdPartyResource? 
Right now there is no restriction, even the RBAC is enabled. 

As for exposing the service, in my case I'm using a internal ELB but we can also use a external ELB or a ingress. [Here](https://github.com/kubernetes-incubator/kube-aws/tree/master/contrib/dex) you have some examples. 

In the future I think could be a good idea to use something like [this](https://gist.github.com/camilb/69008e4dddeb5d663e2e6699b3a93283). It's a public ELB that forward all the traffic to a nginx ingress controller and we can have multiple services using the same ELB. Tectonic it's using it to expose dex.

 I'm thinking having a option in kube-aws to deploy this solution and expose services like dex. This looks good until we have a more simpler solution for a ALB. 

In my deployment I'm also exposing the [example app](https://github.com/coreos/dex/tree/master/cmd/example-app) with the same internal ELB, so the developers can connect easier and get a token. 

And Helm will simplify this more.

c-knowles:
Ah yes, I missed out an important word! Corrected above. OK, sure so it's because we are adding fairly open permissions by default in kube-aws when RBAC is enabled. Oh, I didn't even realise there was documentation there! I think we should put a mention of it in the cluster.yaml so it's easier to find. I'll look through that now as I'll need that bit too.

Yep, as you say if we do https://github.com/kubernetes-incubator/kube-aws/issues/656 for the plugins work then it will be very easy to install a new ingress controller. I'm using Traefik as I prefer it over nginx.

c-knowles:
@camilb any experience exposing using AWS cert manager? I'm getting `/.well-known/openid-configuration: x509: certificate signed by unknown authority` when I try to configure `kubectl` using the command from your contrib README. I have `selfSignedCa: false`.

c-knowles:
Hmmm. The problem above was that I didn't have any `staticClients` defined as I didn't know I needed them but when I updated this config and did a `kube-aws update` it ran successfully but the updates to the ConfigMap were not propagated to Kubernetes for some reason. I assume that is not expected?

Not sure yet if it's still related to my setup but now getting this with a static user I generated - `error: You must be logged in to the server (the server has asked for the client to provide credentials (get nodes))`

c-knowles:
Sorry, updated the above.

camilb:
> @camilb any experience exposing using AWS cert manager? I'm getting /.well-known/openid-configuration: x509: certificate signed by unknown authority when I try to configure kubectl using the command from your contrib README. I have selfSignedCa: false.

@c-knowles Please check your `~/.kube/config` and remove this line if it exists `idp-certificate-authority: /something` On debian I also had to run `update-ca-certificates`.
This should work, please let me know if it fixed the problem for you.

camilb:
> Hmmm. The problem above was that I didn't have any staticClients defined as I didn't know I needed them but when I updated this config and did a kube-aws update it ran successfully but the updates to the ConfigMap were not propagated to Kubernetes for some reason. I assume that is not expected?
Not sure yet if it's still related to my setup but now getting this with a static user I generated - error: You must be logged in to the server (the server has asked for the client to provide credentials (get nodes))

@c-knowles this seems related to your `client-id`. If you monitor your kubernetes API logs, most probably you'll see that there is a error similar to this:

`2017-05-21T22:29:59.847757353Z AUDIT: id="96f1c69f-75d4-435d-aa94-1026ee2e5008" response="200"
E0521 22:30:00.362590       1 authentication.go:58] Unable to authenticate the request due to an error: [invalid bearer token, [invalid bearer token, crypto/rsa: verification error, oidc: JWT claims invalid: invalid claims, 'aud' claim and 'client_id' do not match, aud=kid, client_id=example-app]]`

Please check the `clientId` in `cluster.yaml` and the client id you set in your client app (I'm using the dex "example-app" where you can pass a flag with your client id). Please let me know if this fix the problem.

camilb:
> The problem above was that I didn't have any staticClients defined as I didn't know I needed them but when I updated this config and did a kube-aws update it ran successfully but the updates to the ConfigMap were not propagated to Kubernetes for some reason.

Here maybe is unclear, the current design will not update the ConfigMap if exists. I thought that maybe you want to customise  your ConfigMap after deployment, for example configuring the integration with LDAP, which at the moment isn't possible to do it  thru `kube-aws`, then when you run `kube-aws update`didn't want to erase all of these changes.

c-knowles:
@camilb Great, taking `idp-certificate-authority` out of kubeconfig has solved that one issue. Thanks! I had seen the example app from Dex but also that some people run it inside Kubernetes which I thought was what we would be doing. So I just misunderstood part of the flow. No worries, I will make sure we cover this stuff with updated documentation and I think I understand it better now, thank you.

Regarding updating the config, the other parts of kube-aws either have self-managed or kube-aws managed. I think we should probably avoid having a mixture of both "create it, but don't update it" which will probably be helped by the later Helm updates.

c-knowles:
@camilb got this working after quite a bit of fiddling around, the good news is I have some notes so we can improve the instructions which I'll post in an issue later.

I have one question about how you secure the `--client-secret` passed to the example app? Do you distribute that freely to developers since it needs to be in the kube config for the token to refresh? I'm just wondering how secure it has to be but it seems in this scenario it has to be distributed to employees if their token is to not expire. The other option is deploy the example app or similar and have them request a new one from there when needed. I'm not sure what the default expiry time is now?

camilb:
@c-knowles I have the "example-app" deployed inside kube-system namespace and exposed using an internal ELB. The developers can access this service using a VPN (same for the Kubernetes API) and refresh their token whenever they like.  This way I don't have to pass them the app and the credentials. I'm working on a blog post with more informations on how to configure kube-aws and I'll also explain the dex part. Right now I didn't make public the git repo, but  just created a temporary one for the example-app part, if you like to take a look: https://github.com/camilb/kid

c-knowles:
Ah right so then they have to refresh the token themselves when it expires. I checked the expiry and it seems to be 24 hours at least by default.

camilb:
This can be controlled thru gRPC. You can force a token to expire in case of emergency and some other things. Tectonic it's using it like this.

c-knowles:
Sure, I'm not quite at the stage of building out a client for this - just wanting to see it working a bit first.

Hopefully final question here - can't wait for our Slack channel - do you know if `dex.clientId` and `dex.staticClients[].id` should typically be the same? It seems to be the case in all the examples I saw when I searched around.

camilb:
@c-knowles
Yes, should be the same. Otherwise you'll get  the error above in API (`'client_id' do not match...`)

c-knowles:
Cool, thanks. I thought that was the case, just wanted to double check my understanding. Having gone through all this setup and figured out the moving parts, with a single ID provider I can remove Dex and just use the kube-aws apiserver Dex settings you added to configure it like [here](https://www.skuid.com/blog/reduce-administrative-toil-with-kubernetes-1-3/). So far I can't spot any disadvantages; both ways need a small app to bootstrap user tokens, the app I linked can also be moved in cluster quite easily. The obvious advantage is one less thing to go wrong (Dex). I kind of assume you are using multiple connectors or I've totally missed some other advantage to using Dex.

camilb:
Yes, in our case the main advantage at the moment is the possibility to use multiple connectors, but configuring dex adds more complexity. The rest of the advantages dex provides are not that easy to use. Thinking on keeping only the the configuration for OIDC that can be configured with different providers directly and also with dex. And move dex to helm.  This will also solve #668.

PR #575, opened 2017-04-21
Quote security group refs for etcd, controller, and apiendpoints

soellman:
I noticed that the recent work to extend glue security groups to etcd/controllers/apiendpoints were missing the double quotes when being rendered into the stack json. I think I addressed all instances of this case.

I was getting the following error when enabling security groups for the controllers:
```
Error while rendering template : Error while rendering stack template : invalid character 's' looking for beginning of value:
json syntax error (offset=17339), in this region:
-------
,
        "SecurityGroups": [
          
          
          sg-af3da9d6
          
          ,
          {"Ref":"SecurityGroupController"}

-------

```

### DIFF
--- a/model/controller.go
+++ b/model/controller.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"errors"
+	"fmt"
 )
 
 // TODO Merge this with NodePoolConfig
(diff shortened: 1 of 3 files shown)


soellman:
I signed the CLA

mumoshu:
@soellman My tests on the feature were definitely insufficient 😢 Thanks for the fix!

PR #576, opened 2017-04-21
Set --storage-backend to etcd2 if not using etcd3

cheungpat:
Since Kubernetes 1.6, the default storage backend is etcd3 if
not specified. If etcd version is set to '2' in cluster config,
we need to set --storage-backend=etcd2.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1243,6 +1243,8 @@ write_files:
           - --secure-port=443
           {{if .Etcd.Version.Is3}}
           - --storage-backend=etcd3
+          {{else}}
+          - --storage-backend=etcd2
           {{end}}
           - --kubelet-preferred-address-types=InternalIP,Hostname,ExternalIP
           {{ if .AuthTokensConfig.HasTokens }}


cheungpat:
@k8s-ci-robot

mumoshu:
@cheungpat Good catch 👍 Thanks for the fix.
Btw, have you successfully run your cluster with k8s 1.6 + etcd2?
I'd appreciate it if you could share us your experience so that I can determine whether we should recommend the combination or not.

cheungpat:
@mumoshu Yes I am running k8s 1.6 with etcd2 and it is working. I guess I will not recommend it for new cluster though.

My cluster was created with v0.9.4-rc.X and I decided to keep the etcd2 instance because I prefer not to re-create the cluster when upgrading beyond v0.9.5. Basically what I did is to export the stack templates using `kube-aws up --export` and replace the etcd3 ASG with the old etcd2 instance.

Since I did not recreate the cluster from scratch I don’t know if kube-aws can create a working k8s 1.6 + etcd2 cluster.

mumoshu:
@cheungpat I see. Thanks for the info 👍

Issue #574, opened 2017-04-21
Enable controller nodes to be labeled


mumoshu:
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/170#issuecomment-288405316

Currently, `experimental.nodeLabels` is commented in cluster.yaml like:

> Kubernetes node labels to be added to controller nodes

which doesn't seem to work.

Will also help #566 by allowing to label controller nodes for utilizing nodeAffinity

mumoshu:
Resolved by #577. Thanks for the PR @cheungpat 😄

PR #577, opened 2017-04-23
Add controller node labels if specified

cheungpat:
It appears that the userdata of controller is missing --node-labels on kubelet
even if it is specified in cluster config.


cheungpat:
I just realised there is an issue: #574

mumoshu:
@cheungpat Great! Thanks for your contribution 👍

PR #581, opened 2017-04-24
Update kubelet flags

### LINKED ISSUE #543: Update kubelet deprecated flags
While looking into the panic from https://github.com/kubernetes-incubator/kube-aws/issues/527#issuecomment-293148617, I found that the `api-servers` kubelet flag that kube-aws is using is [deprecated](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L127) some time ago in https://github.com/kubernetes/kubernetes/commit/a66828d73e0f8ca411daf93b6bc7c9ba029956ff. Perhaps there are other flags marked with `MarkDeprecated` as well. This task is to update all kubelet deprecated flags so we are ready for their removal.

mumoshu:
Resolves #543 
cc @c-knowles

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -189,7 +189,7 @@ coreos:
         --rkt-stage1-image=coreos.com/rkt/stage1-coreos \
         {{if .Experimental.NodeLabels.Enabled}}--node-labels {{.Experimental.NodeLabels.String}} \
         {{end}}--register-node=true \
-        {{if .Experimental.Taints}}--register-schedulable=false \
+        {{if .Experimental.Taints}}--register-with-taints={{.Experimental.Taints.String}}\
         {{end}}--allow-privileged=true \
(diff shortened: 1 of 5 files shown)


core/controlplane/config/templates/cloud-config-worker
```diff
@@ -718,32 +696,6 @@ write_files:
...
-           kubectl="/kubectl --server={{.APIEndpointURL}}:443 --kubeconfig=/etc/kubernetes/worker-kubeconfig.yaml"
-           taint="$kubectl taint node --overwrite"
-           for t in ${taints[@]}; do
-             $taint "$hostname" "$t"
-           done
-           echo "done."
-           echo "uncordoning this node."
-           $kubectl uncordon $hostname
-           echo "done."'
-
```
mumoshu: Probably we'd want to bring this back to keep supporting k8s 1.5.x?

mumoshu:
E2E passed. Merging.

PR #582, opened 2017-04-24
Update kube-dns to 1.14.1

### LINKED ISSUE #542: Update kube-dns to latest (v1.14.1)
It looks like [kube-dns has move on a bit](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns) since kube-aws last updated.

Also see the [kube-dns releases page](https://github.com/kubernetes/dns/releases).

At least these things have changed:
1. Moved from `gcr.io/google_containers/kubedns-amd64` that we are using in kube-aws to `gcr.io/google_containers/k8s-dns-kube-dns-amd64`
1. Various other changes to the Deployment spec
1. Added config map and service account

Not sure yet what we need to add into kube-aws but at least we cannot get the latest version without updating the repo.

mumoshu:
Resolves #542


mumoshu:
E2E passed. Merging 💨

PR #589, opened 2017-04-26
Bump to calico 2.1.4

redbaron:


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -690,6 +690,8 @@ write_files:
         selector:
           matchLabels:
             k8s-app: calico-node
+        updateStrategy:
+          type: RollingUpdate
```
mumoshu: Good to know that k8s finally support rolling updates of daemonsets!

mumoshu:
LGTM. Thanks for your continuous support 👍

core/controlplane/config/templates/cloud-config-controller
c-knowles: Are we fully removing support for k8s 1.5? This will be rejected on k8s < 1.6 based on what I've seen before. It's not a problem for my use cases but some other users seems to want/need to use slightly older versions still.

mumoshu: @c-knowles Thanks for chiming in! Yes, this will break the 1.5.x support.

I have never intended to keep supporting k8s 1.5.x myself. In that regard, this change is OK for me. However, I know some people still likes to go with 1.5.x.

I wish someone helps me bringing back the 1.5.x support but if no one is going to, I'll do, probably after listing up all the changes required to re-add 1.5.x support  😃

mumoshu: Opened https://github.com/kubernetes-incubator/kube-aws/issues/599 for that

PR #594, opened 2017-04-26
Fix syntax error

### LINKED ISSUE #593: Typo in install-kube-system causes error when TLS bootstrapping is enabled
Found [this](https://github.com/kubernetes-incubator/kube-aws/blob/master/core/controlplane/config/templates/cloud-config-controller#L607) bash syntax error which prevents the install script to terminate successfully when TLS bootstrapping feature is enabled.

danielfm:
Closes #593.


mumoshu:
LGTM. Thanks for the fix 🎉

PR #602, opened 2017-04-27
Forgot to bump calico ctl to 1.1.3

redbaron:


mumoshu:
LGTM according to http://docs.projectcalico.org/v2.1/releases/. Thanks as always @redbaron :bow:

Issue #586, opened 2017-04-25
With loadbalancer option is correct? apiEndpoints or controller?


Vrtak-CZ:
After introducing apiEndpoints there is two loadbalancer subsection in `cluster.yaml` one is in `apiEndpoints` and another one is in `clontroller`. With one is used and correct?

kube-aws - `v0.9.6-rc.4`

camilb:
@Vrtak-CZ I'm using  only the `apiEndpoints.loadBalancer` and it works fine. Had the same doubt.

mumoshu:
@Vrtak-CZ @camilb Sorry for the confusion and thanks for reporting 👍 
Yea - `controller.loadBalancer` is deprecated in favor of recently added `apiEndpoints[].loadBalancer`.
I wonder if I added a deprecation warning when you've specified `controller.loadBalancer`

mumoshu:
@Vrtak-CZ Can we close this? 😄

Vrtak-CZ:
fixed by #603

PR #603, opened 2017-04-27
controller.loadBalancer is deprecated use apiEndpoints[].loadBalancer

Vrtak-CZ:
fixies https://github.com/kubernetes-incubator/kube-aws/issues/586

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -191,6 +191,7 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #    # References subnets defined under the top-level `subnets` key by their names
 #    - name: ManagedPublicSubnet1
 #    - name: ManagedPublicSubnet2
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -191,20 +191,6 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #    # References subnets defined under the top-level `subnets` key by their names
 #    - name: ManagedPublicSubnet1
 #    - name: ManagedPublicSubnet2
-#  loadBalancer:
```
mumoshu: Could you comment `CAUTION: Deprecated and will be removed in v0.9.7. Please use apiEndpoints[].loadBalancer instead` on this like we do for other deprecated keys?
so that our users won't lose track to new keys

Vrtak-CZ: done

mumoshu:
@Vrtak-CZ Thanks for your contribution 👍 🍻

Issue #295, opened 2017-02-02
TLS handshake errors caused by Controller ELB running tcp healthchecks to 443


sasso:
We are seeing a lot of "TLS handshake error" on kube-apiserver and authproxy

![screenshot 2017-02-02 09 29 08](https://cloud.githubusercontent.com/assets/8398242/22560662/4512e5d0-e92a-11e6-9988-b436134d5d5d.png)

Anyone have any ideas why this is happening?

redbaron:
check from the client side, there should be  more informative error

sasso:
Client side works and I dont see any errors

redbaron:
there must be, these IPs belong to pods right? and something there failed to established connection

sasso:
huh, ok I think I can pin it to a service and ELB healthcheck. ty!

sasso:
![screenshot 2017-02-02 11 36 48](https://cloud.githubusercontent.com/assets/8398242/22565138/f36372c4-e93b-11e6-8bab-6927420e7d66.png)
![screenshot 2017-02-02 11 36 58](https://cloud.githubusercontent.com/assets/8398242/22565139/f37921fa-e93b-11e6-8b50-d8ecf28c832a.png)

This happens because we are doing TCP healthcheck on SSL listener

whereisaaron:
The apiserver requires a client certificate the ELB doesn't have, so the ELB can only do a TCP connect healthcheck, and that connect/disconnect triggers those errors in the logs. I get them too. It would be nice if there were someway to suppress them, or not log on a zero-data connect/disconnect.

mumoshu:
Thanks for the info @whereisaaron!
How about allowing unauthenticated http accesses to apiserver from ELB in a security group so that we can use http health checks instead of tcp checks?

whereisaaron:
Enabling any unauthenticated/anonymous access to the API makes be a bit uneasy @mumoshu !

It is possible for [failed client-cert requests to fall back to anonymous](https://kubernetes.io/docs/admin/authentication/#anonymous-requests), but I don't see that you can restrict those anonymous requests to just the ELB, so everyone with access to the API (from either side of the ELB) could make anonymous requests. These would map to the `system:unauthenticated` RBAC group and so access could be restricted by RBAC. But if you didn't have RBAC enabled, you'd have the whole cluster open for unauthenticated admin access.

[...]

PR #604, opened 2017-04-27
Change API endpoint ELB health check to SSL:443

cheungpat:
The API endpoint ELB health check is set to "TCP:443". This causes
"TLS handshake error" in apiserver logs. Changing it to "SSL:443" will
perform the proper SSL handshake and make the error disappear from the log.

```
I0427 03:57:55.059255       1 logs.go:41] http: TLS handshake error from 172.24.14.156:11370: EOF
I0427 03:57:55.059952       1 logs.go:41] http: TLS handshake error from 172.24.2.59:6348: EOF
```

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -786,7 +786,7 @@
         "HealthCheck" : {
           "HealthyThreshold" : "3",
           "Interval" : "10",
-          "Target" : "TCP:443",
+          "Target" : "SSL:443",
           "Timeout" : "8",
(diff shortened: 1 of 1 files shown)


mumoshu:
This resolves #295

mumoshu:
@cheungpat Wow, good to know ELB supports `SSL` healthchecks! It did work for me - the errors disappeared from apiserver logs.
Thanks for the great job 👍

whereisaaron:
Cool @cheungpat @mumoshu I didn't realise SSL health checks without the client certificate would work. I'll have to try this myself on the next deployment.

Issue #340, opened 2017-02-22
Feature: Add custom IAM policies to worker node pools


c-knowles:
I'm using some [route 53 tooling](https://github.com/fairfaxmedia/area53) which requires additions to the worker IAM role. I'd like to be able to able to specify custom policies without altering the stack templates so much.

We've added some similar items like in https://github.com/coreos/kube-aws/pull/297 for common use cases but I feel like each one is tool specific and not sure we should actively support them all. Since it's [not possible to attach multiple IAM roles to a single instance](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) that means there's no way to "glue" them like we do with SGs. As a consequence the only neat way to do this without customising the stack templates every time is for kube-aws to provide some sort of policy pass through. Could we investigate a generic way for users to add custom parts to the IAM roles, specifically workers?

redbaron:
here we go, our `cluster.yaml` continues it's journey to become `cloudformation.yaml` :)

on serious note, with nodepool mega PR merged, we now have facilities to embed nested CF stacks. If we have `user-stack.json` which is managed by user and never touched by kube-aws apart form including it as a nested stack, will it work for you?

redbaron:
@c-knowles , out of curiosity, did you evaluate https://github.com/kubernetes/kops/tree/master/dns-controller before choosing  https://github.com/fairfaxmedia/area53 ?

c-knowles:
@redbaron I agree with you on the cluster.yaml aspect. I'd prefer if there is a way to link custom pieces into the stack. That's instead of kube-aws directly managing it or needing to edit the kube-aws driven stack templates each time. Editing that way is usually complex and error prone since they change frequently, are hard to read/parse and often it's difficult to have knowledge of every incoming changes kube-aws provides.

For nested CF stacks, I'm not sure exactly how that could work as the role can only be a single one per instance. However, [multiple policies can be included](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#aws-resource-iam-role-syntax) in a single role so if there's a way to link in there that'd work.

[...]

PR #607, opened 2017-04-27
More flexible configuration for IAM and stable naming for roles

Fsero:
this PR aims to fix issue #500 allowing Node Pools to use an existing instance profile. It also includes a brief improvement for IAM Roles managed by kube-aws supporting customer managed policies (ref #340), it also makes IAM Roles fixed and not depending on the stack name if you have configured managedIAMRoleName.

If kube-aws manages the IAM Role it will create a customer managed policy for controllers and for workers, then it will create an IAM Role with this policy attached, that enables policy updates without the need to recreating the whole stack and kube-aws users can tailor the policies to their current needs.

if kube-aws uses an existing InstanceProfile it will not create an IAM Role and the InstanceProfile used should be attached to an IAM Role with the enough level of grants to make the cluster work.

By default if you don't configure an iam block in your worker/controller  kube-aws would manage the role for you.

if you set iam.managedIamRoleName, the IAM Role will be created with the name {AWS::Region}-$managedIamRoleName.

if you set iam.existingInstanceProfile, kube-aws will use that instance profile.

if you set both, kube-aws would fail as these options are incompatible.

```
#  #iam:
#       # This will create a named IAM Role managed by kube-aws with the name {AWS::Region}-$managedIamRoleName.
#       # It will have attached a customer Managed Policy that you can modify afterwards if you need more permissions for your cluster.
#       # Be careful with the Statements you modify because an update could overwrite your own.
#       # the Statements included in the ManagedPolicy are the minimun ones required for the Controllers to run.
#       # managedIamRoleName: "yourManagedRole"
#       # ATTENTION: if you decide to use an existing InstanceProfile you need to ensure it has the needed permissions for kube-aws to run.
#       # kube-aws will not manage and never manage this InstanceProfile. IAM Roles will not be created by kube-aws in this case.
#       # existingInstanceProfile: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILEID"
```

probably we can expand this to define custom iam policies in cluster.yaml, i think this model brings flexibility without confusing cluster.yaml more, but i'm happy to hear your thoughts.

i will be very glad if this is merged, @jpb (this is related to your #556 work), @c-knowles this is implementing customer managed policies.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -185,7 +185,26 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #  # ATTENTION: Consider limiting number of characters in clusterName and managedIamRoleName to avoid the resulting IAM
 #  # role name's length from exceeding the AWS limit: 64
 #  # See https://github.com/kubernetes-incubator/kube-aws/issues/347
-#  managedIamRoleName: "yourManagedRole"
+#  # if you omit this block kube-aws would create an IAM Role and a customer managed policy with a random name.
+#  #iam:
+#    # role:
+#       # This will create a named IAM Role managed by kube-aws with the name {AWS::Region}-$managedIamRoleName.
+#       # It will have attached a customer Managed Policy that you can modify afterwards if you need more permissions for your cluster.
+#       # Be careful with the Statements you modify because an update could overwrite your own.
+#       # the Statements included in the ManagedPolicy are the minimun ones required for the Controllers to run.
+#       # name: "yourManagedRole"
+#       # if you set managedPolicies here it will be attached in addition to the created managedPolicy in kube-aws for the cluster.
+#       # CAUTION: if you attach a more restrictive policy in some resources (i.e ec2:* Deny) you can make kube-aws fail.
+#       # managedPolicies:
+#       # -  arn: "arn:aws:iam::aws:policy/AdministratorAccess"
+#       # -  arn: "arn:aws:iam::YOURACCOUNTID:policy/YOURPOLICYNAME"
+
+#       # if you set an InstanceProfile kube-aws will NOT create any IAM Role and will use the configured instanceProfile.
+#       # CAUTION: you must ensure that the IAM Role linked to the listed InstanceProfile has enough permissions to ensure kube-aws to run.
+#       # if you dont know which permissions are required is recommended to create a cluster with a managed role.
+#       # instanceProfile:
+#       #   arn: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILENAME"
+
 #  # If omitted, public subnets are created by kube-aws and used for controller nodes
 #  subnets:
 #    # References subnets defined under the top-level `subnets` key by their names
@@ -282,6 +301,25 @@ worker:
 #      # If not specified, defaults to `worker.apiEndpointName`
 #      apiEndpointName: versionedPublic
 #
+#     # if you omit this block kube-aws would create an IAM Role and a customer managed policy with a random name.
+#      #iam:
+#       # role:
+#       # This will create a named IAM Role managed by kube-aws with the name {AWS::Region}-$managedIamRoleName.
+#       # It will have attached a customer Managed Policy that you can modify afterwards if you need more permissions for your cluster.
+#       # Be careful with the Statements you modify because an update could overwrite your own.
+#       # the Statements included in the ManagedPolicy are the minimun ones required for the Controllers to run.
+#       #   name: "yourManagedRole"
+#       # if you set managedPolicies here it will be attached in addition to the created managedPolicy in kube-aws for the cluster.
+#       # CAUTION: if you attach a more restrictive policy in some resources (i.e ec2:* Deny) you can make kube-aws fail.
+#       #   managedPolicies:
+#       #     - arn: "arn:aws:iam::aws:policy/AdministratorAccess"
+#       #     - arn: "arn:aws:iam::YOURACCOUNTID:policy/YOURPOLICYNAME"
+
+#       # if you set an InstanceProfile kube-aws will NOT create any IAM Role and will use the configured instanceProfile.
+#       # CAUTION: you must ensure that the IAM Role linked to the listed InstanceProfile has enough permissions to ensure kube-aws to run.
+#       # if you dont know which permissions are required is recommended to create a cluster with a managed role.
+#       # instanceProfile:
+#       #   arn: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILENAME"
 #      # Configuration for external managed ALBs Target Groups for worker nodes
 #      targetGroup:
 #        enabled: true
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -130,6 +130,7 @@
       },
       "DependsOn": ["{{$.Etcd.LogicalName}}{{minus $.Etcd.Count 1}}"]
     },
+    {{ if not .Controller.IAMConfig.InstanceProfile.Arn }}
     "IAMInstanceProfileController": {
       "Properties": {
         "Path": "/",
@@ -141,6 +142,7 @@
       },
       "Type": "AWS::IAM::InstanceProfile"
     },
+    {{end}}
     "IAMInstanceProfileEtcd": {
       "Properties": {
         "Path": "/",
@@ -152,32 +154,15 @@
       },
       "Type": "AWS::IAM::InstanceProfile"
     },
-    "IAMRoleController": {
-      "Properties": {
-        "AssumeRolePolicyDocument": {
+    {{ if not .Controller.IAMConfig.InstanceProfile.Arn }}
+    "IAMManagedPolicyController" : {
+      "Type" : "AWS::IAM::ManagedPolicy",
+      "Properties" : {
+        "Description" : "Policy for managing kube-aws k8s controllers",
+        "Path" : "/",
+        "PolicyDocument" :   {
+          "Version":"2012-10-17",
           "Statement": [
-            {
-              "Action": [
-                "sts:AssumeRole"
-              ],
-              "Effect": "Allow",
-              "Principal": {
-                "Service": [
-                  "ec2.{{.Region.PublicDomainName}}"
-                ]
-              }
-            }
-          ],
-          "Version": "2012-10-17"
-        },
-        "Path": "/",
-        {{if .Controller.ManagedIamRoleName }}
-        "RoleName":  {"Fn::Join": ["",[{"Ref": "AWS::StackName"},"-",{"Ref": "AWS::Region"},"-","{{.Controller.ManagedIamRoleName}}"]]},
-        {{end}}
-        "Policies": [
-          {
-            "PolicyDocument": {
-              "Statement": [
                 {
                   "Action": "ec2:*",
                   "Effect": "Allow",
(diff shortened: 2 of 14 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -185,7 +185,16 @@ kmsKeyArn: "{{.KMSKeyARN}}"
...
+#  # if you omit this block kube-aws would create an IAM Role and a customer managed policy with a random name.
+#  #iam:
+#       # This will create a named IAM Role managed by kube-aws with the name {AWS::Region}-$managedIamRoleName.
+#       # It will have attached a customer Managed Policy that you can modify afterwards if you need more permissions for your cluster.
+#       # Be careful with the Statements you modify because an update could overwrite your own.
+#       # the Statements included in the ManagedPolicy are the minimun ones required for the Controllers to run.
+#       # managedIamRoleName: "yourManagedRole"
+#       # ATTENTION: if you decide to use an existing InstanceProfile you need to ensure it has the needed permissions for kube-aws to run.
+#       # kube-aws will not manage and never manage this InstanceProfile. IAM Roles will not be created by kube-aws in this case.
+#       # existingInstanceProfile: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILEID"
```
mumoshu: I guess `iam` should be `iamInstanceProfile` so that it is even more clearer that it is used to customize `IamInstanceProfile` associated to nodes.

If we go that way, `iam.existingInstanceProfile` could be just `iamInstanceProfile.arn`. It will correspond more to e.g. `subnets[].id` or `apiEndpoints[].loadBalancer.id` which is there for reusing existing AWS resources.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -185,7 +185,16 @@ kmsKeyArn: "{{.KMSKeyARN}}"
...
 #  # role name's length from exceeding the AWS limit: 64
 #  # See https://github.com/kubernetes-incubator/kube-aws/issues/347
-#  managedIamRoleName: "yourManagedRole"
+#  # if you omit this block kube-aws would create an IAM Role and a customer managed policy with a random name.
+#  #iam:
+#       # This will create a named IAM Role managed by kube-aws with the name {AWS::Region}-$managedIamRoleName.
+#       # It will have attached a customer Managed Policy that you can modify afterwards if you need more permissions for your cluster.
+#       # Be careful with the Statements you modify because an update could overwrite your own.
+#       # the Statements included in the ManagedPolicy are the minimun ones required for the Controllers to run.
+#       # managedIamRoleName: "yourManagedRole"
```
mumoshu: If we go [this](https://github.com/kubernetes-incubator/kube-aws/pull/607/files#r113839166) way, `iam.managedIamRoleName` could be `iamInstanceProfile.roleName` which will correspond more to the `RoleName` key in Cfn's IAMRole resources.

core/controlplane/config/templates/stack-template.json
```diff
@@ -257,15 +242,37 @@
...
+              }
+            }
+          ],
+          "Version": "2012-10-17"
+        },
+        "Path": "/",
+        {{if .Controller.IAMConfig.ManagedIamRoleName }}
+        "RoleName":  {"Fn::Join": ["",[{"Ref": "AWS::Region"},"-","{{.Controller.IAMConfig.ManagedIamRoleName}}"]]},
+        {{end}}
+        "ManagedPolicyArns": [ {"Ref": "IAMManagedPolicyController"} ]
```
mumoshu: Not sure it should be addressed in this PR or not but..
I'd appreciate it if I could have `iamInstanceProfile.managedPolicyArns` to add "additional managed policy arns" to the `ManagedPolicyArns` array 😄 
Please see https://github.com/kubernetes-incubator/kube-aws/issues/340
cc @c-knowles

mumoshu: And I just realized that you addressed it in your description of the PR

> It also includes a brief improvement for IAM Roles managed by kube-aws supporting customer managed policies (ref #340)

However it doesn't seem to be implemented yet?

mumoshu:
@Fsero Thanks for your efforts here!

I did add some comments but

> i think this model brings flexibility without confusing cluster.yaml more

Definitely.
Great job 👍

core/controlplane/config/templates/stack-template.json
Fsero: i can change the config block, so instead of being

```
iam:
   managedIamRoleName
   existingInstanceProfile
```

could be:

```
   iam:
       role: 
          iamRoleName: "your role"
          managedPolicyArns:
             - "arn:aws:iam:YOURACCOUNTID/policy/POLICYID"
      instanceProfile:
          arn:  "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILEID"
```
that would be better @mumoshu ?

Regarding #340 i now realize is half implemented yes, if we agree on that syntax i can adapt the PR.

Thanks for your comments @mumoshu :smiley:

core/controlplane/config/templates/cluster.yaml
Fsero: managedIamRoleName changes an IAM Role not the attached IAM Instance Profile, so renaming it to iamInstanceProfile.roleName might be confusing to people, but i don't have hard feelings about it if that key name looks more appealing, i will rename it.

mumoshu: @Fsero I think so too. Then, how about `iamInstanceProfile.role.name`?
My idea is that it will be explicit about:
* `iamInstanceProfile` is used to customize what the instance profile is attached to nodes
* `iamInstanceProfile.role` is used to customize what the role is associated to the instance profile
* `iamInstanceProfile.role.name` is used to customize the name of the role associated to the instance profile

// I neither have hard feelings but just thought a bit of discussions would result in less explanations required to our users afterwards!

core/controlplane/config/templates/cluster.yaml
mumoshu: See my [comment](https://github.com/kubernetes-incubator/kube-aws/pull/607#discussion_r113850051) above. I'm now suggesting `iamInstanceProfile.role.name`

core/controlplane/config/templates/stack-template.json
mumoshu: Sounds great. I'm amost convinced to your idea!

Nit but:
* Would you mind renaming `iamRoleName` to just `name` so that the whole path would look like `iam.role.name` rather than `iam.role.iamRoleName`, which results in less duplication?
* Which of them do you think clearer, `iam.instanceProfile.role` or `iam.role`? My motivation to nest `role` under `instanceProfile` is based on the fact that the role is associated to the instance profile. However I'd really like to hear your idea.

core/controlplane/config/templates/cluster.yaml
mumoshu: See my [comment](https://github.com/kubernetes-incubator/kube-aws/pull/607#discussion_r113850904) below.
I'm like your idea of `iam.instanceProfile`.

core/controlplane/config/templates/stack-template.json
Fsero: i don't mind at all, i like `iam.role.name`. Regarding your second point,

between

```
iam:
     role:
           name: "yourRoleName"
           managedPolicyArns:
                - "arn:aws:iam:YOURACCOUNTID/policy/POLICYID"
     instanceProfile:
          arn: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILEID"
```
and 
```
iam:
     instanceProfile:
        role:
           name: "yourRoleName"
           managedPolicyArns:
                - "arn:aws:iam:YOURACCOUNTID/policy/POLICYID"
        arn: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILEID"
```
 there is a link between the role and the instance profile i like the first slightly more because the option is more clear either you go with a role or either you set an instance profile but you cannot do both. 

That said, both can work, so if you feel that the second makes more sense i'll do that.

Thanks @mumoshu :bowing_man:

mumoshu: @Fsero Thanks for the detailed reply :bow:

> i like the first slightly more because the option is more clear either you go with a role or either you set an instance profile but you cannot do both.

It does make sense! Now, I like the first more, too. Let's go ahead with it 👍

Fsero:
@mumoshu i've implemented the proposed changes in syntax for cluster.yaml and included the features from #340. 

cluster with existing instance profile
selection_056
cluster with existing instance profile detail from workers
selection_057
cluster with fixed role name and managed policies
selection_058
cluster with fixed role name and managed policies detail from workers
selection_059

i did some validation to prevent mistakes writing the cluster.yaml, i hope this is enough to got it merged :)

let me know wdyt @mumoshu

mumoshu:
@Fsero Really an awesome work! 👍 
LGTM overall and this will be merged after v0.9.6 is cut shortly.
I may add nits but they won't prevent this from merging 😄

model/node_pool_config.go
```diff
@@ -99,6 +101,13 @@ func (c NodePoolConfig) Valid() error {
 		fmt.Println(`WARNING: instance types "t2.nano" and "t2.micro" are not recommended. See https://github.com/kubernetes-incubator/kube-aws/issues/258 for more information`)
 	}
 
+	if c.IAMConfig.InstanceProfile.Arn != "" && (c.IAMConfig.Role.Name != "" || c.DeprecatedNodePoolManagedIamRoleName != "") {
+		return errors.New("Incompatible options either kube-aws attachs instances to an existing InstanceProfile or manages an IAM Role")
```
mumoshu: nit:`s/attachs/attaches/

mumoshu: nit: "incompatible options: kube-aws either attaches ~~ or manages ~~" may be easier to read/consistent with other error messages?

core/controlplane/config/templates/stack-template.json
```diff
@@ -716,9 +728,13 @@
             }
           }
         ],
+        {{if .Controller.IAMConfig.InstanceProfile.Arn }}
+        "IamInstanceProfile": "{{.Controller.IAMConfig.InstanceProfile.Arn}}",
+        {{else}}
         "IamInstanceProfile": {
-          "Ref": "IAMInstanceProfileController"
+          "Ref": "IAMInstanceProfileController"
```
mumoshu: nit: redundant whitespace at the end of line?

Fsero: fixed, yep it was extra

model/node_pool_config.go
Fsero: fixed, setting up an error message more similar to others.

model/iamconfig.go
```diff
@@ -0,0 +1,44 @@
...
+		if !managedPolicyRegexp.MatchString(policy.Arn) {
+			return fmt.Errorf("invalid managed policy arn, your managed policy must match this (=arn:aws:iam::(YOURACCOUNTID|aws):policy/POLICYNAME), provided this (%s)", policy.Arn)
+		}
+	}
+	if c.InstanceProfile.Arn != "" {
+		if !instanceProfileRegexp.MatchString(c.InstanceProfile.Arn) {
+			return fmt.Errorf("invalid instance profile, your instance profile must match (=arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILENAME), provided (%s)", c.InstanceProfile.Arn)
+		}
+
+	}
```
mumoshu: nit: We'd better emit a validation error also when both `instanceProfile.arn` and `managePolicies[].arn`s are specified.

mumoshu: Seems like we already had it only for controller nodes above https://github.com/kubernetes-incubator/kube-aws/pull/607/files#diff-75f5c0a1828df862096ed45c755c2990R73

Fsero: you are right, i'd forgot about it. Fixed now

mumoshu:
Wow, thanks for the quick turnaround @Fsero!
LGTM and I'm now waiting for the travis build to finish.

mumoshu:
@Fsero Merged. Sry if you've force-pushed one more commit after I've confirmed the build status and merged the PR. If you had further changes, I'd appreciate it if you could open up an another PR!

Thanks again for your efforts 🍻

Fsero:
Thanks for your detailed review @mumoshu and for the efforts for making kube-aws great. See you in another PR :beers:

Issue #584, opened 2017-04-24
configmap and serviceaccount for kube-dns


mumoshu:
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/542
Although we've upgraded kube-dns to the latest 1.14.1 via #582, we're still missing (1)configmaps mounted as volumes and (2) serviceaccount for kube-dns.
Survey the upstream setups for those and adapt if necessary.
cc @c-knowles

c-knowles:
What do we lose out on without this? At least RBAC won't work right?

danielfm:
I'm fixing a regression bug in TLS bootstrapping, and since TLS bootstrapping should go along with RBAC, I'm also attacking this in the same PR.

mgreenwald-pm:
@mumoshu Does this permit usage of this feature?

http://blog.kubernetes.io/2017/04/configuring-private-dns-zones-upstream-nameservers-kubernetes.html

mumoshu:
@mgreenwald-pm Yes, it should 👍 Just edit the kube-dns confgimap to add settings mentioned in the article.

mgreenwald-pm:
@mumoshu Yes, thank you and it worked!

PR #612, opened 2017-04-28
Convert kube-proxy to DaemonSet + various RBAC fixes

### LINKED ISSUE #588: Warning in kube-proxy due to unspecified cli parameter
I just noticed this log entry in my kube-proxy pods:

```
W0417 15:16:05.916727       1 proxier.go:254] clusterCIDR not specified, unable to distinguish between internal and external traffic
```

It happens because we don't specify the `cluster-cidr` flag to kube-proxy. The question is: should I be worried about this? Is there anything preventing us to specify this flag?

danielfm:
These are the changes implemented by this PR:

- Synchronizes the kube-dns config with changes in [upstream](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns)
- Converts kube-proxy to a DaemonSet. Since the manifest sets a toleration to master taints, the kube-proxy is scheduled to every node in the cluster. Also, the kube-proxy pods now connect to the API server (respecting each node pools' API endpoint URL) with a dedicated ServiceAccount associated with the `system:node-proxier` group
- Sets the `--cluster-cidr` cli parameter on kube-proxy pods to the value of `{{.PodCIDR}}` in order to avoid the warning message shown in #588
- Makes the `kube-worker` user behave like a member of `system:nodes` group (needed when TLS bootstrapping is disabled) by associating this user with the `system:node` cluster role. This effectively removes the association between the `kube-worker` user and the `cluster-admin` role, thus improving security by only granting the permissions nodes are supposed to have
- Prefixes custom roles and bindings with `kube-aws:*`, following the conventions of other tools, like `kubeadm`
- Adds support for `NoExecute` taint effect introduced in Kubernetes 1.6
- Adds more comprehensive taint validation that looks for duplicate key/effect pairs in controller/node pools; duplicate key/effect taint pairs results in broken clusters

TODO

- [x] Address review comments
- [x] Test: node labeling, draining, taints, etc
- [x] Test: create a cluster with both RBAC and TLS bootstrapping enabled
- [x] Test: create a test cluster with RBAC enabled, but TLS bootstrapping disabled
- [x] Test: create a test cluster with both RBAC and TLS bootstrapping disabled

Closes #588 and hopefully #584.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1108,8 +1108,8 @@ write_files:
           name: cluster-admin
           apiGroup: rbac.authorization.k8s.io
 
-  # Makes kube-worker user behave like a regular member of system:node group, needed when
-  # TLS bootstrapping is disabled
+  # Makes kube-worker user behave like a regular member of system:nodes group,
+  # needed when TLS bootstrapping is disabled
   - path: /srv/kubernetes/rbac/cluster-role-bindings/node.yaml
     content: |
         kind: ClusterRoleBinding
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1135,7 +1135,7 @@ write_files:
             name: system:nodes
         roleRef:
           kind: ClusterRole
-          name: system:proxier
+          name: system:node-proxier
           apiGroup: rbac.authorization.k8s.io
 
 {{ if .Experimental.TLSBootstrap.Enabled }}
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -602,7 +602,7 @@ write_files:
 
       kubectl apply -f "${mfdir}/cluster-roles/bootstrapped-node.yaml"
 
-      for manifest in {system-worker,kube-worker,kube-admin,bootstrapped-node}; do
+      for manifest in {system-worker,kube-worker,kube-admin,bootstrapped-node,bootstrapped-node-proxier}; do
           kubectl apply -f "${mfdir}/cluster-role-bindings/$manifest.yaml"
       done
 
@@ -1119,6 +1119,23 @@ write_files:
         roleRef:
           kind: ClusterRole
           name: bootstrapped-node
+          apiGroup: rbac.authorization.k8s.io
+
+  # Since the kube-proxy is run via the pod manifest mechanism, it's not
+  # possible to inject a ServiceAccount into the kube-proxy pods, so kube-proxy
+  # pods access the API using the node identity
+  - path: /srv/kubernetes/rbac/cluster-role-bindings/bootstrapped-node-proxier.yaml
+    content: |
+        kind: ClusterRoleBinding
+        apiVersion: rbac.authorization.k8s.io/v1beta1
+        metadata:
+          name: bootstrapped-node-proxier
+        subjects:
+          - kind: Group
+            name: system:nodes
+        roleRef:
+          kind: ClusterRole
+          name: system:proxier
           apiGroup: rbac.authorization.k8s.io
 
 {{ if .Experimental.TLSBootstrap.Enabled }}
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -612,8 +612,8 @@ write_files:
       done
 
       {{ if .Experimental.TLSBootstrap.Enabled }}
-      kubectl create -f "${mfdir}/cluster-roles/kubelet-bootstrap.yaml" | echo 'Failed to create the cluster role named "kubelet-bootstrap". Skipping'
-      kubectl create -f "${mfdir}/cluster-role-bindings/kubelet-bootstrap.yaml" | echo 'Failed to create the cluster role binding "kubelet-bootstrap". Skipping'
+      kubectl create -f "${mfdir}/cluster-roles/node-bootstrap.yaml" | echo 'Failed to create the cluster role named "kube-aws:node-bootstrap". Skipping'
+      kubectl create -f "${mfdir}/cluster-role-bindings/node-bootstrap.yaml" | echo 'Failed to create the cluster role binding "kube-aws:node-bootstrap". Skipping'
       {{ end }}
       {{ end }}
 
@@ -1159,12 +1159,12 @@ write_files:
 
 {{ if .Experimental.TLSBootstrap.Enabled }}
   # Only allows certificate signing requests to be performed with the bootstrap token
-  - path: /srv/kubernetes/rbac/cluster-roles/kubelet-bootstrap.yaml
+  - path: /srv/kubernetes/rbac/cluster-roles/node-bootstrap.yaml
     content: |
         kind: ClusterRole
         apiVersion: rbac.authorization.k8s.io/v1beta1
         metadata:
-          name: kube-aws:kubelet-bootstrap
+          name: kube-aws:node-bootstrap
         rules:
           - apiGroups:
               - '*'
@@ -1176,12 +1176,12 @@ write_files:
             - list
             - watch
 
-  - path: /srv/kubernetes/rbac/cluster-role-bindings/kubelet-bootstrap.yaml
+  - path: /srv/kubernetes/rbac/cluster-role-bindings/node-bootstrap.yaml
     content: |
         kind: ClusterRoleBinding
         apiVersion: rbac.authorization.k8s.io/v1beta1
         metadata:
-          name: kube-aws:kubelet-bootstrap
+          name: kube-aws:node-bootstrap
         subjects:
           - kind: Group
             namespace: '*'
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1232,6 +1232,16 @@ write_files:
                 effect: "NoSchedule"
               - key: "CriticalAddonsOnly"
                 operator: "Exists"
+              # Since kube-proxy needs to run on every node, it must tolerate
+              # all taints listed in cluster configuration file
+              {{- range $taint := .Experimental.Taints.UniqKeys }}
+              - key: {{ .Key }}
+                operator: "Exists"
+              {{- end }}
+              {{- range $taint := .NodeTaints.UniqKeys }}
+              - key: {{ .Key }}
+                operator: "Exists"
+              {{- end }}
               volumes:
               - name: kube
                 hostPath:
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1122,8 +1122,8 @@ write_files:
           apiGroup: rbac.authorization.k8s.io
 
   # Since the kube-proxy is run via the pod manifest mechanism, it's not
-  # possible to inject a ServiceAccount into the kube-proxy pods, so kube-proxy
-  # pods access the API using the node identity
+  # possible to inject a ServiceAccount into the kube-proxy pods, so they access
+  # the API using the node identity
   - path: /srv/kubernetes/rbac/cluster-role-bindings/bootstrapped-node-proxier.yaml
     content: |
         kind: ClusterRoleBinding
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1040,6 +1040,12 @@ write_files:
           annotations:
             rbac.authorization.kubernetes.io/autoupdate: "false"
         rules:
+          - apiGroups:
+            - extensions
+            resources:
+            - daemonsets
+            verbs:
+            - get
           - apiGroups: ["*"]
             resources:
             - nodes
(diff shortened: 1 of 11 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1056,65 +1070,82 @@ write_files:
...
         roleRef:
           kind: ClusterRole
-          name: cluster-admin
+          name: kube-aws:node
           apiGroup: rbac.authorization.k8s.io
 
-  # Associates the add-on group `bootstrapped-node` to all nodes, so that extra kube-aws
-  # features like node draining work as expected
-  - path: /srv/kubernetes/rbac/cluster-role-bindings/bootstrapped-node.yaml
+  # Since the kube-proxy is run via the pod manifest mechanism, it's not
```
redbaron: No for this PR, but can't we run them via DaemonSets then? I don't think that working kube-proxy is a prerequisite for kube to be able to deploy DaemonSets

core/nodepool/config/config.go
```diff
@@ -39,6 +39,7 @@ type ProvidedConfig struct {
 	cfg.KubeClusterSettings `yaml:",inline"`
 	WorkerNodePoolConfig    `yaml:",inline"`
 	DeploymentSettings      `yaml:",inline"`
+	cfg.FlannelSettings     `yaml:",inline"`
```
redbaron: where is it used?

danielfm: In `userdata/cloud-config-worker`.

core/controlplane/config/templates/cloud-config-controller
danielfm: > I don't think that working kube-proxy is a prerequisite for kube to be able to deploy DaemonSets

No, it's not.

> No for this PR, but can't we run them via DaemonSets then?

In order to create kube-proxy as a DaemonSet, it would have to be created in controller's `/opt/bin/install-kube-system`, which is not a problem. In fact, I was going to convert the kube-proxy to a DaemonSet in this PR, but since kube-aws now supports using different API endpoint URLs for different node pools, there's no way (I think) to make the DaemonSet-based kube-proxy pods to access different API endpoint URLS in different node pools.

danielfm:
Ready for review.

redbaron:
how bad is it when upgrading existing cluster with non-daemonset kube-proxy already running?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1136,60 +1176,73 @@ write_files:
...
-              name: dbus
-              readOnly: false
-          volumes:
-          - hostPath:
-              path: /usr/share/ca-certificates
-            name: ssl-certs-host
-          - hostPath:
-              path: /var/run/dbus
-            name: dbus
+          selector:
```
redbaron: can be omitted, default selectors will be generated instead

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1136,60 +1176,73 @@ write_files:
...
+            matchLabels:
+              k8s-app: kube-proxy
+          template:
```
redbaron: `updateStrategy: RollingUpdate` ?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1136,60 +1176,73 @@ write_files:
...
+              containers:
+              - name: kube-proxy
+                image: {{.HyperkubeImage.RepoWithTag}}
+                imagePullPolicy: IfNotPresent
+                command:
+                - /usr/local/bin/kube-proxy
+                - --cluster-cidr={{.PodCIDR}}
+                - --kubeconfig=/etc/kubernetes/proxy-kubeconfig.yaml
+                securityContext:
+                  privileged: true
```
redbaron: wouldn't CAP_NET_ADMIN be enough?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1136,60 +1176,73 @@ write_files:
...
+                volumeMounts:
+                - mountPath: /etc/kubernetes
+                  name: kube
+                - mountPath: /var/run/dbus
```
redbaron: got puzzled why kube-proxy needs dbus. checked code, turned out it is used to integrate with *firewalld* which is not used on CoreOS so probably can be safely removed

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1136,60 +1176,73 @@ write_files:
...
+                image: {{.HyperkubeImage.RepoWithTag}}
+                imagePullPolicy: IfNotPresent
+                command:
+                - /usr/local/bin/kube-proxy
+                - --cluster-cidr={{.PodCIDR}}
+                - --kubeconfig=/etc/kubernetes/proxy-kubeconfig.yaml
+                securityContext:
+                  privileged: true
+                volumeMounts:
+                - mountPath: /etc/kubernetes
```
redbaron: there are certs in this directory which kube-proxy doesn't need. can `proxy-kubeconfig.yaml` be generated somewhere in `/var/run` and only that directory is mount-binded

redbaron:
looking forward to DeaonSet scheduller and  controller-manager! :)

core/controlplane/config/templates/cloud-config-controller
danielfm: This is a [new feature](https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/) in Kubernetes 1.6.

core/controlplane/config/templates/cloud-config-controller
danielfm: I don't know, got the manifest from [here](https://github.com/kubernetes/kubernetes/blob/194a06917f07aa11567c8ba5b2f7ae869ad031e8/cmd/kubeadm/app/phases/addons/manifests.go#L51).

What is CAP_NET_ADMIN?

core/controlplane/config/templates/cloud-config-controller
redbaron: I know, I was suggesting to add it :)

core/controlplane/config/templates/cloud-config-controller
danielfm: I could move these to `/etc/kubernetes/kubeconfig/`, which would contain the kubeconfig files for the kubelet and kube-proxy. 👍

core/controlplane/config/templates/cloud-config-controller
redbaron: it is a linux capability which gives control over network subsystem. Basically instead of running fully privileged container we can run it with:

```
securityContext:
      capabilities:
        add:
        - CAP_NET_ADMIN
```

and should be enough

core/controlplane/config/templates/cloud-config-controller
danielfm: Oh, I thought it was already there! I'll do it. 👍

core/controlplane/config/templates/cloud-config-controller
danielfm: I'll give it a try.

danielfm:
> how bad is it when upgrading existing cluster with non-daemonset kube-proxy already running?

Need to test it, but I guess the only problem would be that the iptables rules will not be updated for the time between the old pod is deleted and the new one is up and running.

redbaron:
> Need to test it, but I guess the only problem would be that the iptables rules will not be updated for the time between the old pod is deleted and the new one is up and running.

Controllers are updated first, they install DaemonSet for the first time, last worker node is rebooted ~1hour later. All this time one that worker node new kube-proxy will coexist with old kube-proxy, right?

core/controlplane/config/templates/cloud-config-controller
danielfm: @redbaron couldn't make it work with this. This is the error logged by kube-proxy pods:

```
F0429 20:47:38.905145       1 server.go:244] Unable to create proxier: can't set sysctl net/ipv4/conf/all/route_localnet: open /proc/sys/net/ipv4/conf/all/route_localnet: read-only file system
```

redbaron: so it is adjusting sysctl params too, that requires `--privileged` flag unfortunately. :( so it couldn't be replaced

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -594,17 +605,15 @@ write_files:
          echo 'Failed to create the cluster role named "cluster-admin". It is ok if you are using k8s newer than 1.5'
       fi
 
-      for manifest in {bootstrapped-node,}; do
-          kubectl apply -f "${mfdir}/cluster-roles/$manifest.yaml"
-      done
+      kubectl apply -f "${mfdir}/cluster-roles/node-extensions.yaml"
```
redbaron: will it be beneficial to install RBAC roles before installing kube-dns and kube-proxy?

cheungpat:
I have a question about converting kube-proxy to a Daemonset: If I create a node pool with taints, does it mean kube-proxy will not be scheduled on the tainted nodes?

redbaron:
> I have a question about converting kube-proxy to a Daemonset: If I create a node pool with taints, does it mean kube-proxy will not be scheduled on the tainted nodes?

Valid point. We'd need to make kube-proxy to tolerate all possible taints specified in cluster.yaml.

core/controlplane/config/templates/cloud-config-controller
danielfm: I believe it won't make a difference.
The kube-dns and kube-proxy pods will only be scheduled when the first worker node comes up; at that point, all required roles and bindings will already be available.

The kube-proxy running in controllers access the API server in localhost, so it bypasses the auth* mechanism entirely

danielfm:
@cheungpat Thanks for bringing this up, I didn't think of this. I'll come up with something.

danielfm:
@redbaron 

> Controllers are updated first, they install DaemonSet for the first time, last worker node is rebooted ~1hour later. All this time one that worker node new kube-proxy will coexist with old kube-proxy, right?

I think so, but don't know what are the implications of having multiple kube-proxy pods running at the same time.

core/root/config/config.go
```diff
@@ -88,6 +88,14 @@ func ConfigFromBytes(data []byte) (*Config, error) {
 	}
 
 	for i, np := range nodePools {
+		if err := np.Experimental.Taints.Valid(); err != nil {
+			return nil, fmt.Errorf("invalid taints for node pool at index %d: %v", i, err)
+		}
+
+		// Controllers must be aware of node pools' taints so they know which tolerations
+		// to add to pods that are supposed to run on every node (ex: kube-proxy)
+		cpCluster.ComputedNodeSettings.NodeTaints = append(cpCluster.ComputedNodeSettings.NodeTaints, np.Experimental.Taints...)
```
mumoshu: This sounds great as long as users don't taint nodes after they're created.
Should we warn users not to run `kubectl taint` after `kube-aws up` or they may suddenly lose kube-proxy?
Or can we keep the support for kube-proxy deployment via static pods as before and mark daemonset-based deployment experimental?

c-knowles:
Can I ask why we wish to swap kube proxy to Daemon Set? Using a service account as per the description seems a disjointed change we could have without changing how proxy is deployed (which affects other useability like taints). Maybe there's a reason we want to swap. The last times I've run other setup tools including GCP clusters they didn't seem to run proxy as a Daemon Set but maybe that's the way it's going? My preference is to stick to a similar setup to upstream.

mumoshu:
@c-knowles AFAIK, there's an [upstream issue for migrating kube-proxy to daemonsets](https://github.com/kubernetes/kubernetes/issues/23225) planned for k8s 1.7. However I myself am not aware of specific reasons if we want to migrate to daemonsets before upstream!
I'm ok as long as it doesn't break for cases like [my above comment](https://github.com/kubernetes-incubator/kube-aws/pull/612#discussion_r114089548)

mumoshu:
> Using a service account as per the description seems a disjointed change we could have without changing how proxy is deployed (which affects other useability like taints).

Agreed :) A separate PR would be indeed better from perspective of a maintainer but I'm ok if daemonset-based kube-proxy just works and doing so in the same PR is easier for you contributors!

core/root/config/config.go
danielfm: From what I've been testing, running kube-proxy as a DaemonSet as of now may indeed not be the best choice after all. If you guys think it's best to leave this to the future, I'm okay with that!

danielfm:
> Can I ask why we wish to swap kube proxy to Daemon Set?

One of use cases for DaemonSets are for running pods in all nodes, which is the case for kube-proxy. Also, [kubeadm](https://github.com/kubernetes/kubernetes/blob/v1.6.2/cmd/kubeadm/app/phases/addons/manifests.go) already deploys kube-proxy as a DaemonSet.

It would also allow us to set up proper permissions by using a dedicated ServiceAccount for kube-proxy, just like documented [here](https://kubernetes.io/docs/admin/authorization/rbac/#core-component-roles); since I was fixing RBAC issues, it made sense to make that jump in the same PR.

> Using a service account as per the description seems a disjointed change we could have without changing how proxy is deployed (which affects other useability like taints).

It's just not possible to use service accounts in pods created via the pod manifest mechanism.

danielfm:
I looks like I got carried away in this PR, so I'll break this into smaller ones and let you guys review and approve/deny each individual change.

c-knowles:
@danielfm ok sure, thank you for your explanation. It makes sense and if upstream are switching then I think we should also switch.

PR #616, opened 2017-05-01
Improve taints validation

danielfm:
This PR prevents users from creating clusters with invalid taints in both the controller and node pools.

These are the rules:

- Key must be provided
- Effect must be: `NoSchedule`, `PreferNoSchedule`, and `NoExecute`
- Key/effect pair must be unique


mumoshu:
@danielfm Wow, this is a great improvement 👍 Thanks for your continuous efforts :bow:

PR #618, opened 2017-05-02
RBAC setup improvements

danielfm:
- Prefixes custom roles and bindings with `kube-aws:*`, following the conventions of other tools, like `kubeadm`
- Makes the `kube-worker` user behave like a member of `system:nodes` group by associating this user with the `system:node` cluster role. This effectively removes the association between the `kube-worker` user and the `cluster-admin` role, thus improving security by only granting the permissions nodes are supposed to have
- Associates the `kube-worker` user and the `system:nodes` group to `system:node-proxier` so that the kube-proxy works with TLS bootstrapping on and off. This is needed since nodes no longer have admin permissions (see previous bullet)

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -589,22 +589,16 @@ write_files:
 
       {{if .Experimental.Plugins.Rbac.Enabled}}
       mfdir=/srv/kubernetes/rbac
-      # k8s > 1.5 already have this clusterRole, so we accept http.code != 200
-      if ! kubectl create -f "${mfdir}/cluster-roles/cluster-admin.yaml"; then
-         echo 'Failed to create the cluster role named "cluster-admin". It is ok if you are using k8s newer than 1.5'
-      fi
 
-      for manifest in {bootstrapped-node,}; do
-          kubectl apply -f "${mfdir}/cluster-roles/$manifest.yaml"
-      done
+      kubectl apply -f "${mfdir}/cluster-roles/node-extensions.yaml"
 
-      for manifest in {system-worker,kube-worker,kube-admin,bootstrapped-node}; do
+      for manifest in {kube-admin,system-worker,node,node-proxier,node-extensions}; do
           kubectl apply -f "${mfdir}/cluster-role-bindings/$manifest.yaml"
       done
 
       {{ if .Experimental.TLSBootstrap.Enabled }}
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1071,18 +1070,24 @@ write_files:
...
+  - path: /srv/kubernetes/rbac/cluster-role-bindings/node-proxier.yaml
     content: |
         kind: ClusterRoleBinding
         apiVersion: rbac.authorization.k8s.io/v1beta1
         metadata:
-          name: kube-worker
+          name: kube-aws:node-proxier
         subjects:
           - kind: User
             name: kube-worker
```
mumoshu: Can we just remove the subject for `kube-worker` here? If I understand correctly, the `system:nodes` group, which the `kube-worker` user is already associated in [the above](https://github.com/kubernetes-incubator/kube-aws/pull/618/files#diff-ef25536c536667a40b993d4d24ab7567R1012),  already seem to have all the necessary roles assigned.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1101,30 +1106,32 @@ write_files:
...
+  - path: /srv/kubernetes/rbac/cluster-role-bindings/node-extensions.yaml
     content: |
         kind: ClusterRoleBinding
         apiVersion: rbac.authorization.k8s.io/v1beta1
         metadata:
-          name: bootstrapped-node
+          name: kube-aws:node-extensions
         subjects:
+          - kind: User
+            name: kube-worker
```
mumoshu: Same as [the above comment](https://github.com/kubernetes-incubator/kube-aws/pull/618/files#r115162048)

core/controlplane/config/templates/cloud-config-controller
danielfm: @mumoshu I tried that, but for some reason, it didn't work. My guess is that when you set the `system:nodes` group, you'd have to set the Common Name to `system:node:<node-hostname>` for it to work (at least, this is how the certificates generated via CSR are set).

mumoshu: @danielfm Thanks for the clarification. That's good to know! In the meantime, how about add some comment to note about the work-around that kube-worker user is bound to system:nodes group just for the case kubelet bootstrapping is disabled?

danielfm: Done. Let me know if you want any other changes.

mumoshu:
@danielfm LGTM. Thanks for your continuous efforts to make the kubelet bootstrapping support great 👍

PR #624, opened 2017-05-04
Update default dashboard to 1.6.0

Vrtak-CZ:


mumoshu:
@Vrtak-CZ LGTM. Thanks as always for your support 👍

PR #625, opened 2017-05-04
Fix dashboard version in labels and RC name

Vrtak-CZ:


Vrtak-CZ:
Btw @mumoshu is there any reason for version in RC name for dashboard?

mumoshu:
@Vrtak-CZ Nothing specific on my side - Even [the corresponding kube-dashboard manifest in the upstream repo](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-controller.yaml) doesn't contain version numbers in its names.
I guess we can safely remove the version numbers if you'd (also) like to do so.

Vrtak-CZ:
@mumoshu rebased and removed :-)

mumoshu:
@Vrtak-CZ LGTM. Thanks for your support 👍

PR #151, opened 2016-12-11
cluster-autoscaler support

mumoshu:
CloudFormation stacks created via kube-aws can now be tagged with the following keys with the values containing source of truth for cluster-autoscaler to update their autoscaling target and its min/max size.

* `kube-aws:cluster-autoscaler:logical-name` (Currently fixed to `AutoScaleWorker` which corresponds to the logical name of the autoscaling group for worker nodes defined in `stack-template.json`)
* `kube-aws:cluster-autoscaler:max-size`
* `kube-aws:cluster-autoscaler:min-size`

Adding tags to stacks instead of to autoscaling groups allows us to support autoscaling of spot fleet via cluster-autoscaler, in the future.

Currently, enabling this configuration doesn't automatically deploy cluster-autoscaler itself.
Please read [the original issue](coreos#148) carefully for more information.

ref #148 
depends #150 

This will be rebased once #150 is merged.

PR #629, opened 2017-05-09
Re: cluster-autoscaler support

mumoshu:
In #151, we've introduced the initial, incomplete, almost theoretical support for cluster-autoscaler. The situation has changed considerably since then - Now we can finally complete the cluster-autoscaler support.

## Features

* Automatically scale out a node group by adding node(s) when one or more pods in the group become unschedulable due to insufficient resource.
* Automatically scales in a node group by removing node(s) that is safe to be done so. Basically, a node group is safe to be removed when it is not running a critical k8s component

## kube-aws scope changes

* CA is now deployed automatically
  * More concretely, a k8s deployment for cluster-autoscaler is automatically created on a k8s cluster when cluster-autoscaler addon is enabled in cluster.yaml

## Configuration changes

A valid cluster.yaml for a CA-enabled cluster would now look like:
```yaml
# cluster-autoscaler addon is deployed and enabled to watch node groups so that it can scale in/out node group(s) that is registered as an autoscaling target.
addons:
  clusterAutoscaler:
    enabled: true
worker:
  nodePools:
  - name: scaled
    # Make this node pool an autoscaling target
    autoscaling:
      clusterAutoscaler:
        enabled: true
  - name: notScaled
    # This node pool is not an autoscaling target
```

* The former `experimental.clusterAutoscalerSupport.enabled` is dropped for controller nodes in favor of `addons.clusterAutoscaler.enabled`
* `worker.nodePools[].clusterAutoscaler.minSize` and `maxSize` are dropped in favor of `worker.nodePools[].autoscaling.clustserAutoscaler.enabled` and the auto discovery feature of cluster-autoscaler
* `worker.nodePools[].clusterAutoscalerSupport` is kept as-is, but not necessarily be `true` because when you've enabled `addons.clusterAutoscaler`, kube-aws by default gives enough IAM permissions to only controller nodes and CA is scheduled there.
* This work currently relies on the docker image built from a fork of cluster-autoscaler which supports the automatic node group discovery feature

## cloud-config-controller changes

* Add cluster-autoscaler deployment to be created when the CA addon is enabled

## Go changes

* The former `ClusterAutoscalerImage` is renamed to `ClusterProportionalAutoscalerImage`
* Introduce `ClusterAutoscalerImage`(`clusterAutoscalerImage` in cluster.yaml) for the cluster-autoscaler docker image reference
* `ClusterAutoscalerSupport` is no-op in controller nodes, used only for worker nodes. `Addons.ClusterAutoscaler` is used instead to give controller nodes appropriate IAM permissions and deploy CA to them.
* Most of CA related types and funcs are moved from `core/controlplane/config` to the `model` package

## IAM changes

* `autoscaling:DescribeTags` is allowed in IAM to allow enabling the automatic node group discovery feature of cluster-autoscaler

## Gotchas

Note that a node running cluster-autoscaler or kube-resources-autosave can not be scaled in:

```
09 05:29:21.523426       1 cluster.go:74] Fast evaluation: ip-10-0-0-68.ap-northeast-1.compute.internal for removal
I0509 05:29:21.523467       1 cluster.go:88] Fast evaluation: node ip-10-0-0-68.ap-northeast-1.compute.internal cannot be removed: non-deamons set, non-mirrored, kube-system pod present: cluster-autoscaler-998591511-thcpj
I0509 05:29:21.523479       1 cluster.go:74] Fast evaluation: ip-10-0-0-133.ap-northeast-1.compute.internal for removal
I0509 05:29:21.523488       1 cluster.go:103] Fast evaluation: node ip-10-0-0-133.ap-northeast-1.compute.internal may be removed
I0509 05:29:21.523493       1 cluster.go:74] Fast evaluation: ip-10-0-0-150.ap-northeast-1.compute.internal for removal
I0509 05:29:21.523530       1 cluster.go:88] Fast evaluation: node ip-10-0-0-150.ap-northeast-1.compute.internal cannot be removed: non-deamons set, non-mirrored, kube-system pod present: kube-resources-autosave-2845171460-1cbs5
```

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -197,7 +197,7 @@ coreos:
         --container-runtime={{.ContainerRuntime}} \
         --rkt-path=/usr/bin/rkt \
         --rkt-stage1-image=coreos.com/rkt/stage1-coreos \
-        --node-labels node-role.kubernetes.io/master{{if .Experimental.NodeLabels.Enabled}},{{.Experimental.NodeLabels.String}} \
+        --node-labels node-role.kubernetes.io/master{{if .NodeLabels.Enabled}},{{.NodeLabels.String}} \
         {{end}} \
         --register-with-taints=node.alpha.kubernetes.io/role=master:NoSchedule \
         --allow-privileged=true \
@@ -578,7 +578,7 @@ write_files:
       kubectl apply -f "${mfdir}/kube-dns-sa.yaml"
 
       # Deployments
-      for manifest in {kube-dns-de,kube-dns-autoscaler-de,heapster-de{{ if .KubeResourcesAutosave.Enabled }},kube-resources-autosave{{ end }}}.yaml; do
+      for manifest in {kube-dns-de,kube-dns-autoscaler-de,cluster-autoscaler-de,heapster-de{{ if .KubeResourcesAutosave.Enabled }},kube-resources-autosave{{ end }}}.yaml; do
           kubectl apply -f "${mfdir}/$manifest"
       done
 
@@ -1530,7 +1530,7 @@ write_files:
                 operator: "Exists"
               containers:
               - name: autoscaler
-                image: {{ .ClusterAutoscalerImage.RepoWithTag }}
+                image: {{ .ClusterProportionalAutoscalerImage.RepoWithTag }}
                 resources:
                     requests:
                         cpu: "20m"
@@ -1794,6 +1794,64 @@ write_files:
                     - --container=heapster
                     - --poll-period=300000
                     - --estimator=exponential
+
+  {{if .Addons.ClusterAutoscaler.Enabled}}
+  - path: /srv/kubernetes/manifests/cluster-autoscaler-de.yaml
+    content: |
+        apiVersion: extensions/v1beta1
+        kind: Deployment
+        metadata:
+          name: cluster-autoscaler
+          namespace: kube-system
+          labels:
+            app: cluster-autoscaler
+        spec:
+          replicas: 1
+          selector:
+            matchLabels:
+              app: cluster-autoscaler
+          template:
+            metadata:
+              labels:
+                app: cluster-autoscaler
+            spec:
+              affinity:
+                nodeAffinity:
+                  requiredDuringSchedulingIgnoredDuringExecution:
+                    nodeSelectorTerms:
+                    - matchExpressions:
+                      - key: "kube-aws.coreos.com/cluster-autoscaler-supported"
+                        operator: "In"
+                        values:
+                        - "true"
+              tolerations:
+              - key: "node.alpha.kubernetes.io/role"
+                operator: "Equal"
+                value: "master"
+                effect: "NoSchedule"
+              containers:
+                - image: {{ .ClusterAutoscalerImage.RepoWithTag }}
+                  name: cluster-autoscaler
+                  resources:
+                    limits:
+                      cpu: 100m
+                      memory: 300Mi
+                    requests:
+                      cpu: 100m
+                      memory: 300Mi
+                  command:
+                    - ./cluster-autoscaler
+                    - --v=4
+                    - --stderrthreshold=info
+                    - --cloud-provider=aws
+                    - --skip-nodes-with-local-storage=false
+                    - --expander=least-waste
+                    - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled
+                  env:
+                    - name: AWS_REGION
+                      value: {{.Region}}
+                  imagePullPolicy: "Always"
+  {{end}}
 
   - path: /srv/kubernetes/manifests/heapster-svc.yaml
     content: |
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -203,12 +203,13 @@
                   "Resource": [ "*" ]
                 },
                 {{end}}
-                {{if .Experimental.ClusterAutoscalerSupport.Enabled}}
+                {{if .Addons.ClusterAutoscaler.Enabled}}
                 {
                 "Effect": "Allow",
                   "Action": [
                     "autoscaling:DescribeAutoScalingGroups",
                     "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeTags",
                     "autoscaling:SetDesiredCapacity",
                     "autoscaling:TerminateInstanceInAutoScalingGroup"
                   ],
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -145,6 +145,13 @@
         ],
         "MinSize": "{{.MinCount}}",
         "Tags": [
+          {{if .Autoscaling.ClusterAutoscaler.Enabled}}
+          {
+            "Key": "{{.Autoscaling.ClusterAutoscaler.AutoDiscoveryTagKey}}",
+            "PropagateAtLaunch": "false",
+            "Value": ""
+          },
+          {{end}}
           {
             "Key": "kubernetes.io/cluster/{{ .ClusterName }}",
             "PropagateAtLaunch": "true",
@@ -355,15 +362,25 @@
                   "Resource": [ "*" ]
                 },
                 {{end}}
-                {{if .ClusterAutoscalerSupport.Enabled}}
+                {{if .Addons.ClusterAutoscaler.Enabled}}
                 {
-                "Effect": "Allow",
                   "Action": [
                     "autoscaling:DescribeAutoScalingGroups",
                     "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeTags",
+                  ],
+                  "Effect": "Allow",
+                  "Resource": "*"
+                },
+                {
+                  "Action": [
                     "autoscaling:SetDesiredCapacity",
                     "autoscaling:TerminateInstanceInAutoScalingGroup"
                   ],
+                  "Condition": {
+                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
+                  },
+                  "Effect": "Allow",
                   "Resource": "*"
                 },
                 {{end}}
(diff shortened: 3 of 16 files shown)


mumoshu:
I'm trying to complete the cluster-autoscaler support according to our roadmap for v0.9.7 https://github.com/kubernetes-incubator/kube-aws/blob/master/ROADMAP.md#v097

mumoshu:
This work depends on https://github.com/kubernetes/autoscaler/pull/11

e2e/run
```diff
@@ -156,13 +163,17 @@ customize_cluster_yaml() {
 worker:
   nodePools:
   - name: asg1
+    clusterAutoscalerSupport:
+      enabled: true
```
mumoshu: `clusterAutoscalerSupport` is meant to provide enough permissions to call AWS APIs required to host cluster-autoscaler. Then, probably we should provide a nodeSelector for cluster-autoscaler so that we can ensure CA to be scheduled to nodes with the enough permissions?

core/nodepool/config/templates/stack-template.json
```diff
@@ -145,6 +145,13 @@
         ],
         "MinSize": "{{.MinCount}}",
         "Tags": [
+          {{if gt .ClusterAutoscaler.MaxSize 0}}
```
mumoshu: Should be `if .ClusterAutoscaler.Enabled`

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1781,6 +1781,56 @@ write_files:
...
+                    - --stderrthreshold=info
+                    - --cloud-provider=aws
+                    - --skip-nodes-with-local-storage=false
+                    - --expander=least-waste
+                    - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled
+                  env:
+                    - name: AWS_REGION
+                      value: {{.Region}}
+                  volumeMounts:
+                    - name: ssl-certs
```
mumoshu: This volumeMount and the corresponding volume will be unnecessary once https://github.com/kubernetes/autoscaler/pull/48 is merged to CA

e2e/run
mumoshu: We should add a validation which emits an error when there are no worker node pool or controller node whose `clusterAutoscalerSupport` is enabled.
Otherwise cluster-autoscaler is unable to work/be scheduled due to missing IAM permissions.

redbaron: can't we run CA on controller nodes? they already have elevated permissions, adding more wont make it much worse

mumoshu: @redbaron Yes, we can. kube-aws as of today already supports `controller.clusterAutoscalerSupport.enabled` to provide appropriate iam permissions to controller nodes. It would be a matter of just adding appropriate node labels and node selectors to stick cluster-autoscaler to whatever nodes(worker or controller) clusterAutoscalerSupport is enabled.

core/controlplane/config/templates/cloud-config-controller
mumoshu: The PR is merged

mumoshu:
Assuming I have no control over when a docker image for cluster-autoscaler containing improvements on which this PR depends is released, I have opened https://github.com/kube-aws/autoscaler and https://quay.io/repository/kube-aws/cluster-autoscaler for hosting our own.

mumoshu:
kubernetes/autoscaler#11 is now merged but a docker image containing it isn't released yet.

core/controlplane/config/templates/stack-template.json
```diff
@@ -203,12 +203,13 @@
                   "Resource": [ "*" ]
                 },
                 {{end}}
-                {{if .Experimental.ClusterAutoscalerSupport.Enabled}}
+                {{if .Addons.ClusterAutoscaler.Enabled}}
```
redbaron:
```
            {
                 "Effect": "Allow",
                   "Action": [
                    "autoscaling:DescribeAutoScalingInstances"
                   ],
                  "Resource": "*"
                },
                {
                  "Action": [
                  ],
                  "Resource": "*"
                  "Condition": {
                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
                  },
                  "Resource": "*",
                  "Effect": "Allow"
                },
```
```

File: core/nodepool/config/templates/stack-template.json [hunk 5]
```diff
@@ -361,6 +368,7 @@
                   "Action": [
                     "autoscaling:DescribeAutoScalingGroups",
```
redbaron: I still don't understand, what is the benfit of running cluster autoscaler on worker nodes?

model/node_labels.go
```diff
@@ -0,0 +1,28 @@
...
+// Returns key=value pairs separated by ',' to be passed to kubelet's `--node-labels` flag
+func (l NodeLabels) String() string {
+	labels := []string{}
+	keys := []string{}
+	for k, _ := range l {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		v := l[k]
```
redbaron:
```
v := l[k]
if len(v) > 0 {
  labels = append(labels, fmt.Sprintf("%s=%s", k, v))
}
else {
  labels = append(labels, fmt.Sprintf("%s", k))
}
```

mumoshu: @redbaron Thanks!
Is a node label spec like `mykey=` is invalid?

redbaron: no it isn't, but seeing args on a command line with empty `=` is a little confusing, at least for me. end result is the same

core/nodepool/config/templates/stack-template.json
mumoshu: @redbaron Thanks, good question 👍 
The bigger a k8s cluster is, the more CA's resource usage is.
So, I guess one wants a dedicated node pool with a single worker node which can be recreated easier than controller nodes and used to run CA which is possibly scaled-up/down by a vertical-pod-autoscaler.
That's why I tried to support running CA in worker nodes, even though it isn't the default setup.
What do you think? 😃

mumoshu: For an usual use-case, running CA in a controller node is recommended and that's the default setup.

core/controlplane/config/templates/stack-template.json
mumoshu: What is the `Null` condition? Could you point a doc for that for me?

core/nodepool/config/templates/stack-template.json
redbaron: On larger clusters you'd have large controller nodes too :) Anyway, I see the point and I guess adding this support isn't much of the work

model/node_labels.go
mumoshu: Thanks for the confirmation. Ok, I will make the suggested change 👍

core/controlplane/config/templates/stack-template.json
redbaron: Conditional keys supported by AWS Autoscaling:
http://docs.aws.amazon.com/autoscaling/latest/userguide/control-access-using-iam.html


Null condition:
http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements.html#Conditions_Null

core/nodepool/config/templates/stack-template.json
mumoshu: @redbaron 
> On larger clusters you'd have large controller nodes too :)

Yea, that's certainly true!
Do you think we can just recommend users to make controller large enough, instead of complication cluster.yaml like this?

core/controlplane/config/templates/stack-template.json
mumoshu: Thanks! So it permits an action only when the targeted resource has the tag? Great - I will incorporate it.

mumoshu: Done for controller nodes.
I'm still unsure if we'd want to support the customization to schedule CA to worker nodes https://github.com/kubernetes-incubator/kube-aws/pull/629#discussion_r117418053

model/node_labels.go
mumoshu: Done

mumoshu:
I've updated the PR description accordingly to the current state of the work.

mumoshu:
Added a cluster-autoscaling section in the doc

mumoshu:
Testing E2E after merging with the master branch locally

mumoshu:
Whoa, all the tests have passed.

c-knowles:
`cluster.yaml` defaults could do with a bit of an update here, it's still listing config which is not read any more.

VinceMD:
Sorry to comment on a closed ticket (happy to open a new one if required).

What happens if  the condition is missing for the IAM role:

`            "Condition": {
                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
                  },` 

I don't have much control over IAM roles and my roles are shared between all my clusters.
I tried using a wildcard / start but it is not accepted by AWS.
What would you recommend for a case where the role is shared between many clusters?
What happens if I omit the condition altogether? The upstream project does not have it.
Is there a more generic tag I should check instead of being specific to a cluster?

mumoshu:
@VinceMD If it was missing for `Describe*` operations, it would be ok. I'm afraid you would end up leaking more autoscaling group names when one of your pods are hijacked by attackers but it won't be a huge problem anyway?

However, if it was missing for `TerminateInstancesInAutoscalingGroups`, CA can theoretically terminate EC2 instances outside of K8S clusters when e.g. CA had bug(s) result in doing so. #800 is an example of that.

In your case, my suggestion is to tag EC2 instances with something like `"PartOfKubernetes": "true"` and use it in the condition of the IAM policy. So that CA simply can't terminate EC2 instances  not part of K8S cluster(s). I suppose it isn't the best thing to do - the best would be create IAM role per cluster - but better than nothing.

Does my explanation make sense? 😃

VinceMD:
@mumoshu thanks for the clear explanation. I got it.

Good news is that my production Kubernetes cluster will have its own IAM role, so I can be specific and use the conditions properly (can only terminate/setDesired for the prod cluster based on specific tag) 

It's only my non prod clusters, which are non public facing and sharing same IAM role at the moment. I will work with a tag, as suggested. 

Problem solved!

PR #632, opened 2017-05-09
Fix unable to backup namespaced resources

cheungpat:
The kube-resources-autosave script is unable to find the list of namespaces
and so it tries to backup resources from `null` namespace.

Log from kube-resources-autosave-dumper:

```
(...snip...)
++ jq -r .metadata.name
+ for ns in '$(jq -r '\''.metadata.name'\'' < ${DUMP_DIR}/namespaces.json)'
+ echo 'Searching in namespace: null'
+ mkdir -p /kube-resources-autosave/tmp/2017-05-09_15-18-57/null
Searching in namespace: null
(...snip...)
```


jollinshead:
@cheungpat Nice find 👍 LGTM

mumoshu:
Good catch 👍 Thanks for the fix @cheungpat!

PR #639, opened 2017-05-10
Avoid unnecessary node replacements when TLS bootstrapping is enabled

### LINKED ISSUE #633: Re-encrypting kubelet token causes worker node pools to be recreated needlessly
Successive calls to `kube-aws update` are causing worker nodes to be recreated needlessly when TLS bootstrapping is enabled due to re-encrypting the token used for the bootstrapping process.

This is the same issue solved by the good old `.enc` files for certificates.

danielfm:
Simplifies the code by re-using the same encryption/caching/fingerprinting logic both to cert files and TLS bootstrapping / auth tokens, and keeps the TLS bootstrap token in its own file instead of storing it directly in `./credentials/tokens.csv`.

Concretely, instead of using the `./credentials/tokens.csv` file to store the TLS bootstrap token (and re-encrypting it every time, causing the bug in #633), we now store the bootstrap token in its own file, thus not causing unnecessary node pool replacement at every `kube-aws update`.

TODO

- [x] Test creation/update workflow with TLS bootstrapping disabled, without custom auth tokens
- [x] Test creation/update workflow with TLS bootstrapping disabled, with custom auth tokens
- [x] Test creation/update workflow with TLS bootstrapping enabled, without custom auth tokens
- [x] Test creation/update workflow with TLS bootstrapping enabled, with custom auth tokens
- [x] Address review comments and perform the last round of tests

## Actions Required

If this get merged, we must not forget to inform users of the experimental TLS Bootstrap feature that the following actions are required:

1. Run `kube-aws render stack` to update controller/worker user data templates
2. Move the bootstrap token from `./credentials/tokens.csv` to its own file `./credentials/kubelet-tls-bootstrap-token`
   
   ```
   # Remove the following line from ./credentials/tokens.csv, and move <token>
   # (with no leading/trailing blank chars and lines) to it's own file
   # ./credentials/kubelet-tls-bootstrap-token
   <token>,kubelet-bootstrap,10001,system:kubelet-bootstrap
   ```

3. Run `kube-aws update` to update the cluster. This operation will cause controllers and workers to be replaced

Fixes #633.


danielfm:
This PR is ready for review.

core/root/render/credentials.go
```diff
@@ -25,9 +25,9 @@ func NewCredentialsRenderer(c *config.Cluster) CredentialsRenderer {
 	}
 }
 
-func (r credentialsRendererImpl) RenderTLSCerts(renderCredentialsOpts config.CredentialsOptions) error {
+func (r credentialsRendererImpl) RenderCerts(renderCredentialsOpts config.CredentialsOptions) error {
```
mumoshu: Perhaps `RenderCredentials` should be a better name here as it generates not only certs but also tls-bootstrap token?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1001,6 +1001,20 @@ write_files:
...
+           {{- if .AssetsConfig.HasAnyAuthTokens }}
+           authDir=/etc/kubernetes/auth
+           echo generating $authDir/computed-tokens.csv
+           echo > $authDir/computed-tokens.csv
+
+           {{- if .AssetsConfig.HasTLSBootstrapToken }}
+           echo $(cat $authDir/tls-bootstrap.token),kubelet-bootstrap,10001,system:kubelet-bootstrap >> $authDir/computed-tokens.csv
+           {{- end }}
+           {{- if .AssetsConfig.HasAuthTokens }}
+           cat $authDir/tokens.csv >> $authDir/computed-tokens.csv
```
mumoshu: nit but my preference is first write a work file like `tokens.csv` to `/var/run/coreos/tokens.csv` and keep the final one named `$authDir/tokens.csv` rather than `computed-tokens.csv`
so that someone peeked into `$authDir` finds only `tokens.csv` and don't get confused like `which one is the "final" tokens file passed to apiserver, computed-tokens.csv or tokens.csv?`

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1326,7 +1340,7 @@ write_files:
         - hostPath:
             path: /usr/share/ca-certificates
           name: ssl-certs-host
-        {{if .AuthTokensConfig.HasTokens}}
+        {{if .AssetsConfig.HasAnyAuthTokens}}
```
mumoshu: BTW I like the function naming 👍

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -634,9 +634,9 @@ write_files:
              mv -f $f ${encKey%.enc}
            done;
 
-           {{ if .Experimental.TLSBootstrap.Enabled }}
+           {{ if and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken }}
            echo injecting token into the kubelet bootstrap kubeconfig file
-           bootstrap_token=$(cat /etc/kubernetes/auth/kubelet-bootstrap.token);
+           bootstrap_token=$(cat /etc/kubernetes/auth/tls-bootstrap.token);
```
mumoshu: Just for my education but why did you rename it from kubelet- to tls-?
I believe you tried to match it against other relevant symbol names(TLSBootstrap, HasTLSBootstrapToken).
However, the naming in kubelet is `kubelet-bootstrap` rather than `tls-bootstrap`.
I guess you assumed (like me) that it is originally meant for `kubelet-(tls-)bootstrap` hence named `tls-bootstrap`?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1001,6 +1001,20 @@ write_files:
...
              mv -f $f ${encKey%.enc}
            done;
+
+           {{- if .AssetsConfig.HasAnyAuthTokens }}
+           authDir=/etc/kubernetes/auth
+           echo generating $authDir/computed-tokens.csv
+           echo > $authDir/computed-tokens.csv
+
+           {{- if .AssetsConfig.HasTLSBootstrapToken }}
+           echo $(cat $authDir/tls-bootstrap.token),kubelet-bootstrap,10001,system:kubelet-bootstrap >> $authDir/computed-tokens.csv
```
mumoshu: Sorry for being late in commenting something like this but
* my preference after read through code is, the token file named `tls-bootstrap-token` rather than `tls-bootstrap.token` so that we can be consistent in the word `token` contained in basenames of both files `tls-bootstrap-token` and `tokens.csv`
  * It can be named like `tls-bootstra-token.txt` but I feels that it is too much 😃 and we already have various txt files without the `.txt` extentions anyway

mumoshu:
@danielfm Thanks as always for your continuous efforts on this area :bow:

I've added several comments but LGTM overall.
Would you mind addressing these?

Anyway IMHO, what the most important thing in this PR is the code reduced by 500 lines!
Great job 👍

danielfm:
@mumoshu sure, I'll try to address these ASAP.

core/controlplane/config/templates/cloud-config-worker
danielfm: Yes, my reasoning was to make it match with the symbols in Go code, and what this feature is called in upstream [Kubernetes documentation](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/), but I think I forgot to replace a few other names in templates.

Do you think it's the right thing to do here? Keep the old `kublet-bootstrap` names throughout the templates, or replace them all by `tls-bootstrap` (or even `kubelet-tls-bootstrap`)?

mumoshu: Thanks for the clarification!

Yes. Renaming to tls-bootstrap makes more sense for me.

Also, I began to slightly prefer kubelet-tls-bootstrap because it produces no ambiguity even in the context of kube-aws. "kube-aws tls bootstrap" vs "kube-aws kubelet tls bootstrap" - the latter seems to better represent that it is for kubelet for me.

danielfm: Done.

mumoshu:
@danielfm Thanks a lot for your efforts! LGTM 👍

Issue #640, opened 2017-05-10
Etcd management does not clean up old/stopped docker containers


jeremyd:
Etcdadm runs docker for awscli actions periodically.  It does not use --rm so these stopped containers grow over time and eventually will fill the root disk.  We just need to use --rm on the docker run to prevent this.  PR incoming.

mumoshu:
Closing this as the fix #641 has been merged.
@jeremyd Thanks for the report and the fix 👍

PR #641, opened 2017-05-10
adding --rm to docker run aws_cli commands in etcdadm

jeremyd:
fix for #640


mumoshu:
@jeremyd LGTM. Thanks as always for your support 👍

PR #645, opened 2017-05-11
NVIDIA driver installation support on GPU instances

everpeace:
Hello community!  Thank you for developing the great tool!  I'm glad to have opportunity to contribute this project because my colleague @mumoshu always encourages me.

As everybody know, AWS offers Nvidia GPU ready instance type families (P2 and G2).  And, of course Kubernetes supports GPU resource scheduling since 1.6.  However Nvidia drivers is not installed in default coreos ami used in kube-aws.  Then, let's support it!

This PR implements auto installation support of Nvidia GPU driver.  I borrowed some driver installation script from [/Clarifai/coreos-nvidia](https://github.com/Clarifai/coreos-nvidia/).

#### Configuration and what will happen
New configuration for this feature is really simple.  `worker.nodePool[i].gpu.nvidia.{enabled,version}` is introduced in `cluster.yaml`.

- default value of `enabled` is false.
- user will be warned if 
  - user set `enabled: true` when `instanceType` doesn't support GPU.  In this case the configuration will be ignored.
  - user set `enabled: false` when `instanceType` does support GPU
- when `enabled: true` on GPU supported instance type, 
  - nvidia driver will be installed automatically in each node in the nodepool.  
  - The installation will happen just before `kubelet.service` starting (see below).  
  - And, `kubelet` will start with [`--feature-gates="Accelerators=true"`](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L212-L214)
  - then container can mount nvidia driver [like this](https://gist.github.com/everpeace/9e03050467d5ef5f66b7ce96b5fefa72#file-pod-yaml-L30-L53)
- several tags are assigned to the node for enabling schedule on appropriate GPU model and its driver version by using `nodeAffinity`.
  - `alpha.kubernetes.io/nvidia-gpu-name=<GPU hardware type name>`
  - `kube-aws.coreos.com/gpu=nvidia`,
  - `kube-aws.coreos.com/nvidia-gpu-version=<version>`
  -  Because substitution are not used in unit definition, I introduced `/etc/default/kubectl` for defining these label values in [this commit](https://github.com/kubernetes-incubator/kube-aws/commit/5c59944a15b46d6ca9ecaf589bc41d4cfe52fedf).

#### Driver installation process
Most of installation script is borrowed from [/Clarifai/coreos-nvidia](https://github.com/Clarifai/coreos-nvidia/).  Especially, for device node installation, I referenced to Clarifai/coreos-nvidia#4 .  I just described summary of installation process.

- [`kubelet.service`](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L144-L147) ruires [`nvidia-start.service`](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L456-L471)
- [`nvidia-start.service`](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L456-L471)  invokes [`build-and-install.sh`](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L918-L947), which installs nvidia drivers and kernel module files, via `ExecStartPre`.  `nvidia-start.service` will create device nodes(`nvidiactl` and `nvidia0,1,...`).  Other dynamic device nodes are controlled by`udevadam` (configuration is in [this rule file](https://github.com/everpeace/kube-aws/blob/feature/nvidia-gpu/core/controlplane/config/templates/cloud-config-worker#L905-L939))
  - `nvidia-start.service` is `type=oneshot` because `kubelet.service` should wait until `nvidia-start.sh` completely succeeded.
  - `Restart` policy cannot be used with`type=oneshot`.  `nvidia-start.service` doesn't use systemd's retry feature is not used but manual `retry.sh` is used.
- [nvidia-persistenced](https://docs.nvidia.com/deploy/driver-persistence/#persistence-daemon) is also enabled for speeding up startup.  this service is started/stopped via `udevadam` too.

### How to try
1. build `kube-aws` on this branch
2. `kube-aws up` with minimal nodepool configuration below
   ```
   worker:
    nodePools:
     - name: p2xlarge
       count: 1
       instanceType: p2.xlarge
       rootVolume:
         size: 30
         type: gp2
       gpu:
         nvidia:
           enabled: true
           version: "375.66"
    ```
3.  check `kubectl get nodes --show-labels`.  Then you'll see one node with gpu related labels.
4. try starting this [pod](https://gist.github.com/everpeace/9e03050467d5ef5f66b7ce96b5fefa72#file-pod-yaml)
   ```
   kubectl create -f pod.yaml
   ```
5. log reports sample matrix multiplication is computed on gpus.
   ```
   kubectl logs gpu-pod
   ```
Feedbacks are always welcome!!

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -142,7 +142,7 @@ coreos:
         Wants=flanneld.service cfn-etcd-environment.service
         After=cfn-etcd-environment.service
         {{- if .Gpu.Nvidia.IsEnabledOn .InstanceType }}
-        Wants=nvidia-start.service
+        Requires=nvidia-start.service
         After=nvidia-start.service
         {{- end }}
         [Service]
@@ -184,9 +184,6 @@ coreos:
         {{if .UseCalico -}}
         ExecStartPre=/usr/bin/docker run --rm -e SLEEP=false -v /opt/cni/bin:/host/opt/cni/bin {{ .CalicoCniImage.RepoWithTag }} /install-cni.sh
         {{end -}}
-        {{- if .Gpu.Nvidia.IsEnabledOn .InstanceType }}
-        ExecStartPre=/usr/bin/systemctl is-active nvidia-start.service
-        {{- end }}
         ExecStart=/usr/lib/coreos/kubelet-wrapper \
         --cni-conf-dir=/etc/kubernetes/cni/net.d \
         {{/* Work-around until https://github.com/kubernetes/kubernetes/issues/43967 is fixed via https://github.com/kubernetes/kubernetes/pull/43995 */ -}}
@@ -462,30 +459,13 @@ coreos:
         [Unit]
         Description=Load NVIDIA module
         After=local-fs.target
-        Wants=nvidia-install.service
-        After=nvidia-install.service
 
         [Service]
         Type=oneshot
         RemainAfterExit=true
-        ExecStartPre=/usr/bin/systemctl is-active nvidia-install.service
-        ExecStart=/opt/nvidia/current/bin/nvidia-start.sh
-
-        [Install]
-        WantedBy=multi-user.target
-
-    - name: nvidia-install.service
-      enable: false
-      content: |
-        [Unit]
-        Description=Install NVIDIA driver and modules
-        After=local-fs.target
-
-        [Service]
-        Type=oneshot
-        RemainAfterExit=true
-        ExecStart=/opt/nvidia-build/util/retry.sh 0 /opt/nvidia-build/build-and-install.sh
+        ExecStartPre=/opt/nvidia-build/util/retry.sh 0 /opt/nvidia-build/build-and-install.sh
         TimeoutStartSec=900
+        ExecStart=/opt/nvidia/current/bin/nvidia-start.sh
 
         [Install]
         WantedBy=multi-user.target
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -452,6 +452,34 @@ coreos:
         Type={{.Experimental.EphemeralImageStorage.Filesystem}}
 {{end}}
 
+{{if .Gpu.Nvidia.IsEnabledOn .InstanceType}}
+    - name: nvidia-start.service
+      enable: false
+      content: |
+        [Unit]
+        Description=Load NVIDIA module
+        After=local-fs.target
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=true
+        ExecStart=/opt/nvidia/current/bin/nvidia-start.sh
+
+        [Install]
+        WantedBy=multi-user.target
+    - name: nvidia-persistenced.service
+      enable: false
+      content: |
+        [Unit]
+        Description=NVIDIA Persistence Daemon
+        Wants=local-fs.target
+
+        [Service]
+        Type=forking
+        ExecStart=/opt/nvidia/current/bin/nvidia-persistenced --user nvidia-persistenced --no-persistence-mode --verbose
+        ExecStopPost=/bin/rm -rf /var/run/nvidia-persistenced
+{{end}}
+
 {{if .SSHAuthorizedKeys}}
 ssh_authorized_keys:
   {{range $sshkey := .SSHAuthorizedKeys}}
@@ -479,6 +507,15 @@ ssh_authorized_keys:
         [Install]
         WantedBy=kubelet.service
 {{end}}
+
+{{if .Gpu.Nvidia.IsEnabledOn .InstanceType}}
+users:
+  - name: nvidia-persistenced
+    gecos: NVIDIA Persistence Daemon
+    homedir: /
+    shell: /sbin/nologin
+{{end}}
+
 write_files:
 {{- if .CustomFiles}}
   {{- range $w := .CustomFiles}}
@@ -1140,35 +1177,6 @@ write_files:
       # This script is borrowed from https://github.com/Clarifai/coreos-nvidia/pull/4
       /usr/sbin/insmod /opt/nvidia/current/lib/modules/$(uname -r)/video/$1
 
-  - path: /opt/nvidia-build/nvidia-persistenced.service
-    owner: root:root
-    permissions: 0644
-    content: |
-      [Unit]
-      Description=NVIDIA Persistence Daemon
-      Wants=local-fs.target
-
-      [Service]
-      Type=forking
-      ExecStart=/opt/nvidia/current/bin/nvidia-persistenced --user nvidia-persistenced --no-persistence-mode --verbose
-      ExecStopPost=/bin/rm -rf /var/run/nvidia-persistenced
-
-  - path: /opt/nvidia-build/nvidia-start.service
-    owner: root:root
-    permissions: 0644
-    content: |
-      [Unit]
-      Description=Load NVIDIA module
-      After=local-fs.target
-
-      [Service]
-      Type=oneshot
-      RemainAfterExit=true
-      ExecStart=/opt/nvidia/current/bin/nvidia-start.sh
-
-      [Install]
-      WantedBy=multi-user.target
-
   - path: /opt/nvidia-build/nvidia-start.sh
     owner: root:root
     permissions: 0755
@@ -1261,11 +1269,6 @@ write_files:
       echo "/opt/nvidia/current/lib64" > /etc/ld.so.conf.d/nvidia.conf
       ldconfig
 
-      echo "Configuring nvidia persistence user"
-      id -u nvidia-persistenced >/dev/null 2>&1 || \
-      useradd --system --home '/' --shell '/sbin/nologin' -c 'NVIDIA Persistence Daemon' nvidia-persistenced
-
-      cp *.service /etc/systemd/system/
       systemctl daemon-reload
       systemctl enable nvidia-start.service
       systemctl start nvidia-start.service
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -143,6 +143,7 @@ coreos:
         After=cfn-etcd-environment.service
         [Service]
         EnvironmentFile=-/etc/etcd-environment
+        EnvironmentFile=-/etc/default/kubelet
         Environment=KUBELET_IMAGE_TAG={{.K8sVer}}
         Environment=KUBELET_IMAGE_URL={{.HyperkubeImage.RktRepoWithoutTag}}
         Environment="RKT_RUN_ARGS=--volume dns,kind=host,source=/etc/resolv.conf {{.HyperkubeImage.Options}}\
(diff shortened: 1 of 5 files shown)


mumoshu:
cc @jollinshead @redbaron I believe you're currently running batch workloads on your kube-aws clusters. Are you also willing to run machine learning workloads utilizing GPUs? 😃

core/controlplane/config/templates/cluster.yaml
```diff
@@ -387,6 +387,16 @@ worker:
 #      # Documentation: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/dedicated-instance.html
 #      tenancy: default
 #
+#      # (Experimental) GPU Driver installation support
+#      # Currently, only Nvidia driver is supported.
+#      # This setting takes effect only when node's instance family is p2 of g2.
+#      # Otherwise, installation will be skipped even if enabled.
```
mumoshu: According to https://github.com/kubernetes-incubator/kube-aws/pull/645/files#diff-5e5dcac90c0e906cb335a42b0352ce9cR47?, it seems like kube-aws emits a validation error when the gpu support is enabled on a node pool with instance type other than p2 or g2?

mumoshu: Ah, sorry! I've missed that there is only a warning.

Anyway, I believe we'd better make it an error rather than a warning because an user does seem to intend to enable the GPU support but kube-aws was unable to do so.
WDYT?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -845,3 +860,414 @@ write_files:
...
+      [Unit]
+      Description=NVIDIA Persistence Daemon
+      Wants=local-fs.target
+
+      [Service]
+      Type=forking
+      ExecStart=/opt/nvidia/current/bin/nvidia-persistenced --user nvidia-persistenced --no-persistence-mode --verbose
+      ExecStopPost=/bin/rm -rf /var/run/nvidia-persistenced
+
+  - path: /opt/nvidia-build/nvidia-start.service
```
mumoshu: Just for my education, would you mind sharing me how this systemd unit gets installed to systemd?
Can we set up the systemd unit via the `units` section of cloud-config like others?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -845,3 +860,414 @@ write_files:
...
+
+      TOOLS="nvidia-debugdump nvidia-cuda-mps-control nvidia-xconfig nvidia-modprobe nvidia-smi nvidia-cuda-mps-server
+      nvidia-persistenced nvidia-settings"
+
+      # Create archives with no paths
+      tar -C ${ARTIFACT_DIR} -cvj $(basename -a ${ARTIFACT_DIR}/*.so.*) > libraries-${VERSION}.tar.bz2
+      tar -C ${ARTIFACT_DIR} -cvj ${TOOLS} > tools-${VERSION}.tar.bz2
+      tar -C ${ARTIFACT_DIR}/kernel -cvj $(basename -a ${ARTIFACT_DIR}/kernel/*.ko) > modules-${COMBINED_VERSION}.tar.bz2
+
+  - path: /opt/nvidia-build/build.sh
```
mumoshu: AFAICS, `coreos-nvidia` supports [cross-building against any version of Container Linux](https://github.com/Clarifai/coreos-nvidia/#coreos-nvidia).
Then, just curious, but can we run `build.sh` locally in e.g. a Vagrant machine hosting Container Linux and then export built assets to the disk, then embed to cloud-config or put to S3 for faster startup of GPU-enabled nodes?

model/gpu.go
```diff
@@ -0,0 +1,56 @@
...
+	return Gpu{
+		Nvidia: NvidiaSetting{
+			Enabled: false,
+			Version: "",
+		},
+	}
+}
+
+// This function is used when rendering cloud-config-worker
+func (c NvidiaSetting) IsEnabledOn(instanceType string) bool {
```
mumoshu: I like the nice naming 👍

core/controlplane/config/templates/cluster.yaml
```diff
@@ -387,6 +387,16 @@ worker:
...
+#      # Make sure to choose 'docker' as container runtime when enabled this feature.
+#      gpu:
+#        nvidia:
+#          enabled: true
```
mumoshu: I guess an installed driver can be unusable once Container Linux is updated afterwards, due to an updated kernel.
// Do you think so too?
Then, I believe we'd better document it - maybe something like:

> Ensure that automatic Container Linux is disabled(it is disabled by default btw). Otherwise the installed driver may stop working when an OS update resulted in an updated kernel

would work.

core/controlplane/config/templates/cluster.yaml
everpeace: oh, ok. yes, I agree with you.  kube-aws should prohibit 'enabled:true' with instance types which doesn't support GPU.  I will update my code.

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -845,3 +860,414 @@ write_files:
...
+      ldconfig
+
+      echo "Configuring nvidia persistence user"
+      id -u nvidia-persistenced >/dev/null 2>&1 || \
+      useradd --system --home '/' --shell '/sbin/nologin' -c 'NVIDIA Persistence Daemon' nvidia-persistenced
+
+      cp *.service /etc/systemd/system/
+      systemctl daemon-reload
+      systemctl enable nvidia-start.service
+      systemctl start nvidia-start.service
```
everpeace: `nvidia-install.sh` install `nvidia-start.service` and `nvidia-persitenced.service` to the systemd.  and this script only stats `nvidia-start.service`.  the unit `insmod` nvidia module and then udevadam will spawn several actions in `71-nvidia.rules` which includes `nvidia-persisnteced.service`.

core/controlplane/config/templates/cloud-config-worker
everpeace: please see [the comment](https://github.com/kubernetes-incubator/kube-aws/pull/645/files#r116218676).  `nvidia-install.sh` does.

core/controlplane/config/templates/cluster.yaml
everpeace: Yes, you're absolutely right.  I will put this to the comment.

core/controlplane/config/templates/cloud-config-worker
mumoshu: Thanks! So, basically we're copying systemd unit files written via `write_files` into `/etc/systemd/system`?
Then, it is possible to just write those units directly into the `units` section rather than copy via `write_files`, right?

core/controlplane/config/templates/cloud-config-worker
everpeace: yeah, probably. I will try this my local vagrant!  If it could, would it better to distribute pre-build driver files to nodes from s3 or somewhere??

core/controlplane/config/templates/cloud-config-worker
everpeace: if we put `nvidia-start.service` by `write_files`, can we control when the service start?   Unless we don't define `enabled` explicitly in unit definition, this doesn't automatically, right?

mumoshu: I guess so.

You can even enable it by default.
More concretely, my idea is modifying `nvidia-start.service` to something like:

```
[Unit]
Description=Start NVIDIA daemon(?)
After=local-fs.target
Before=kublet.service
[Service]
Type=oneshot
RemainAfterExit=true
ExecStartPre=/opt/bin/retry /opt/nvidia/current/bin/nvidia-start.sh
ExecStart=/bin/true
[Install]
RequiredBy=kubelet.service
```

And trigger it via a systemd dependency from a newly introduced `nvidia-install.service`:

```
[Unit]
Description=Start NVIDIA daemon(?)
After=local-fs.target
Before=nvidia-start.service
[Service]
Type=oneshot
RemainAfterExit=true
ExecStartPre=/opt/bin/retry /opt/nvidia/current/bin/nvidia-install.sh
ExecStart=/bin/true
[Install]
RequiredBy=nvidia-start.service
```

where `/opt/bin/retry` is:

```bash
#/usr/bin/bash

set -e

# could be improved to a finite loop
while true; do
  if "$@"; then
    exit 0
  fi
  echo retrying "$@"
  # could be improved to an exponential backoff
  sleep 1
done
````

and omit `ExecStartPre=/opt/nvidia-build/build-and-install.sh` from kubelet.service.

File: core/controlplane/config/templates/cloud-config-worker (hunk 3 above)
Comment (everpeace) [inline core/controlplane/config/templates/cloud-config-worker]: As you pointed out, building libraries, kernel modules and nvidia tools succeeded in vagrant!
Puting pre-build binaries to gpu nodes directly makes  it startup faster.  However `kube-aws up` process will be a more complex instead and it requires users to install vagrant and virtualbox.

Honestly speaking, building process takes several minutes (it probably 5 ~10 min. which depends on speed of downloading coreos dev container and nvidia installer).  I believe this duration would be acceptable for many users because gpu node pool usually doesn't need to scale so quickly like normal nodepool which hosts service pods.

What do you think?? Do you prefer local build & put pre-build binaries directly for faster startup??

Comment (mumoshu) [inline core/controlplane/config/templates/cloud-config-worker]: > I believe this duration would be acceptable for many users because gpu node pool usually doesn't need to scale so quickly like normal nodepool which hosts service pods.

I completely agree with you here 👍 
The local build feature could be an extra thing we may or may not add in the future.

File: core/controlplane/config/templates/cloud-config-worker [hunk 7]
```diff
@@ -845,3 +897,380 @@ write_files:
...
+      cp -f 71-nvidia.rules /etc/udev/rules.d/
+      udevadm control --reload-rules
+
+      mkdir -p /etc/ld.so.conf.d/ 2>/dev/null
+      echo "/opt/nvidia/current/lib64" > /etc/ld.so.conf.d/nvidia.conf
+      ldconfig
+
+      systemctl daemon-reload
+      systemctl enable nvidia-start.service
+      systemctl start nvidia-start.service
```
mumoshu: It may sound like a nit but -

If we could transform the `build-and-install.sh` in ExecStartPre to a systemd unit like suggested in https://github.com/kubernetes-incubator/kube-aws/pull/645#discussion_r116232098, `systemctl daemon-reload` and `enable`, `start` can be omitted altogether and such dependencies can be handled completely by systemd rather than by a bash script?

everpeace: @mumoshu 
Yes, sounds nice!  I'll fix this.

everpeace:
I updated 'Driver installation process' in this PR summary.

everpeace:
I updated 'Driver installation process' in this PR summary because of https://github.com/kubernetes-incubator/kube-aws/pull/645/commits/2710606a171536a6fe292b7fd1d9b6b3a757b839 and https://github.com/kubernetes-incubator/kube-aws/pull/645/commits/0546e94264488a98f05e4eec33ee2ac704549d89

everpeace:
@mumoshu I updated systemd units' dependencies.  I'm glad if you could take a look!

mumoshu:
@everpeace LGTM. Thanks for your efforts on the great feature 👍

kylegato:
Has anyone tested this w/ the new G3 instances yet?

PR #650, opened 2017-05-14
Label masters with 'node-role.kubernetes.io/master' label

### LINKED ISSUE #649: Controller nodes should be labeled `node-role.kubernetes.io/master`
Currently we taint, but do not label masters. Label set by `kubeadm` is official and used at least here:

https://github.com/kubernetes/kubernetes/blob/f542fe12837560d991218289e8297dac7362c7ba/pkg/controller/service/servicecontroller.go#L627

redbaron:
Fixes #649


mumoshu:
@redbaron LGTM. Thanks for the improvement! 👍

PR #651, opened 2017-05-15
WIP: ALB as a k8s API endpoint load balancer

### LINKED ISSUE #608: ALB as a k8s API endpoint load balancer
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/598#issuecomment-297921158

mumoshu:
* `apiEndpoints[].classic` is added. It must be explicitly set to `false` to use ALB instead of ELB
* `apiEndpoints[].certificateArn` is required when ALB is enabled.
  * Run `aws iam upload-server-certificate --server-certificate-name kube1 --certificate-body file://credentials/apiserver.pem --certificate-chain file://credentials/ca.pem --private-key file://credentials/apiserver-key.pem` after `kube-aws render credentials` to upload certs and obtain the ARN.
* One big concern is that when one of API endpoints is backed by ALB, kube-apiserver must bind the insecure-port to be more open to the world(from `127.0.0.1` to `0.0.0.0`) for ALB health-checks.

Closes #608

TODO:

- [ ] Make `classic` true by default

### DIFF
--- a/core/controlplane/cluster/describer.go
+++ b/core/controlplane/cluster/describer.go
@@ -6,6 +6,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elbv2"
 )
 
 type ClusterDescriber interface {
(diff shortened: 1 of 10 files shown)


mumoshu:
A test is failing:

```
    --- FAIL: TestMainClusterConfig/WithMultiAPIEndpoints (8.79s)
        --- PASS: TestMainClusterConfig/WithMultiAPIEndpoints/AssertConfig (0.00s)
        --- FAIL: TestMainClusterConfig/WithMultiAPIEndpoints/AssertCluster (0.00s)
        	maincluster_test.go:1736: unexpected managed ELB logical names: 
        --- PASS: TestMainClusterConfig/WithMultiAPIEndpoints/ValidateUserData (4.00s)
        --- PASS: TestMainClusterConfig/WithMultiAPIEndpoints/ValidateTemplates (0.20s)
```

redbaron:
Regarding insecure port for healthchecks,  ALB spec has this param:
```
Matcher
The HTTP codes to use when checking for a successful response from a target. You can specify values or ranges of values between 200 and 499. The default value is 200.
```

we can set it to code returned by API server on unauthorized access

mumoshu:
@redbaron Thanks! Good point - I tried and found it definitely works.

mumoshu:
It may or may no be obvious but as ALB doesn't support HTTPS pass-through(by definition?), we now need to use an authentication mechanism other than the default TLS mutual authentication using credentials/admin.pm or any k8s API invocation from kubelet or admin machin ends up with:

> error: You must be logged in to the server (the server has asked for the client to provide credentials)

mumoshu:
// I wish if ALB could pass tls client cert like nginx and haproxy does

mumoshu:
So - I can make it conditionally use token authentication to secure kublet-to-apiserver communication like what we do today for the kublet-tls-bootstrap feature.
However, I wonder if anyone is really going to use ALB as API load balancer when it turned out you can no longer rely on TLS mutual auth to secure kublet-to-apiserver communication?? Isn't it just insecure without no concrete, clear benefit gained?

redbaron:
there is mutual auth: client continues to validate server cert and server validates client using bearer token. I agree though, that ALB isn't that useful in current kube-aws setup which relies on certs

mumoshu:
@redbaron Thanks for the reply 👍 

> client continues to validate server cert and server validates client using bearer token

That's true. Just to be extra clear, it isn't "TLS" mutual authentication, right? (The server identity is proven but the client identity is not(=identity provided by admin-key.pem is not taken into account hence [the error](https://github.com/kubernetes-incubator/kube-aws/pull/651#issuecomment-301688304)?)

redbaron:
> it isn't "TLS" mutual authentication, right?

Right, kubelet service need to be using configuration, where token and not certs are used to be able to communicate with API Server

mumoshu:
@redbaron Thanks for the confirmation 🍵

PR #652, opened 2017-05-15
Use new K8S 1.6 AWS tagging scheme

redbaron:
KubernetesCluster is deprecated in favour of `kubernetes.io/cluster/$NAME`


mumoshu:
Thanks for the improvement @redbaron!
I'm assuming that this is based on https://github.com/kubernetes/kubernetes/commit/0b5ae5391ef9eeba337f1dfa62b6f3125e28c86c.
Am I correct?

redbaron:
@mumoshu , yes, based on the code in https://github.com/kubernetes/kubernetes/blob/release-1.6/pkg/cloudprovider/providers/aws/tags.go

mumoshu:
@redbaron Thanks. LGTM 👍

PR #659, opened 2017-05-17
Update Kubernetes dashboard to v1.6.1. Update calico to v2.2.1.

### LINKED ISSUE #627: Migrating kube-dashboard from rc to deployment
According to the [corresponding manifest in the upstream repo](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-controller.yaml)

We're still deploying it via a replication controller(not even a replica set!)
https://github.com/kubernetes-incubator/kube-aws/blob/d3f5986/core/controlplane/config/templates/cloud-config-controller#L1753

camilb:
* update Kubernetes dashboard to [v1.6.1](https://github.com/kubernetes/dashboard/releases/tag/v1.6.1)
* update calico to [v2.2.1](http://docs.projectcalico.org/v2.2/releases/)
* replace kubernetes dashboard replicationcontroller with a deployment

Closes #627


mumoshu:
@camilb Thanks for the updates 👍 LGTM

PR #663, opened 2017-05-18
Make kubelet flags more consistent

redbaron:
It seems that kubelet silently converts `cluster_dns` into
`cluster-dns`, so nothing was broken, it is just a minor cleanup


mumoshu:
@redbaron Good catch 👍 Thanks for the fix!

PR #671, opened 2017-05-22
Fix "install-kube-system" script when "clusterAutoscaler" is disabled.

camilb:
Hi @mumoshu it looks that leaving the 

      clusterAutoscaler:
        enabled: false

will break the "install-kube-system" script. This is the error in journald: 

    May 22 13:42:19 ip-10-0-1-49.ec2.internal install-kube-system[16493]: error: the path "/srv/kubernetes/manifests/cluster-autoscaler-de.yaml" does not exist


mumoshu:
@camilb Ah, my bad. Thanks a lot for the fix!
I will cut rc.2 with the fix asap.

PR #674, opened 2017-05-22
Improved node drainer

### LINKED ISSUE #673: Node draining not working when RBAC is enabled
The problem is in the `kube-aws:node-extensions` role, where the `apiGroups` attribute is set to the wrong value (`"policy"`, where it's supposed to be `""`). This causes the nodes to be cordoned as expected, but pods fail to be evicted.

**Edit:** I already tested the fix, I'll send a PR in the next few hours.

danielfm:
**This PR is ready to review.**

This PR introduces a rewrite of the node draining logic (based on ASG lifecycle hooks) + a fix for a bug in RBAC-enabled clusters that causes pods not to be evicted (see #673).

Now, if the node drainer is enabled, the install script creates a DaemonSet that executes a control loop implemented in bash that checks the current state of each instance and, if the auto scaling group signals some instance must be terminated, the script will detect the change and will perform the node draining - with retries in case the drain command takes too long to finish.

The old node draining systemd unit was removed, as they don't really work in practice: since the node draining could take more than a few seconds to finish, it don't usually have time to finish draining the node before AWS halts the instance. (In case the operator wants to terminate some arbitrary instance, it's up to the operator to drain the node before terminating it)

Fixes #673.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1072,7 +1072,7 @@ write_files:
             - replicationcontrollers
             verbs:
             - get
-          - apiGroups: ["policy"]
+          - apiGroups: [""]
             resources:
             - pods/eviction
             verbs:
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -571,6 +571,10 @@ write_files:
       kubectl apply -f "${mfdir}/calico.yaml"
       {{ end }}
 
+      {{ if .Experimental.NodeDrainer.Enabled }}
+      kubectl apply -f "${mfdir}/kube-node-drainer-asg-ds.yaml"
+      {{ end }}
+
       # Configmaps
       kubectl apply -f "${mfdir}/kube-dns-cm.yaml"
 
@@ -1020,6 +1024,94 @@ write_files:
 
       rkt rm --uuid-file=/var/run/coreos/decrypt-assets.uuid || :
 {{ end }}
+
+{{if .Experimental.NodeDrainer.Enabled}}
+  - path: /srv/kubernetes/manifests/kube-node-drainer-asg-ds.yaml
+    content: |
+        kind: DaemonSet
+        apiVersion: extensions/v1beta1
+        metadata:
+          name: kube-node-drainer-sg-ds
+          namespace: kube-system
+          labels:
+            k8s-app: kube-node-drainer-sg-ds
+        spec:
+          updateStrategy:
+            type: RollingUpdate
+          template:
+            metadata:
+              labels:
+                k8s-app: kube-node-drainer-sg-ds
+              annotations:
+                scheduler.alpha.kubernetes.io/critical-pod: ''
+            spec:
+              tolerations:
+              - operator: Exists
+                effect: NoSchedule
+              - operator: Exists
+                effect: NoExecute
+              - operator: Exists
+                key: CriticalAddonsOnly
+              initContainers:
+                - name: hyperkube
+                  image: {{.HyperkubeImage.RepoWithTag}}
+                  command:
+                  - /bin/cp
+                  - -f
+                  - /hyperkube
+                  - /workdir/hyperkube
+                  volumeMounts:
+                  - mountPath: /workdir
+                    name: workdir
+              containers:
+                - name: main
+                  image: {{.AWSCliImage.RepoWithTag}}
+                  env:
+                  - name: NODE_NAME
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: spec.nodeName
+                  command:
+                  - /bin/sh
+                  - -xec
+                  - |
+                    kubectl()  { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
+                    metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
+                    asg()      { aws --region="${REGION}" autoscaling "$@"; }
+
+                    INSTANCE_ID=$(metadata meta-data/instance-id)
+                    REGION=$(metadata dynamic/instance-identity/document | jq -r .region)
+                    [ -n "${REGION}" ]
+
+                    while sleep 30; do
+                      STATE=$(asg describe-auto-scaling-instances --region "${REGION}" --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
+                      if [ ! "${STATE}" = Terminating:Wait ]; then
+                        continue
+                      fi
+                      echo Node is in Terminating:Wait state, draining it
+
+                      if ! kubectl drain --ignore-daemonsets=true --delete-local-data=true --force=true --timeout=60s "${NODE_NAME}"; then
+                        echo Not all pods on this host can be evicted, will try again
+                        continue
+                      fi
+
+                      echo All evictable pods are gone, notifying AutoScalingGroup that instance ${INSTANCE_ID} can be shutdown
+                      ASG_NAME=$(asg describe-auto-scaling-instances --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].AutoScalingGroupName')
+                      HOOK_NAME=$(asg describe-lifecycle-hooks --auto-scaling-group-name "${ASG_NAME}" | jq -r '.LifecycleHooks[].LifecycleHookName' | grep -i nodedrainer)
+
+                      echo Sending notification to ASG_NAME=${ASG_NAME} HOOK_NAME=${HOOK_NAME}
+                      asg complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id "${INSTANCE_ID}" --lifecycle-hook-name "${HOOK_NAME}" --auto-scaling-group-name "${ASG_NAME}"
+
+                      # sleep 5 mins + 1 mins, expecting that instance will be shut down in this time
+                      sleep 300
+                    done
+                  volumeMounts:
+                  - mountPath: /opt/bin
+                    name: workdir
+              volumes:
+                - name: workdir
+                  emptyDir: {}
+{{end}}
 
 {{if .Experimental.Plugins.Rbac.Enabled }}
   # TODO: remove the following binding once the TLS Bootstrapping feature is enabled by default, see:
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1075,9 +1075,13 @@ write_files:
                   - /bin/sh
                   - -xec
                   - |
-                    kubectl()  { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
                     metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
                     asg()      { aws --region="${REGION}" autoscaling "$@"; }
+
+                    # Hyperkube binary is not statically linked, so we need to use
+                    # the musl interpreter to be able to run it in this image
+                    # See: https://github.com/kubernetes-incubator/kube-aws/pull/674#discussion_r118889687
+                    kubectl() { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
 
                     INSTANCE_ID=$(metadata meta-data/instance-id)
                     REGION=$(metadata dynamic/instance-identity/document | jq -r .region)
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -230,6 +230,27 @@
                   "Resource" : "{{.KMSKeyARN}}"
                 },
                 {{end}}
+                {{if .Experimental.NodeDrainer.Enabled }}
+                {
+                  "Action": "sns:Publish",
+                  "Effect": "Allow",
+                  "Resource": { "Ref": "{{.Controller.LogicalName}}NodeDrainerLHT" }
+                },
+                {
+                  "Action": [
+                    "autoscaling:CompleteLifecycleAction",
+                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeLifecycleHooks"
+                  ],
+                  "Condition": {
+                    "StringEquals": {
+                      "autoscaling:ResourceTag/aws:cloudformation:stack-id": { "Ref": "AWS::StackId" }
+                    }
+                  },
+                  "Effect": "Allow",
+                  "Resource": "*"
+                },
+                {{end}}
                 {
                   "Action": [
                     "ecr:GetAuthorizationToken",
@@ -258,7 +279,8 @@
               "Effect": "Allow",
               "Principal": {
                 "Service": [
-                  "ec2.{{.Region.PublicDomainName}}"
+                  "ec2.{{.Region.PublicDomainName}}",
+                  "autoscaling.{{.Region.PublicDomainName}}"
                 ]
               }
             }
@@ -713,6 +735,27 @@
         ]]}}
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration"
+    },
+    {{end}}
+    {{if .Experimental.NodeDrainer.Enabled }}
+    "{{.Controller.LogicalName}}NodeDrainerLHT" : {
+      "Properties" : {
+        "DisplayName" : "SNS topic for the {{.Controller.LogicalName}} node drainer ASG lifecycle hook"
+      },
+      "Type" : "AWS::SNS::Topic"
+    },
+    "{{.Controller.LogicalName}}NodeDrainerLH" : {
+      "Properties" : {
+        "AutoScalingGroupName" : {
+          "Ref": "{{.Controller.LogicalName}}"
+        },
+        "DefaultResult" : "CONTINUE",
+        "HeartbeatTimeout" : "{{.Experimental.NodeDrainer.DrainTimeoutInSeconds}}",
+        "LifecycleTransition" : "autoscaling:EC2_INSTANCE_TERMINATING",
+        "NotificationTargetARN" : { "Ref": "{{.Controller.LogicalName}}NodeDrainerLHT" },
+        "RoleARN" : { "Fn::GetAtt": [ "IAMRoleController", "Arn" ] }
+      },
+      "Type" : "AWS::AutoScaling::LifecycleHook"
     },
     {{end}}
     "{{.Controller.LogicalName}}LC": {
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1,6 +1,16 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
   "Description": "kube-aws Kubernetes cluster {{.ClusterName}}",
+  "Parameters": {
+    "NotificationTargetARN": {
+      "Type": "String",
+      "Description": "ASG LifecycleHook notification target ARN"
+    },
+    "NotificationRoleARN": {
+      "Type": "String",
+      "Description": "Role to be used by ASG LifecycleHook to publish notification"
+    }
+  },
   "Resources": {
     "{{.Controller.LogicalName}}": {
       "Type": "AWS::AutoScaling::AutoScalingGroup",
@@ -232,22 +242,21 @@
                 {{end}}
                 {{if .Experimental.NodeDrainer.Enabled }}
                 {
-                  "Action": "sns:Publish",
-                  "Effect": "Allow",
-                  "Resource": { "Ref": "{{.Controller.LogicalName}}NodeDrainerLHT" }
-                },
-                {
                   "Action": [
-                    "autoscaling:CompleteLifecycleAction",
                     "autoscaling:DescribeAutoScalingInstances",
                     "autoscaling:DescribeLifecycleHooks"
                   ],
-                  "Condition": {
-                    "StringEquals": {
-                      "autoscaling:ResourceTag/aws:cloudformation:stack-id": { "Ref": "AWS::StackId" }
-                    }
-                  },
                   "Effect": "Allow",
+                  "Resource": "*"
+                },
+                {
+                  "Action": [
+                    "autoscaling:CompleteLifecycleAction"
+                  ],
+                  "Effect": "Allow",
+                  "Condition": {
+                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
+                  },
                   "Resource": "*"
                 },
                 {{end}}
@@ -279,8 +288,7 @@
               "Effect": "Allow",
               "Principal": {
                 "Service": [
-                  "ec2.{{.Region.PublicDomainName}}",
-                  "autoscaling.{{.Region.PublicDomainName}}"
+                  "ec2.{{.Region.PublicDomainName}}"
                 ]
               }
             }
@@ -738,12 +746,6 @@
     },
     {{end}}
     {{if .Experimental.NodeDrainer.Enabled }}
-    "{{.Controller.LogicalName}}NodeDrainerLHT" : {
-      "Properties" : {
-        "DisplayName" : "SNS topic for the {{.Controller.LogicalName}} node drainer ASG lifecycle hook"
-      },
-      "Type" : "AWS::SNS::Topic"
-    },
     "{{.Controller.LogicalName}}NodeDrainerLH" : {
       "Properties" : {
         "AutoScalingGroupName" : {
@@ -752,8 +754,8 @@
         "DefaultResult" : "CONTINUE",
         "HeartbeatTimeout" : "{{.Experimental.NodeDrainer.DrainTimeoutInSeconds}}",
         "LifecycleTransition" : "autoscaling:EC2_INSTANCE_TERMINATING",
-        "NotificationTargetARN" : { "Ref": "{{.Controller.LogicalName}}NodeDrainerLHT" },
-        "RoleARN" : { "Fn::GetAtt": [ "IAMRoleController", "Arn" ] }
+        "NotificationTargetARN" : { "Ref": "NotificationTargetARN" },
+        "RoleARN" : { "Ref": "NotificationRoleARN" }
       },
       "Type" : "AWS::AutoScaling::LifecycleHook"
     },
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -221,12 +221,6 @@
       "Metadata": {{template "Metadata" .}}
     },
     {{if .NodeDrainer.Enabled }}
-    "{{.LogicalName}}NodeDrainerLHT" : {
-      "Properties" : {
-        "DisplayName" : "SNS topic for the {{.LogicalName}} node drainer ASG lifecycle hook"
-      },
-      "Type" : "AWS::SNS::Topic"
-    },
     "{{.LogicalName}}NodeDrainerLH" : {
       "Properties" : {
         "AutoScalingGroupName" : {
@@ -235,8 +229,8 @@
         "DefaultResult" : "CONTINUE",
         "HeartbeatTimeout" : "{{.NodeDrainer.DrainTimeoutInSeconds}}",
         "LifecycleTransition" : "autoscaling:EC2_INSTANCE_TERMINATING",
-        "NotificationTargetARN" : { "Ref": "{{.LogicalName}}NodeDrainerLHT" },
-        "RoleARN" : { "Fn::GetAtt": [ "IAMRoleWorker", "Arn" ] }
+        "NotificationTargetARN" : { "Ref": "NotificationTargetARN" },
+        "RoleARN" : { "Ref": "NotificationRoleARN" }
       },
       "Type" : "AWS::AutoScaling::LifecycleHook"
     },
@@ -421,23 +415,20 @@
                 {{if .NodeDrainer.Enabled }}
                 {
                   "Action": [
-                    "sns:Publish"
-                  ],
-                  "Effect": "Allow",
-                  "Resource": { "Ref": "{{.LogicalName}}NodeDrainerLHT" }
-                },
-                {
-                  "Action": [
-                    "autoscaling:CompleteLifecycleAction",
                     "autoscaling:DescribeAutoScalingInstances",
                     "autoscaling:DescribeLifecycleHooks"
                   ],
-                  "Condition": {
-                    "StringEquals": {
-                      "autoscaling:ResourceTag/aws:cloudformation:stack-id": { "Ref": "AWS::StackId" }
-                    }
-                  },
                   "Effect": "Allow",
+                  "Resource": "*"
+                },
+                {
+                  "Action": [
+                    "autoscaling:CompleteLifecycleAction"
+                  ],
+                  "Effect": "Allow",
+                  "Condition": {
+                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
+                  },
                   "Resource": "*"
                 },
                 {{end}}
@@ -469,8 +460,7 @@
               "Effect": "Allow",
               "Principal": {
                 "Service": [
-                  "ec2.{{.Region.PublicDomainName}}",
-                  "autoscaling.{{.Region.PublicDomainName}}"
+                  "ec2.{{.Region.PublicDomainName}}"
                 ]
               }
             }
@@ -498,6 +488,14 @@
     "ControlPlaneStackName": {
       "Type": "String",
       "Description": "The name of a control-plane stack used to import values into this stack"
+    },
+    "NotificationTargetARN": {
+      "Type": "String",
+      "Description": "ASG LifecycleHook notification target ARN"
+    },
+    "NotificationRoleARN": {
+      "Type": "String",
+      "Description": "Role to be used by ASG LifecycleHook to publish notification"
     }
   },
   "Resources": {
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -220,22 +220,22 @@
       },
       "Metadata": {{template "Metadata" .}}
     },
-    {{if and .NodeDrainer.Enabled (gt .NodeDrainer.DrainInterval 0) }}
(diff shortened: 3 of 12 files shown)


mumoshu:
@danielfm Thanks for the cool enhancement!
How did it work for you - Does setting `drainInterval` to a fairly large value, say `600`, really keep an EC2 instance into `Terminating` state for such a long period?
In other words, what would be the max value for it in practice?
If you are aware of it, I'd appreciate it you could share 👍

danielfm:
I'm not so sure right now, but I'm gonna test a few rollouts with different values to see what happens.

danielfm:
@mumoshu after testing a few things, I had to change the implementation to leverage ASG Lifecycle Hooks in order to implement the drain interval feature.

In summary, we attach a lifecycle hook to each auto scaling group that causes instances to hang for the time configured in `drainInterval`. Instances can remain in a wait state for at most one hour (this time can be prolonged, but I think it's pretty reasonable for our use case), so I put a validation check to ensure only a value between 0 and 60 can be used.

redbaron:
@danielfm , thanks for taking action on it. Few notes:

1. ASG need to be configured to make use of new update strategy
2. cycling rkt container every 10 seconds can be taxing,we have problems with it before, one yet to be confirmed described here: https://github.com/coreos/bugs/issues/1927 
3. you need to to notify ASG then it can proceed with termination once node is drained
4. node drain can legitimately fail when PodDisruptionBudget doesn't allow more pods to be evicted. therefore draining should be repeated.
5. IMHO having this logic running as daemonset will improve visibility and make it more standard . In general more we move to kube the merrier, so having another set of systemd unit files is something I'd like to avoid.

Please have a look at our internal change which is yet to be properly tested, but can serve as a good starting point:

```
commit 97fbf20c25aa71438314c725494f764fe693cec2
Author: Maxim Ivanov <ivanov.maxim@gmail.com>
Date:   Tue May 16 17:14:51 2017 +0100

    RUN-444 reliable node drain

diff --git a/asg-roller-ds.yaml b/asg-roller-ds.yaml
new file mode 100644
index 0000000..6016715
--- /dev/null
+++ b/asg-roller-ds.yaml
@@ -0,0 +1,85 @@
+---
+kind: DaemonSet
+apiVersion: extensions/v1beta1
+metadata:
+  name: kube-aws-asg-roller
+  namespace: kube-system
+  labels:
+    k8s-app: kube-aws-asg-roller
+spec:
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-aws-asg-roller
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      tolerations:
+      - key: "node.alpha.kubernetes.io/role"
+        operator: "Equal"
+        value: "master"
+        effect: "NoSchedule"
+      - key: "CriticalAddonsOnly"
+        operator: "Exists"
+      initContainers:
+        - name: hyperkube
+          #image: {{.HyperkubeImage.RepoWithTag}}
+          image: quay.io/coreos/hyperkube:v1.6.3_coreos.0
+          command:
+          - /bin/cp
+          - -f
+          - /hyperkube
+          - /workdir/hyperkube
+          volumeMounts:
+          - mountPath: /workdir
+            name: workdir
+      containers:
+        - name: main
+          #image: {{.AWSCliImage.RepoWithTag}}
+          image: quay.io/coreos/awscli
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          command:
+          - /bin/sh
+          - -xec
+          - |
+            kubectl() { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
+            metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
+            asg() { aws --region="${REGION}" autoscaling "$@"; }
+
+            INSTANCE_ID=$(metadata meta-data/instance-id)
+            REGION=$(metadata dynamic/instance-identity/document | jq -r .region)
+            [ -n "${REGION}" ]
+
+            while sleep 60; do
+              STATE=$(asg describe-auto-scaling-instances --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
+              if [ ! "x${STATE}" = "xTerminating:Wait" ]; then
+                continue
+              fi
+              echo "Node is in Terminating:Wait state, draining it"
+
+              if ! kubectl drain --ignore-daemonsets=true --delete-local-data=true --force=true "$(NODE_NAME)" --timeout=120s; then
+                echo "Not all PODs on this host can be evicted. Will try again soon."
+                continue
+              fi
+
+              echo "All evictable PODs are gone, notifying AutoScalingGroup that instance  $INSTANCE_ID can be shutdown"
+              ASG_NAME=$(asg describe-auto-scaling-instances --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].AutoScalingGroupName')
+              HOOK_NAME=$(asg describe-lifecycle-hooks --auto-scaling-group-name "${ASG_NAME}" | jq -r '.LifecycleHooks[].LifecycleHookName' | grep -i kubeaws)
+
+              echo "Sending notification to ASG_NAME=${ASG_NAME} HOOK_NAME=${HOOK_NAME}"
+              sleep 60  # TODO: remove me, it is here for debugging
+              asg complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id "${INSTANCE_ID}" --lifecycle-hook-name "${HOOK_NAME}" --auto-scaling-group-name "${ASG_NAME}"
+              sleep 300  # sleep 5 mins + 1 mins, expecting that instance will be shut down in this time
+            done
+          volumeMounts:
+          - mountPath: /opt/bin
+            name: workdir
+      volumes:
+        - name: workdir
+          emptyDir: {}
diff --git a/cluster.yaml b/cluster.yaml
index dc6ee56..c669d68 100644
--- a/cluster.yaml
+++ b/cluster.yaml
@@ -26,6 +26,7 @@ customSettings: &customSettings
  vault:
   address: https://vault.lab.hcom/
  HostedZoneId: *HostedZoneId
+ ASGReplacingUpdate: true
 
 # Additional keys to preload on the coreos account (keep this to a minimum)
 sshAuthorizedKeys:
@@ -52,13 +53,21 @@ controller:
   loadBalancer:
     private: true
 
+experimental: &features
+  clusterAutoscalerSupport:
+    enable: true
+  kube2IamSupport:
+    enabled: true
+  nodeDrainer:
+    enabled: true
+
 worker:
   nodePools:
     - name: nodepool-a
       subnets:
         - name: sn-backend-main-us-west-2a
       autoScalingGroup:
-        minSize: 1
+        minSize: 2
         maxSize: 3
         rollingUpdateMinInstancesInService: 0
       customSettings: *customSettings
@@ -109,14 +118,6 @@ waitSignal:
   enabled: true
   maxBatchSize: 1
 
-experimental:
-  clusterAutoscalerSupport:
-    enable: true
-  kube2IamSupport:
-    enabled: true
-  nodeDrainer:
-    enabled: true
-
 # AWS Tags for cloudformation stack resources
 stackTags:
   Team: "RUNP"
diff --git a/stack-templates/control-plane.json.tmpl b/stack-templates/control-plane.json.tmpl
index f368e6a..c4eac19 100644
--- a/stack-templates/control-plane.json.tmpl
+++ b/stack-templates/control-plane.json.tmpl
@@ -232,11 +232,20 @@
                 "Effect": "Allow",
                   "Action": [
-                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeAutoScalingInstances"
+                   ],
+                  "Resource": "*"
+                },
+                {
                   ],
-                  "Resource": "*"
+                  "Condition": {
+                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
+                  },
+                  "Resource": "*",
+                  "Effect": "Allow"
                 },
                 {{end}}
                 {{if .Experimental.Kube2IamSupport.Enabled }}
@@ -253,6 +262,28 @@
                   "Resource" : "{{.KMSKeyARN}}"
                 },
                 {{end}}
+                {{ if .CustomSettings.ASGReplacingUpdate }}
+                {
+                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeLifecycleHooks"
+                  ],
+                  "Resource": "*"
+                },
+                {
+                    "autoscaling:CompleteLifecycleAction"
+                  ],
+                  "Condition": {
+                    "StringEquals": {
+                      "autoscaling:ResourceTag/aws:cloudformation:stack-id": { "Ref": "AWS::StackId" }
+                    }
+                  },
+                  "Resource": "*"
+                },
+                {{ end }}
                 {
                   "Action": [
                     "ecr:GetAuthorizationToken",
@@ -553,6 +584,7 @@
       },
       {{end}}
       "UpdatePolicy" : {
+{{ if not .CustomSettings.ASGReplacingUpdate }}
         "AutoScalingRollingUpdate" : {
           "MinInstancesInService" : "0",
           "MaxBatchSize" : "1",
@@ -563,6 +595,11 @@
           "PauseTime": "PT2M"
           {{end}}
         }
+{{else}}
+        "AutoScalingReplacingUpdate" : {
+          "WillReplace": true
+        }
+{{end}}
       },
       "Metadata" : {
         "AWS::CloudFormation::Init" : {
diff --git a/stack-templates/node-pool.json.tmpl b/stack-templates/node-pool.json.tmpl
index e9c529d..c91c781 100644
--- a/stack-templates/node-pool.json.tmpl
+++ b/stack-templates/node-pool.json.tmpl
@@ -188,6 +188,7 @@
       },
       {{end}}
       "UpdatePolicy" : {
+{{ if not .CustomSettings.ASGReplacingUpdate }}
         "AutoScalingRollingUpdate" : {
           "MinInstancesInService" :
           {{if .SpotPrice}}
@@ -204,10 +205,27 @@
           "PauseTime": "PT2M"
           {{end}}
         }
+{{else}}
+        "AutoScalingReplacingUpdate" : {
+          "WillReplace": true
+        }
+{{end}}
       },
       "Metadata": {{template "Metadata" .}}
     },
-
+{{- if .CustomSettings.ASGReplacingUpdate }}
+    "{{.LogicalName}}KubeAwsLifecycleHook": {
+      "Properties": {
+        "AutoScalingGroupName": { "Ref": "{{.LogicalName}}" },
+        "LifecycleTransition": "autoscaling:EC2_INSTANCE_TERMINATING",
+        "NotificationTargetARN": { "Ref": "NotificationTargetARN" },
+        "RoleARN": { "Ref": "NotificationRoleARN" },
+        "DefaultResult": "CONTINUE",
+        "NotificationMetadata": "{ \"SpotFleet\": false }"
+      },
+      "Type": "AWS::AutoScaling::LifecycleHook"
+    },
+{{- end}}
     "{{.LogicalName}}LC": {
       "Properties": {
         "BlockDeviceMappings": [
@@ -263,6 +281,17 @@
       "Type": "String",
       "Description": "The name of a control-plane stack used to import values into this stack"
     }
+{{- if .CustomSettings.ASGReplacingUpdate }}
+    ,
+    "NotificationTargetARN": {
+      "Type": "String",
+      "Description": "ASG LifecycleHook notification target ARN"
+    },
+    "NotificationRoleARN": {
+      "Type": "String",
+      "Description": "Role to be used by ASG LifecycleHook to publish notification"
+    }
+{{- end}}
   },
   "Resources": {
     {{if .SpotFleet.Enabled}}
@@ -380,11 +409,20 @@
                 "Effect": "Allow",
                   "Action": [
-                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeAutoScalingInstances"
+                   ],
+                  "Resource": "*"
+                },
+                {
                   ],
-                  "Resource": "*"
+                  "Condition": {
+                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
+                  },
+                  "Resource": "*",
+                  "Effect": "Allow"
                 },
                 {{end}}
                 {{if .SpotFleet.Enabled}}
@@ -401,6 +439,28 @@
                 },
                 {{end}}
                 {{end}}
+                {{ if .CustomSettings.ASGReplacingUpdate }}
+                {
+                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeLifecycleHooks"
+                  ],
+                  "Resource": "*"
+                },
+                {
+                    "autoscaling:CompleteLifecycleAction"
+                  ],
+                  "Condition": {
+                    "StringEquals": {
+                      "autoscaling:ResourceTag/aws:cloudformation:stack-id": { "Ref": "AWS::StackId" }
+                    }
+                  },
+                  "Resource": "*"
+                },
+                {{ end }}
                 {
                   "Action": [
                     "ecr:GetAuthorizationToken",
diff --git a/stack-templates/root.json.tmpl b/stack-templates/root.json.tmpl
index 98e869b..2d47dc6 100644
--- a/stack-templates/root.json.tmpl
+++ b/stack-templates/root.json.tmpl
@@ -2,6 +2,44 @@
   "AWSTemplateFormatVersion": "2010-09-09",
   "Description": "kube-aws Kubernetes cluster {{.ClusterName}}",
   "Resources": {
+{{/* Commented bacause .CustomSettings are not available here
+{{- if .CustomSettings.ASGReplacingUpdate }}
+*/}}
+    "ASGNotificationTarget": {
+      "Type": "AWS::SQS::Queue"
+    },
+    "ASGNotificationRole": {
+      "Properties": {
+        "AssumeRolePolicyDocument": {
+          "Statement": [ {
+            "Action": [ "sts:AssumeRole" ],
+            "Effect": "Allow",
+            "Principal": {
+              "Service": [ "autoscaling.amazonaws.com" ]
+            }
+          } ],
+          "Version": "2012-10-17"
+        },
+        "Policies": [{
+          "PolicyName": "root",
+          "PolicyDocument": {
+            "Version": "2012-10-17",
+            "Statement": [ {
+              "Effect": "Allow",
+              "Action": [
+                "sqs:SendMessage",
+                "sqs:GetQueueUrl"
+              ],
+              "Resource": { "Fn::GetAtt": [ "ASGNotificationTarget", "Arn" ] }
+            } ]
+          }
+        }]
+      },
+      "Type": "AWS::IAM::Role"
+    },
+{{/* Commented bacause .CustomSettings are not available here
+{{- end}}
+*/}}
     "{{.ControlPlane.Name}}": {
       "Type" : "AWS::CloudFormation::Stack",
       "Properties" : {
@@ -23,6 +61,15 @@
       "Properties" : {
         "Parameters": {
           "ControlPlaneStackName": {"Fn::GetAtt" : [ "{{$.ControlPlane.Name}}" , "Outputs.StackName" ]}
+{{/* Commented bacause .CustomSettings are not available here
+{{- if .CustomSettings.ASGReplacingUpdate }}
+*/}}
+          ,
+          "NotificationTargetARN": { "Fn::GetAtt": [ "ASGNotificationTarget", "Arn" ] },
+          "NotificationRoleARN": { "Fn::GetAtt": [ "ASGNotificationRole", "Arn" ] }
+{{/* Commented bacause .CustomSettings are not available here
+{{- end}}
+*/}}
         },
         "Tags" : [
           {
@@ -38,9 +85,6 @@
       },
       "DependsOn": [
         "{{$.ControlPlane.Name}}"
-        {{ if $i }}
-        ,"{{ (index $.NodePools (minus $i 1)).Name }}"
-	{{ end }}
       ]
     }{{end}}
   },
diff --git a/userdata/cloud-config-controller b/userdata/cloud-config-controller
index 1fc03c0..c0567b5 100644
--- a/userdata/cloud-config-controller
+++ b/userdata/cloud-config-controller
@@ -1423,6 +1423,9 @@ write_files:
       for manifest in {kube-dns,heapster,kube-dashboard}-svc.yaml;do
           kubectl apply -f "${mfdir}/$manifest"
       done
+      {{- if .CustomSettings.ASGReplacingUpdate }}
+      kubectl apply -f "${mfdir}/asg-roller-ds.yaml"
+      {{- end }}
 
       kubectl apply -f "${mfdir}/tiller.yaml"
 
@@ -1807,6 +1810,96 @@ write_files:
               emptyDir: {}
 {{ end }}
 
+{{- if .CustomSettings.ASGReplacingUpdate }}
+  - path: /srv/kubernetes/manifests/asg-roller-ds.yaml
+    content: |
+      ---
+      kind: DaemonSet
+      apiVersion: extensions/v1beta1
+      metadata:
+        name: kube-aws-asg-roller
+        namespace: kube-system
+        labels:
+          k8s-app: kube-aws-asg-roller
+      spec:
+        updateStrategy:
+          type: RollingUpdate
+        template:
+          metadata:
+            labels:
+              k8s-app: kube-aws-asg-roller
+            annotations:
+              scheduler.alpha.kubernetes.io/critical-pod: ''
+          spec:
+            tolerations:
+            - key: "node.alpha.kubernetes.io/role"
+              operator: "Equal"
+              value: "master"
+              effect: "NoSchedule"
+            - key: "CriticalAddonsOnly"
+              operator: "Exists"
+            initContainers:
+              - name: hyperkube
+                #image: quay.io/coreos/hyperkube:v1.6.3_coreos.0
+                command:
+                - /bin/cp
+                - -f
+                - /hyperkube
+                - /workdir/hyperkube
+                - mountPath: /workdir
+                  name: workdir
+            containers:
+              - name: main
+                image: {{.AWSCliImage.RepoWithTag}}
+                #image: quay.io/coreos/awscli 
+                env:
+                - name: NODE_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: spec.nodeName
+                command:
+                - /bin/sh
+                - -xec
+                - |
+                  kubectl() { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
+                  metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
+                  asg() { aws --region="${REGION}" autoscaling "$@"; }
+
+                  INSTANCE_ID=$(metadata meta-data/instance-id)
+                  REGION=$(metadata dynamic/instance-identity/document | jq -r .region)
+                  [ -n "${REGION}" ]
+
+                  while sleep 60; do
+                    STATE=$(asg describe-auto-scaling-instances --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
+                    if [ ! "x${STATE}" = "xTerminating:Wait" ]; then
+                      continue
+                    fi
+                    echo "Node is in Terminating:Wait state, draining it"
+
+                    if ! kubectl drain --ignore-daemonsets=true --delete-local-data=true --force=true "$(NODE_NAME)" --timeout=120s; then
+                      echo "Not all PODs on this host can be evicted. Will try again soon."
+                      continue
+                    fi
+
+                    echo "All evictable PODs are gone, notifying AutoScalingGroup that instance  $INSTANCE_ID can be shutdown"
+                    ASG_NAME=$(asg describe-auto-scaling-instances --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].AutoScalingGroupName')
+                    HOOK_NAME=$(asg describe-lifecycle-hooks --auto-scaling-group-name "${ASG_NAME}" | jq -r '.LifecycleHooks[].LifecycleHookName' | grep -i kubeaws)
+
+                    echo "Sending notification to ASG_NAME=${ASG_NAME} HOOK_NAME=${HOOK_NAME}"
+                    sleep 60  # TODO: remove me, it is here for debugging
+                    asg complete-lifecycle-action --lifecycle-action-result CONTINUE --instance-id "${INSTANCE_ID}" --lifecycle-hook-name "${HOOK_NAME}" --auto-scaling-group-name "${ASG_NAME}"
+                    sleep 300  # sleep 5 mins + 1 mins, expecting that instance will be shut down in this time
+                  done
+                - mountPath: /opt/bin
+                  name: workdir
+            volumes:
+              - name: workdir
+                emptyDir: {}
+{{- end}}
+
 {{if .AssetsEncryptionEnabled }}
   - path: /opt/bin/decrypt-assets
     owner: root:root
```

danielfm:
@redbaron nice!

> ASG need to be configured to make use of new update strategy

May I ask why is this required?

> You need to to notify ASG then it can proceed with termination once node is drained

I still think it's a good idea to wait some time after the node is drained so we can proceed to the next node (so that all pods running in this instance could have some time to start elsewhere), this is why I introduced the `drainInterval` config. My implementation could be improved so that, instead of always waiting for X minutes after the termination was signaled, we wait for X minutes _after_ the node is drained.

> node drain can legitimately fail when PodDisruptionBudget doesn't allow more pods to be evicted. therefore draining should be repeated.

Nice catch.

> IMHO having this logic running as daemonset will improve visibility and make it more standard . In general more we move to kube the merrier, so having another set of systemd unit files is something I'd like to avoid.

I think, for now, the best way to ensure this pod will run on _every_ node is by deploying it via the manifest directory instead of using a DaemonSet with tolerations; something along those lines were discussed in a previous PR of mine (#612), where I tried to convert kube-proxy to a DaemonSet.

redbaron:
>> ASG need to be configured to make use of new update strategy

> May I ask why is this required?

OK, maybe it doesn't *need* to, but it is just an easy win. Instead of rolling nodes in batches and therefore causing multiple pods restarts, use of http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-replacingupdate policy will ensure that all PODs moved exactly once


> I still think it's a good idea to wait some time after the node is drained in order to proceed to the next node (so that all pods running there could have some time to start elsewhere);

If started pod doesn't report *Ready* state, next eviction can't happen and will be retried, therefore it is safe not to wait. 

> I think, for now, the best way to ensure this pod will run on every node is by deploying it via the manifest directory instead of using a DaemonSet with tolerations; something along those lines were discussed in a previous PR of mine (#612), where I tried to convert kube-proxy to a DaemonSet.

We already have Calico DaemonSet which is then broken for users with manually tainted nodes. Given that there were no reports we can continue migrating unit files to DaemonSets. If you adamant not to run it as DaemonSet (but please do!), at least make it a static POD instead of another set of  magical systemd unit files which are barely observable and next to impossible to fix/update quickly without rolling the cluster.

danielfm:
> OK, maybe it doesn't need to, but it is just an easy win. Instead of rolling nodes in batches and therefore causing multiple pods restarts, use of ... will ensure that all PODs moved exactly once

Makes sense!

> If started pod doesn't report Ready state, next eviction can't happen and will be retried, therefore it is safe not to wait.

Fair enough. I think it would be nice to add a comment next to the node drainer config in `cluster.yaml` to instruct users to use pod disruption budgets if they want to make node rollouts even more safer then.

> We already have Calico DaemonSet which is then broken for users with manually tainted nodes. Given that there were no reports we can continue migrating unit files to DaemonSets. If you adamant not to run it as DaemonSet (but please do!), at least make it a static POD instead of another set of magical systemd unit files which are barely observable and next to impossible to fix/update quickly without rolling the cluster.

Don't worry, I really dig your suggestion of moving away from systemd units. :)

mumoshu:
@redbaron @danielfm Just my two cents on the small part of the great discussion above:

> > I think, for now, the best way to ensure this pod will run on every node is by deploying it via the manifest directory instead of using a DaemonSet with tolerations; something along those lines were discussed in a previous PR of mine (#612), where I tried to convert kube-proxy to a DaemonSet.
>
> We already have Calico DaemonSet which is then broken for users with manually tainted nodes. 

I happened to know that we can tolerate to all the taints with any keys and values by specifying only `operator` and `effect`.
https://github.com/kubernetes/kubernetes/pull/45715/files#diff-1817ac4fa1a04c538e3d9b8cd04bde8aR106

There's no obvious issues in migrating magical systemd units and static pods to daemonsets then?

danielfm:
@mumoshu In my tests, a pod that tolerates all possible taints can be specified like this:

```yaml
              tolerations:
              - operator: Exists
                effect: NoSchedule
              - operator: Exists
                effect: NoExecute
              - operator: Exists
                key: CriticalAddonsOnly
```

This would allow us to use DaemonSet for kube-proxy pods. 😎

danielfm:
Updated this PR to use the solution proposed by @redbaron.

Still testing to make sure I didn't break anything.

danielfm:
I performed several rollouts to test this change (workers-only, controllers-only, and controllers+workers), and they all seem to work as expected. Could you guys give another look?

I updated the PR description according to the new implementation.

Note: I'm planning on creating separate PRs to address other issues, such as adding support for the `AutoScalingReplacingUpdate` update policy, like the one used in the example given by @redbaron.

core/nodepool/config/templates/stack-template.json
```diff
@@ -398,6 +418,29 @@
...
+                    "sns:Publish"
+                  ],
+                  "Effect": "Allow",
+                  "Resource": { "Ref": "{{.LogicalName}}NodeDrainerLHT" }
+                },
+                {
+                  "Action": [
+                    "autoscaling:CompleteLifecycleAction",
+                    "autoscaling:DescribeAutoScalingInstances",
+                    "autoscaling:DescribeLifecycleHooks"
```
redbaron:
```
     "autoscaling:DescribeAutoScalingInstances" 
     "autoscaling:DescribeLifecycleHooks"
```

do not support condition keys: http://docs.aws.amazon.com/autoscaling/latest/userguide/control-access-using-iam.html#resource-level-permissions-auto-scaling

core/controlplane/config/templates/stack-template.json
```diff
@@ -258,7 +279,8 @@
               "Effect": "Allow",
               "Principal": {
                 "Service": [
-                  "ec2.{{.Region.PublicDomainName}}"
+                  "ec2.{{.Region.PublicDomainName}}",
+                  "autoscaling.{{.Region.PublicDomainName}}"
```
redbaron: why is it needed?

danielfm: Without this, the stack creation fails because AWS tries to publish a test message to the topic.

core/controlplane/config/templates/stack-template.json
```diff
@@ -715,6 +737,27 @@
...
+    "{{.Controller.LogicalName}}NodeDrainerLH" : {
+      "Properties" : {
+        "AutoScalingGroupName" : {
+          "Ref": "{{.Controller.LogicalName}}"
+        },
+        "DefaultResult" : "CONTINUE",
+        "HeartbeatTimeout" : "{{.Experimental.NodeDrainer.DrainTimeoutInSeconds}}",
+        "LifecycleTransition" : "autoscaling:EC2_INSTANCE_TERMINATING",
+        "NotificationTargetARN" : { "Ref": "{{.Controller.LogicalName}}NodeDrainerLHT" },
+        "RoleARN" : { "Fn::GetAtt": [ "IAMRoleController", "Arn" ] }
```
redbaron: I think it's better to create separate role and give `autoscaling.amazonaws.com` permission to assume that role, instead of allowing nodes themselves accessing this topic. not that it is dangerous, but simply not needed

redbaron: see use of `ASGNotificationRole` in example

danielfm: This topic is dedicated to that node draining hook, so I don't think allowing nodes themselves to access it would cause problems. Plus, it makes the stack JSON a bit smaller.

However, if you are strongly opposed to this, I can change it.

redbaron: Topic is indeed dedicated to draining hook, but you give `autoscaling.amazonaws.com` permissions to do much more than publishing to that topic.  Tight explicit permissions is always a safe play.

To reduce size you can create single SNS topic and IAM role to publish there in a root template and pass both to controller and worker templates

danielfm: Done.

core/nodepool/config/templates/stack-template.json
danielfm: Done.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1021,6 +1025,94 @@ write_files:
...
+                  env:
+                  - name: NODE_NAME
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: spec.nodeName
+                  command:
+                  - /bin/sh
+                  - -xec
+                  - |
+                    kubectl()  { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
```
mumoshu: I wonder why `/lib/ld-musl-x86=64.so.1` is required.
I'd appreciate it if you could add a comment that explains why 🌷

model/node_drainer_test.go
```diff
@@ -0,0 +1,94 @@
+package model
+
+import (
+	"testing"
+)
+
+func TestDrainTimeoutInSeconds(t *testing.T) {
```
mumoshu: Thanks as always for adding a test 👍

core/root/config/templates/stack-template.json
```diff
@@ -2,9 +2,45 @@
   "AWSTemplateFormatVersion": "2010-09-09",
   "Description": "kube-aws Kubernetes cluster {{.ClusterName}}",
   "Resources": {
+    "ASGNotificationTarget": {
+      "Type": "AWS::SQS::Queue"
+    },
```
mumoshu: I see that we aren't using this queue at all for listening incoming lifecycle events but polling via `aws autoscaling describe-auto-scaling-instances` to detect state transitions of ec2 instances, right?
Assuming my understanding is correct, I guess we'd better add some comment on this resource, so that we can clearly see that this resource is not actually used anywhere from code but just passed to the asg lifecycle resources in the control-plane stack and node pool stacks as required.
Also, naming this resource like `DummyASGNotificationT` would even better clarify the fact?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1072,7 +1164,7 @@ write_files:
             - replicationcontrollers
             verbs:
             - get
-          - apiGroups: ["policy"]
+          - apiGroups: [""]
```
mumoshu: Just for confirmation but you've changed this so that it refers to an existing api resource?
I thought so because I could see no api resource named `pods/eviction` in the [`policy` api group](https://github.com/kubernetes/kubernetes/blob/master/api/swagger-spec/policy.json).

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1021,6 +1025,94 @@ write_files:
...
+                    kubectl()  { /lib/ld-musl-x86_64.so.1 /opt/bin/hyperkube kubectl "$@"; }
+                    metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
+                    asg()      { aws --region="${REGION}" autoscaling "$@"; }
+
+                    INSTANCE_ID=$(metadata meta-data/instance-id)
+                    REGION=$(metadata dynamic/instance-identity/document | jq -r .region)
+                    [ -n "${REGION}" ]
+
+                    while sleep 30; do
+                      STATE=$(asg describe-auto-scaling-instances --region "${REGION}" --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
```
mumoshu: Rather than polling with aws cli, could we try to receive messages from the sqs queue more frequently?
Also, could we run this not in daemonset-managed pods but in a single pod in one of controller nodes managed by a deployment(like cluster-autoscaler and rescheduler do)?
So that we can further save cluster resource(1 drainer pod per node vs 1 drainer pod per cluster) and aws api calls(1 call/30sec/node vs 1 call/30sec/cluster)?

core/controlplane/config/templates/cloud-config-controller
redbaron: instead of publishing new image, we copy `hyperkube` binary from hyperkube image into awscli image. Hyperkube image is based on GNU Libc and awscli is based on musl.  `hyperkube` is not a static binary, therefore it requires ELF interpreter, which when build with GNU Libc is `/lib64/ld-linux-x86-64.so.2`. Under the hood when you run `./hyperkube` what actually running is `/lib64/ld-linux-s86-64.so2 ./hyperkube`, binary is given as an argument to interpreter. 

awscli image is based on Musl and therefore misses `/lib/lid-linux-x86-64.so.2` interpreter, it provides another one `/lib/ld-musl-x86_64.so.1` which we use here to load `hyperkube` binary

core/controlplane/config/templates/cloud-config-controller
redbaron: TLDR;  each node polling for own state makes things MUCH simpler and easier to implement.

Consuming SQS messages isn't trivial, it is an inherently stateful process: once message is consumed it need to be acted on (eviction initiated), on and reported as "consumed" to SQS.  There will be multiple states to juggle, 2 per each message consumed, as each message represents a single node planned for termination and SQS message confirmation state. There is also no selective receive when client can filter for messages it is willing to receive. 

You'd also need to cover cases when controller node goes down at unfortunate time when evictions are in progress and and extra case when controller node attempts to evict itself.  So well behaving ASG Roller would store instance eviction state in ConfigMap and not bother with SQS state management confirming messages as soon as they persisted to ConfigMap. 

This makes ASG Roller essentially a controller which would require it's own github project and development cycle.


... or you don't do it and just let few lines of bash script to poll for status :)

core/controlplane/config/templates/cloud-config-controller
mumoshu: @redbaron It totally make sense now. Thank you very much for the detailed explanation 👍

redbaron:
@danielfm, does it work with RBAC?

core/controlplane/config/templates/cloud-config-controller
mumoshu: @redbaron Thanks for the great explanation! I'm marking my own idea YAGNI now 😄 
Maybe we'd revisit this to implement my idea (with own github repo when one exceeds the rate limit in calling AWS APIs with the current method - which may or may not occur in the future.

danielfm:
@redbaron yes, it does.

core/controlplane/config/templates/cloud-config-controller
danielfm: Exactly. This is what fixes the issue mentioned in the PR description.

core/controlplane/config/templates/cloud-config-controller
danielfm: One thing I could do to alleviate the rate limiting issue is to run the loop less frequently, like every minute. I could change that if you guys think it's worth it.

mumoshu: @danielfm Thanks for the suggestion 👍 I believe we can make it configurable once it turned out that the 30 sec setting is problematic to someone in practice. Until then, we can go ahead with 30 which seems nice to me.

mumoshu:
@danielfm LGTM. Thanks for the awesome work 🎉 
@redbaron Thanks for your very detailed, helpful review 👍

Issue #654, opened 2017-05-16
Provision nodes with kubeadm


mumoshu:
Probably after kubeadm starts supporting multi-master & dedicated-etcd-nodes setup.

Relevant PR(which may or may not be merged): https://github.com/kubernetes/kubernetes/pull/44793

c-knowles:
I'm curious about what this will add to kube-aws given the [docs](https://kubernetes.io/docs/getting-started-guides/kubeadm/) about it. @mumoshu What are your thoughts on what kube-aws gains from this?

mumoshu:
@c-knowles Thanks for chiming in 👍 
My thought is written in #675 - In short, I expect the amount of code for configuring apiserver and kubelets and so on to be reduced when we completely moved to kubeadm.
What do you think?

c-knowles:
I think it seems aligned in some aspects like self hosting. If we are sticking with CloudFormation then kubeadm doesn't seems like 100% a natural fit based on what the docs say. It depends on your view around how much we bake in and how many moving parts on node initialisation.

mumoshu:
Probably adopting kubeadm should prevent issues like #793 in longer term
cc @danielfm

luxas:
> Probably after kubeadm starts supporting multi-master & dedicated-etcd-nodes setup.

kubeadm has supported external etcd clusters pretty much from the get go.
There is no out-of-the-box HA, no, but in situations where you can copy files around and have a loadbalancer, that's not a problem.

> Probably adopting kubeadm should prevent issues like #793 in longer term

That and much more is exactly why kubeadm should be used as the basic building block in a cluster.
See my blog post about this here: http://blog.kubernetes.io/2017/01/stronger-foundation-for-creating-and-managing-kubernetes-clusters.html

I think this should work fine, at least to start speccing out. Do you mind attending the kubeadm adoption working group meeting later today?
cc @justinsb 


See:
https://groups.google.com/forum/#!topic/kubernetes-sig-cluster-lifecycle/HHr7WphU_xE
https://groups.google.com/forum/#!topic/kubernetes-sig-cluster-lifecycle/v2TzSHwYe9I
https://docs.google.com/document/d/1KdXsLYiJYJdiRbtgZsx6qbHF4g_K-gAScB9Zs4avgzg/edit#heading=h.aly6m9xjbivl
[...]

Issue #534, opened 2017-04-13
Dedicated website for kube-aws Documentation


mumoshu:
Extracted from https://github.com/kubernetes-incubator/kube-aws/pull/526#issuecomment-293787922

mumoshu:
@calebamiles Hi caleb!
Would you mind providing us e.g. a permission to publish kube-aws documentation to https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html?

In old days of kube-aws when it was still at coreos/kube-aws, the documentation had been published to there.
I don't have access to there myself but @colhom had been kindly taken the role to do it for us.
Would you mind discussing here how should we maintain and publish the documentation site, after kube-aws is transferred to kubernetes-incubator?
Would you guys at CoreOS still like to have the doc hosted under coreos.com, in the first place? (I personally prefer that because kube-aws heavily rely on various softwares thankfully developed by CoreOS 😄

c-knowles:
@mumoshu what should we do here? I suggest we take the initiative and start our own docs site as we have quite a backlog of users asking for more docs.

mumoshu:
@c-knowles Agree to your suggestion - Would you mind if I created a dedicated site `https://kube-aws.github.io` backed by `kube-aws/docs` github repo while putting all the necessary documentation sources and build scripts in this repo?
This way we can handle whole the process under our control and the url for the doc is pretty 😄

c-knowles:
@mumoshu agree, that's exactly what I would expect and doesn't change our existing workflow so much. I'll start on it very soon.

c-knowles:
I just thought of one downside, documentation cannot be committed in the same PR as the feature. If you think this is ok, let's proceed. Otherwise if we put the docs here the URL without any custom domain would be https://kubernetes-incubator.github.io/kube-aws/.

mumoshu:
@c-knowles I'm assuming that we can still bundle a documentation update and a code improvemente a same PR. That's because I expect to put all the necessary documentation sources and build scripts in this repo but not kube-aws/docs repo.

smrutiranjantripathy:
@mumoshu and @c-knowles I would suggest it would be easier from the point of maintenance and creation to use gitbook.com.
Any thoughts ?

mumoshu:
@c-knowles If you by any chance have e.g. a prototype of the doc site in your laptop, would you mind sharing it in https://github.com/kube-aws/docs so that we can start improving it anyway? 😃
[...]

PR #675, opened 2017-05-24
Update ROADMAP.md

mumoshu:
@c-knowles @danielfm @redbaron @camilb Recently, I've implemented the revised CA support in #629 according to our roadmap.
The revised CA support being the last concrete objective that was achievable in short term, I believe it's time to discuss about the plan for upcoming releases 😃 
Any comments, requests, thoughts on the roadmap?

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -55,6 +55,22 @@ Please file an issue to make suggestions on this roadmap!
       * Dynamic reconfiguration of cluster-autoscaler
       * Automatic discovery of target node pools for cluster-autoscaler
     * Requires much work on CA side
+
+## v0.9.8
+
+  * Experimental support for kube-aws plugins
+  * Tiller installed by default
+    * For use from the plugin support
+
+## v0.9.9
+
+  * RBAC enabled by default
+  * [Optional] Several kube-aws core features as plugins
+
+## v0.9.10
+
+  * kubeadm support to simplify k8s components configuration
+    * while reducing the amount of code required in kube-aws
   
 ## v0.9.x
 


ROADMAP.md
```diff
@@ -55,6 +55,22 @@ Please file an issue to make suggestions on this roadmap!
       * Dynamic reconfiguration of cluster-autoscaler
       * Automatic discovery of target node pools for cluster-autoscaler
     * Requires much work on CA side
+
+## v0.9.8
+
+  * Experimental support for kube-aws plugins
```
mumoshu: Accordingly to #509 
I'm looking forward to @c-knowles's design proposal for this :bow:

ROADMAP.md
```diff
@@ -55,6 +55,22 @@ Please file an issue to make suggestions on this roadmap!
...
+  * Tiller installed by default
+    * For use from the plugin support
+
+## v0.9.9
+
+  * RBAC enabled by default
```
mumoshu: Accordingly to #655 
My plan is to first ensure all the features to be working with RBAC in v0.9.8.
It would be a matter of just start enabling it by default in v0.9.9 then.

ROADMAP.md
```diff
@@ -55,6 +55,22 @@ Please file an issue to make suggestions on this roadmap!
...
+  * [Optional] Several kube-aws core features as plugins
+
+## v0.9.10
+
+  * kubeadm support to simplify k8s components configuration
```
mumoshu: Accordingly to #654 
However, I'm not yet sure what clear and concrete benefits we can gain by doing so, at least at the moment of writing this.

redbaron:
I am working on porting things to Ignition

camilb:
I'm currently testing the integration of these add-ons :
* [external-dns](https://github.com/kubernetes-incubator/external-dns) 
* [aws-nginx-ingress](https://github.com/kubernetes/ingress/tree/master/examples/aws/nginx). This allow using a single ELB for multiple services. Also watching the progress of [alb-ingress-controller](https://github.com/coreos/alb-ingress-controller)
* [kube-lego](https://github.com/jetstack/kube-lego)

Working on some improvements:
* provide RBAC permissions for kube2iam, nginx ingress, external-dns
* simplify and improve security for dex integration

Future plans:
* option to deploy a in-cluster monitoring solution using Prometheus Operator with basic alerts and dashboards configured by default. Thinking to integrate some of the features I implemented in  [prometheus-kubernetes](https://github.com/camilb/prometheus-kubernetes)

redbaron:
My current plan in the order of pain points for us:

 * Ignition
 * Move Flannel and Calico to Kubernetes backends instead of Etcd
 * Self hosting Flannel & Calico via daemonset
 * Safe cluster rolls (coordinate ASG shutdown events and node drain calls)
 * Self-hosted Etcd via bootkube & etcd-operator

danielfm:
@redbaron 

> Safe cluster rolls (coordinate ASG shutdown events and node drain calls)

I'm currently testing this in #674.

redbaron:
@danielfm that improves node drainer behaviour, which is good, but it doesn't prevent 2 nodes going down and causing lets say ZK running in kube to lose quorum.

danielfm:
@redbaron could you elaborate on this in #674 please?

If there's anything I can do to accommodate this in that PR, please let me know.

c-knowles:
I'm mainly concentrating on plugins and will publish some design soon. Also looking at general stability. Several times when I updated kube-aws something broke and not because I did anything different/special or changed any of my setup. In that sense, we should double down on tests and think a little more about the goals of kube-aws and what should _not_ be included. I know @redbaron mentioned this same thing many times ;)

When there are a range of options for implementation of some aspect with no clear winner, what do we do? We'd need to pick one or support multiple options or some generic method to plug them in. How do we decide? I'm interested in why we wish to pick a certain tool over another. Some of the above seem natural choices, mostly the CoreOS projects or other Kubernetes org projects. A counter example is ingress controller choice - there a several, we're using the Traefik Helm chart which also does the job of kube-lego.

danielfm:
@c-knowles I was also thinking what should be the "right" scope for kube-aws.

Is it only supposed to create a working cluster - and the user must use other means to bootstrap the services running on top of it, such as ingress controllers, monitoring, etc -, or is it supposed to be a "batteries included" kind of tool?

I think the first approach is the best, but I wonder if there's any objective ways to determine which approach is the right one.

camilb:
@c-knowles I agree we should freeze the new features at some point and focus on stability. Features where we have multiple options, if possible, we should provide them as plugins for helm, like the nginx ingress + kube-lego or traefik and the user can choose which one is better for his needs. For other tools there is a more clear direction, like, for example,  "external-dns" which is a project that tries to replace all the other tools for automatic DNS configuration and can be directly integrated, deployed using helm or as an add-on. 

The features I'm looking at the moment, can be deployed separately. Maybe we can create a directory for add-ons and charts where we can have multiple examples of self-managed services that can be easy to deploy after the cluster is up. 

The "cluster.yaml" already became quite complex and sometimes there are options that not work together by default (like rbac + kube2iam for example) and will be even harder to test and maintain all the possible setups in the future.

c-knowles:
@danielfm The things I've built have assumed the first also due to what kube-aws was like when I started using it. I agree if there is a more objective way to assess that I'm happy to participate. The metrics I can think of are community support/involvement, ease of maintenance such as testing properly, clear separation of concerns.

@mumoshu listed a bunch of the goals of kube-aws, maybe we should list non-goals as well? If we stick to a CoreOS type ethos then it will be sensible, secure, HA defaults and most things will be modular and usable separately. As @camilb says we already have options that don't work together and the cluster.yaml is winding up towards 1500 lines by default when you start using the tool without doing any modifications!

redbaron:
For me kube-aws ends where working, stable and manageable Kubernetes API starts. nginx-ingress, monitoring,  anything else which runs on top of it already has Helm, where community collaborate on best practices.

mumoshu:
> Is it only supposed to create a working cluster - and the user must use other means to bootstrap the services running on top of it, such as ingress controllers, monitoring, etc -, or is it supposed to be a "batteries included" kind of tool?

Could be the latter, but certainly what would be included into kube-aws should be limited.
kube-aws is all about running production k8s clusters - what we need for a production k8s cluster is high scalability/availability/security/flexibility/customizability.

Especially while incubating, I'm rather happy to include many optional "integrations" or "supports" that contribute to easily achieve any of kube-aws goals above. For example, installing dex on a k8s cluster is not yet a matter of just running `helm install` as of today, and it also contribute to security which is one of kube-aws goals. That's why it made sense for me to include dex integration into kube-aws.

Regarding integrations or supports don't contribute to any of kube-aws goals, I'd like them to not included in kube-aws for now but after kube-aws plugins #509 has landed. Plugins contributing kube-aws goals may reside in `builtinplugins` and others may reside in something like `contrib/plugins` in this repo.

We should investigate and settle on a set of possible/necessary extension points for kube-aws to implement kube-aws plugins #509. Hardening all the extension points on kube-aws side would result in adding more integrations as kube-aws plugins not to break a cluster.

So -

> For me kube-aws ends where working, stable and manageable Kubernetes API starts. nginx-ingress, monitoring, anything else which runs on top of it already has Helm, where community collaborate on best practices.

I agree 👍 

> options that don't work together 

If the options were ones in kube-aws core and builtin plugins, kube-aws maintainers(oh, me) would be responsible to ensure they're working(or at least validated and forbid).
If the options were ones from kube-aws plugins, I believe plugin maintainers would be responsible.

Btw I believe we'd better improve our CI so that we can keep ensuring a lot of options to work.
However we don't yet have budget for e.g. a dedicated AWS account to allow us running many E2E tests in CI.

> cluster.yaml is winding up towards 1500 lines by default when you start using the tool without doing any modifications!

For this problem, I believe a dedicated kube-aws documentation site #534 with a bunch of example `cluster.yaml`s and detailed explanation per each setting key would make the everything-in-the-fat-`clulster.yaml` to be unnecessary. If we had such a doc, the default cluster.yaml should be just several lines. However I'm sure the doc takes much more time to maintain than the current fat `cluster.yaml`. I myself didn't happen to have much time to maintain a dedicated doc site so I have been came along with the fat cluster.yaml until now, even though I'm not happy with that.

----

Please poke me if anything is still unclear!
Thanks for your supports 👍

mumoshu:
@redbaron How are you planning to migrate to ignition?
I have been thinking of a hybrid(ignition calls cloud-config, both executed) at the first stage and gradually move systemd unit definitions from cloud-config to ignition so that each change won't be too big and we can fire and motion.

everpeace:
This is what I'm just thinking, and not short term, But I am wondering to support [Elastic GPU](https://aws.amazon.com/jp/ec2/Elastic-GPUs/) which is still under preview release.

Recently, I implement NVIDIA Driver installation support for GPU instances (#645) because I think easiness to spin up deep-learning-ready k8s cluster is important.  Elastic GPU support provides users spawn more flexible nodepools (medium instance  having huge GPUs or so).

This feature might not worth to be included `builtinplugins` which @mumoshu stated at this point.

c-knowles:
@mumoshu ah yes I should have added to my list documentation! I'm working on that but progress is slow as I'm concentrating on plugins design. I'll be back onto that afterwards.

Anyone interested in [istio](https://cloudplatform.googleblog.com/2017/05/istio-modern-approach-to-developing-and.html) ?

mumoshu:
@c-knowles I'm interested in istio and also linkerd - not sure how they are different in which ways yet

c-knowles:
@mumoshu best places I found with comparisons between istio and linkerd:
https://news.ycombinator.com/item?id=14410533
https://lyft.github.io/envoy/docs/intro/comparison.html

From what I've read so far, istio is definitely our preference.

mumoshu:
Playing with k8s-bot...

PR #676, opened 2017-05-24
Ensure TLSBootstrapToken related code is not written if TLSBootstrap is disabled

jollinshead:
Entries that use the notation **{{ if .AssetsConfig.HasAnyAuthTokens }}** always return true because a TLSBootstrapToken is auto-generated by default.

If there are no AuthTokens and  TLSBootstrap is disabled then the statement **{{ if .AssetsConfig.HasAnyAuthTokens }}**  still returns true. Consequently, code is included in the config-controller file which causes kube-apiserver failures (it cannot find files related to tokens).

To avoid this issue we should ensure the existence of a TLSBootstrapToken is ignored in checks if TLSBootstrap is disabled.


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -991,7 +991,7 @@ write_files:
           -ec \
           'echo decrypting assets
            shopt -s nullglob
-           for encKey in /etc/kubernetes/{ssl,{{if .AssetsConfig.HasAnyAuthTokens}}auth{{end}}}/*.enc; do
+           for encKey in /etc/kubernetes/{ssl,{{ if or (.AssetsConfig.HasAuthTokens) ( and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken) }}auth{{end}}}/*.enc; do
```
danielfm: @jollinshead Does this mean we can get rid of `HasAnyAuthTokens` then?

jollinshead: There would only be one use of ```HasAnyAuthTokens``` (currently on line 680 in config-worker) ```           for encKey in /etc/kubernetes/{ssl,{{if and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasAnyAuthTokens}}auth{{end}}}/*.enc; do``` I left this alone since I wasn't sure if this logic was meant to be implemented as shown. 
If we can replace the logic used here with ```{{ if or (.AssetsConfig.HasAuthTokens) ( and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken) }}```, then we can remove the HasAnyAUthToken method since it would no longer be used.

danielfm: @jollinshead I think this logic can be replaced as well without breaking the intended behavior.

jollinshead:
@danielfm Done

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -677,7 +677,7 @@ write_files:
           -ec \
           'echo decrypting assets
            shopt -s nullglob
-           for encKey in /etc/kubernetes/{ssl,{{if and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasAnyAuthTokens}}auth{{end}}}/*.enc; do
+           for encKey in /etc/kubernetes/{ssl,{{ if and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken }}auth{{end}}}/*.enc; do
```
mumoshu: Really just curious but why extra spaces? 😃

mumoshu:
@jollinshead LGTM. Thanks for the improvement 👍 
@danielfm Thanks for reviewing 👍

PR #678, opened 2017-05-25
Kill dead code and restore DNS config validation

redbaron:


mumoshu:
@redbaron Thanks a lot for the cleanup. It looks much nicer 👍

PR #680, opened 2017-05-26
Remove ancient CoreOS version check which is not relevant anymore

redbaron:


glide.yaml
```diff
@@ -15,10 +15,6 @@ import:
   version: ^1.12.0
   subpackages:
   - config/validate
-- package: github.com/coreos/go-semver
```
mumoshu: Ok so we will see go-semver removed from `vendor/` on next dependency update 👍

mumoshu:
@redbaron Good catch 👍

PR #682, opened 2017-05-29
Allow userdata to be split across multiple parts

redbaron:
PR is split into 2 commits for ease of review: vendor only and code change only.

This change is a pure refactoring of userdata handling with zero
    user-visible changes.

Userdata now is a multipart template, each part can be processed
    independently. In the current form parts can be templated into
    instance UserData and uploaded to S3 bucket.

2 new template functions are available to userdata templates:

  * `extra`: caller can pass in place extra context, which then acessible
      to userdata as a result of `extra` function (see Etcd userdata as example)
  * `self`: different part of userdata template can refer to UserData object
      using `self` function. This makes templats completely self-aware

This PR also makes userdata handling to be better integrated with
    cfnstack assets management. Asset management is responsible for storing assets
    on S3 and therefore acts as a source of truth for all names, S3 URLs etc. Once
    part is added to Asset manager via `AddUserDataPart`, `Asset` field on part is
    then set to contain infromation about location of the given part, it's final
    rendered content etc.

  - [X] tests passing
  - [X] try creating cluster with new code


core/controlplane/config/templates/stack-template.json
```diff
@@ -692,25 +696,11 @@
...
-          "   {{$.AWSCliImage.Options}}{{$.AWSCliImage.RktRepo}} --exec=aws -- s3 --region $REGION  cp {{ $.UserDataEtcdS3URI }} /var/run/coreos/$USERDATA_FILE; do",
-          "sleep 1",
-          "done",
-          "exec /usr/bin/coreos-cloudinit --from-file /var/run/coreos/$USERDATA_FILE"
-        ]]}}
+        "PlacementTenancy": "{{$.Etcd.Tenancy}}"
+        {{- if $.UserDataEtcd.Parts.instance }}
+        ,
+          "UserData": {{ $.UserDataEtcd.Parts.instance.Base64 true (dict "etcdIndex" $etcdIndex) | checkSizeLessThan 16384 | quote }}
+        {{- end}}
```
mumoshu: Suppose we must always have the instance part of user-data, could you omit this `if-else` block like:

```
"PlacementTenancy": "{{$.Etcd.Tenancy}}",
"UserData": {{ $.UserDataEtcd.Parts.instance.Base64 true (dict "etcdIndex" $etcdIndex) | checkSizeLessThan 16384 | quote }}
```

so that the UserData in the stack template won't be accidentally emptied when we broke cloud-config-etcd?

core/controlplane/config/stack_config.go
```diff
@@ -128,35 +55,11 @@ func (c *StackConfig) EtcdSnapshotsS3PrefixRef() (string, error) {
...
 
-func (c *StackConfig) ValidateUserData() error {
-	err := userdatavalidation.Execute([]userdatavalidation.Entry{
-		{Name: "UserDataWorker", Content: c.UserDataWorker},
-		{Name: "UserDataController", Content: c.UserDataController},
-		{Name: "UserDataEtcd", Content: c.UserDataEtcd},
-	})
-
-	return err
-}
```
mumoshu: I believe we'd better not to remove ValidateUserData but update it to validate against the `s3` and `instance` parts of each UserData like:
```
err := userdatavalidation.Execute([]userdatavalidation.Entry{
  {Name: "UserDataWorker", Content: c.UserDataWorker.Parts.instanec},
  {Name: "UserDataWorker", Content: c.UserDataWorker.Parts.s3},
  {Name: "UserDataController", Content: c.UserDataController.Parts.instance},
  {Name: "UserDataController", Content: c.UserDataController.Parts.s3},
  {Name: "UserDataEtcd", Content: c.UserDataEtcd.Parts.instance},
  {Name: "UserDataEtcd", Content: c.UserDataEtcd.Parts.s3},
})
```

model/userdata_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"testing"
+)
+
+const (
+	s3Body = `#cloud-config
+---
+coreos: {}
+`
+	USERDATA_S3       = "s3"
+	USERDATA_INSTANCE = "instance"
```
mumoshu: Should we define these in `userdata.go` instead so that we can use these to avoid hard-coded strings "s3" and "instance"?

model/userdata.go
```diff
@@ -0,0 +1,147 @@
...
+type UserData struct {
+	Parts map[string]UserDataPart
+}
+
+type UserDataPart struct {
+	Type     UserDataPartType
+	Asset    Asset
+	tmpl     *template.Template
+	tmplData interface{}
+}
```
mumoshu: My preference is make `UserDataPart` an interface rather than a struct and implement the interface as `UserDataS3Part` and `UserDataInstancePart` for respective `UserDataPartType`, so that the UserDataPartType and the `switch` block in the `validate` func can be completely omitted(=less cyclomatic complexity)

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1,3 +1,21 @@
+{{ define "instance" -}}
+{{- $S3URI := self.Parts.s3.Asset.URL -}}
```
mumoshu: Assuming the only reference to `self` is this, I prefer `parts.s3.Asset.URL` rather than `self.Parts.s3.Asset.URL` so that we won't expose too much things to the cloud-config template.

Currently, `self` is only used to access `Parts`, right? Then, we can safely remove `self` in favor of an another func for directly accessing `Parts`(would be named `parts` in the FuncMap)?

mumoshu:
@redbaron I've added several comments but overall, this is an EXCELLENT JOB! The code looks much cleaner and easier to understand now 👍

core/controlplane/config/stack_config.go
redbaron: Because template rendering is now lazy, validation is happening at render time inside `.Template()` function

model/userdata.go
redbaron: then you replacing  `switch` in `validate`  with another in `NewUserData` to choose which part type to create + sprinkle interfaces implementations which are pretty much common between all parts.

core/controlplane/config/templates/cloud-config-controller
redbaron: Idea behind `self` was to give access part template to a whole UserData object, so that it doesn't need to depend on things like `.UserDataController.Parts` etc. Whole object is accessible so that if interface grows more complex, all those methods are available for template to use. Right now it is just `self.Parts` though, which I agree looks redundant.

`.Parts` is an artifact of previous iterations on this design. Now that it is settled, I might as well just drop it all together and go with `type UserData map[string]UserDataPart`, then it would be `self.s3.Asset.URL`,  what would you say?

redbaron:
Next refactoring steps (will not part of this PR) which I had to skip due to lack of time, but leaving here somebody wants to pick up broom and continue this cleaup:
 - move Etcd snapshot path handling from `controlplane/config/stack_config.go` to `cfnstack.Assets`, making it the only place in code responsible for our S3 layout.
 - rename `model.UserData` to ` model.TemplatedAsset` and migrate stack.json rendering/validating logic to it
 - now that all `StackConfig` are nothing more that stubs, delete them all
 - with StackConfig gone, clenaup mess in `controller/cluster/cluster.go` where there is ClusterRef with config embedded with StckConfig and Provided/computed config lurking somewhere around. Preferably delete them all :)
 - finally unify `root.Cluster`, `nodepool.Cluster` and `controlplane.Cluster` as a single class. Not an interface with tons of duplications.

model/userdata.go
mumoshu: True 👍 However my best bet would still be moving `switch` to `NewUserData` because then we can test it by just calling `NewUserData`?

core/controlplane/config/templates/cloud-config-controller
mumoshu: I like `self.s3.Asset.URL` 👍 

Btw, I'd prefer the `UserData` type to be a struct with fields like `s3` and `instance` rather than a map. For me. map[string] looks a bit confusing - Even though the only expected keys for it are `s3` and `instance`, `map[string]` doesn't tell that to us.

model/userdata.go
redbaron: Each and single part type interface implementation would need to implement their own `Template()` then right? As it can't be part of a *base* part, because we call `.Validate()` in there which is itself not part of interface. AFAIK, there is no C++-like virtual methods in Go, where base class can call methods on derived classes.

core/controlplane/config/stack_config.go
mumoshu: Thanks! That's awesome 👍

core/controlplane/config/templates/cloud-config-controller
redbaron: It was one of the designs I tried, it ended up being more verbose.

model/userdata.go
mumoshu: I'd try introducing a new type `userDataPart` which has everything currently implemented for `UserDataPart` but `.validate()`. Then, it could be embedded into `UserDataInstancePart` and `UserDataS3Part` like:

```golang
type userDataPart struct {
	Asset    Asset
	tmpl     *template.Template
	tmplData interface{}
}

type UserDataInstancePart struct {
	userDataPart
}

type UserDataS3Part struct {
	userDataPart
}

func (self userDataPart) Base64(compress bool, extra ...map[string]interface{}) (string, error) {
	//...
}

func (self userDataPart) Template(extra ...map[string]interface{}) (string, error) {
	//...
}

func (self UserDataInstancePart) validate(content []byte) error {
	//...
}

func (self UserDataS3Part) validate(content []byte) error {
	//...
}
```

I believe this is still arguably a trade-off between verbosity and testability(?). WDYT?

redbaron: That is what I am talking about, `Template()` won't be able to call `validate()`

see https://play.golang.org/p/3U71sZ86b4

redbaron: I'll change `validate` to be a function pointer which then passed instead  of UserDataPartType value, that way validation can be as complicated as needed, even CF stack can be validated then

redbaron:
@mumoshu , `checkSizeLessThen` complains about number of arguments, yet it should receive 2, do you have any idea why does it fail?

mumoshu:
@redbaron Seems like you need [some initialization](https://github.com/HotelsDotCom/kube-aws/blob/9edd032e0a85d74920559249203bd4ecc8e71e5c/core/controlplane/cluster/cluster.go#L131-L145) before [this line](https://github.com/HotelsDotCom/kube-aws/blob/9edd032e0a85d74920559249203bd4ecc8e71e5c/core/controlplane/config/stack_config_test.go#L46) - `RenderStackTemplateAsString` requires userdata parts "instance" and "s3", which seems to be missing in the failing test case, eventually causing the wrong argument number error due to [the issue in text/template](https://github.com/golang/go/issues/18716).

For example, when "instance" part is missing in the `Parts` map, I had been expecting `.Parts.instance.Base64` to return an empty string produced by `Base64` on an empty UserDataPart struct. However, surprisingly, it actually returns something looks like `nil` of `interface{}`, which is eventually causing the confusing error.

FYI here's my snippet for reproducing the issue: https://play.golang.org/p/YyOI677EjX

core/controlplane/config/templates/cloud-config-controller
mumoshu: Especially after seeing your design decision caused the confusing "wrong number of argument" error in an edge-case, I still prefer avoiding maps in exchange for verbosity. 

However, I completely agree that going ahead with my suggestion would do result in more verbose code!
I'll leave the decision to you.

redbaron: I see, that is default template behaviour where missing map key produces "nothing": https://golang.org/pkg/text/template/#Template.Option

It is more of a case, Cluster objects are produced in a not fully usable state and certain methods must be called in certain order before other methods can be called at all :)

I'll move `Cluster.Assets()` insdie `Cluster` constructor where it belongs to, so that when `Cluster` object is initialized and is ready to be used

mumoshu:
Hi @redbaron, just checking in but did you have any chance to finish this?
The definition of "finish" is up to you but I'd certainly want to merge this once the tests were fixed.

redbaron:
Tests are passing, but they are pretty useless as the don't test whole cluster, just individual subclusters. (and cluster term being HEAVILY overloaded in a codebase :( ). Frankly I just made it to pass, it is still very untidy, most importantly for me, currently there can be `Cluster{}` instances which are not fully initialized. I moved `.Assets()` call inside constructor, but `.Load()` calls are still necessary to fully initialize instances

redbaron:
@mumoshu we have trouble accessing our AWS infrastructure, would you mind running e2e tests to see if change doesn't break anything?

mumoshu:
@redbaron Thanks for taking time! Sure. I'll run it tomorrow and report
back.
2017年6月14日(水) 21:39 Maxim Ivanov <notifications@github.com>:

mumoshu:
Running e2e tests...

mumoshu:
```
Creating AWS resources. Please wait. It may take a few minutes.
Error: Error creating cluster: Stack creation failed: CREATE_FAILED : The following resource(s) failed to create: [Controlplane].

Printing the most recent failed stack events:
CREATE_FAILED AWS::CloudFormation::Stack pr682 The following resource(s) failed to create: [Controlplane].
CREATE_FAILED AWS::CloudFormation::Stack Controlplane Embedded stack arn:aws:cloudformation:ap-northeast-1:738575627980:stack/pr682-Controlplane-14AG446P2P9RF/ac8a0980-515f-11e7-a530-50fa13f400ad was not successfully created: The following resource(s) failed to create: [Etcd0].
```

mumoshu:
Full journal from the etcd node can be seen at https://gist.github.com/mumoshu/b79435fc842a87969676ec4240d284c2

`aws s3 cp` command seems to be consistently failing

```
Jun 15 00:18:48 ip-10-0-0-107.ap-northeast-1.compute.internal bash[935]: [   38.306250] awscli[5]: usage: aws s3 cp <LocalPath> <S3Uri> or <S3Uri> <LocalPath> or <S3Uri> <S3Uri>
Jun 15 00:18:48 ip-10-0-0-107.ap-northeast-1.compute.internal bash[935]: [   38.307025] awscli[5]: Error: Invalid argument type
```

mumoshu:
User-data taken from the etcd node follows:

```
$ curl http://169.254.169.254/latest/user-data | gunzip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   486  100   486    0     0   689k      0 --:--:-- --:--:-- --:--:--  474k
#!/bin/bash -xe
echo 'KUBE_AWS_ETCD_INDEX=0' >> /var/run/coreos/etcd-node.env
 . /etc/environment
export COREOS_PRIVATE_IPV4 COREOS_PRIVATE_IPV6 COREOS_PUBLIC_IPV4 COREOS_PUBLIC_IPV6
REGION=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r '.region')
USERDATA_FILE=userdata-etcd
while ! /usr/bin/rkt run \
   --net=host \
   --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf  \
   --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos \
   --trust-keys-from-https \
   quay.io/coreos/awscli:master --exec=aws -- s3 --region $REGION  cp https://s3.amazonaws.com//  /var/run/coreos/$USERDATA_FILE; do
  sleep 1
done
exec /usr/bin/coreos-cloudinit --from-file /var/run/coreos/$USERDATA_FILE
```

redbaron:
Thank you, I'll have a look and maybe add some tests

redbaron:
@mumoshu , I believe this was fixed and can be E2E tested

mumoshu:
Thanks @redbaron, e2e tests are now running...

mumoshu:
@redbaron It seems like we're still seeing the same-looking error. journald log and the userdat taken from the etcd node can be seen at https://gist.github.com/mumoshu/93ca35ea0f080ccd2e16f39277e11878

mumoshu:
I see `cp https://s3.amazonaws.com/cwtest-kube-aws-templates/clusters/pr682b/kube-aws/clusters/pr682b/exported/stacks/control-plane/userdata-etcd-90c9d79a19d687a807db09eb99b4ce52ecc8a1891204551e2726c42a7e0cd305  /var/run/coreos/$USERDATA_FILE` as arguments for the awscli invocation but shouldn't the first arg to `cp` start with the s3 scheme `s3://`?

redbaron:
@mumoshu how to run e2e tests myself? how AWS account should be set up?

mumoshu:
@redbaron Would the [e2e README](https://github.com/kubernetes-incubator/kube-aws/blob/master/e2e/README.md) help?

For example, I've run e2e against this pull request with:

```
$ git checkout pr682
$ git pull origin pull/682/head
$ cd e2e
$ ETCD_DISASTER_RECOVERY_AUTOMATED=1 ETCD_SNAPSHOT_AUTOMATED=1 ETCD_COUNT=3 CONTROLLER_COUNT=1 KUBE_AWS_CLUSTER_NAME=pr682b bash -c './run all'
```

redbaron:
@mumoshu didn't have chance to try e2e, please check if you have time.  I was rebased and squashed all but vendor update commit

mumoshu:
@redbaron Thanks for the fix 👍 It seems to have progressed but I'm still seeing the error below:

```
Jun 22 02:03:40 ip-10-0-0-149.ap-northeast-1.compute.internal cfn-etcd-environment[949]: [   39.453409] awscli[5]: #!/bin/bash -vxe
Jun 22 02:03:40 ip-10-0-0-149.ap-northeast-1.compute.internal cfn-etcd-environment[949]: [   39.454033] awscli[5]: cfn-init -v -c "etcd-server" --region ap-northeast-1 --resource Etcd$KUBE_AWS_ETCD_INDEX --stack $KUBE_AWS_STACK_NAME
Jun 22 02:03:40 ip-10-0-0-149.ap-northeast-1.compute.internal cfn-etcd-environment[949]: [   39.454488] awscli[5]: + cfn-init -v -c etcd-server --region ap-northeast-1 --resource Etcd0 --stack
Jun 22 02:03:40 ip-10-0-0-149.ap-northeast-1.compute.internal cfn-etcd-environment[949]: [   39.600918] awscli[5]: Usage: cfn-init [options]
Jun 22 02:03:40 ip-10-0-0-149.ap-northeast-1.compute.internal cfn-etcd-environment[949]: [   39.601266] awscli[5]: cfn-init: error: --stack option requires an argument
```

https://gist.github.com/mumoshu/a9985bca2abfa7054aec3fa59717a6bd

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -6,14 +6,28 @@ echo '{{.EtcdIndexEnvVarName}}={{extra.etcdIndex}}' >> {{.EtcdNodeEnvFileName}}
...
+  done
+}
+aws s3 --region $REGION  cp {{ $S3URI }}  /var/run/coreos/$USERDATA_FILE
+
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+STACK_NAME=$(aws ec2 describe-tags --region "$REGION" --filters \
+  "Name=resource-id,Values=$INSTANCE_ID" \
+  "Name=key,Values=aws:cloudformation:stack-name" \
+  --output json \
+| jq -r ".Tags[].Value")
```
mumoshu: @redbaron Now jq seems to be complaining like:
```
jq: error (at <stdin>:1): Cannot index array with string "Tags"
parse error: Invalid numeric literal at line 1, column 22
```

I wondered why it happened at first glance but running the aws command alone revealed the issue:

```
# /usr/bin/rkt run --net=host --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos --trust-keys-from-https quay.io/coreos/awscli:master --exec=aws -- ec2 describe-tags --region ap-northeast-1 --filters Name=resource-id,Values=i-0e4224cb0ffdb4ce7 Name=key,Values=aws:cloudformation:stack-name --output json
[ 1506.407457] awscli[5]: {
[ 1506.408016] awscli[5]:     "Tags": [
[ 1506.408338] awscli[5]:         {
[ 1506.408610] awscli[5]:             "ResourceType": "instance",
[ 1506.408878] awscli[5]:             "ResourceId": "i-0e4224cb0ffdb4ce7",
[ 1506.409197] awscli[5]:             "Value": "pr682a-Controlplane-11TWK5UX7W875",
[ 1506.409551] awscli[5]:             "Key": "aws:cloudformation:stack-name"
[ 1506.409863] awscli[5]:         }
[ 1506.410205] awscli[5]:     ]
[ 1506.410519] awscli[5]: }
```

Adding the `--interactive` flag to `rkt run` seems to resolve the issue:

```
/usr/bin/rkt run --interactive --net=host --volume=dns,kind=host,source=/etc/resolv.conf,readOnly=true --mount volume=dns,target=/etc/resolv.conf --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false --mount volume=awsenv,target=/var/run/coreos --trust-keys-from-https quay.io/coreos/awscli:master --exec=aws -- ec2 describe-tags --region ap-northeast-1 --filters Name=resource-id,Values=i-0e4224cb0ffdb4ce7 Name=key,Values=aws:cloudformation:stack-name --output json
{
    "Tags": [
        {
            "ResourceType": "instance",
            "ResourceId": "i-0e4224cb0ffdb4ce7",
            "Value": "pr682a-Controlplane-11TWK5UX7W875",
            "Key": "aws:cloudformation:stack-name"
        }
    ]
}
```

redbaron:
@mumoshu I have created/updated existing clusters with this PR merged, seems to be coming up OK as expected, feel free to e2e test it (probably worth doing together with cloudwatch logs PR)

mumoshu:
@redbaron E2E passed & LGTM. Thank you very much for your efforts 👍

redbaron:
@mumoshu thanks for your patience and review!

Issue #685, opened 2017-05-30
Add Heapster service account


c-knowles:
Our Heapster deployment should probably be using a separate service account similar to the [Heapster repo](https://github.com/kubernetes/heapster/tree/master/deploy/kube-config) and [kops](https://github.com/kubernetes/kops/blob/master/addons/monitoring-standalone/v1.6.0.yaml)

c-knowles:
I tried adding the RBAC setup above to a dev cluster and got the same issue as https://github.com/kubernetes/heapster/issues/1589, https://github.com/kubernetes/kubeadm/issues/248.

I can no longer find the ref to the heapster-nanny so wondering if it's still needed. @mumoshu any ideas about that?

c-knowles:
I've done the changes locally, will submit a PR but the nanny still isn't working (doesn't seem to work already with RBAC).

c-knowles:
PR above, still an issue with the nanny sidecar though.

mumoshu:
@c-knowles Thanks for your help here 👍 
Sry if i'm not answering your question correctly, but AFAIK the nanny sidecar is used toy constantly update a heapster deployment to set more/less resource limits according to cluster size.

Relevant code in nanny: https://github.com/kubernetes/contrib/blob/master/addon-resizer/nanny/nanny_lib.go#L109

So, it makes sense for me to provide the heapster role permission(s) required to update a heapster deployment.

c-knowles:
@mumoshu I think the problem is then heapster will have a privilege escalation path. Do you know of other examples of nanny being used? I was trying to find some other similar examples to see how they are tackling this issue.

mumoshu:
@c-knowles Sorry but I'm not aware of any other example usage of nanny. I'm looking forward to see @piosz's reply in the upstream issue, too.

piosz:
cc @mwielgus @x13n

x13n:
Right now only heapster and eventer are using nanny for vertical scaling. But since this is all a single deployment, this problem does not show up anywhere else. As I wrote in https://github.com/kubernetes/heapster/issues/1595#issuecomment-307470310, short term solution would be just to give those permissions to Heapster.

c-knowles:
I found https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/heapster-rbac.yaml which seems like it's what we need.

[...]

PR #686, opened 2017-05-31
Heapster service account and cluster role binding

c-knowles:
For https://github.com/kubernetes-incubator/kube-aws/issues/685.

After this PR we still have these logs:

```
heapster-v1.3.0-2152587771-9cg3w heapster-nanny E0531 09:40:59.332643       1 nanny_lib.go:110] the server does not allow access to the requested resource (get deployments.extensions heapster-v1.3.0)
```

I agree with the comment https://github.com/kubernetes/heapster/issues/1589#issuecomment-291184395, it seems unwise to allow heapster to modify it's own config.

However, without this PR heapster constantly cycles replica sets when RBAC is enabled so at least this way it does not churn the pod and we have heapster running more constantly which is a start.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1249,6 +1249,22 @@ write_files:
           name: kube-aws:node-extensions
           apiGroup: rbac.authorization.k8s.io
 
+  # Allow heapster access to the built in cluster role via its service account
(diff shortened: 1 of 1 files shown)


mumoshu:
@c-knowles LGTM. Thanks a lot for the contribution 👍 
I've also left a comment about nanny in the original issue.

PR #687, opened 2017-06-01
calico-node - DaemonSet tolerating all taints

jeffersongirao:
When using taints in Node Pools together with calico the stack
always fail in Cloudformation since cfn-signal sytemd unit checks
if calico is up before signaling success


mumoshu:
@jeffersongirao Thank you very much for the good catch!
The change LGTM. I'll merge this once you've singned CLA.

jeffersongirao:
@mumoshu thanks! I've just signed, I guess it will re-validate it soon.

mumoshu:
@jeffersongirao Thanks for your effort 👍

PR #700, opened 2017-06-13
Enhance cluster.yaml documentation

ytsarev:
* Change confusing example for controller nodeLabels
* Remove outdated and scary statement for rbac enablement


mumoshu:
@ytsarev Thanks for the good catch 👍

Issue #691, opened 2017-06-02
ETCd crash after few days


Vrtak-CZ:
Hello,

we experience crash 2 of 3 etcd servers see logs bellow. But what is really bad we have no access to whole k8s cluster due etcd errors (timeout).

```
core@ip-10-101-3-236 ~ $ journalctl -u etcdadm-check.service -f

-- Logs begin at Wed 2017-05-31 22:08:13 UTC. --
Jun 02 15:38:09 ip-10-101-3-236.eu-west-1.compute.internal systemd[1]: etcdadm-check.service: Unit entered failed state.
Jun 02 15:38:09 ip-10-101-3-236.eu-west-1.compute.internal systemd[1]: etcdadm-check.service: Failed with result 'timeout'.
Jun 02 15:38:21 ip-10-101-3-236.eu-west-1.compute.internal systemd[1]: Starting etcd health check...
Jun 02 15:38:24 ip-10-101-3-236.eu-west-1.compute.internal systemctl[11792]: active
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_CACERT="/etc/ssl/certs/ca.pem"
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_CA_FILE="/etc/ssl/certs/ca.pem"
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_CERT="/etc/ssl/certs/etcd-client.pem"
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_CERT_FILE="/etc/ssl/certs/etcd-client.pem"
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_KEY="/etc/ssl/certs/etcd-client-key.pem"
Jun 02 15:38:27 ip-10-101-3-236.eu-west-1.compute.internal etcdadm[11806]: declare -x ETCDCTL_KEY_FILE="/etc/ssl/certs/etcd-client-key.pem"
Jun 02 15:40:24 ip-10-101-3-236.eu-west-1.compute.internal systemd[1]: etcdadm-check.service: Start operation timed out. Terminating.
[...]

PR #705, opened 2017-06-16
Use docker instead of rkt for regular etcdadm tasks (#4)

ytsarev:
Motivation is to avoid serious memory leak on etcd nodes as in
* https://github.com/coreos/bugs/issues/1927

Also it potentially mitigates issues https://github.com/kubernetes-incubator/kube-aws/issues/691 and https://github.com/kubernetes-incubator/kube-aws/issues/660 as the symptoms looks similar to our case.

### DIFF
--- a/etcdadm/etcdadm
+++ b/etcdadm/etcdadm
@@ -493,20 +493,14 @@ member_restore_from_local_snapshot() {
     rm -rf "$restored_dir"
   fi
 
-  _run_as_root rkt run \
-      --insecure-options=image \
-      --set-env ETCDCTL_API=3 \
-      --dns=host \
-      --net=host \
-      --volume $(member_snapshots_dir_name),kind=host,source="$(member_host_snapshots_dir_path)" \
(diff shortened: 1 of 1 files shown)


mumoshu:
@ytsarev Thank you very much for your effort on this 👍 I've read through/connected all the referenced links in mind and this change looks good to me.

billyteves:
@mumoshu is the issue also exists in v0.9.6?

mumoshu:
Yes, I believe so, unfortunately.
2017年6月21日(水) 18:25 billy ray teves <notifications@github.com>:

VinceMD:
If I select containerRuntime: docker in cluster.yml, is that supposed to use docker for etcd setup? or this cluster.yml option is only used for Workers? What about Controllers?

mumoshu:
@VinceMD It is supposed to use docker only for containers managed by kubelets when `containerRuntime: docker` is set. Both controllers and workers runs kubelets for scheduling various pods, so controllers do get affected by the option.

On the other hand, etcd isn't managed by k8s/kubelets hence it is always run by rkt.

PR #706, opened 2017-06-16
Do not leave thousands of exited containers from awscli

ytsarev:


etcdadm/etcdadm
```diff
@@ -332,9 +332,9 @@ _docker_awscli_command() {
   dir=$(dirname "$(config_state_dir)")
   echo docker
   echo run
+  echo --rm
   printf "%s\n" "-e"
   echo "AWS_DEFAULT_REGION=$aws_region"
-  echo "--rm"
```
mumoshu: Just for my education but why does moving the echo command from here to the above resolve the issue?

mumoshu: @ytsarev Sorry for bothering but ping!

redbaron: it changes nothing, probably was part of experiments which got commited

ytsarev: it *should* change nothing but it actually solved the issue in my case, i'm 100% sure.  I will need some time for reproduction to explain that :) The way how the command line gets constructed here is a bit confusing

mumoshu: @ytsarev Thanks for the confirmation 👍
I tried to reproduce the issue by running `docker run -e FOO=BAR --rm ...` and `docker run --rm -e FOO=BAR` on my docker v1.11.1 running in minikubeVM but had no luck so far.
Probably there may be unknown variable(s) other than the order of command-line args causing the issue.

ytsarev:
Can't reproduce it, probably was a coincidence with parallel activities around `etcdadm`. Sorry for confusion

PR #710, opened 2017-06-21
Fix node drain error when trying to evict pods from jobs

danielfm:
Without this permission, the `kube-node-drainer` unit fails when the node being drained contain job pods:

```
User "system:node:<name>" cannot get jobs.batch in the namespace "<ns>". (get jobs.batch <job-name>): <pod-name>; User "system:node:<name>" cannot get jobs.batch in the namespace "<ns>". (get jobs.batch <job-name>): <pod-name>
```


mumoshu:
@danielfm Again, good catch! Thanks for the fix 👍

PR #711, opened 2017-06-21
Remove unused sysctl override for nf_conntrack_max

### LINKED ISSUE #708: sysctl parameter net.netfilter.nf_conntrack_max not being loaded correctly
If we enable Calico, the following files are created:

```
# /etc/modules-load.d/nf.conf
nf_conntrack

# /etc/sysctl.d/nf.conf
net.netfilter.nf_conntrack_max=1000000
```

But, for some reason, this value is not being applied unless we manually restart the `systemd-sysctl` service:

```
# Before manually restarting the service
$ sysctl net.netfilter.nf_conntrack_max
net.netfilter.nf_conntrack_max = 131072

$ sudo systemctl restart systemd-sysctl
$ sysctl net.netfilter.nf_conntrack_max
net.netfilter.nf_conntrack_max = 1000000
```

My guess is the following snippet - which apparently has the job of restarting the `systemd-sysctl` service in order to load the custom `nf_conntrack_max` value - is not working:

```
    # https://github.com/coreos/docs/blob/5d7b1cccb8286185275b07db1495828be9fdb0ea/os/other-settings.md#tuning-sysctl-parameters
    - name: systemd-modules-load.service
      command: restart
    - name: systemd-sysctl.service
      command: restart
```

danielfm:
Kube-proxy, by default, [automatically sets](https://kubernetes.io/docs/admin/kube-proxy/) `nf_conntrack_max` to `<n-cores> * <value of conntrack-max-per-core>` when it starts, so it's useless to try to override `nf_conntrack_max` in some systemd unit.

If we wish to override this value, the best way to do so is by providing the `--conntrack-min` (or `--conntrack-max-per-core`) argument to kube-proxy, which also takes care of setting a proportional value to conntrack `hashsize`.

Closes #708

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -38,14 +38,6 @@ coreos:
         ExecStartPre=/opt/bin/cfn-etcd-environment
         ExecStart=/usr/bin/mv -f /var/run/coreos/etcd-environment /etc/etcd-environment
 
-{{if .UseCalico }}
-    # https://github.com/coreos/docs/blob/5d7b1cccb8286185275b07db1495828be9fdb0ea/os/other-settings.md#tuning-sysctl-parameters
-    - name: systemd-modules-load.service
-      command: restart
-    - name: systemd-sysctl.service
-      command: restart
-{{ end }}
-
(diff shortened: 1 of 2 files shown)


mumoshu:
Good catch @danielfm! Thanks for the fix 👍

PR #713, opened 2017-06-27
Fix kube-resources-autosave when kube2iam is enabled

camilb:
When `kube2iam` support is enabled and the DaemonSet was deployed,
`kube-resources-autosave` will fail to save the cluster state to S3 due
to missing role.

I added a condition to assign the `controller` role to
`kube-resources-autosave` deployment and also created examples for
`kube2iam`configuration.

### DIFF
--- a/contrib/kube2iam/kube2iam-calico-ds.yaml
+++ b/contrib/kube2iam/kube2iam-calico-ds.yaml
@@ -1,0 +1,35 @@
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: kube2iam
+  namespace: kube-system
+  labels:
+    app: kube2iam
+spec:
+  template:
+    metadata:
+      labels:
+        name: kube2iam
+    spec:
+      serviceAccountName: kube2iam
+      hostNetwork: true
+      containers:
+        - image: jtblin/kube2iam:latest
+          name: kube2iam
+          args:
+            - "--app-port=8282"
+            - "--base-role-arn=arn:aws:iam::YOUR_ACCOUNT_ID:role/"
+            - "--iptables=true"
+            - "--host-ip=$(HOST_IP)"
+            - "--host-interface=cali+"
+          env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+          ports:
+            - containerPort: 8282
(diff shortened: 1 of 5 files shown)


camilb:
To work correctly have to use `.WorkerNodePoolConfig.IAMConfig.Role.Name`  instead of `.Controller.IAMConfig.Role.Name` but could not find a elegant way to get this value in `cloud-config-controller` template.  Anyone has an idea how to do it? Thanks!

mumoshu:
@camilb Thanks for the fix 👍 
Regarding the question, I've been thinking that we would need to do that someday 😄
I know it won't look clean but somehow (1) adding `Worker root.config.Worker `yaml:"worker,omitempty"` to `controlplane.config.Cluster` and then (2) copying `Worker` from the root config to the control-plane config [around here](https://github.com/kubernetes-incubator/kube-aws/blob/master/core/root/config/config.go#L60-L63) would do the job for now.
Perhaps (2) implies that you need to rename `root.config.Worker` to `model.Worker` while extracting the type into `model/worker.go`

camilb:
Thanks @mumoshu , I'll check it.

mumoshu:
@camilb Hi, thanks for taking your time on this 👍 
Would you mind if I merged this as-is and extract what was stated in [my prev comment](https://github.com/kubernetes-incubator/kube-aws/pull/713#issuecomment-311546728) into an another issue?

redbaron:
@camilb , am I right that it is necessary to have that manual step mentioned in a comment, because of *YOUR_ACCOUNT_ID* ? In other words, if kube-aws templates it when file is created, then kube2iam DS can be created automatically, right?

camilb:
@mumoshu Sorry for  the delay, just got back to Spain today. I think it can be merged, first let me check  if can easily  automate the kube2iam DS deployment mentioned in the previous comment.

@redbaron You are right. Basically there are 3 parameters we should take in count:  `YOUR_ACCOUNT_ID`,  `host-interface` if calico is disabled/enabled  and  `serviceAccountName` if RBAC is disabled/enabled.

mumoshu:
@camilb @redbaron Created #805 and #806 to track another issues extracted from our discussion.

mumoshu:
@camilb Thanks a lot for your continuous contributions in this area 👍 Keep up the good work 🎉

PR #715, opened 2017-06-28
Add "Role" tag to all nodes.

camilb:
Hi @mumoshu, @redbaron, I remember the "Role" key tag was used for workers in `kube-aws` and removed at some point.  I'm already using the "Role" tag for all of the EC2 instances and this PR will help organising them better. If there was a specific reason for removing the "Role" tag and you find it unnecessary, I'll just use another method to create the tags automatically. 

Also, if anyone interested, I created a simple solution to add tags to instances in a spot fleet.
Here is the [repo](https://github.com/camilb/kube-aws-spot-fleet-tags)

Thanks!

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -46,6 +46,11 @@
             "Value": "{{.ClusterName}}-{{.StackName}}-kube-aws-controller"
           },
           {
+            "Key": "Role",
+            "PropagateAtLaunch": "true",
+            "Value": "Master"
+          },
+          {
             "Key": "kubernetes.io/role/master",
             "PropagateAtLaunch": "true",
             "Value": ""
@@ -626,6 +631,11 @@
             "Key": "Name",
             "PropagateAtLaunch": "true",
             "Value": "{{$.ClusterName}}-{{$.StackName}}-kube-aws-etcd-{{$etcdIndex}}"
+          },
+          {
+            "Key": "Role",
+            "PropagateAtLaunch": "true",
+            "Value": "Etcd"
           },
           {
             "Key": "kube-aws:role",
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -148,6 +148,11 @@
             "Key": "Name",
             "PropagateAtLaunch": "true",
             "Value": "{{.ClusterName}}-{{.StackName}}-kube-aws-worker"
+          },
+          {
+            "Key": "Role",
+            "PropagateAtLaunch": "true",
+            "Value": "Worker"
           }
         ],
         {{if .LoadBalancer.Enabled}}


redbaron:
dont' we  have more standard `kubernetes.io/role` tags already?

camilb:
`kubernetes.io/role` is only present for controller nodes and only has a key as `kubernetes.io/role/master` without a value and doesn't allow to filter the instances. 
Then etcd nodes have `kube-aws:role=etcd`.

redbaron:
Wold you agree to a more kube-specific tag name? There might be many instances running in AWS account, which had no relation to kube, yet tagged Role: Master.

camilb:
Of course,  you have something specific in mind? . My use case currently is to filter first using `stackTags` or cluster name, then see the roles for each of the nodes.

redbaron:
I am fine with `kube-aws:role=etcd|master|nodepool1|nodepool2|nodepool3`, what do you think?

camilb:
There is already one for nodepools `kube-aws:node-pool:name=your-nodepool`. I think we only have to add for `etcd|master|worker`. My current environment forces me to add "Role" to all instances, but I can use other tools to achieve this. Just let me know if it's ok to add the fallowing tags ` kube-aws:role=etcd|master|worker` and I'll make the changes.

mumoshu:
@camilb Thanks for your suggestion - yes, I'm fine with `kube-aws:role=etcd|master|worker`.
cc @redbaron

redbaron:
+1 for `kube-aws:role=etcd|master|worker`, but what seems we actually need is configurable set of extra tags per nodepool, controllers and etcd pools, with defaults being `kube-aws:role=etcd|master|worker`

mumoshu:
Thanks!
Basically I agree with you but beware that kube-aws:role=etcd is required
by H/A etcd. Therefore, it wouldn't make sense to allow overriding whole
the set of tags.
Allowing to add extra tags per pool would make sense.
2017年6月29日(木) 14:32 Maxim Ivanov <notifications@github.com>:

camilb:
>+1 for kube-aws:role=etcd|master|worker, but what seems we actually need is configurable set of extra tags per nodepool, controllers and etcd pools, with defaults being kube-aws:role=etcd|master|worker

@mumoshu @redbaron I think configurable tags make more sense for the must use cases, including mine, where I have to add some specific tags.

mumoshu:
@camilb How about adding the brand-new setting keys in cluster.yaml like `worker.nodePools[].additionalTags`, `controller.additionalTags` and `etcd.additionalTags` respectively for that?

```yaml
controller:
  # Additional tags associated to the ASG managing controller nodes for administrative purpose
  additionalTags:
  - key: kube-aws:role
    value: controller

etcd:
  # Additional tags associated to the ASG managing etcd nodes for administrative purpose
  additionalTags:
  - key: kube-aws:role
    value: etcd
```

camilb:
@mumoshu Yes, that's exactly what I was thinking.

mumoshu:
Thanks for the confirmation!
Would you mind updating this PR with that, or in an another PR, or you'd
like me to do?
2017年7月5日(水) 23:37 Camil Blanaru <notifications@github.com>:

camilb:
@mumoshu If you find it simple enough and it will not take too much time for you. My coding skills are a little bit rust .

camilb:
@mumoshu I'll try to work on it these days. I'll close this PR for now and open another one.

PR #722, opened 2017-07-02
Propagate etcd version from cluster.yaml to etcdadm

ytsarev:
Without this additional propagation `etcdadm` falls back to default `3.1.3` version as in
https://github.com/iflix-letsplay/kube-aws/blob/master/etcdadm/etcdadm#L88

In case when any other version is specified in `cluster.yaml` it will create the deviation between running etcd version and the one that is used for periodic snapshotting, e.g. running `3.1.8` in the main container and still using default `3.1.3` for `etcdadm` based operations.

This PR fixes the deviation.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -725,6 +725,9 @@
                   "'\n",
                   "ETCDADM_MEMBER_INDEX='",
                     "{{$etcdIndex}}",
+                  "'\n",
+                  "ETCD_VERSION='",
+                    "{{$.Etcd.Version}}",
                   "'\n"
                 ]]}
               }


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -618,6 +618,7 @@ write_files:
       ETCD_ADVERTISE_CLIENT_URLS=https://$advertised_hostname:2379
       ETCD_LISTEN_PEER_URLS=https://$private_ip:2380
       ETCD_INITIAL_ADVERTISE_PEER_URLS=https://$advertised_hostname:2380" >> /var/run/coreos/etcd-environment
+      ETCD_VERSION={{.Etcd.Version}}
```
redbaron: there is an easy to miss redirect happening in this file, line you add should be inside that redirect to have an effect

ytsarev: not sure i get you properly,  can you clarify with code example ?

mumoshu: Sorry this is confusing but given what we have so far:

```
echo "KUBE_AWS_ASSUMED_HOSTNAME=$advertised_hostname
*snip*
ETCD_INITIAL_ADVERTISE_PEER_URLS=https://$advertised_hostname:2380" >> /var/run/coreos/etcd-environment
ETCD_VERSION={{.Etcd.Version}}
```

It should be the something like the below:

```
echo "KUBE_AWS_ASSUMED_HOSTNAME=$advertised_hostname
*snip*
ETCD_INITIAL_ADVERTISE_PEER_URLS=https://$advertised_hostname:2380
ETCD_VERSION={{.Etcd.Version}}" >> /var/run/coreos/etcd-environment
```

or `ETCD_VERSION` would be just thrown away.

ytsarev: Thanks for spotting it guys. Actually first commit was both wrong and redundant, i removed it.
Cloud formation based propagation is enough to fix the problem

mumoshu:
@ytsarev Ah, that made sense! Thanks for your contribution 👍

PR #724, opened 2017-07-03
Add critical-pod annotation to CA

### LINKED ISSUE #683: Cluster autoscaler missing critical pod annotation
According to https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#deployment we should mark the cluster autoscaler with the `scheduler.alpha.kubernetes.io/critical-pod` annotation. While we may by default deploy to controllers only I think it's probably better to have the annotation to mirror the other Deployments we have?

mumoshu:
Resolves #683

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1981,6 +1981,8 @@ write_files:
             metadata:
               labels:
                 app: cluster-autoscaler
+              annotations:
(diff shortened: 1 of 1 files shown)


redbaron:
Don't we need to have tolerations `[{"key":"CriticalAddonsOnly", "operator":"Exists"}]` specified there as well?

mumoshu:
@redbaron Thanks! Just submitted a follow-up PR for that at #738

Issue #580, opened 2017-04-24
etcdadm scripts should be written to files at `render` phase, not `up`


redbaron:
right now `etcdadm` script is embedded in kube-aws binary and "materializes" itself only when we run "kube-aws up/update" which makes it harder to track full kluster config in git

mumoshu:
@redbaron Thanks for the feedback 👍 
Yes, it should be. It was even in the "remaining issues" list in “the original PR](https://github.com/kubernetes-incubator/kube-aws/pull/417) which introduced etcdadm.

How about `scripts/` directory to store scripts like etcdadm?
Any preference on the naming of the directory?

redbaron:
why `scripts` is everything else is in `write_files`?

mumoshu:
@redbaron My intention was to ensure users place only scripts inside the directory. The same as `credentials` for `/etc/kubernetes/ssl`.

Would you like to embed `etcdadm` directly into cloud-config-etcd like:
```yaml
  - path: /opt/bin/etcdadm
    content: |
      // the content of etcdadm here
```
or anything else?

redbaron:
either way works for me. `scripts` is more convenient, but probably would require more work

PR #730, opened 2017-07-04
Install Amazon SSM Agent onto EC2 instances

andromedarabbit:
Amazon SSM agent make us to quickly and easily execute remote commands or scripts against one or more EC2 instances: https://github.com/aws/amazon-ssm-agent

This feature is disabled by default and configurable in cluster.yaml:

```
# When enabled, all nodes will have Amazon SSM Agent installed.
# It is disabled by default.
#amazonSsmAgent:
#  enabled: true
```

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -113,6 +113,11 @@ func NewDefaultCluster() *Cluster {
 			MapPublicIPs:       true,
 			Experimental:       experimental,
 			ManageCertificates: true,
+			AmazonSsmAgent: AmazonSsmAgent{
+				Enabled:     false,
+				DownloadUrl: "",
+				Sha1Sum:     "",
+			},
 			CloudWatchLogging: CloudWatchLogging{
 				Enabled:         false,
 				RetentionInDays: 7,
@@ -427,6 +432,7 @@ type DeploymentSettings struct {
 	ManageCertificates     bool              `yaml:"manageCertificates,omitempty"`
 	WaitSignal             WaitSignal        `yaml:"waitSignal"`
 	CloudWatchLogging      `yaml:"cloudWatchLogging,omitempty"`
+	AmazonSsmAgent         `yaml:"amazonSsmAgent,omitempty"`
 
 	// Images repository
 	HyperkubeImage                     model.Image `yaml:"hyperkubeImage,omitempty"`
@@ -578,6 +584,12 @@ type Kube2IamSupport struct {
 type KubeResourcesAutosave struct {
 	Enabled bool `yaml:"enabled"`
 	S3Path  string
+}
+
+type AmazonSsmAgent struct {
+	Enabled     bool   `yaml:"enabled"`
+	DownloadUrl string `yaml:"downloadUrl"`
+	Sha1Sum     string `yaml:"sha1sum"`
 }
 
 type CloudWatchLogging struct {
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -41,6 +41,27 @@ coreos:
         {{ $l }}
         {{- end }}
 {{- end}}
+{{if and (.AmazonSsmAgent.Enabled) (ne .AmazonSsmAgent.DownloadUrl "")}}
+    - name: amazon-ssm-agent.service
+      command: start
+      enable: true
+      content: |
+        [Unit]
+        Description=amazon-ssm-agent
+        Requires=network-online.target
+        After=network-online.target
+
+        [Service]
+        Type=simple
+        ExecStartPre=/opt/ssm/bin/install-ssm-agent.sh
+        ExecStart=/opt/ssm/bin/amazon-ssm-agent
+        KillMode=controll-group
+        Restart=on-failure
+        RestartSec=1min
+
+        [Install]
+        WantedBy=network-online.target
+{{end}}
 {{if .CloudWatchLogging.Enabled}}
     - name: journald-cloudwatch-logs.service
       command: start
@@ -461,6 +482,41 @@ write_files:
     content: {{$w.GzippedBase64Content}}
   {{- end }}
 {{- end }}
+{{if and (.AmazonSsmAgent.Enabled) (ne .AmazonSsmAgent.DownloadUrl "")}}
+  - path: "/opt/ssm/bin/install-ssm-agent.sh"
+    permissions: 0700
+    content: |
+      #!/bin/bash
+      set -e
+
+      TARGET_DIR=/opt/ssm
+      if [[ -f "${TARGET_DIR}"/bin/amazon-ssm-agent ]]; then
+        exit 0
+      fi
+
+      TMP_DIR=$(mktemp -d)
+      trap "rm -rf ${TMP_DIR}" EXIT
+
+      TAR_FILE=ssm.linux-amd64.tar.gz
+      CHECKSUM_FILE="${TAR_FILE}.sha1"
+
+      echo -n "{{ .AmazonSsmAgent.Sha1Sum }} ${TMP_DIR}/${TAR_FILE}" > "${TMP_DIR}/${CHECKSUM_FILE}"
+
+      curl --silent -L -o "${TMP_DIR}/${TAR_FILE}" "{{ .AmazonSsmAgent.DownloadUrl }}"
+
+      sha1sum --quiet -c "${TMP_DIR}/${CHECKSUM_FILE}"
+
+      tar zfx "${TMP_DIR}"/"${TAR_FILE}" -C "${TMP_DIR}"
+      chown -R root:root "${TMP_DIR}"/ssm
+
+      CONFIG_DIR=/etc/amazon/ssm
+      mkdir -p "${CONFIG_DIR}"
+      mv -f "${TMP_DIR}"/ssm/amazon-ssm-agent.json "${CONFIG_DIR}"/amazon-ssm-agent.json
+      mv -f "${TMP_DIR}"/ssm/seelog_unix.xml "${CONFIG_DIR}"/seelog.xml
+
+      mv -f "${TMP_DIR}"/ssm/* "${TARGET_DIR}"/bin/
+
+{{end}}
 {{if .Experimental.DisableSecurityGroupIngress}}
   - path: /etc/kubernetes/additional-configs/cloud.config
     owner: root:root
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -73,6 +73,27 @@ coreos:
 
         [Install]
         RequiredBy=format-etcd2-volume.service
+{{if and (.AmazonSsmAgent.Enabled) (ne .AmazonSsmAgent.DownloadUrl "")}}
+    - name: amazon-ssm-agent.service
+      command: start
+      enable: true
+      content: |
+        [Unit]
+        Description=amazon-ssm-agent
+        Requires=network-online.target
+        After=network-online.target
+
+        [Service]
+        Type=simple
+        ExecStartPre=/opt/ssm/bin/install-ssm-agent.sh
+        ExecStart=/opt/ssm/bin/amazon-ssm-agent
+        KillMode=control-group
+        Restart=on-failure
+        RestartSec=1min
+
+        [Install]
+        WantedBy=network-online.target
+{{end}}
 {{if .CloudWatchLogging.Enabled}}
     - name: journald-cloudwatch-logs.service
       command: start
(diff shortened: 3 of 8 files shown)


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -73,6 +73,28 @@ coreos:
...
+      enable: true
+      content: |
+        [Unit]
+        Description=amazon-ssm-agent
+        Requires=network-online.target
+        After=network-online.target
+
+        [Service]
+        Type=simple
+        WorkingDirectory=/opt/bin/ssm
```
mumoshu: Does running `aws ssm send-command` result in running commands in this working directory?
Any reason we won't make it `WorkingDirectory=/`?

mumoshu:
@andromedarabbit Thanks for the PR!
Wow, this is what I had dreamed of 😃 

Although I believe duplicating `/opt/bin/ssm/install-ssm-agent.sh` across all the roles(etcd, worker, controller) isn't the best thing to do, probably we don't have a better option at this point. We should introduce a directory dedicated to define common files included in two or more cloud-configs. See #580.

So, changes LGTM overall, only a minor comment for now!

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: No special reasons. No objection if you think that setting `WorkingDirectory` to `/` is preferable.

c-knowles:
Would symlinks work or are they not picked up when we create the instance files?

Can we make the default download URL the official AWS one?

core/controlplane/config/config.go
```diff
@@ -113,6 +113,10 @@ func NewDefaultCluster() *Cluster {
 			MapPublicIPs:       true,
 			Experimental:       experimental,
 			ManageCertificates: true,
+			AmazonSsmAgent: AmazonSsmAgent{
+				Enabled:     false,
+				DownloadUrl: "https://github.com/DailyHotel/amazon-ssm-agent/releases/download/v2.0.805.1/ssm.linux-amd64.tar.gz",
```
ankon: This URL looks suspicious and doesn't seem to be needed in the code, as it does also appear in the configuration. I rather not have a binary on my system that wants to download from "somewhat arbitrary" github places by default for configuring my production environments.

Would it not be possible to set the default to an empty string, and make the user explicitly acknowledge the download location by having to uncomment it in the configuration file?

andromedarabbit: @ankon @c-knowles  First, AWS does not distribute AMD64 binaries for CoreOS so you have to build it by yourself. Second, URL might look suspicious but the company is very well-known in Korea, Republic of. This is one of the biggest company among the mobile hotel booking services. Anyway, as @ankon suggested, setting the default to an empty string might be better.

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Thanks!
Just for my education but do you have any experience for my first question?

andromedarabbit: Yes, as I tested and confirmed. `aws ssm send-command` result in running commands in this working directory, in this case `/opt/bin/ssm`. However, you can set working directory with `aws ssm send-command`.

Anyhow, I am writing a fix for this and will commit it once I am done with testing.

andromedarabbit:
@mumoshu @ankon I've made some changes according to the feedbacks received: https://github.com/kubernetes-incubator/kube-aws/pull/730/commits/48978b188b2d1581aac2d2f8c813f96154c7a08f

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -355,6 +377,43 @@ write_files:
...
+      #!/bin/bash
+      set -eo pipefail
+
+      function finish {
+        if [ "$(pwd)" == "/opt/bin/ssm" ]; then
+          popd > /dev/null 2>&1
+        fi
+      }
+
+      trap finish EXIT
```
redbaron: all `finish` is doing is changing directory just before exiting, whats the point of it?

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -355,6 +377,43 @@ write_files:
...
+
+      pushd /opt/bin/ssm > /dev/null 2>&1
+
+      TAR_FILE=ssm.linux-amd64.tar.gz
+      CHECKSUM_FILE="${TAR_FILE}.sha1"
+
+      curl --silent -L -o ${TAR_FILE} "{{ .AmazonSsmAgent.DownloadUrl }}"
+      if [[ -f "${CHECKSUM_FILE}" ]]; then
+        ( sha1sum --status -c "${CHECKSUM_FILE}" && exit 0 ) || true
```
redbaron: whats the role of `CHECKSUM_FILE`? 

1. it is generated from TAR_FILE itself at the end of the script. Usually signed checksums are downloaded from same location where tarball is, to verify integrity of a tarball
2. it is never used on first download
3. checksum check results are ignored

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -355,6 +377,43 @@ write_files:
     content: {{$w.GzippedBase64Content}}
   {{- end }}
 {{- end }}
+{{if and (.AmazonSsmAgent.Enabled) (ne .AmazonSsmAgent.DownloadUrl "")}}
+  - path: "/opt/bin/ssm/install-ssm-agent.sh"
+    permissions: 0700
+    content: |
+      #!/bin/bash
+      set -eo pipefail
```
redbaron: there is no pipes in a script, and pipefail is dangerous anyway

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -355,6 +377,43 @@ write_files:
...
+
+      function finish {
+        if [ "$(pwd)" == "/opt/bin/ssm" ]; then
+          popd > /dev/null 2>&1
+        fi
+      }
+
+      trap finish EXIT
+
+      pushd /opt/bin/ssm > /dev/null 2>&1
```
redbaron: so executables are in `/opt/bin/ssm/<name>?` It is subjective, but I think `/opt/ssm/bin/<name>` is more natural

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: This can be convenient when you run this script to install the agent by yourself for whatever reasons.

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: This is for upgrading the agent. Sometimes you might want to upgrade your agent without running `kube-aws update`. In such a case, you can overwrite the gzipped file on the download url and then reboot EC2 instances to reinstall the agent:

1. `sha1sum --status -c "${CHECKSUM_FILE}"` will fail in this case
2. then the following lines will be executed and the agent will be overwritten and upgraded

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: Can you explain to me the reasons why `pipefail` is dangerous? Just for my own education.

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: Yes, `/opt/ssm/bin` is a better place to put it in.

core/controlplane/config/templates/cloud-config-etcd
redbaron: Changing directory in a script you run from she'll, doesn’t change directory in your shell

core/controlplane/config/templates/cloud-config-etcd
redbaron: Just a note of warning: Changing node config using agent is a slippery slope, where whole set of nodes can be managed out of bounds for the sake of simplicity only to find out that ASG might restart nodes and new ones come up with unaltered “old” config. It is easy to take it further and turn your cluster into a snowflake, where ‘ kube-aws update’ is not enough to bring up k8s cluster as some set of changes need to be applied which are not recorded in SCM where you do kube-aws update from.
 
 
Now back to your script. Just try updating URL and try running it yourself. ‘( .... && exit 0 )’ isn't doing what you think it is doing

core/controlplane/config/templates/cloud-config-etcd
redbaron: try `/bin/bash -c 'set -o pipefail; cat /bigtextfile | head -n1 '; echo $?`

in a nutshell, it fails expressions which do not consume whole input, such as `head`, `grep -q` and many others. It is happening because writers which try to write to closed stdout receive SIGPIPE signal, which in 99.999% of cases is unhandled and therefore triggers process shutdown with non 0 exit code. `pipefail` then uses first non-0 exit code in a pipe as a result of whole expression, which leads to surprising errors together with `set -e` which you also set

mumoshu:
Hi @andromedarabbit, just checking in but did you have any chance to discuss on/address review comments left by @redbaron?
As this is already LGTM overall, I guess I can adapt your work and improve it on our side if you're short on time.
Please let me know your preference!

andromedarabbit:
@mumoshu I was busy for my personal business. Now things are settled and I am going to push my fixes ASAP.

andromedarabbit:
@mumoshu @redbaron @ankon I've rewritten installation script in the https://github.com/kubernetes-incubator/kube-aws/pull/730/commits/e57e3c7275866824d9745d653d6def915af0cbf1 :

- Binaries are placed into `/opt/ssm/bin`, instead of `/opt/bin/ssm`
- Simple but robust installation process

andromedarabbit:
@ankon 

- I just have realized `-o` is not required. My bad.
- In case of `curl -f` fails, `ssm.linux-amd64.tar.gz` is still created with the content `Not Found`.  I need the better solution but has not found it yet.
- `/bin/rm` is just my old habbit and can be fixed in the next commit.

andromedarabbit:
With https://github.com/kubernetes-incubator/kube-aws/pull/730/commits/02f56cb030b370a8ee47d529e609c0a36926c61d, the integrity of downloaded file is verified with `sha1sum`.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -461,6 +483,41 @@ write_files:
...
+    content: |
+      #!/bin/bash
+      set -e
+
+      TARGET_DIR=/opt/ssm
+      if [[ -f "${TARGET_DIR}"/bin/amazon-ssm-agent ]]; then
+        exit 0
+      fi
+
+      TMP_DIR=$(mktemp -d)
```
redbaron: You can add `trap` so this directory is removed even if script failed:

```
trap "rm -rf ${TMP_DIR}" EXIT
```

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -73,6 +73,28 @@ coreos:
...
+        Description=amazon-ssm-agent
+        Requires=network-online.target
+        After=network-online.target
+
+        [Service]
+        Type=simple
+        WorkingDirectory=/
+        ExecStartPre=/opt/ssm/bin/install-ssm-agent.sh
+        ExecStart=/opt/ssm/bin/amazon-ssm-agent
+        KillMode=process
```
redbaron: I am not familiar with Amazon SSM agent, but is there a reason for this KillMode? According to docs it kills top level process only, leaving anything it spawns to stay alive.

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -73,6 +73,28 @@ coreos:
...
+      enable: true
+      content: |
+        [Unit]
+        Description=amazon-ssm-agent
+        Requires=network-online.target
+        After=network-online.target
+
+        [Service]
+        Type=simple
+        WorkingDirectory=/
```
redbaron: isn't it a default value already?

core/controlplane/config/templates/stack-template.json
```diff
@@ -225,6 +225,87 @@
...
+                {
+                  "Effect": "Allow",
+                  "Action": [
+                      "s3:PutObject",
+                      "s3:GetObject",
+                      "s3:AbortMultipartUpload",
+                      "s3:ListMultipartUploadParts",
+                      "s3:ListBucketMultipartUploads"
+                  ],
+                  "Resource": "*"
```
redbaron: this is *really* permissive

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: @redbaron I am also not an expert. I just followed the official source file here:  https://github.com/aws/amazon-ssm-agent/blob/master/packaging/linux/amazon-ssm-agent.service

mumoshu: I'm not an expert either 😉

Perhaps this should be `KillMode=control-group` if you'd want all the processes created for user commands run via the agent to be killed altogether?
I don't have a strong opinion here - am ok with either `process` or `control-group`.

@redbaron @andromedarabbit WDYT?

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Seems so according to https://www.freedesktop.org/software/systemd/man/systemd.exec.html#WorkingDirectory=

core/controlplane/config/templates/stack-template.json
mumoshu: AFAIK, s3 access is not necessary to a SSM agent.

@andromedarabbit Perhaps you added this to support possible use-case(s) with SSM like uploading logs from ec2 instances to s3?

Anyway, if it is an optional feature, can we remove these optional iam policies and add doc/comment to encourage users to add policies by their own?

core/controlplane/config/templates/cloud-config-etcd
andromedarabbit: Thanks for the advice. I will look into this topic.

core/controlplane/config/templates/stack-template.json
andromedarabbit: This is permissive because usually when people need tools like this they are in desperate needs in controlling whatever they need to do. Anyway I also concern that providing this as a default might be harmful.

mumoshu:
@andromedarabbit Sorry, etcd iam policies started conflicting - I've extracted iam policy into its own cfn resource for better customizability. Please feel free to ask for me to resolve the conflict if you had difficulty doing so.

core/controlplane/config/templates/stack-template.json
mumoshu: @andromedarabbit Thanks for the explanation 👍 
Yea, I think it might be harmful, too.

Could we make the default policies inside `{{if and (.AmazonSsmAgent.Enabled) (ne .AmazonSsmAgent.DownloadUrl "")}}` to be minimal requirements for ssm agents to run?
Then, we can open an another github issue to discuss easier way to provide "necessary" iam plicies to nodes running ssm agents.

Personally, I'd suggest everyone to define their own sets of iam policies required for their use-cases of ssm agents as "managed policies" and provide those policies to `(controller|etcd|worker).iam.role.managedPolicies[].arn`.

WDYT?

andromedarabbit: @redbaron @mumoshu I realized that this a bit of permissive IAM policy might be the minimum requirement. This configuration comes from [AmazonEC2RoleforSSM](http://docs.aws.amazon.com/systems-manager/latest/userguide/automation-permissions.html). 

As @mumoshu suggested, letting people to set up their own managed policies would be the best, I believe.

andromedarabbit:
With https://github.com/kubernetes-incubator/kube-aws/pull/730/commits/3096f975034032c25ad5ebf405e3ca1b64a7e5b9, permissive IAM policies are removed. Other changes are:

- `KillMode=control-group` instead of `KillMode=process`
- `WorkingDirectory=/` are removed since it is already default.

e2e/run
```diff
@@ -273,6 +273,8 @@ addons:
     enabled: true
 cloudWatchLogging:
   enabled: true
+ssmAgent:
```
redbaron: isn't it `amazonSsmAgent`?

redbaron:
Once e2e param fixed, please squash and rebase on top of latest master to have one nice well formed commit to be merged.

e2e/run
andromedarabbit: You are right. Wow, it was written at the first commit but have not been found until this. Thanks!

redbaron:
@andromedarabbit, you can use `git rebase -i` to squash all the commits into one

andromedarabbit:
All the commits are squashed. I appreciated all the helps I've gotten.

redbaron:
seems that you cherry-picked other commits into your branch, now they show up in this PR. can you please remove them?

andromedarabbit:
@redbaron I should have been more careful. A squashed commit is put onto the top of latest master branch.

redbaron:
Thanks for contributing!
,

Issue #585, opened 2017-04-24
Failing-fast or warn when tried to update cluster created with older kube-aws


mumoshu:
According to #455, you need to recreate your cluster to upgrade kube-aws as of today.
However, I believe some people are still getting into trouble trying to run `kube-aws update` with newer kube-aws on their old clusters.
Making kube-aws failing fast when tried to do so would allow people to notice the fact.

danielfm:
Couldn't we add a metadata to `cluster.yaml` to specify the kube-aws version used to generate it, and then only allow further commands, like `render`, `up` and `update` to go through if the versions match?

mumoshu:
@danielfm Thanks for the suggestion. That sounds good to me 👍 
Actually, I was initially thinking about adding a version tag to the root cfn stack so that we can use it to prevent `kube-aws update` to be run. Probably yours is better though.

c-knowles:
I think users may just update it in the yaml if it's in there? Especially if they use the git flow style mechanism of updating. Having said that, ideally we only want to prevent updates where we know running update will break.

mumoshu:
@c-knowles 
> I think users may just update it in the yaml if it's in there? Especially if they use the git flow style mechanism of updating. 

Good point. Then, we'd better add a cfn tag via a template snippet like `"Value": "{{.KubeAwsVersion}}"` so that the git workflow like described in #462 #238 won't accidentally update the version number associated to a cluster.

c-knowles:
@mumoshu if we do that though, how do we support upgrade path for cases that do work? We reject by default unless an update is allowed? Or allow by default unless an update is disallowed?

mumoshu:
@c-knowles I guess the former is better because then we can ensure that only upgrade paths tested by contributors can be allowed explicitly in code.

redbaron:
{{ .KubeAwsVersion }} is not informative enough as it mightbe just  a git commit of a private fork :)

mumoshu:
@redbaron Thanks for the feedback 👍 

1. How about tagging some commits in your private fork with version numbers so that you can explicitly control one version is updatable or not in code?
2. Or I guess you can just leave it just as a git commit id so that kube-aws can prevent `kube-aws update` among different commit ids by default?
3. If none of the above sounds good, how about adding an another flag in cluster.yaml to disable update prevention at all...? 😃

redbaron:
how kube-aws itself defines which version is updatable and which is not?

[...]

PR #731, opened 2017-07-04
Add cluster kube-aws version to outputs

Vrtak-CZ:
It's just an idea. We have few clusters of kubernetes which is managed by kube-aws but there is no way how to check with version of kube-aws is used for creating or updating this cluster. So we don't know if we need proceed update on this clusters.

Btw usage of CloudFormation output is just an idea we should use CloudFormation tags needer.

PS: I have no experience with Go so I'm just trying if this works.

### DIFF
--- a/core/root/config/templates/stack-template.json
+++ b/core/root/config/templates/stack-template.json
@@ -97,6 +97,11 @@
     }{{end}}
   },
   "Outputs": {
+    "KubeAwsVersion": {
+      "Description": "The version number of kube-aws which was used to create this cluster",
+      "Value": "{{$.KubeAwsVersion}}",
+      "Export": { "Name": { "Fn::Sub": "${AWS::StackName}-KubeAwsVersion" } }
+    },
     {{ if $.ControlPlane.NeedToExportIAMroles }}
     "ControllerIAMRoleArn": {
       "Description": "The IAM Role ARN for controller nodes",
--- a/core/root/template_params.go
+++ b/core/root/template_params.go
@@ -16,6 +16,10 @@ func (p TemplateParams) ClusterName() string {
 	return p.cluster.controlPlane.ClusterName
 }
 
+func (p TemplateParams) KubeAwsVersion() string {
+	return controlplane.VERSION
+}
+
 func (p TemplateParams) CloudWatchLogging() config.CloudWatchLogging {
 	return p.cluster.controlPlane.CloudWatchLogging
 }


core/root/config/templates/stack-template.json
```diff
@@ -97,6 +97,11 @@
     }{{end}}
   },
   "Outputs": {
+    "KubeAwsVersion": {
+      "Description": "This stack is created by kube-aws version",
```
mumoshu: Suggestion: `The version number of kube-aws which was used to create this cluster`

mumoshu:
@Vrtak-CZ I like your idea 👍 
Actually we need this sooner or later for #585
Please poke me if this is ready to be reviewed.

core/root/template_params.go
```diff
@@ -16,6 +16,10 @@ func (p TemplateParams) ClusterName() string {
 	return p.cluster.controlPlane.ClusterName
 }
 
+func (p TemplateParams) ClusterVersion() string {
+	return p.cluster.version
```
mumoshu: Perhaps you can just return `controlplane.VERSION` here for now?

Vrtak-CZ:
@mumoshu fixed and rebased

core/root/template_params.go
```diff
@@ -16,6 +16,10 @@ func (p TemplateParams) ClusterName() string {
 	return p.cluster.controlPlane.ClusterName
 }
 
+func (p TemplateParams) ClusterVersion() string {
```
redbaron: I'd rename it to `KubeAwsVersion` to be 100% sure what value represents

mumoshu: So that we won't confuse this with k8s version? Makes sense! I'll make a follow-up PR for it.

mumoshu:
@Vrtak-CZ Oh, btw,

> Btw usage of CloudFormation output is just an idea we should use CloudFormation tags needer.

Do you prefer tags over outputs? Would you like me to improve your work to use tags instead?

Vrtak-CZ:
> Do you prefer tags over outputs?
Not really because tags does not have description. 

> Would you like me to improve your work to use tags instead?
Feel free to improve it.

mumoshu:
> Not really because tags does not have description.

Good point 👍 Then, I prefer leaving it as a cfn output.

mumoshu:
LGTM. Thanks for your contribution @Vrtak-CZ 🍻

Issue #727, opened 2017-07-03
Nodepools created with spot instances fail to register in cluster.


camilb:
The new method of getting the `KUBE_AWS_STACK_NAME` from `aws:cloudformation:stack-name` tag  doesn't work with Spot Instances because these instances are not tagged by Cloudformation. 

`cfn-etcd-environment.service` fails to start without `KUBE_AWS_STACK_NAME` value in `/etc/environment`

mumoshu:
@camilb Thanks for the good catch 👍
As we need to tag spot instances manually, relying on `aws:cloudformation:stack-name` wouldn't be an universal option as it does result in the chicken-and-the-egg problem.
We need to somehow pass the stack name from `{"Ref":"AWS::StackName"}` to the userdata...

mumoshu:
I have considered to use multi-part cloud-config so that we can use the first part to populate the env var with the stack name as the value. However, unfortunately, coreos-cloudinit doesn't support it.
https://github.com/coreos/bugs/issues/912

mumoshu:
A possible solution(which doesn't need to revert the recent improvements on userdata made by @redbaron) came up in my mind is to utilizing ignition with:

```
"UserData": {{ .UserDataWorker.Parts.instance.Base64 true | checkSizeLessThan 16384 | quote }}
```

modified to:

```
"UserData": {{ .UserDataWorker.Parts.instance.IgnitionConfigRef }}
```

where the IgnitionConfig func is:

```golang
func quoteForCfn(jsonLike string) string {
  return strings.Replace(strings.Replace(jsonLike, "\n", "\\n", -1), ```"```, ```\"```, -1)
}

func (self UserDataPart) IgnitionConfigRef() string {
  splits := []string{
    quoteForCfn(```{
  "ignition": {
    "version": "2.0.0"
  },
  "storage": {
    "files": [
      {
        "filesystem": "root",
        "path": "/var/run/coreos/kube-aws-stack-name",
        "contents": {
          "source": "data:,```),
    base64StackNameRef,
    quoteForCfn(fmt.Sprintf(```"
        }
      },
      {
        "filesystem": "root",
        "path": "/opt/bin/bootstrap-worker-node",
        "contents": {
          "source": "data:,%s"
        }
      }
    ]
  },
  "systemd": {
    "units": [
      {
        "name": "bootstrap-worker-node",
        "enable": true,
[...]

Issue #728, opened 2017-07-04
Migrate to Ignition


mumoshu:
Once https://github.com/coreos/ignition/pull/382 lands on Container Linux Stable.

See my comment in https://github.com/kubernetes-incubator/kube-aws/issues/252#issuecomment-312761217

redbaron:
yeah, I've been following that PR for a while, I have WIP userdata conversion and kube-aws code changes to support ignition, waiting for CoreOS alpha release with Ignition 0.17 included, then I can test a create a PR.

Then we wait for a Ignition 0.17 to land on stable, which might take couple of months.

mumoshu:
Great! Really looking forward to see your PR 👍

mumoshu:
https://github.com/kubernetes-incubator/kube-aws/pull/675#issuecomment-303888310

mumoshu:
@redbaron Ignition 0.17 seems to have landed to [the latest alpha of container linux 1465.0.0](https://coreos.com/releases/#1465.0.0). Could we support enabling the ignition integration only for clusters created with that or later versions of container linux?

redbaron:
In theory yes, there are 2 ways:
 
 1. Maintain 2 sets of userdata
 1. Transform one form of userdata to another on the fly. Probably from cloud-init to ignition and have separate “instance” templates for old and new way of doing it
 
Realistically I wouldn’t bother, not with current codebase at least. I’ll test my branch and submit WIP PR which then can be rebased until Ignition lands to master in ~6 months or so

mumoshu:
@redbaron FYI, the stable channel is now pointed to 1465.6.0 with ignition 0.17.2.

Regarding the 2 ways I slightly prefer 1 with starting by calling out to coreos-cloudinit from ignition while gradually migrating our templating logics inside cloudinit-based cloud-config-(worker|etcd|controller) to their ignition alternatives. WDYT?
Anyway, I'm ok with either way as long as it works until we fully migrate to ignition though 😃

mumoshu:
@redbaron Do you think you can work on this towards v0.9.10-rc.1?

redbaron:
Our internal kube-aws fork just finally caught up with master, I can resurrect branch and see how it goes. Have no ETA, but I have very little time left to work on kube-aws at all, so need to pick my battles wisely :) At worst I'll push incomplete version as a branch for somebody else to pick up.

pawelprazak:
[...]

PR #732, opened 2017-07-04
Fix spot fleets after the multi-part userdata improvement

mumoshu:
This is currently a POC of the possible fix for #727 with ignition, probably simplified/replaced/enhanced with a more simpler solution without ignition.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -22,11 +22,7 @@ run aws s3 --region $REGION  cp {{ $S3URI }} /var/run/coreos/$USERDATA_FILE
 
 INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
 
-run /bin/sh -c 'echo {{.StackNameEnvVarName}}=$(aws ec2 describe-tags --region "'$REGION'" --filters \
-  "Name=resource-id,Values='$INSTANCE_ID'" \
-  "Name=key,Values=aws:cloudformation:stack-name" \
-  --output json \
-| jq -r ".Tags[].Value") >> {{.StackNameEnvFileName}}'
+echo "{{.StackNameEnvVarName}}=$(cat /kube-aws-stack-name)" >> {{.StackNameEnvFileName}}
 
 exec /usr/bin/coreos-cloudinit --from-file /var/run/coreos/$USERDATA_FILE
 {{ end }}
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -101,7 +101,7 @@
               {{end}}
             ],
             "SubnetId": {{$workerSubnet.Ref}},
-            "UserData": {{ $.UserDataWorker.Parts.instance.Base64 true | checkSizeLessThan 16384 | quote }}
+            "UserData": {{ $.UserDataWorker.Parts.instance.Base64IgnitionConfigRef }}
           }
           {{end}}
           {{end}}
@@ -261,7 +261,7 @@
         {{else}}
         "PlacementTenancy": "{{.Tenancy}}",
         {{end}}
-        "UserData": {{ .UserDataWorker.Parts.instance.Base64 true | checkSizeLessThan 16384 | quote }}
+        "UserData": {{ .UserDataWorker.Parts.instance.Base64IgnitionConfigRef }}
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration"
     {{if not .IAMConfig.InstanceProfile.Arn}}
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -101,7 +101,7 @@
               {{end}}
             ],
             "SubnetId": {{$workerSubnet.Ref}},
-            "UserData": {{ $.UserDataWorker.Parts.instance.Base64IgnitionConfigRef }}
+            "UserData": {{ $.UserDataWorker.Parts.instance.Base64UserDataRef }}
           }
           {{end}}
           {{end}}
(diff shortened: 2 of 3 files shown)


model/userdata.go
```diff
@@ -147,3 +148,71 @@ func validateCoreosCloudInit(content []byte) error {
...
+		return "", fmt.Errorf("failed to generate ignition config: %v", err)
+	}
+
+	// TODO Append user-provided ignition config via `ignition.config.append.source`
+	// once it supports sourcing from a s3 object
+	// https://coreos.com/ignition/docs/0.14.0/configuration-v2_0.html
+
+	head, err := json.Marshal(`{
+  "ignition": {
+    "version": "2.0.0"
```
mumoshu: In the future, to allow customizing the ignition config on user-side i.e. retain the flexibility of having all configs in a text form, we can add `ignition.config.append.source` which points to the user-provided ignition config json uploaded to s3.

model/userdata.go
```diff
@@ -147,3 +148,71 @@ func validateCoreosCloudInit(content []byte) error {
...
+	base64Part, err := self.Base64(false)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate ignition config: %v", err)
+	}
+
+	// TODO Append user-provided ignition config via `ignition.config.append.source`
+	// once it supports sourcing from a s3 object
+	// https://coreos.com/ignition/docs/0.14.0/configuration-v2_0.html
+
+	head, err := json.Marshal(`{
```
mumoshu: @redbaron I've thrown away the `quoteForCfn` func 😉

model/userdata.go
mumoshu: Then, we can also utilize the user-customizable ignition config to gradually migrate from our current cloud-configs to ignition.

mumoshu:
Verified that this does fix #727 😉

core/nodepool/config/templates/stack-template.json
```diff
@@ -101,7 +101,7 @@
               {{end}}
             ],
             "SubnetId": {{$workerSubnet.Ref}},
-            "UserData": {{ $.UserDataWorker.Parts.instance.Base64 true | checkSizeLessThan 16384 | quote }}
+            "UserData": {{ $.UserDataWorker.Parts.instance.Base64UserDataRef }}
```
mumoshu: I've introduced `Base64UserDataRef` to inject `{"Ref":"AWS::StackName"}` into nodes by modifying the user-provided cloud-config(bash script)

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -22,11 +22,7 @@ run aws s3 --region $REGION  cp {{ $S3URI }} /var/run/coreos/$USERDATA_FILE
 
 INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
 
-run /bin/sh -c 'echo {{.StackNameEnvVarName}}=$(aws ec2 describe-tags --region "'$REGION'" --filters \
-  "Name=resource-id,Values='$INSTANCE_ID'" \
-  "Name=key,Values=aws:cloudformation:stack-name" \
-  --output json \
-| jq -r ".Tags[].Value") >> {{.StackNameEnvFileName}}'
+echo "{{.StackNameEnvVarName}}=$(cat /kube-aws-stack-name)" >> {{.StackNameEnvFileName}}
```
mumoshu: The instance part of userdata now reads the stack name from a text file which is written by a bash snippet automatically injected by kube-aws

model/userdata.go
```diff
@@ -147,3 +148,105 @@ func validateCoreosCloudInit(content []byte) error {
...
+  "ignition": {
+    "version": "2.0.0"
+  },
+  "storage": {
+    "files": [
+      {
+        "filesystem": "root",
+        "path": "/kube-aws-stack-name",
```
mumoshu: The path should be somewhere else like `/var/run/coreos/kube-aws-stack-name`. However, `/var/run` is a tmpfs other than the root filesystem, which requires me to provide ignition an another filesystem name other than `"root"`.
The problem doing so is that I'm not sure that names of filesystems other than `"root"` is stable i.e. safe to be hard-coded.
So, I'm rather wanting to leave this undecided for now.

model/userdata.go
```diff
@@ -147,3 +148,105 @@ func validateCoreosCloudInit(content []byte) error {
 func validateNone([]byte) error {
 	return nil
 }
+
+func (self UserDataPart) Base64UserDataRef() (string, error) {
+	// Change this to Base64IgnitionConfigRef to test with ignition
+	return self.Base64BashScriptRef()
```
mumoshu: Ignition is not used by default for now. However, I like to leave the ignition support able to be enabled in code so that we can start testing with it toward #728 🤓

model/userdata.go
```diff
@@ -147,3 +148,105 @@ func validateCoreosCloudInit(content []byte) error {
...
+			return "", fmt.Errorf("failed to generate userdata: failed to marshal \"%s\" into json: %v", o, err)
+		}
+		jsonStrings = append(jsonStrings, string(escaped))
+	}
+
+	splits := []string{}
+	splits = append(splits, jsonStrings[:1]...)
+	splits = append(splits, `"echo \""`)
+	splits = append(splits, `{"Ref":"AWS::StackName"}`)
+	splits = append(splits, `"\" >> /kube-aws-stack-name\n"`)
```
mumoshu: This func is used to inject `{"Ref":"AWS::StackName"}` into nodes by modifying the "instance" parts of userdata.

mumoshu:
Implemented, enabled, and verified the simpler fix without ignition.

@redbaron Could you review this? This solution with `Base64UserDataRef` seemed to be a minimum possible fix for #727 for now. My originally proposed solution with ignition is disabled by default and can be enabled only by modifying code and rebuilding kube-aws.

Also, please let me sync with you that I'd like to address #727 before releasing v0.9.7 final 🙇

mumoshu:
Closing this in favor of #735

PR #735, opened 2017-07-05
Inject stack name  into userdata for nodepool workers

### LINKED ISSUE #727: Nodepools created with spot instances fail to register in cluster.

redbaron:
Fixes #727

### DIFF
--- a/filereader/texttemplate/texttemplate.go
+++ b/filereader/texttemplate/texttemplate.go
@@ -2,6 +2,7 @@ package texttemplate
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"github.com/Masterminds/sprig"
 	"io/ioutil"
@@ -21,9 +22,27 @@ func Parse(filename string, funcs template.FuncMap) (*template.Template, error) 
 			}
 			return content, nil
 		},
+		"toJSON": func(v interface{}) (string, error) {
+			data, err := json.Marshal(v)
+			return string(data), err
+		},
+		"execTemplate": func(name string, ctx interface{}) (string, error) {
+			panic("[bug] Stub 'execTemplate' was not replaced")
+		},
 	}
 
-	return template.New(filename).Funcs(sprig.HermeticTxtFuncMap()).Funcs(funcs).Funcs(funcs2).Parse(string(raw))
+	t, err := template.New(filename).Funcs(sprig.HermeticTxtFuncMap()).Funcs(funcs).Funcs(funcs2).Parse(string(raw))
+	if err == nil {
+		t = t.Funcs(template.FuncMap{
+			"execTemplate": func(name string, ctx interface{}) (string, error) {
+				b := bytes.Buffer{}
+				err := t.ExecuteTemplate(&b, name, ctx)
+				return b.String(), err
+			},
+		})
+	}
+	return t, err
+
 }
 
 func GetBytesBuffer(filename string, data interface{}) (*bytes.Buffer, error) {
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -1,6 +1,5 @@
-{{ define "instance" -}}
+{{ define "instance-script" -}}
 {{- $S3URI := self.Parts.s3.Asset.S3URL -}}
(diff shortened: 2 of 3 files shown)


mumoshu:
@redbaron Thanks!
This looks nice, too :+1:
Perhaps this method can also be enhanced to use ignition in the future? Then, I prefer this over #732.

redbaron:
Whole UserData refactor was done as a preparation for Ingition

mumoshu:
@redbaron Oh, I had not realized that! Thank you very much 👍
Just confirming but have you verified this to work with ASG-based node pools? (SpotFleet-based ones are ok as they were verified by @camilb)

filereader/texttemplate/texttemplate.go
```diff
@@ -21,9 +22,27 @@ func Parse(filename string, funcs template.FuncMap) (*template.Template, error)
 			}
 			return content, nil
 		},
+		"toJSON": func(v interface{}) (string, error) {
```
mumoshu: `toJson` is already available through `Masterminds/sprig` according to https://github.com/Masterminds/sprig/blob/dba49a8d3a46ae8522f79e2c0241842d6ab613ca/defaults.go#L66
Could we use the one from sprig instead?

filereader/texttemplate/texttemplate.go
```diff
@@ -21,9 +22,27 @@ func Parse(filename string, funcs template.FuncMap) (*template.Template, error)
...
+			data, err := json.Marshal(v)
+			return string(data), err
+		},
+		"execTemplate": func(name string, ctx interface{}) (string, error) {
```
mumoshu: Could we use [the `template` func](https://golang.org/pkg/text/template/#hdr-Nested_template_definitions) bundled in `text/template` instead?

filereader/texttemplate/texttemplate.go
redbaron: I didn't check if it exists in  a code, because it is undocumented on their godoc page: https://godoc.org/github.com/Masterminds/sprig
 
so I am not sure we can rely on it being available all the time

filereader/texttemplate/texttemplate.go
redbaron: that is a workaround for `template` not being a real function :)  You can not pipe or capture its result

filereader/texttemplate/texttemplate.go
mumoshu: @redbaron I believe we can expect it to remain to be existing because it is used widely in many helm charts.

filereader/texttemplate/texttemplate.go
mumoshu: Oh, good to know! Thanks 😄

redbaron:
> Just confirming but have you verified this to work with ASG-based node pools?

No I didn't verify it. Your question made me check for UserData calls and I found that I forgot to update normal ASG worker userdata calls. It should be in sync now and givent that template they are calling is identical it should work, but I'd like to ask for e2e test  or apply it to a test cluster if you have it running locally.

filereader/texttemplate/texttemplate.go
redbaron: they use their own version: https://github.com/kubernetes/helm/blob/dece57e0baa94abdba22c0e3ced0b6ea64a83afd/pkg/engine/engine.go#L82

mumoshu: Thanks for the good catch 👍 Let me also add that both sprig's and helm's implementations of `toJson` seems to swallow errors, which sounds problematic to me. I'm now convinced to have our own implementation!

mumoshu:
@redbaron Thanks for the fix. Sure, I'll kick the E2E script before I go to sleep 🛌

mumoshu:
E2E passed. Thanks for your contribution as always, @redbaron 👍

PR #737, opened 2017-07-05
Drain node on spot instance termination notice as well

### LINKED ISSUE #734: Drain node on spot instance termination notice
We need to improve our new node drainer daemonset to also handle spot instance termination notice; when the instance is scheduled for termination, we need to drain the node, just like we do when instances are being rotated out of the auto scaling group.

danielfm:
Refactored node drainer code to also run in response to a spot instance termination notice.

Closes #734.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -610,7 +610,7 @@ write_files:
       {{ end }}
 
       {{ if .Experimental.NodeDrainer.Enabled }}
-      kubectl apply -f "${mfdir}/kube-node-drainer-asg-ds.yaml"
+      kubectl apply -f "${mfdir}/kube-node-drainer-ds.yaml"
       {{ end }}
 
       # Configmaps
@@ -1065,7 +1065,7 @@ write_files:
 {{ end }}
 
 {{if .Experimental.NodeDrainer.Enabled}}
-  - path: /srv/kubernetes/manifests/kube-node-drainer-asg-ds.yaml
+  - path: /srv/kubernetes/manifests/kube-node-drainer-ds.yaml
     content: |
         kind: DaemonSet
         apiVersion: extensions/v1beta1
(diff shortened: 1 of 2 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1126,26 +1126,59 @@ write_files:
...
+                      secs=$((secs + SPOT_POLL_INTERVAL))
+
+                      # Spot instance termination check
+                      http_status=$(curl -o /dev/null -w '%{http_code}' -sL http://169.254.169.254/latest/meta-data/spot/termination-time)
+                      if [ "${http_status}" -eq 200 ]; then
+                        termination_source=spot
+                        break
+                      fi
+
+                      if [ $((secs % SG_POLL_INTERVAL)) -eq 0 ]; then
```
redbaron: when it can possibly be not true?

danielfm: The idea is to run both checks in the same loop; the first check runs in every iteration (5s), whereas the second check runs every 6 iterations (30s).

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1126,26 +1126,59 @@ write_files:
...
+                    SG_POLL_INTERVAL=30
+
+                    # Used to control when to execute each termination check
+                    secs=0
+
+                    # Used to identify which was the source that requested the instance termination ('spot' or 'asg', for now)
+                    termination_source=''
+
+                    # Instance termination detection loop
+                    while sleep ${SPOT_POLL_INTERVAL}; do
```
redbaron: what do you think about splitting it into multiple wait loop, depending on where you run?

```
wait_on_asg() {
  echo "Awaiting ASG lifecycle hook for drain signal"
  while sleep 30; do
    state=$(asg describe-auto-scaling-instances --region "${REGION}" --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
    if [ "${state}" = Terminating:Wait ]; then
      break
    fi
done
}

wait_on_spotfleet() {
  echo "Awaiting spotfleet notification for drain signal"
  while sleep "${SPOT_POLL_INTERVAL}"; do
    http_status=$(curl -o /dev/null -w '%{http_code}' -sL http://169.254.169.254/latest/meta-data/spot/termination-time)
    if [ "${http_status}" -eq 200 ]; then
      break;
    fi
done
}

if [ "$(curl -o /dev/null -w '%{http_code}' -sL http://169.254.169.254/latest/meta-data/spot/)" -eq 200 ]; then
  termination_source=spot
  wait_func=wait_on_spotfleet
else
 termination_source=asg
 wait_func=wait_on_asg
fi

while ! $wait_func; do
 echo "[bug] wait func returned unexpected error, not draining." >&2
done

while true; do
  echo Node is terminating, draining it...
  ...
done
```

danielfm: I created one spot instance to mess around with those `/latest/meta-data/spot/*` endpoints, but in my tests this `/latest/meta-data/spot/` returns 404, so I think there's no way of distinguishing between a spot and an on-demand/reserved instance at this level.

core/controlplane/config/templates/cloud-config-controller
redbaron: oh, I missed that you increment by SPOT_POLL_INTERVAL but check for modulo SG_POLL_INTERVAL. make sense

core/controlplane/config/templates/cloud-config-controller
redbaron: it is sad :(

redbaron:
Thanks for you contribution!

PR #746, opened 2017-07-06
Update Kubernetes to v1.7.0

### LINKED ISSUE #744: Update Kubernetes to v1.7.0

mumoshu:
Closes #744

### DIFF
--- a/contrib/bump-version
+++ b/contrib/bump-version
@@ -6,11 +6,11 @@
 
 if [ $# -ne 1 ] || [ `expr $1 : ".*_.*"` == 0 ]; then
     echo "USAGE: $0 <target-version>"
-    echo "  example: $0 'v1.6.3_coreos.0'"
+    echo "  example: $0 'v1.7.0_coreos.0'"
     exit 1
 fi
 
-CURRENT_VERSION=${CURRENT_VERSION:-"v1.6.3_coreos.0"}
+CURRENT_VERSION=${CURRENT_VERSION:-"v1.7.0_coreos.0"}
 TARGET_VERSION=${1}
 
 CURRENT_VERSION_BASE=${CURRENT_VERSION%%_*}
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	k8sVer = "v1.6.3_coreos.0"
+	k8sVer = "v1.7.0_coreos.0"
 
 	credentialsDir = "credentials"
 	userDataDir    = "userdata"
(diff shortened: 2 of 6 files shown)


mumoshu:
Running E2E...

mumoshu:
Nanny is complaining:

```
I0706 07:50:20.831711       1 pod_nanny.go:63] Invoked by [/pod_nanny --cpu=80m --extra-cpu=4m --memory=200Mi --extra-memory=4Mi --threshold=5 --deployment=heapster-v1.3.0 --container=heapster --poll-period=300000 --estimator=exponential]
unknown flag: --threshold
Usage of /pod_nanny:
      --acceptance-offset=20: A number from range 0-100. The dependent's resources are rewritten when they deviate from expected by a percentage that is higher than this threshold. Can't be lower than recommendation-offset.
      --alsologtostderr[=false]: log to standard error as well as files
      --container="pod-nanny": The name of the container to watch. This defaults to the nanny itself.
      --cpu="MISSING": The base CPU resource requirement.
      --deployment="": The name of the deployment being monitored. This is required.
      --extra-cpu="0": The amount of CPU to add per node.
      --extra-memory="0Mi": The amount of memory to add per node.
      --extra-storage="0Gi": The amount of storage to add per node.
      --log-flush-frequency=5s: Maximum number of seconds between log flushes
      --log_backtrace_at=:0: when logging hits line file:N, emit a stack trace
      --log_dir="": If non-empty, write log files in this directory
      --logtostderr[=true]: log to standard error instead of files
      --memory="MISSING": The base memory resource requirement.
      --namespace="kube-system": The namespace of the ward. This defaults to the nanny pod's own namespace.
      --pod="heapster-v1.3.0-2885230499-6j11r": The name of the pod to watch. This defaults to the nanny's own pod.
      --poll-period=10000: The time, in milliseconds, to poll the dependent container.
      --recommendation-offset=10: A number from range 0-100. When the dependent's resources are rewritten, they are set to the closer end of the range defined by this percentage threshold.
      --stderrthreshold=2: logs at or above this threshold go to stderr
      --storage="MISSING": The base storage resource requirement.
      --v=0: log level for V logs
      --vmodule=: comma-separated list of pattern=N settings for file-filtered logging
```

mumoshu:
`--threshold` doesn't seem to be necessary now https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/cluster-monitoring/influxdb/heapster-controller.yaml

mumoshu:
We had to remove the `--estimator` flag, too. Re-running E2E...

mumoshu:
E2E passed on a cluster provisioned without tls bootstraping feature. TLS bootstrapping might have broken but it is an another story. See #744.

Issue #721, opened 2017-06-30
Support Kubernetes 1.7 new TLS bootstrapping features


danielfm:
More details [here](https://kubernetes.io/docs/admin/kubelet-tls-bootstrapping/#approval-controller) and [here](https://kubernetes.io/docs/admin/authorization/node/).

mumoshu:
Thanks for bringing this up!
I'm especially looking forward to the creds rotation alpha feature 😃

mumoshu:
@danielfm Can we close this as resolved now?

* For the cert rotation feature, I've created #881 
* Probably we'd want to enable this feature by default at some point. If you agree, could you create a separate issue for that?

danielfm:
Of course, closing!

PR #749, opened 2017-07-07
Fix kubelet bootstrap for Kubernetes 1.7

danielfm:
This PR fixes the automatic approval of CSR requests made by worker nodes during the bootstrap process in Kubernetes 1.7 when RBAC is enabled.

Quoting the documentation:

> In 1.7 the experimental “group auto approver” controller is dropped in favor of the new csrapproving controller that ships as part of kube-controller-manager and is enabled by default. The controller uses the SubjectAccessReview API to determine if a given user is authorized to request a CSR, then approves based on the authorization outcome. To prevent conflicts with other approvers, the builtin approver doesn’t explicitly deny CSRs, only ignoring unauthorized requests.

So, for the kubelet bootstrap to work in Kubernetes 1.7, we need to add a cluster role+binding to instruct the CSR approver to auto-approve requests made on behalf of the kubelet bootstrap group `system:kubelet-bootstrap-token`.

As of now, we cannot enable the Node authorizer and the NodeRestriction admission controller since it will break our node drainer implementation. (more info:  https://github.com/kubernetes/kubernetes/issues/48666)

Ref.: #721

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1134,8 +1134,7 @@ addons:
 #   # certificate signing requests (csr) made to the API server using the bootstrap token. It's recommended to
 #   # also enable the rbac plugin in order to limit requests using the bootstrap token to only be able to make
 #   # requests related to certificate provisioning.
-#   # The bootstrap token must be defined in ./credentials/tokens.csv; kube-aws creates this file automatically
-#   # with a randomly generated token when this setting is enabled.
+#   # The bootstrap token is automatically generated in ./credentials/kubelet-tls-bootstrap-token.
 #   tlsBootstrap:
 #     enabled: true
 #   # This option has not yet been tested with rkt as container runtime
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -650,8 +650,13 @@ write_files:
       done
 
       {{ if .Experimental.TLSBootstrap.Enabled }}
-      kubectl create -f "${mfdir}/cluster-roles/node-bootstrapper.yaml" | echo 'Failed to create the cluster role named "kube-aws:node-bootstrapper". Skipping'
-      kubectl create -f "${mfdir}/cluster-role-bindings/node-bootstrapper.yaml" | echo 'Failed to create the cluster role binding "kube-aws:node-bootstrapper". Skipping'
+      for manifest in {node-bootstrapper,kubelet-certificate-bootstrap}; do
+          kubectl apply -f "${mfdir}/cluster-roles/$manifest.yaml"
+      done
+
+      for manifest in {node-bootstrapper,kubelet-certificate-bootstrap}; do
+          kubectl apply -f "${mfdir}/cluster-role-bindings/$manifest.yaml"
+      done
       {{ end }}
       {{ end }}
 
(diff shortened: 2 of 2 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1187,10 +1187,15 @@ addons:
...
 #   # also enable the rbac plugin in order to limit requests using the bootstrap token to only be able to make
 #   # requests related to certificate provisioning.
-#   # The bootstrap token must be defined in ./credentials/tokens.csv; kube-aws creates this file automatically
-#   # with a randomly generated token when this setting is enabled.
+#   # The bootstrap token is automatically generated in ./credentials/kubelet-tls-bootstrap-token.
 #   tlsBootstrap:
 #     enabled: true
+#     # If enabled, instructs the API server to enable the Node authorizer.
+#     # Only supported in Kubernetes 1.7.
+#     nodeAuthorizer: false
```
mumoshu: Is the node authorization feature part of the TLS bootstrapping feature?
If not, should be extract the `nodeAuthorizer` setting to the upper level:
```yaml
tlsBoostrap;
  enabled: true
# If enabled, instructs the API server to enable the Node authorizer.
# Only supported in Kubernetes 1.7.
# Requires `tlsBoostrap` to be enabled
nodeAuthorizer:
  enabled: true
```

danielfm: Sorry for the delay.

These are unrelated features, but in the context of kube-aws, having TLS bootstrap enabled is a requirement for the Node authorizer to work because it expects nodes to use credentials that conform with the naming scheme (where username = `system:node:<node name>`, and group = `system:nodes`).

mumoshu: @danielfm Thanks for the confirmation 👍 
Does it imply that we can improve the situation by modifying kube-aws to generate worker credentials with username of e.g. `system:node:worker`(`worker` part could be anything)  and group `system:nodes`?

danielfm: Unfortunately, I don't think so. The `<name>` part of `system:node:<name>` must match the node name, which currently is based on the assigned private IP, which we have no control over.

mumoshu: Thanks, I got the point 👍 
Then, I'm ok with the nesting.

mumoshu:
@danielfm Please notify me once this gets ready to be reviewed 😄

danielfm:
@mumoshu I repurposed this PR to address the things I can for the time being. Unfortunately other cool 1.7 features would have to wait. 😢

mumoshu:
@danielfm Thanks for taking your time 👍 
Just for my education but what would be the cool 1.7 features left? Enabling the NodeRestriction admission control probably after the upstream is released with your improvement included?

redbaron:
Thanks for polishing all the rough edges around RBAC and Kube 1.7 upgrade!

redbaron:
BTW, I remember there were some roles created in 1.5 time, are they still in place? Do we still need them?

danielfm:
@mumoshu 

> Just for my education but what would be the cool 1.7 features left?

NodeRestriction admission controller + Node authorizer + Kubelet’s credential rotation. I'm a simple man; when I see security improvements, I like. 😄

danielfm:
@redbaron 

> BTW, I remember there were some roles created in 1.5 time, are they still in place?

I think most (all?) 1.5-specific roles and bindings were removed, but I can take another look.

mumoshu:
@danielfm Thanks for you confirmation.

> NodeRestriction admission controller + Node authorizer + Kubelet’s credential rotation. I'm a simple man; when I see security improvements, I like. 😄

Just watched where the upstream PR ended up - your fix seems not being back-ported to k8s 1.7.x but instead, perhaps it is going to be included in k8s 1.8?
If you think so too, I'd like to mention about "NodeRestriction admission controller + Node authorizer + Kubelet’s credential rotation" in our roadmap (probably in kube-aws v0.9.9) because it is such a great set of features 😉

danielfm:
@mumoshu if Kubernetes does not introduce any other change that breaks things for us, that should be no problem adding support for them in v0.9.9. I'll get back on this as soon as Kube v1.8.0 is released.

redbaron:
This change is going to be in Kubernetes 1.7.3 (or maybe 1.7.4), I managed to convince maintainers that it is a bug we are talking about, not an improvement.

PR #751, opened 2017-07-09
Initial plugins proposal

c-knowles:
For https://github.com/kubernetes-incubator/kube-aws/issues/509

### DIFF
--- a/proposals/plugins.md
+++ b/proposals/plugins.md
@@ -1,0 +1,146 @@
+# Introduction
+
+A draft design for [kube-aws plugins](https://github.com/kubernetes-incubator/kube-aws/issues/509). Plugins are initially defined as additions to a kube-aws generated
+cluster that are not part of the core feature set of providing an easy to use, highly available,
+best practice cluster on AWS. Plugins can also later be used for built-in functionality to provide a more modular deployment and ability to test.
+
+# Initial Use Cases
+
+1. A user wishes to deploy a component by default to their cluster instances outside of Kubernetes, e.g. a 
+monitoring agent.
+1. A user wishes to deploy a component into their cluster by default that is not part of core
+Kubernetes such as cluster wide services, addons, ingress controller or even persistent volumes.
+Reasons can vary from company restrictions to always having this component form part of every 
+cluster. e.g. [the rescheduler](https://github.com/kubernetes-incubator/kube-aws/issues/118)
+1. A user wishes to add [customised parts to the CloudFormation template](https://github.com/kubernetes-incubator/kube-aws/issues/96#issuecomment-263835808) 
+produced by kube-aws. Anything they add would need access to the same parameters and template 
+engine that already exists.
+1. A user wishes to customise existing parts of the CloudFormation template produced by kube-aws
+without necessarily having to diff all the files for changes with there sometimes being large
+conflicts or huge updates due to kube-aws changes that can be difficult to digest while
+being able to merge in the simple changes they need. Examples 
+[here](https://github.com/kubernetes-incubator/kube-aws/issues/340) and 
+[here](https://github.com/kubernetes-incubator/kube-aws/issues/338)
+
+# Specific Examples
+
+* Deployment outside of Kubernetes - Dex uses some additional API server parameter, Datadog agent still has more functionality if not using the containerised version
+* Deployment to Kubernetes - rescheduler, kube2iam
+* Add new parts to CloudFormation - bastion host
+* Customisations to existing parts of CloudFormation - IAM roles/groups
+
+# Plugin Principles
+
+1. Plugin functionality should first be added in places where kube-aws needs it most, i.e. where
+kube-aws receives more feature requests which do not fit with the core feature set.
+1. kube-aws should dogfood the plugins system, i.e. use the plugins system where possible on existing functionality
+1. Everything that is pluggable should be kept as modular as possible, from lists of IAM roles to kube API server parameters.
+1. Where possible, checks will be added to kube-aws to ensure the integrity of the cluster via validation or disallowing particular changes. e.g. disallow overriding node pool init metadata
+   * Note, this could be gradually added rather than forming part of the first implementation
+   * Some of this will likely be provided by CloudFormation but kube-aws could possibly replicas the checks to save time for the stack create/update to fail
+
+# Design
+
+## Plugin Types
+
+* Built-in plugins versus ad-hoc user specific plugins
+  * Some plugins will be provided by kube-aws and some will be specific to a user
+  * Both will use the same mechanisms to hook into kube-aws/deploy etc
+  * Both types of plugins will be enabled and configured in the same way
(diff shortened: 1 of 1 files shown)


proposals/plugins.md
```diff
@@ -0,0 +1,146 @@
...
+## Configuration
+
+All plugins will be enabled and configured via the `cluster.yaml` plus a set of plugin specific
+configuration files. However, as above, the whole configuration model will be passed to all plugins
+so they can access parameters from another plugin if necessary.
+
+## Plugin Locations
+
+* Built-in plugins can either be part of the kube-aws binary or downloadable from a secure source 
+if/when the user enables that plugin. The second is recommended.
```
mumoshu: I consider that *built-in* plugins should be part of the kube-aws binary hence "built-in",
while calling downloadable plugins *contrib* plugins.

mumoshu: I was also considering that:

* *contrib* plugins can be downloaded from `kubernetes-incubator/kube-aws/contrib/plugins/<PLUGIN_NAME>` to `<USER_DIRECTORY>/plugins/<PLUGIN_NAME>`
* The user can also install or create an adhoc plugin under `<USER_DIRECTORY>/plugins/<PLUGIN_NAME>`.

proposals/plugins.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Design
+
+## Plugin Types
+
+* Built-in plugins versus ad-hoc user specific plugins
+  * Some plugins will be provided by kube-aws and some will be specific to a user
+  * Both will use the same mechanisms to hook into kube-aws/deploy etc
+  * Both types of plugins will be enabled and configured in the same way
+* Built-in plugins versus contrib plugins
```
mumoshu: Please let me also add that *contrib* plugins is hosted remotely in the kube-aws repo which can be installed locally and then work similarly to *adhoc* plugins.

Related: https://github.com/kubernetes-incubator/kube-aws/pull/751/files#r126319557

proposals/plugins.md
```diff
@@ -0,0 +1,146 @@
...
+1. render
+   * Stack files are gathered and rendered locally, most plugins will perform a no-op here but some may choose to output additional CloudFormation stacks or perform additional validation. A core plugin could grab all the primary constructs and create the main CloudFormation or a kube API server plugin could validate none of the parameters it has set are conflicting.
+   * Input - output of `initialise` stage
+   * Output - CloudFormation file(s), other file reference(s) such as Helm Charts
+1. deploy
+   * Take the CloudFormation and file(s), upload the files to S3 and deploy the template(s) to CloudFormation. This is an opportunity for plugins to perform waiting tasks and/or actions that need to be done on deploy/install, for example post to an endpoint created during deploy.
+   * Input - output of `render` stage, stack name
+   * Output - stack details such as stack name, endpoints, certs
+1. validate
```
mumoshu: How about naming this `validatePostConditions` as it takes the output of `deploy` stage as its input?

mumoshu: The motivation for suffixing it with "post" is that at first glance I misunderstood that `validate` corresponds to `kube-aws validate` which validates pre-conditions before running `kube-aws up` and `kube-aws update`.

proposals/plugins.md
c-knowles: I was considering built-in to mean they can be configured out of the box a bit easier but I think it's fine if they are part of the same package, I don't much of a preference either way for initial implementation.

I like your idea about contrib plugins going to the same place as the adhoc plugins.

proposals/plugins.md
c-knowles: Agreed, we can add that.

proposals/plugins.md
c-knowles: Happy to rename. I agree we should avoid name clashes in cases we expose this to the users, in this case it will be exposed directly in our operator documentation so worth picking something more descriptive. How about `postDeployValidate`? Or simply `postDeploy` is an option as well if we wish to allow plugins to perform other "post" actions.

mumoshu: Thanks, I like your idea of `postDeployValidate` 👍

proposals/plugins.md
mumoshu: @c-knowles Thanks for the confirmation 👍 
I'm even considering that a *built-in* plugin, when it is enabled, should be extracted from the kube-aws binary and then *rendered* into the working tree like *adhoc* plugins, so that it can be customized freely by the user without rebuilding the whole kube-aws binary.
What do you think?

cc @redbaron

mumoshu: > I was considering built-in to mean they can be configured out of the box a bit easier

Btw I agree on this 👍

mumoshu:
@c-knowles I'm still reconsidering/updating my implementation proposal of the plugin system at https://gist.github.com/mumoshu/3fa94cee7be23f5a8765c10dd6a91861#file-example-plugin-yaml.
For now, I'm rather going to provide the user `<plugin name>/plugin.yaml` for defining all the customizations to cluster.yaml, stack-templates, assets, etc, other than logics(like validations and defaulting which seems not feasible in yaml)
WDYT?

c-knowles:
@mumoshu if you think that's a good in road to get started I'm happy to roll with that and see where we get to.

proposals/plugins.md
```diff
@@ -0,0 +1,146 @@
...
+  battle tested; they will still be configured in the same way
+* Plugins will be allowed to act across the whole cluster or act on a specific set of node pools, this should form part of the base configuration for which plugins to run
+
+## Stages
+
+The workflow for kube-aws should consist of the following stages, each of which will be called on the plugins in turn. Each stage will complete across all plugins prior to proceeding to the next stage. All plugins will receive input based on the whole setup, not just their own. This will allow them to hook into parameters from other plugins or core logic. The stages are not intended to become kube-aws commands and should mostly be transparent to a kube-aws user.
+
+1. initialise
+   * Perform any prerequisite setup required such as calls to external system. Configuration, presence of files etc can be validated.
+   * Input - configuration
```
redbaron: how plugins communicate with core? will it be YAML? JSON? binary protocols? which communication channel is used? pipes? socket? passed file descriptors?

mumoshu:
Good point!

I have been scratching my head about that but I'm now considering grpc+yaml
by using hashicorp/go-plugin as a plugin system over grpc while marshalling
complex params as yaml.

I don't like to implement a yet another plugin system of my own, at least.
I considered golang's native plugin but it does't work on macOS until go
1.9.

WDYT?
2017年7月13日(木) 20:10 Maxim Ivanov <notifications@github.com>:

c-knowles:
My vote goes for the hashicorp plugin because it's available now, although I've never used it before so I don't actually mind picking another. I'd like to switch to native go plugins when it's available. Grpc seems a natural fit as well.

mumoshu:
@c-knowles Thanks for your efforts on this! #791 should provide the first impl of this.

PR #752, opened 2017-07-09
Deny direct command execution on privileged containers

ytsarev:
Deny host access through privileged containers with the help of
https://kubernetes.io/docs/admin/admission-controllers/#denyescalatingexec

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1468,7 +1468,7 @@ write_files:
           - --authentication-token-webhook-cache-ttl={{ .Experimental.Authentication.Webhook.CacheTTL }}
           {{ end }}
           - --advertise-address=$private_ipv4
-          - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass{{if .Experimental.Admission.PodSecurityPolicy.Enabled}},PodSecurityPolicy{{ end }},ResourceQuota
+          - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass{{if .Experimental.Admission.PodSecurityPolicy.Enabled}},PodSecurityPolicy{{ end }},ResourceQuota,{{if .Experimental.Admission.DenyEscalatingExec.Enabled}}DenyEscalatingExec{{end}}
           - --anonymous-auth=false
           {{if .Experimental.Dex.Enabled}}
           - --oidc-issuer-url={{.Experimental.Dex.Url}}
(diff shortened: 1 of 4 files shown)


mumoshu:
@ytsarev Thanks for the PR 👍 
I basically like this change except one concern about debugging.
I have been using `kubectl exec/attach` to debug privileged pods - how can we do that after this change? Logging-in to the node via `ssh` and then running  e.g. `docker exec`?

danielfm:
This is a nice feature, but I think it should be configurable by the user (and disabled by default).

ytsarev:
@mumoshu yes, debugging privileged would be quite inconvenient as you described.
So I will make it as a config option as @danielfm suggested

mumoshu:
@ytsarev @danielfm Thanks! Yes, I agree with you. I'd suggest adding `experimental.admission. denyEscalatingExec.enabled` to `cluster.yaml` for now.

ytsarev:
@mumoshu configuration option added, please review.
Also gave it an e2e test, the option works as expected
```
 kubectl exec -it kube-system-kube2iam-kube2iam-5gk4n -n kube-system -- sh
Error from server (InternalError): Internal error occurred: [cannot exec into or attach to a privileged container, object does not implement the Object interfaces]
```

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1468,7 +1468,7 @@ write_files:
           - --authentication-token-webhook-cache-ttl={{ .Experimental.Authentication.Webhook.CacheTTL }}
           {{ end }}
           - --advertise-address=$private_ipv4
-          - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass{{if .Experimental.Admission.PodSecurityPolicy.Enabled}},PodSecurityPolicy{{ end }},ResourceQuota
+          - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass{{if .Experimental.Admission.PodSecurityPolicy.Enabled}},PodSecurityPolicy{{ end }},ResourceQuota,{{if .Experimental.Admission.DenyEscalatingExec.Enabled}}DenyEscalatingExec{{end}}
```
mumoshu: I guess that the comma should go inside the block like `{{if .Experimental.Admission.DenyEscalatingExec.Enabled}},DenyEscalatingExec{{end}}`.
Anyway, I'll fix it myself myif my assumption turned to be correct 😉

mumoshu:
LGTM. @ytsarev Thank you very much for your contribution 👍

core/controlplane/config/templates/cloud-config-controller
mumoshu: JFYI: The trailing comma was ok. My apiserver is running without any problem with:
```
--admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,ResourceQuota,
```

ytsarev: Good to know, thanks

PR #755, opened 2017-07-11
Rename experimental.nodeLabels to controller.nodeLabels

mumoshu:
Closes #578

This is verified manually by running E2E

PR #756, opened 2017-07-11
Explicitly disallow tainting controller nodes

### LINKED ISSUE #754: Disallow specifying `experimental.taints`
The `experimental.taints` key had been added to cluster.yaml via #132 in the end of last year for tainting controller nodes.
As the name doesn't express the fact that it is only for controller nodes very well, and also it has been several releases since it is added as an experimental feature, I suggest renaming.

cc @redbaron @c-knowles @danielfm @camilb

mumoshu:
This PR depends on #755. Please see the last commit for changes made solely in this PR.

Resolves #754

This is verified manually by running E2E

### DIFF
--- a/core/nodepool/config/config.go
+++ b/core/nodepool/config/config.go
@@ -332,6 +332,10 @@ func (c ProvidedConfig) valid() error {
 		return err
 	}
 
+	if err := c.NodeSettings.Validate(); err != nil {
+		return err
+	}
+
 	clusterNamePlaceholder := "<my-cluster-name>"
 	nestedStackNamePlaceHolder := "<my-nested-stack-name>"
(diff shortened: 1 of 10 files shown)


core/nodepool/config/config.go
```diff
@@ -332,6 +332,10 @@ func (c ProvidedConfig) valid() error {
 		return err
 	}
 
+	if err := c.NodeSettings.Validate(); err != nil {
```
danielfm: nit for a later PR: standardize the name of these validation functions.

mumoshu: Must do! Submitted the issue #759 to track it.

mumoshu:
@danielfm Thanks for your review 👍

PR #763, opened 2017-07-12
Fix node labeling to allow scheduling cluster-autoscaler to workers

### LINKED ISSUE #758: Unable to schedule cluster-autoscaler to worker nodes
cluster-autoscaler(CA) is scheduled to controller nodes by default.
It is designed to be configured to be scheduled to worker nodes by setting `worker.nodePools[].clusterAutoscalerSupport.enabled` to `true` in `cluster.yaml`. 

However, my manual testing revealed that the setting doesn't work at all.
More concretely, a `kube-aws.coreos.com/cluster-autoscaler-supported` node label which is expected to be added to nodes when the support is enabled is just missing from the nodes.

mumoshu:
Fixes #758

### DIFF
--- a/core/nodepool/config/config.go
+++ b/core/nodepool/config/config.go
@@ -267,6 +267,14 @@ func (c ProvidedConfig) Config() (*ComputedConfig, error) {
 	return &config, nil
 }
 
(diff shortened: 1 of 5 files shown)


mumoshu:
This is tested manually by running E2E

PR #770, opened 2017-07-16
Containerized test run

ytsarev:
Local `make test` fails for me miserably so
`containerized-test-run` was created in a similar
fashion to `containerized-build-release-binaries`


mumoshu:
@ytsarev Wow, this looks great! Thanks for your contribution 👍

Issue #747, opened 2017-07-06
Reusing existing IAM roles for etcd nodes and ASG lifecycle notification


mclarke47:
Really happy to see [607](https://github.com/kubernetes-incubator/kube-aws/pull/607) make it into 0.9.7 but it would be great to have it in for etcd nodes too.

mumoshu:
@mclarke47 Thanks for the feedback 👍 Just curious but would you mind clarifying what's your use-case for having the flexible IAM configuration for etcd nodes?

mumoshu:
I just noticed that you've modified the subject to contain "ASG notification role" - but what do you mean by that?

mclarke47:
Frequently developers do not have direct control over IAM and networking in their aws account. Kube-aws has solved the networking issue by allowing users to define VPCs/subnets which already exist. iAM however seems to be a bit half implemented, you can define existing instan

VinceMD:
I'm in the same scenario as mclarke47. My AWS user cannot alter anything IAM. Instead I have prepared the roles and instanceProfile and handed over to my SEC/cloud department so they can set them up as one-off after review.

But for etcd I have to manually edit the cloudformation script before applying via console. I'm not sure how it will work with kube-aws update.

Separate but related topic, I've worked out minimum set of policies.  ec2:* and elasticloadbalancing:* was too broad and not accepted where I work. I've created a smaller set of policies in a form of a cloudformation script. I took inspiration from kops guy who are also working on that problem. My policies seem to work for kube-aws but wouldn't mind getting it reviewed and part of the kube-aws documentation for others to reuse. What's the best way to publish my work?

mumoshu:
@mclarke47 Thanks for the confirmation!

> would you agree that we should add support for existing IAM roles/instanceprofiles?

Definitely 👍 
I'll try to address this in v0.9.8-rc.1 - would you mind testing it if I could made it into the release?

mumoshu:
It is WIP at #761

VinceMD:
I can test too. In the same spirit, is there already a PR in progress for allowing usage of a pre-created ASGNotificationRole? ASGNotificationRole is always attempted to be created, even with autoscaler disabled

mumoshu:
@VinceMD Thanks!

Regarding the reusability of existing notification roles, we don't have one for that yet.

I'm even not sure how we make it customizable.
Would you mind discussing about the spec with me?

[...]

PR #775, opened 2017-07-18
Allow reusing an existing autoscaling notification target

### LINKED ISSUE #771: Reusing an existing autoscaling notification target
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/747#issuecomment-314756229
cc @VinceMD @mclarke47

mumoshu:
Resolves #771

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1214,6 +1214,8 @@ addons:
 #    role:
 #      # When specified, kube-aws reuses an existing IAM role as the notification role of a autoscaling lifecycle hook
 #      arn: "arn:aws:iam::YOURACCOUNTID:role/ROLENAME"
+#  target:
+#    arn: "arn:aws:sqs::YOURACCOUNTID:QUEUENAME"
 
 # User-provided YAML map available in control-plane's stack-template.json
 #customSettings:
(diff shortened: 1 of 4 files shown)


redbaron:
Does node drainer work without specified notification target all together? If so we should remove it instead of making it customizable

mumoshu:
@redbaron Thanks for the comment!
Would you mind taking a look into [my comment](https://github.com/kubernetes-incubator/kube-aws/issues/747#issuecomment-314984908)?

redbaron:
Queue was used to make lifecycle hook happy. If it can now work without it, there is no need for the queue. I can see no meaningful way it can be used for debugging, as only user of the queue was ASG placing termination messages which are then never consumed. so only information you could extract from it is whether ASG wants instance to be terminated, you can get that info just by querying  asg status

mumoshu:
@redbaron Thanks for the clarification.

> so only information you could extract from it is whether ASG wants instance to be terminated, you can get that info just by querying asg status

This did make sense to me 👍

danielfm:
> Queue was used to make lifecycle hook happy. If it can now work without it, there is no need for the queue.

Exactly, it was not possible to create the ASG lifecycle hook via CloudFormation without a target SQS queue/SNS topic, so that's the only reason it exists.

mumoshu:
@redbaron @danielfm Thanks! Closing this in favor of #780

PR #780, opened 2017-07-19
Stop using unnecessary autoscaling notification target/role

mumoshu:
The target/role was originally introduced while improving the node draining feature because CloudFormation seemed to require the target/role for an autoscaling lifecycle hook.
However, a [recent improvement](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html) on CloudFormation resulted in the target/role to be optional parameters. As our node draining doesn't actually rely on them, we can safely remove them now.

This makes #747 and #771 unnecessary

cc @redbaron @danielfm


VinceMD:
I was able to spin up a cluster with this revision and it did not create SQS nor associated role. 
The cloud formation and kube-aws indicate that the creation was completed successfully.

Would this be sufficient as a test or do we want to do more?

redbaron:
@VinceMD did you use nodeDrainer feature? Check if node-drainer daemon set is running. If you make CloudFormation to roll nodes (lets say add space somewhere in worker's userdata), and tail node-drainer logs on some of the node which is about to restart, do you see that it picked up node status and performs draining?

redbaron:
I wonder if anything can be deleted from root CF stack,  wouldn't it cause update failure?

danielfm:
@mumoshu nice! I edited your PR description to provide the link to CloudFormation changelog, just for reference.

danielfm:
> I wonder if anything can be deleted from root CF stack, wouldn't it cause update failure?

Just tested this in a pet cluster, and the update went through without problems.

redbaron:
@danielfm , @mumoshu , @VinceMD , thanks for contributing to this improvement!

VinceMD:
That's awesome. This is the last missing piece: now, kube-aws can work in a restricted AWS account (no ability to dynamically create IAM roles) without having to alter templates or generated CF.

PR #786, opened 2017-07-20
WIP: An attempt to refactor UserData assets management

mumoshu:
I went for a ramble in the codebase and came up with a possible improvement like this for less coupling between the `cfnstack/assets` package and the `model` package, less fields in `UserDataPart`, etc.

The major difficulty preventing this work from being finished is that:

```
"UserData": {{ $.UserDataEtcd.Parts.instance.Base64 true (dict "etcdIndex" $etcdIndex) |
```

in the stack template doesn't work anymore due to that I changed UserDataPart to render the template on its creation hence the `Base64` func to not take the second argument.

Anyway, I'm submitting this so that anyone who is interested hopefully comes and leaves comments 😉

### DIFF
--- a/model/userdata.go
+++ b/model/userdata.go
@@ -3,17 +3,20 @@ package model
 import (
 	"github.com/coreos/coreos-cloudinit/config/validate"
 	"github.com/kubernetes-incubator/kube-aws/filereader/texttemplate"
+	"github.com/kubernetes-incubator/kube-aws/fingerprint"
 	"github.com/kubernetes-incubator/kube-aws/gzipcompressor"
 
 	"bytes"
 	"encoding/base64"
 	"fmt"
+	"path"
 	"strings"
 	"text/template"
 )
 
 // UserDataValidateFunc returns error if templated Part content doesn't pass validation
 type UserDataValidateFunc func(content []byte) error
+type UserDataTemplateExtraParams map[string]interface{}
 
 const (
 	USERDATA_S3       = "s3"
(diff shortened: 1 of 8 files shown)


redbaron:
@mumoshu , I'll have a look at it in next couple of days

model/userdata.go
```diff
@@ -83,33 +106,48 @@ func NewUserData(templateFile string, context interface{}, opts ...UserDataOptio
...
 	}
 	return v, nil
 }
 
-func (self UserDataPart) Base64(compress bool, extra ...map[string]interface{}) (string, error) {
-	content, err := self.Template(extra...)
-	if err != nil {
-		return "", err
-	}
+func (self *UserData) templateFileBaseName() string {
```
redbaron: where these funcs are going to be used and what for?

model/userdata.go
```diff
@@ -22,14 +25,13 @@ const (
...
 // UserData represents userdata which might be split across multiple storage types
 type UserData struct {
-	Parts map[string]*UserDataPart
+	Parts            map[string]*UserDataPart
+	templateFilePath string
 }
 
 type UserDataPart struct {
-	Asset    Asset
-	tmpl     *template.Template
```
redbaron: keeping original template allows UserDataPart to be lazy which is a requirement for `extraContext` calls in CF stack template

model/userdata.go
```diff
@@ -22,14 +25,13 @@ const (
 
 // UserData represents userdata which might be split across multiple storage types
 type UserData struct {
-	Parts map[string]*UserDataPart
+	Parts            map[string]*UserDataPart
+	templateFilePath string
```
redbaron: why does it need to carry template file path?

mumoshu:
Closing this due to inactivity(of me!)

Issue #118, opened 2016-12-05
Feature: Rescheduler add-on


c-knowles:
Do we want to support the [rescheduler add-on](http://kubernetes.io/docs/admin/rescheduler/) out of the box? Useful links - [k8s](https://github.com/kubernetes/kubernetes/pull/30870), [kops](https://github.com/kubernetes/kops/issues/355)

mumoshu:
@c-knowles Thanks for bringing this up 👍 

Yes. I even believe it should be enabled by default.

pieterlange:
Lets feature (experimental) flag it first

mumoshu:
Note: [Rescheduler e2e test](https://github.com/piosz/kubernetes/blob/4a6123df0aceff6ae9e5522c836d62128782250c/test/e2e/rescheduler.go) should be run against kube-aws created clusters in https://github.com/coreos/kube-aws/tree/master/e2e

mumoshu:
@c-knowles Do you have any time to address this?

c-knowles:
Yes, can do. What's kube-aws's philosophy when it comes to which add-ons we install? For example, I'd like to add [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) too plus some of the [others](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons) maybe even using the add-on manager. We can likely add all that but make them optional where k8s doesn't require it to run (the DNS is an add-on but most people will expect it to be running since the k8s docs list how to use it and say it's built in)

mumoshu:
I'd like to keep kube-aws focused on:

* High availability
* High scalability
* Security
* Flexibility

Accordingly, I'd like to provide special treatment(e.g. can be enabled easily via `cluster.yaml`) for addons like rescheduler because it allows users achieve high availability, one of the focuses of kube-aws.

For addons don't directly contribute to scalability, availability and security, I'd like not to support them specially in e.g. cluster.yaml but rather provide a way to provide arbitrary addons to be installed during `kube-aws up`.

It would be something like `addons/` directory under kube-aws state directory to which an user feed arbitrary manifest yaml files.

mumoshu:
Btw, if you like the idea of `addons/` directory, I can submit a pull request containing the PoC for it.

c-knowles:
If we do the add-ons directory I think we should use the add-on manager as that seems to be the normal way to do this in k8s.

mumoshu:
Yes.

[...]

PR #787, opened 2017-07-20
Rescheduler logs now piped to docker

c-knowles:
Rather than being stored inside the container instance. For https://github.com/kubernetes-incubator/kube-aws/issues/118.

v0.3.1 has the fix for https://github.com/kubernetes/contrib/issues/2518.

I've deployed this on a dev cluster and I can see rescheduler logs in the cluster logs again.


redbaron:
Thanks for your contribution!

Issue #462, opened 2017-03-24
document possible git workflow


redbaron:
there are many requests coming in for feature which are alredy possible to be implemented without  a change to kube-aws. Deepest level of customisation can be achieved with simple git workflow:

Initial setup:
 1. `git checkout --orphan vanilla` creates `vanilla` branch with no history
 1. `kube-aws init --cluster-name CHANGEME ....  && git add .&& git commit -m "update with $(kube-aws version)"` - create initial structure and commit it with no modifications/customizations. All mandatory values are `CHANGEME` placeholders
 1. `git checkout -b cluster/main1 vanilla` - create new branch based on latest vanilla
 1. `vim cluster.yaml; vim userdata/....; git commit -m "my changes to cluster1"` - apply any customization to your templates,userdata, cluster.yaml etc. Feel free to change security groups, IAM roles,add ebs volumes, whatever...

Updating to new kube-aws:
 1. `git checkout vanilla` - go back to vanilla branch.
 2. `rm -rf * && kube-aws init --cluster-name CHANGEME && git add .  && git commit -m "update with $(kube-aws version)"` - create new version of vanilla commit, replacing entire content of previous one. Because there were no modifications to it, no need to carefully merge changes or anything
 3. `git checkout cluster/main1 && git merge vanilla` - pull changes from recent vanilla into your cluster branch, resolve conflicts, remove obsoleted bits etc.

Important thing is that every commit to vanilla must be a direct result of `kube-aws` calls,no editing afterwards.they use power of git to propagate those changes to your heavily customized clusters.

If you have multiple cluster branches, you can use `git rerere` to avoid redoing conflicts resolution for each one of them when it is possible.

If there are many cluster branches, it is benefitial to have additional level in the form of `tailored` branch,which is created from `vanilla` ,but contains customizations needed by all clusters. Cluster branches then pull changes not from `vanilla`, but from `tailored`.

Maybe we should add suggested workflow like that to official docs?

mumoshu:
Definitely! How about adding `Documentation/kubernetes-on-aws-workflow.md` with a title like "Recommended workflow for managing multiple kube-aws clusters in code"? The shorter the better though 😃

PR #789, opened 2017-07-23
Ability to propagate custom options to kubelet

ytsarev:
Main motivation for us is to utilize
https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/

Anything from https://kubernetes.io/docs/admin/kubelet/
can be potentially used as a value

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -622,7 +622,7 @@ write_files:
     permissions: 0755
     owner: root:root
     content: |
-      KUBELET_OPTS=""
+      KUBELET_OPTS="{{.Experimental.KubeletOpts}}"
 
   - path: /etc/kubernetes/cni/docker_opts_cni.env
     content: |
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -468,6 +468,12 @@ worker:
 #      kube2IamSupport:
 #        enabled: true
 #
+#      # Propagate custom CLI options to kubelet
+#      # Provided example enables docker image garbage collection as in https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/
+#      # Anything from https://kubernetes.io/docs/admin/kubelet/ can be used as a value
+#
+#      kubeletOpts: --image-gc-low-threshold=50 --image-gc-high-threshold=65
+#
 #      # Kubernetes node labels to be added to worker nodes
 #      nodeLabels:
 #        kube-aws.coreos.com/role: worker
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -469,7 +469,8 @@ worker:
 #        enabled: true
 #
 #      # Propagate custom CLI options to kubelet
-#      # Provided example enables docker image garbage collection as in https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/
+#      # Provided example overrides default docker image garbage collection limits
+#      # as in https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/
 #      # Anything from https://kubernetes.io/docs/admin/kubelet/ can be used as a value
 #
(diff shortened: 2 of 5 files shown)


mumoshu:
@ytsarev Thank you very much for your contribution 👍 
Not saying this shouldn't be merged but just curious - what would be pros of exposing this via `cluster.yaml`?
Like DOCKER_OPTS passed to dockerd, we've been recommending users to edit and version-control files generated by kube-aws so that they can make all the kinds of changes possible via not only cluster.yaml but cloud-configs and stack templates.
Do you find exposing it via cluster.yaml better for some reasons?

ytsarev:
@mumoshu actually we intentionally never modify cloud formation templates. We version control them , but modification go strictly through `kube-aws render stack`. This way we avoid any ad-hoc modifcations and possible incompatibility with a tool. I believe such approach brings reliability I would like to stick with it.

mumoshu:
@ytsarev Thanks, I understand the motivation 👍

mumoshu:
Hey @redbaron, WDYT on this?

I believe we can't prioritize these kinds of improvements over other scalability, availability, reliability things we're focusing on - creating a cloudformation.yaml isn't our focus :wink:
However, I also believe that we can accept contributions of these kinds as long as it doesn't badly affect our main focus. I understand @ytsarev's motivation on this and it doesn't seem to affect our focus badly. It is useful and orthogonal to our focus.

So, I'm wanting this to be merged. WDYT?

core/controlplane/config/templates/cluster.yaml
```diff
@@ -468,6 +468,12 @@ worker:
 #      kube2IamSupport:
 #        enabled: true
 #
+#      # Propagate custom CLI options to kubelet
+#      # Provided example enables docker image garbage collection as in https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/
```
redbaron: is it correct? Isn't GC enabled by default, just with different thresholds?

redbaron:
@ytsarev , there is an issue about git workflow where it is possible to keep your customizations and continue receiving updates from kube-aws: https://github.com/kubernetes-incubator/kube-aws/issues/462

@mumoshu , PR looks allright, cluster.yaml already grew beyond point of no return :)

core/controlplane/config/templates/cluster.yaml
ytsarev: @redbaron definitely not enabled by default, at least not explicitly

redbaron: From your link:

```
User Configuration
Users can adjust the following thresholds to tune image garbage collection with the following kubelet flags :

image-gc-high-threshold, the percent of disk usage which triggers image garbage collection. Default is 90%.

image-gc-low-threshold, the percent of disk usage to which image garbage collection attempts to free. Default is 80%.
```

ytsarev: ok, then they are too high by default ,right. Want me to change formulation in the comment ?

redbaron: yes please. we have many people starting playing with kube using kube-aws, providing accurate information is important. Original formulation gives impression that GC is switched off, which might cause panic :)

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -622,7 +622,7 @@ write_files:
     permissions: 0755
     owner: root:root
     content: |
-      KUBELET_OPTS=""
+      KUBELET_OPTS="{{.Experimental.KubeletOpts}}"
```
redbaron: I think both controller and worker kubelets should be using KubeletOpts

ytsarev: Indeed, i will amend it

mumoshu:
@redbaron 

> cluster.yaml already grew beyond point of no return :)

True :rofl:

core/controlplane/config/templates/cloud-config-worker
ytsarev: @redbaron PR amended, please review

redbaron:
Looks good now, thanks for your contribution!

PR #791, opened 2017-07-24
Plugin System

mumoshu:
This is an initial implementation of the plugin system #509 as proposed in #751. Not all but most of knobs mentioned in the proposal except pre/post-cluster-creation validations are implemented.

Basically, it allows the user to define a set of customizations to various aspects of resources created and managed by kube-aws as a "kube-aws plugin" and reuse it.
The set of customizations is defined in a separate file other than a `cluster.yaml` for reusability.

More concretely, provide `<your project root>/plugins/<your-plugin-name>/plugin.yaml` like seen in test/integration/plugin_test.go to extend a kube-aws cluster from many aspects including:

- additional iam policy statements per node role(worker/controller/etcd)
- additional cfn stack template resources per stack(root/control-plane/node-pool)
- additional systemd units/custom files per node role(worker/controller/etcd)
- additional kubelet feature gates for worker kubelets
- additional node labels for worker/controller kubelets

and so on.

The new plugin system is not used to implement core features of kube-aws yet. Therefore, I believe we don't need to worry much about breaking things via this change.

At least one core feature implemented as a plugin is planned in the next version of kube-aws v0.9.9, as noted in our roadmap.

Non-TODO:

* Plugin binaries to be used for pre/post validations

TODO:

- [x] Custom systemd units
  - [x] For worker nodes
  - [x] For controller nodes
  - [x] For etcd nodes
- [x] Kubelet options
  - [x] Node Labels
    - [x] For worker nodes
    - [x] For controller nodes
  - [x] Feature Gates
    - [x] For worker nodes
- [x] Kubernetes manifests
- [x] Helm charts
- [x] API server options
- [x] Custom files
  - [x] For worker nodes
  - [x] For controller nodes
  - [x] For etcd nodes
- [x] Custom IAM policies
  - [x] For root stacks
  - [x] For control-plane stacks
  - [x] For worker node pool stacks
- [x] Refactoring: Consistency tweaks(naming, modularizion, etc)
- [ ] More access to configuration API in templates available in `plugin.yaml`
- [ ] More tests
- [ ] Split PRs if necessary

TODO after this PR:

- [ ] Rename the package `model` to `clustermodel`
- [ ] Rename the setting key `plugins` to something else in cluster.yaml
- [ ] Rename the setting key `kubeAwsPlugins` to `plugins` in cluster.yaml
- [ ] https://github.com/kubernetes-incubator/kube-aws/pull/791/files#r132388872
- [ ] https://github.com/kubernetes-incubator/kube-aws/pull/791#discussion_r132382700
- [ ] Utilize CustomFiles more https://github.com/kubernetes-incubator/kube-aws/pull/791#discussion_r132122572 https://github.com/kubernetes-incubator/kube-aws/pull/791#discussion_r132122593 https://github.com/kubernetes-incubator/kube-aws/pull/791#discussion_r132122644


mumoshu:
We'll be using the helm docker image whose code is hosted in the kube-aws org https://github.com/kube-aws/docker-helm because helm doesn't have an official docker image release.

mumoshu:
@c-knowles @redbaron @camilb @danielfm I guess I'm close to finish this - I'd appreciate it if you could review this and leave any comment 😃 

This is an initial implementation of our plugin system #509 as proposed in #751. Not all but most of nobs mentioned in proposal except pre/post validations are implemented.

Basically, it allows you to provide `<your project root>/plugins/<your-plugin-name>/plugin.yaml` like [this](https://github.com/kubernetes-incubator/kube-aws/pull/791/files#diff-941dbd01d3b0e4ed2e77ddcc4e23c43fR67) to extend a kube-aws cluster from many aspects:

- additional iam policy statements per node role(worker/controller/etcd)
- additional cfn stack template resources per stack(root/control-plane/node-pool)
- additional systemd units/custom files per node role(worker/controller/etcd)
- additional kubelet feature gates for worker kubelets
- additional node labels for worker/controller kubelets
- and so on...

The new plugin system is not used to implement core features of kube-aws yet. So I believe we don't need to worry much about breaking things.

Btw, at least one core feature implemented as a plugin is planned in the next version of kube-aws v0.9.9, as noted in our roadmap.

mumoshu:
Note: I tried my best not to further bloat existing parts of the kube-aws code-base by implementing the plugin system in brand-new packages and go sources whereas possible.

The plugin system implemented under the packages `plugin/*` are responsible to load `plugin.yaml`s, generates extra configuration appended/merged to the cluster config generated from `cluster.yaml`. The kube-aws core calls out to the plugin system only for fetching extra configs appended/merged to the cluster config.

test/integration/plugin_test.go
```diff
@@ -0,0 +1,528 @@
...
+						"assets/etcd/baz.txt":       "etcd-baz",
+						"assets/worker/baz.txt":     "worker-baz",
+					},
+					Yaml: `
+metadata:
+  name: my-plugin
+  version: 0.0.1
+spec:
+  configuration:
+    values:
```
mumoshu: This is the defaults for the values passed to templates like `cloudformation.stacks.{controlPlane,nodePool,root}.resources.append` and `kubernetes.apiserer.flags[].value`.

The defaults can be overridden from `cluster.yaml` like done in https://github.com/kubernetes-incubator/kube-aws/pull/791/files#diff-941dbd01d3b0e4ed2e77ddcc4e23c43fR52

test/integration/plugin_test.go
```diff
@@ -0,0 +1,528 @@
...
+    queue:
+      name: baz1
+    oidc:
+      issuer:
+        url: "https://login.example.com/"
+
+worker:
+  nodePools:
+  - name: pool1
+    kubeAwsPlugins:
```
mumoshu: I'd like to rename this to just `plugins` if we could agree with renaming the current `plugins` used solely for `rbac` to something else(perhaps `kubernetes` and `kubernetes.rbac` so that it nicely represents the fact it is used to enable `rbac` across the k8s cluster?)

test/integration/plugin_test.go
```diff
@@ -0,0 +1,528 @@
...
+		context       string
+		clusterYaml   string
+		plugins       []helper.TestPlugin
+		assertConfig  []ConfigTester
+		assertCluster []ClusterTester
+	}{
+		{
+			context: "WithAddons",
+			clusterYaml: minimalValidConfigYaml + `
+kubeAwsPlugins:
```
mumoshu: Same as https://github.com/kubernetes-incubator/kube-aws/pull/791/files#r132120639

test/integration/plugin_test.go
```diff
@@ -0,0 +1,528 @@
...
+`
+	validCases := []struct {
+		context       string
+		clusterYaml   string
+		plugins       []helper.TestPlugin
+		assertConfig  []ConfigTester
+		assertCluster []ClusterTester
+	}{
+		{
+			context: "WithAddons",
```
mumoshu: The name of test case doesn't seem valid 😉

mumoshu: Btw, I'm not very happy with this single-big-test-case-for-everything.

mumoshu:
I began to believe that today's `model` package could be renamed to `clusterapi` so that we can better represent that it is for defining a kube-aws cluster, whereas `pluginapi` introduced in this PR is for defining a kube-aws plugin.
Any thoughts?

plugin/pluginapi/config.go
```diff
@@ -0,0 +1,290 @@
...
+	return nil
+}
+
+// Spec is the specification of a kube-aws plugin
+// A spec consists of two parts: Configuration and Command
+type Spec struct {
+	// Configuration is the configuration part of a plugin which is used to append arbitrary configs into various resources managed by kube-aws
+	Configuration `yaml:"configuration,omitempty"`
+	// Command is the command part of a plugin which is used to hook arbitrary code into various extension points provided by kube-aws
+	Command CommandSpec `yaml:"command,omitempty"`
```
mumoshu: Currently unused. Could be removed before merging.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2220,6 +2285,33 @@ write_files:
...
+    encoding: gzip+base64
+    content: {{$f.Content.ToGzip.ToBase64}}
+{{ end }}
+
+  - path: {{.HelmReleasePlugin.ReleaseListFile.Path}}
+    encoding: gzip+base64
+    content: {{.HelmReleasePlugin.ReleaseListFile.Content.ToGzip.ToBase64}}
+
+
+{{ range $r := .HelmReleasePlugin.Releases }}
```
mumoshu: Could be merged into `.CustomFiles`

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2220,6 +2285,33 @@ write_files:
         status:
           loadBalancer: {}
 
+  - path: {{.KubernetesManifestPlugin.ManifestListFile.Path}}
+    encoding: gzip+base64
+    content: {{.KubernetesManifestPlugin.ManifestListFile.Content.ToGzip.ToBase64}}
+
+{{ range $m := .KubernetesManifestPlugin.Manifests }}
```
mumoshu: Could be merged into `.CustomFiles`

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2220,6 +2285,33 @@ write_files:
...
+{{ $f := $m.ManifestFile }}
+  - path: {{$f.Path}}
+    encoding: gzip+base64
+    content: {{$f.Content.ToGzip.ToBase64}}
+{{ end }}
+
+  - path: {{.HelmReleasePlugin.ReleaseListFile.Path}}
```
mumoshu: Could be merged into `.CustomFiles`

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2220,6 +2285,33 @@ write_files:
         status:
           loadBalancer: {}
 
+  - path: {{.KubernetesManifestPlugin.ManifestListFile.Path}}
```
mumoshu: Could be merged into `.CustomFiles`

c-knowles:
The original reason for introducing the model package was to provide clearer separation between the main processing logic and the data model. Hopefully it's made it easier to see what can be used into templates. Seems fine to rename it, clusterapi to me sounds like it's related to kube API but not sure of a better name.

mumoshu:
@c-knowles Thanks for the reply 👍 

> The original reason for introducing the model package was to provide clearer separation between the main processing logic and the data model

Yes, thank you very much for introducing it 👍
The separation worked very well so far and therefore I'd like to keep it.
That's why I've introduced the `pluginapi` package to clearly separate the data model and the logic around the plugin system.

> clusterapi to me sounds like it's related to kube API

I feel the same way.

I believe the root of the issue is that I myself tend to mix up two meanings of the word "cluster" - the surrounding kube-aws "cluster" composed of multiple cfn stacks, and the k8s "cluster" built on top of the kube-aws cluster. I'm still unsure how we could better differentiate those two meanings of "cluster" in our code base.

Or can we just call those "pluginmodel" and "clustermodel" rather than "pluginapi" and "clusterapi" so that they aren't confused with k8s "api"? WDYT?

mumoshu:
Slightly off-topic but I suppose that the model and the logic inside `core/{controlplane,nodepool,root}/config.go` should also be clearly separated.
Probably the most part of the model should go into the `clustermodel` package, whereas all the logic should remain where it is today.

c-knowles:
@mumoshu `pluginmodel` or even `model/plugin` seems clearer. I too use "cluster" to mean both of those, there's even a few instances of that in the new documentation.

mumoshu:
@c-knowles Thanks for the confirmation and the suggestion 👍 
I really appreciate having this kind of discussion because then we can better build up consensus on our code base.

Not trying to argue or insist anything but just to sync up with you, I prefer `pluginmodel` rather than `plugin/model` the same way `ioutil` is named so rather than `io/util`. The combination of `cluster/model` and `plugin/model`, when imported, could introduce local name collisions in code under the `core` package. (Oh, btw, the package name "core" won't be a good name once we moved to `clustermodel` and `pluginmodel` 😉)

> Note: There is no concept of sub packages in Go. This is why the ioutil package is called ioutil, not util with an import path of io/util. This avoids local namespace collisions.
> https://dave.cheney.net/2014/12/01/five-suggestions-for-setting-up-a-go-project

c-knowles:
Ah right yes, I'm not an expert in all things golang yet so that's my mistake regarding sub-packages. For the cluster naming my suggestion is we call the AWS infrastructure the "cluster" and we call the part on top of it just "Kubernetes". e.g. "Deploy Kubernetes on an AWS cluster using kube-aws". Either way I don't mind, I will try to make the docs consistent once there is consensus.

mumoshu:
@c-knowles Thanks again! I'm not an expert either; Honestly, I came to the conclusion on the package naming thing TODAY, even though I've seen many of examples while writing in golang 😉

Your suggestion on disambiguating "cluster" sounds good.

One of things I have not yet organized though is that, according to my observation, [the collocation "kubernetes cluster" seems to have some popularity](https://www.google.com/search?q="kubernetes+cluster") and the "cluster" in the collocation doesn't seem to clearly represent either a "k8s" cluster or a set of cloudprovider resources.
Also, ["kubernetes cluster addons" in the upstream repository](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons) is not named "kubernetes addons".

core/nodepool/config/config.go
```diff
@@ -275,6 +278,14 @@ func (c ProvidedConfig) NodeLabels() model.NodeLabels {
 	return labels
 }
 
+func (c ProvidedConfig) FeatureGates() model.FeatureGates {
```
mumoshu: Should be moved to `ComputedConfig` because it is where all the logic should go, whereas  `ProvidedConfig` is the part of the model.

core/nodepool/cluster/cluster.go
```diff
@@ -64,17 +66,44 @@ func NewClusterRef(cfg *config.ProvidedConfig, awsDebug bool) *ClusterRef {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load node pool stack extras from plugins: %v", err)
+	}
+	c.StackConfig.ExtraCfnResources = extraStack.Resources
+
+	extraWorker, err := extras.Worker()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load controller node extras from plugins: %v", err)
+	}
+	c.StackConfig.CustomSystemdUnits = append(c.StackConfig.CustomSystemdUnits, extraWorker.SystemdUnits...)
```
mumoshu: * This should be moved and implemented as `func (c *Cluster) CustomSystemdUnits() []model.CustomSystemdUnit` while adding `extraWorker` as a field of the type `Cluster`
  * Mutating `StackConfig` here doesn't seem intuitive as it spreads the logic what should be included in `StackConfig`
  * Even better, probably `StackConfig` should be merged into `Cluster`

mumoshu:
I had been on vacation until yesterday.

Now, this is open quite long.
Given the pile of changes stacked for v0.9.8-rc.1, I'm going to cut this as-is into the release so that we won't delay rc.1 further 😉

mumoshu:
e887558 includes a fix for https://github.com/kubernetes-incubator/kube-aws/pull/835#discussion_r134234477

mumoshu:
Running E2E tests against several differently configured clusters...

mumoshu:
E2E passed.

PR #792, opened 2017-07-24
Make PODs to resolve DNS names via locally running dnsmasq

dvdthms:
Workaround fix for https://github.com/kubernetes/kubernetes/issues/32749

Credit to @redbaron

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -438,7 +438,7 @@ type DeploymentSettings struct {
 	CloudWatchLogging       `yaml:"cloudWatchLogging,omitempty"`
 	AmazonSsmAgent          `yaml:"amazonSsmAgent,omitempty"`
 	CloudFormationStreaming bool `yaml:"cloudFormationStreaming,omitempty"`
-	KubeDnsMasq            `yaml:"kubeDnsMasq,omitempty"`
+	KubeDnsMasq             `yaml:"kubeDnsMasq,omitempty"`
 
 	// Images repository
 	HyperkubeImage                     model.Image `yaml:"hyperkubeImage,omitempty"`
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -127,6 +127,9 @@ func NewDefaultCluster() *Cluster {
 					interval: 60,
 				},
 			},
+			KubeDnsMasq: KubeDnsMasq{
+				DaemonSet: false,
+			},
 			HyperkubeImage:                     model.Image{Repo: "quay.io/coreos/hyperkube", Tag: k8sVer, RktPullDocker: false},
 			AWSCliImage:                        model.Image{Repo: "quay.io/coreos/awscli", Tag: "master", RktPullDocker: false},
 			CalicoNodeImage:                    model.Image{Repo: "quay.io/calico/node", Tag: "v1.2.1", RktPullDocker: false},
@@ -433,6 +436,7 @@ type DeploymentSettings struct {
 	WaitSignal             WaitSignal        `yaml:"waitSignal"`
 	CloudWatchLogging      `yaml:"cloudWatchLogging,omitempty"`
 	AmazonSsmAgent         `yaml:"amazonSsmAgent,omitempty"`
+	KubeDnsMasq            `yaml:"kubeDnsMasq,omitempty"`
 
 	// Images repository
 	HyperkubeImage                     model.Image `yaml:"hyperkubeImage,omitempty"`
@@ -634,6 +638,16 @@ type Plugins struct {
 
 type Rbac struct {
 	Enabled bool `yaml:"enabled"`
+}
+
+type KubeDnsMasq struct {
+	DaemonSet bool `yaml:"daemonSet"`
+}
+
+func (c *KubeDnsMasq) MergeIfEmpty(other KubeDnsMasq) {
+	if c.DaemonSet == false {
+		c.DaemonSet = other.DaemonSet
+	}
 }
 
 type WaitSignal struct {
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -205,6 +205,7 @@ coreos:
         Wants=flanneld.service cfn-etcd-environment.service
         After=cfn-etcd-environment.service
         [Service]
+        EnvironmentFile=/etc/environment
         EnvironmentFile=-/etc/etcd-environment
         Environment=KUBELET_IMAGE_TAG={{.K8sVer}}
         Environment=KUBELET_IMAGE_URL={{ .HyperkubeImage.RktRepoWithoutTag }}
@@ -261,7 +262,7 @@ coreos:
         --register-with-taints=node.alpha.kubernetes.io/role=master:NoSchedule \
         --allow-privileged=true \
         --pod-manifest-path=/etc/kubernetes/manifests \
-        --cluster-dns={{.DNSServiceIP}} \
+        --cluster-dns=${COREOS_PRIVATE_IPV4} \
         --cluster-domain=cluster.local \
         --cloud-provider=aws
         Restart=always
@@ -678,6 +679,9 @@ write_files:
 
       # Install tiller by default
       kubectl apply -f "${mfdir}/tiller.yaml"
+
+      # DNS Masq Fix
+      kubectl apply -f "${mfdir}/dnsmasq-node-ds.yaml"
 
       # Deployments
       for manifest in {kube-dns-de,kube-dns-autoscaler-de,{{ if .Addons.ClusterAutoscaler.Enabled }}cluster-autoscaler-de,{{ end }}heapster-de{{ if .KubeResourcesAutosave.Enabled }},kube-resources-autosave{{ end }}}.yaml; do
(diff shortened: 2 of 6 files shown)


dvdthms:
CLA now signed.

c-knowles:
Are we sure we want it before the upstream issue is agreed upon?

mumoshu:
@dvdthms Thank you very much for the PR 👍 This looks very interesting.

Not saying if this should be merged or not at this point but anyway, pros of this method mentioned in the upstream issue makes sense to me:

> * External lookups won't fail if kube-dns/endpoint controller/kube-proxy are down
> * Internal lookups will fail fast.

For instance, I myself have a poor app running in my production cluster which fails when dns lookups from its pods to our internal datastore service managed outside of k8s took 5 seconds each, probably due to intermittent failures in Amazon DNS.

I wish I could prevent failures like this at the infrastructure level and for me this seems to do that.
Would my understanding be correct?

mumoshu:
@c-knowles I'm not against moving in advance of the upstream if it helps users to run their production clusters in regards to availability, security, scalability, etc., which are goals of kube-aws.

I'd suggest to submit feedback(s) to the upstream once the workaround turned to work well for us, though!

danielfm:
@mumoshu also, should this be enabled by default?

mumoshu:
@danielfm Thanks for chiming in 👍
No, I'd like this to be disabled by default and configurable in cluster.yaml for now.

redbaron:
Running dnsmasq locally is a best practice in many non-kubernetes setups: it provides robustness and speed where it is needed at a very minimal cost of running small dnsmasq daemon.

We achieve that in kube-way by utilising DaemonSet instead of classic approach of a local systemd unit, but in a nutshell it is still an implementation of well known and battle tested approach.  Kube adds few additional daemons (sidecar containers), which may or may not be there, but I decided to keep them so that all kube-dns features such as monitoring or ConfigMap driven upstream DNS server configuration can continue to work in the same way as for kube-dns.

Additional benefit specific for Kubernetes in AWS case is better AvailabilityZone failure tolerance. In-kube traffic is not AZ-aware, therefore even if you spread kube-dns across AZs, problems with one of them will halts DNS resolution for all apps in the whole cluster.

I can't think of a single reason not to have dnsmasq locally on nodes, so this change doesn't have a corresponding cluster.yaml entry.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -205,6 +205,7 @@ coreos:
         Wants=flanneld.service cfn-etcd-environment.service
         After=cfn-etcd-environment.service
         [Service]
+        EnvironmentFile=/etc/environment
```
mumoshu: Is this probably an improvement irrelevant to the local dnsmasq thing?
If that's the case, would you mind submitting an another PR dedicated for it, or at least splitting commits?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1829,6 +1833,118 @@ write_files:
...
+              - name: healthz
+                image: {{ .ExecHealthzImage.RepoWithTag }}
+                resources:
+                  limits:
+                    memory: 50Mi
+                  requests:
+                    cpu: 10m
+                    memory: 50Mi
+                args:
+                - --cmd=nslookup ec2.amazonaws.com 127.0.0.1 >/dev/null
```
mumoshu: AFAIK, this eventually requires Amazon DNS dedicated to your VPC responds to the query, because it is Amazon DNS who resolves the general endpoint `ec2.amazonaws.com` to a region-specific EC2 endpoint.
If my assumption is correct, it won't work for users who doesn't use Amazon DNS.

So, assuming we don't allow disabling this feature at all, I guess we should at least allow users to customize the endpoint used for health-checking.

mumoshu:
@redbaron Thank you very much for the detailed explanation 👍 

Honestly,

* I agree with every point you've made
* I'll enable this feature in my production clusters, ASAP
* I'd like this feature to be HIGHLY recommended in our doc
* I'd like this feature to be enabled by default in the near future, probably starting v0.9.9

However, I'm still wanting this to be disabled by default for NOW.

It is just that I can't ensure this being supported by broad use-cases of kube-aws.
I'm afraid that there should be edge-cases like https://github.com/kubernetes-incubator/kube-aws/pull/792#discussion_r129461458 (not saying my comment is true or not), which would delay merging this PR if we'd like to enable it by default.

I'd like this to be merged earlier as an highly recommended but yet experimental feature so that we can start battle-testing it without worrying about edge-cases.
WDYT?

core/controlplane/config/templates/cloud-config-controller
redbaron: no, `${COREOS_IPV4}` value is read from there

core/controlplane/config/templates/cloud-config-controller
redbaron: `ec2.amazonaws.com` is resolvable even from your laptop, so it should be fine in any setup.

I just found that in kube-aws kube-dns runs with empty healthz container, which is completely pointless. I'll make healthz to look exactly the same as it is now in kube-aws , then' we'll address what to do with healthz in a separate PR

core/controlplane/config/templates/cloud-config-controller
mumoshu: Got it. Thanks for the clarification 👍

mumoshu: TODO Comment about its usage https://github.com/kubernetes-incubator/kube-aws/pull/792#discussion_r129494580 (It wasn't very obvious to me)

mumoshu:
@redbaron Thanks for the fix about dns health-checking 👍 
I'm ok with including it in this PR as long as the commit is separated.

Also, maybe you're in process of writing it but could you express your thoughts on https://github.com/kubernetes-incubator/kube-aws/pull/792#issuecomment-317916624, so that we can proceed (regardless of my wants are respected or not 😉)

It is certainly true I like the improvement in general. So, if you still think that we can go ahead with enabling this by default even after reading my thoughts, I'm ok.

redbaron:
yes, we'll work on cluster.yaml switch for this feature

mumoshu:
@redbaron Thank you very much for your cooperation 👍 I wish the switch works well for us all.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -1095,6 +1095,11 @@ worker:
 #  downloadUrl: https://github.com/DailyHotel/amazon-ssm-agent/releases/download/v2.0.805.1/ssm.linux-amd64.tar.gz
 #  sha1sum: a6fff8f9839d5905934e7e3df3123b54020a1f5e
 
+# When enabled, will enable a DNS-masq DaemonSet to make PODs to resolve DNS names via locally running dnsmasq
+# It is disabled by default.
+#kubeDnsMasq:
```
redbaron: I'd name it `kubeDns` after component name this feature is related to

redbaron:
@mumoshu do you have any feedback on this PR? @dvdthms waits for more suggestions to come (if there are any) so that all of them can be addressed in one go

core/controlplane/config/config.go
```diff
@@ -435,6 +438,7 @@ type DeploymentSettings struct {
 	CloudWatchLogging       `yaml:"cloudWatchLogging,omitempty"`
 	AmazonSsmAgent          `yaml:"amazonSsmAgent,omitempty"`
 	CloudFormationStreaming bool `yaml:"cloudFormationStreaming,omitempty"`
+	KubeDnsMasq             `yaml:"kubeDnsMasq,omitempty"`
```
mumoshu: How about renaming things so that the whole expression look like `NodeLocalKubeDnsMasq.Enabled` rather than `KubeDnsMasq.DaemonSet`?
I guess it will then better/more directly represents the fact that a dns masq pod is created for and per each node.

mumoshu:
@redbaron Thanks! Just added a comment about naming. Otherwise LGTM.

core/controlplane/config/config.go
redbaron: My thinking was that we should try to place new switches into top level groups , so that we can add more related options should we need to do so.

```
kubeDns:  # kube-dns related opts follow
  nodeLocalResolver: true    # run dnsmasq locally on every node
  monitorDomainsResoveTime:  # domains to be picked up by Prometheus
     -  ec2.amazonaws.com
     -  mycompany.internal.domain
     -  google.com
```

mumoshu: It sounds awesome! I like yours more than mine :+1:
If I could add one thing, `kubeDns.nodeLocalResolver` should be an object rather than a bool so that we can differentiate configs(like cache size, which isn't configurable right now) between node local dnsmasq and cluster-wide dnsmasq?
The node local resolver could be enabled via `kubeDns.nodeLocalResolver.enabled` then.

redbaron: itsn't it too much of a detail for no so much important feature? all dnsmasq should be configured equally, so just a flag under kube-dns should be enough, no? Otherwise  depth of config is comparable with nodepools :)

mumoshu: I'm ok with the depth if it provides long term api stability & extensibility but anyways,

Should all dnsmasq configured equally?(I just didn't know! 

Then, `kubeDns.nodeLocalResolverEnabled` would be good from my perspective.
If it doesn't look ok to you, I'm ok with `kubeDns.nodeLocalResolver` as a bool setting, as you've originally suggested!

dvdthms:
Updated the name :)

mumoshu:
As I've stated above, a noun phrase key for a bool setting doesn't sound very natural to me but probably I can live with as multiple contributors are saying that's better!

mumoshu:
@dvdthms Thanks for the fix :+1:

redbaron:
Thanks for your contribution, keep it coming! :)

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -261,9 +263,10 @@ coreos:
         --register-with-taints=node.alpha.kubernetes.io/role=master:NoSchedule \
         --allow-privileged=true \
         --pod-manifest-path=/etc/kubernetes/manifests \
-        --cluster-dns={{.DNSServiceIP}} \
-        --cluster-domain=cluster.local \
-        --cloud-provider=aws \
+        {{ if .KubeDns.NodeLocalResolver }}--cluster-dns=${COREOS_PRIVATE_IPV4} \
+        {{ else }}--cluster-dns={{.DNSServiceIP}} \
+        {{ end }}--cluster-domain=cluster.local \
+        --cloud-provider=aws
```
c-knowles: @redbaron some merge must have gone wrong here - the `\` has been removed so latest master updates fail right now.

Issue #782, opened 2017-07-19
Optimize usage of AWS AutoScaling API


danielfm:
I'm consistently being throttled on `DescribeAutoScalingInstances` and `DescribeAutoScalingGroups` requests to the AWS API, most likely due to having both the cluster autoscaler and node drainer features enabled.

For the cluster autoscaler, we can provide a way for users to customize the scan interval in order to compensate for larger clusters with a high number of auto scaling groups. I'm not familiar with the cluster autoscaler code to know if there are anything we can optimize there, so I'm not commenting on that.

Now, regarding the node drainer, we can probably make the current implementation much more efficient by running it in a single pod (instead of as a DaemonSet), that does something like this:

```bash
# This would be the node draining loop:

# 1. Take the externalID (which maps to the EC2 instance ID) from all nodes
# (We need to return some `externalID -> node name` mapping though, so we know which node to drain based on its EC2 instance ID)
$ kubectl get nodes -o jsonpath='{range .items[*]}"{.spec.externalID}" {end}'

# 2. Use a single call to retrieve the current status for all nodes
# (We can only specify 50 IDs at a time, so we need to be careful in order to handle large clusters properly)
$ aws autoscaling describe-auto-scaling-instances --instance-ids <ids>

# 3. For all nodes that are flagged as being terminated
$ kubectl drain --force --ignore-daemonset --delete-local-data <node name>

# 4. Sleep for a few seconds and repeat
```

Note this is not the whole program, just some snippets to exemplify the idea. We can maybe take this opportunity to rewrite this thing (again) in Go, with some unit tests and everything.

What do you guys think?

redbaron:
Out of interest, how big is your cluster?

My only concern is what happens when node which performs drain looping over multiple nodes tries to drain itself. I guess "drain controller" pod will be moved elsewhere and will discover state of ASGs and continue drain looping, right? If so then yes, it will be a good optimization to do.

Alternatively single pod can query ASG and update configmap, which is then updated in node-drainer-daemonset pods by Kube. Each node-drainer then checks configmap and figures out if it need to be drained and terminated. This way each nodes continues to be responsible for draining itself, but on the downside we now have two control loops: one updating configmap, another checking configmap and draining the node.

[...]

PR #794, opened 2017-07-24
Refactor node drainer implementation

danielfm:
This PR separates the node drainer into two components: the ASG watcher, and the node drainer itself.

The ASG watcher is the component (that runs as a Deployment) responsible for using the AWS AutoScaling API to check whether there are any instances waiting for termination; if anything is changed in the status of any ASG-controlled instances, the script updates the ASG status ConfigMap with the IDs of the instances being terminated. This ConfigMap is mounted in every node, and is used by the node drainer pods, which run as a DaemonSet.

The node drainer pods are responsible for checking whether the instance ID corresponding to the node they are running in are listed in the ASG status ConfigMap (it does that by mounting the ConfigMap as a volume in order to avoid hitting the API Server directly), and draining their own nodes if necessary.

This design allows us to significantly reduce both the API Server and the AutoScaling API usage -
 and reduce the risk of being throttled in the process - which is a good idea especially for large clusters.

Fixes: #782

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -666,7 +666,9 @@ write_files:
       {{ end }}
 
       {{ if .Experimental.NodeDrainer.Enabled }}
-      kubectl apply -f "${mfdir}/kube-node-drainer-ds.yaml"
+      for manifest in {kube-node-drainer-ds,kube-node-drainer-asg-status-updater-de}; do
+          kubectl apply -f "${mfdir}/$manifest.yaml"
+      done
       {{ end }}
 
       # Configmaps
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1211,16 +1285,10 @@ write_files:
...
-                        if [ "${state}" = Terminating:Wait ]; then
-                          termination_source=asg
-                          break
-                        fi
+                      # Termination annotation check
+                      termination_source=$(kubectl get nodes "${NODE_NAME}" -o json | jq -r '.metadata.annotations["kube-aws.coreos.com/node-termination-source"]')
+                      if [ "${termination_source}" == asg ]; then
+                        break
                       fi
                     done
```
redbaron: can we use `while sleep 10; do` or something like that so that it doesn't try to drain again immediately after `asg complete-lifecycle-action ` ?

It should still be a loop in case aws command failed or something else goes wrong and we need to repeat draining cycle.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1211,16 +1285,10 @@ write_files:
...
-                        secs=0
-
-                        # AutoScalingGroup lifecycle check
-                        state=$(asg describe-auto-scaling-instances --region "${REGION}" --instance-ids "${INSTANCE_ID}" | jq -r '.AutoScalingInstances[].LifecycleState')
-                        if [ "${state}" = Terminating:Wait ]; then
-                          termination_source=asg
-                          break
-                        fi
+                      # Termination annotation check
+                      termination_source=$(kubectl get nodes "${NODE_NAME}" -o json | jq -r '.metadata.annotations["kube-aws.coreos.com/node-termination-source"]')
```
redbaron: have no exact numbers to prove, but feels like this can be taxing for big enough clusters, while reading configmap is cheap as it is updated by kubelet.

core/controlplane/config/templates/cloud-config-controller
danielfm: If the `complete-lifecycle-action` command fails, the only side effect is that we'd have to wait for the drain timeout to complete before proceeding. Also, the loop will restart after 300s, where it will try to evict the node again (this is basically a no-op since all evictable pods were already moved to other nodes at this point), but yes, I agree that we could wrap the `complete-life-cycle` command in its own loop to avoid waiting too long before proceeding in case the command fails the first time.

core/controlplane/config/templates/cloud-config-controller
danielfm: I don't understand how a `kubectl get node <n>` would be more costly than a `kubectl get configmap <cm>`, but I'm not that familiar with the internals.

core/controlplane/config/templates/cloud-config-controller
redbaron: oh ,sorry, missed `sleep 300` at the bottom. then yes it fine. I was afraid it would go in a tight loop if anything goes wrong or unexpected,

core/controlplane/config/templates/cloud-config-controller
danielfm: Also, how using a configmap would make the "wait for signal" simpler?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1129,22 +1131,99 @@ write_files:
...
+                    [ -n "${REGION}" ]
+
+                    # Not customizable, for now
+                    POLL_INTERVAL=30
+
+                    # Instance termination detection loop
+                    while sleep ${POLL_INTERVAL}; do
+
+                      # Fetch the list of instances being terminated by their respective ASGs
+                      instances_to_drain=$(asg describe-auto-scaling-groups | jq -r '.AutoScalingGroups[] | select(.Tags[].Key | contains("kube-aws")) | .Instances[] | select(.LifecycleState == "Terminating:Wait") | .InstanceId')
```
danielfm: At least in my current cluster, I could not find any tags that I could use to further restrict the ASGs returned by the `describe-auto-scaling-groups` command.

We can use the `aws ec2 describe-instance` command to retrieve the `vpcId` corresponding to the current VPC, but I couldn't find any way of filtering ASGs by VPC in `describe-auto-scaling-groups`. 😢

danielfm: Thinking again, I could probably also filter by the `kubernetes.io/cluster/{{ .ClusterName }}` tag; this will only return ASGs associated with the current k8s cluster.

core/controlplane/config/templates/cloud-config-controller
redbaron: ConfigMap can be written to a file by kubelet and be updated anytime it changes in APIServer via https://kubernetes.io/docs/concepts/storage/volumes/#example-pod-with-a-secret-a-downward-api-and-a-configmap

wait for signal is then just stat and jq check, something like `[ "$(stat -c %y configmap)" != $PREV_MTIME ] && jq -e '(len(select(.path.to.node == $mynode))) > 0'`

danielfm: Nice, I'll give it a try.

danielfm:
Thanks @redbaron. Please do not merge this just yet; I'll try to test the ConfigMap thing. 😄

core/controlplane/config/templates/cloud-config-controller
danielfm: @redbaron I changed the scripts to use a ConfigMap for keeping the nodes that need to be drained. This ConfigMap is mounted in every node, and the draining script now periodically checks the volume mount for a file named after the node, instead of querying the API server.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1192,12 +1201,13 @@ write_files:
                       # Fetch the list of instances being terminated by their respective ASGs
                       instances_to_drain=$(asg describe-auto-scaling-groups | jq -r '.AutoScalingGroups[] | select((.Tags[].Key | contains("kube-aws")) and (.Tags[].Key | contains("kubernetes.io/cluster/{{.ClusterName}}"))) | .Instances[] | select(.LifecycleState == "Terminating:Wait") | .InstanceId')
```
redbaron: I don't understand why `node=` bit and patching is needed.  Deploy should just dump all pending ASG termination into configmap, this way it has one owner and daemonsets dont need to clean it up. It also makes it just a single write call to apiserver. 

To me it looks like it can be just:
`
instances_to_drain=$(... | jq  '[.AutoScalingGroups[] | ... | .InstanceId] | sort | join("\n") ' # build json escaped string

cat >yaml  <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
name: kube-node-drainer-status
namespace: kube-system
data:
  asg: ${instances_to_drain}
EOF

# TODO don't apply if $instances_to_drain didn't change since last iteration
kubectl apply -f yaml
```

Then on daemonset side:

```
INSTANCE_ID=$(....)
if [ -e "/etc/kube-node-drainer/from-asg" ] && grep -q $INSTANCE_ID /etc/kube-node-drainer/from-asg; then
   termination_source=asg; break
fi
```
and
```
             - name: kube-node-drainer-status
                projected:
                  sources:
                  - configMap:
                      name: kube-node-drainer-status
                      items:
                       - key: asg
                         path: from-asg
```

danielfm: Makes sense.

danielfm: Just pushed a commit to address your comments, let me know if there's anything else you'd like to change.

PR #801, opened 2017-07-26
CloudFormation events stream to stdout during kube-aws up/update

jollinshead:
This feature will allow CouldFormation stack events to be streamed to the console (stdout) during a `kube-aws up` or `kube-aws update` operation.
The format of the messages are:
```
TimePassed   StackName   ResourceStatus    LogicalResourceId   StatusReason
```
For example:
```
+00:02:45	Nodepoolb	CREATE_IN_PROGRESS      		WorkersLC             	"Resource creation Initiated"
+00:02:45	Nodepoolb	CREATE_COMPLETE         		WorkersLC
+00:02:48	Nodepoolb	CREATE_IN_PROGRESS      		Workers
```

### DIFF
--- a/cfnstack/provisioner.go
+++ b/cfnstack/provisioner.go
@@ -248,7 +248,7 @@ func (c *Provisioner) StreamEventsNested(q chan struct{}, f *cloudformation.Clou
 				&cloudformation.DescribeStackEventsInput{StackName: &stackId},
 				func(page *cloudformation.DescribeStackEventsOutput, lastPage bool) bool {
 					for _, e := range page.StackEvents {
-						if (*e.Timestamp).Before(t) {
+						if (e.Timestamp).Before(t) {
 							return false
 						}
 						if *e.EventId == lastSeenEventId {
--- a/cfnstack/provisioner.go
+++ b/cfnstack/provisioner.go
@@ -232,7 +232,7 @@ func (c *Destroyer) Destroy() error {
 	return err
 }
 
-func (c *Provisioner) StreamCloudFormationNested(q chan struct{}, f *cloudformation.CloudFormation, stackId string, headStackName string, t time.Time) error {
+func (c *Provisioner) StreamEventsNested(q chan struct{}, f *cloudformation.CloudFormation, stackId string, headStackName string, t time.Time) error {
 	nestedStacks := make(map[string]bool)
 	nestedQuit := make(chan struct{}, 1)
 	var lastSeenEventId string
@@ -263,7 +263,7 @@ func (c *Provisioner) StreamCloudFormationNested(q chan struct{}, f *cloudformat
 				e := events[i]
 				if *e.ResourceType == "AWS::CloudFormation::Stack" && *e.PhysicalResourceId != *e.StackId && !nestedStacks[*e.PhysicalResourceId] {
 					nestedStacks[*e.PhysicalResourceId] = true
-					go c.StreamCloudFormationNested(nestedQuit, f, *e.PhysicalResourceId, headStackName, t)
+					go c.StreamEventsNested(nestedQuit, f, *e.PhysicalResourceId, headStackName, t)
 				}
 				eventPrettyPrint(e, headStackName, t)
 				lastSeenEventId = *e.EventId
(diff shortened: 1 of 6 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1095,6 +1095,9 @@ worker:
 #  downloadUrl: https://github.com/DailyHotel/amazon-ssm-agent/releases/download/v2.0.805.1/ssm.linux-amd64.tar.gz
 #  sha1sum: a6fff8f9839d5905934e7e3df3123b54020a1f5e
 
+# When enabled, CloudFormation events will stream to stdout during kube-aws 'apply | up'.
```
mumoshu: nit: `update | up`

core/root/cluster.go
```diff
@@ -186,13 +186,17 @@ func (c clusterImpl) Create() error {
...
+
 	if c.controlPlane.CloudWatchLogging.Enabled && c.controlPlane.CloudWatchLogging.LocalStreaming.Enabled {
-		// Return Journald logs in a separate GoRoutine
-		q := make(chan struct{})
-		defer func() { q <- struct{}{} }()
 		go streamJournaldLogs(c, q)
 	}
 
+	if c.controlPlane.CloudFormationStreaming {
+		go streamCloudFormation(c, cfSvc, q)
```
mumoshu: Nit: I'd prefer `streamStackEvents` so that it would better be consistent with `stackProvisioner` immediately after this line, while better representing it streams not general info of a cfn stack but just events.

cfnstack/provisioner.go
```diff
@@ -231,3 +231,67 @@ func (c *Destroyer) Destroy() error {
 	_, err := cfSvc.DeleteStack(dreq)
 	return err
 }
+
+func (c *Provisioner) StreamCloudFormationNested(q chan struct{}, f *cloudformation.CloudFormation, stackId string, headStackName string, t time.Time) error {
```
mumoshu: Would you mind commenting this func so that we can understand the scope at a glance?
For example, `StreamCloudFormationNested` streams all the events from the root, the contorl-plane, and worker node pool stacks`

mumoshu: Nit: I'd like its name to be `StreamEventsNested` after https://github.com/kubernetes-incubator/kube-aws/pull/801/files#r129498094

jollinshead:
@mumoshu Done.

cfnstack/provisioner.go
```diff
@@ -231,3 +231,67 @@ func (c *Destroyer) Destroy() error {
...
+		case <-q:
+			return nil
+		case <-time.After(1 * time.Second):
+			events := make([]cloudformation.StackEvent, 0)
+
+			_ = f.DescribeStackEventsPages(
+				&cloudformation.DescribeStackEventsInput{StackName: &stackId},
+				func(page *cloudformation.DescribeStackEventsOutput, lastPage bool) bool {
+					for _, e := range page.StackEvents {
+						if (*e.Timestamp).Before(t) {
```
redbaron: I wonder if pointer dereference is actually needed here and throughout the function? Doesn't Go dereference them on access and implicit dereferencing is needed only on assignments?

core/root/cluster.go
```diff
@@ -408,3 +416,9 @@ func streamJournaldLogs(c clusterImpl, q chan struct{}) error {
 		}
 	}
 }
+
+func streamStackEvents(c clusterImpl, cfSvc *cloudformation.CloudFormation, q chan struct{}) error {
+	fmt.Printf("Streaming CloudFormation events for the cluster '%s'...\n", c.controlPlane.ClusterName)
+	// StreamEventsNested streams all the events from the root, the contorl-plane, and worker node pool stacks
```
redbaron: move this comment to a function declaration, it will be more visible there

jollinshead:
@redbaron Done.

c-knowles:
I've been doing some documentation work and wondered if we favour YAML or CLI flags for this sort of feature, any preferences?

jollinshead:
As it stands currently, using a CLI flag approach is feasible; however, since improvements are anticipated (https://github.com/kubernetes-incubator/kube-aws/issues/802) which may complicate the CLI flag approach, my preference is on yaml. 

I'm all for discussing other pros/cons of both methods though.

Issue #805, opened 2017-07-31
Accessing the worker iam role name from cloud-config-controller


mumoshu:
So that we can fully automate configuration of kube2iam.

Extracted from https://github.com/kubernetes-incubator/kube-aws/pull/713#issuecomment-311415479

cc @camilb

camilb:
@mumoshu This is not necessary anymore as by using `--auto-discover-default-role` in kube2iam and not providing the role annotation in "kube-resources-autosave"  deployment it will automatically fallback to the default role. Fixed in #809.

mumoshu:
@camilb Great 👍 Thanks for the improvement.

PR #809, opened 2017-08-01
Automatically configure kube2iam.

### LINKED ISSUE #806: Injecting the AWS account ID into a kube2iam k8s manifest
Probably via kube-aws' templating so that we can fully automate creation of [the kube2iam daemonset](https://github.com/kubernetes-incubator/kube-aws/pull/713/files#diff-5186d04bb20c1e42f69eab17092849d2).

Extracted from https://github.com/kubernetes-incubator/kube-aws/pull/713#issuecomment-317328034

cc @camilb @redbaron

camilb:
This PR completely automates `kube2iam and resolves #806.


mumoshu:
@camilb Thanks for your effort 👍 
FYI, the build is failing due to an index-out-of-range error in the newly added func as per https://travis-ci.org/kubernetes-incubator/kube-aws/builds/259985892#L1070

core/controlplane/config/config.go
```diff
@@ -1482,3 +1482,9 @@ func withHostedZoneIDPrefix(id string) string {
 	}
 	return id
 }
+
+func (c Config) AWSAccountID() string {
+	re := regexp.MustCompile("(?:[0-9]{12})")
+	id := re.FindStringSubmatch(c.KMSKeyARN)
+	return id[0]
```
mumoshu: If this is inducing an index-out-of-range error in a test, perhaps we have wrongly formatted kmsKeyArn in our test data?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -751,6 +751,14 @@ write_files:
       fi
       {{ end }}
 
+      {{if .Experimental.Kube2IamSupport.Enabled }}
+        mfdir=/srv/kubernetes/manifests
+       {{if .Experimental.Plugins.Rbac.Enabled}}
```
mumoshu: Thank you very much for your efforts to support RBAC 👍

core/controlplane/config/config.go
camilb: Found a quick fix for it

c-knowles:
kube2iam is available via Helm, we already install Tiller by default so it seems we have some divergence here of our plan. @mumoshu is the Tiller part ready for use yet?

Some things I would consider regardless of install method:
* `auto-discover-base-arn` and `auto-discover-default-role` args are useful
* We should add `RollingUpdate` strategy to any daemon sets we deploy
* I think this daemon set should be annotated with `scheduler.alpha.kubernetes.io/critical-pod: ''` and any other critical pod annotations
* Can we define reasonable resource limits for anything we're adding in kube-aws. I'm currently using this config for kube2iam:
```
resources:
  limits:
    cpu: 10m
    memory: 32Mi
  requests:
    cpu: 10m
    memory: 32Mi
```

mumoshu:
@c-knowles Thanks for reviewing 👍
I'd say tiller is free to be used as it is installed by default.
However I'd use it only when necessary.

I agreed to install tiller by default because it was considered to be necessary for some complex k8s apps (if I recall correctly).
Use of helm/tiller isn't a requirement, right?
I'm ok with either way of the deployment - a k8s manifest or a helm chart - as long as important points like you've mentioned in the prev comment are addressed in the deployment.
Does kube2iam chart addresses those points? If so, we'd better go with reusing the chart. However, it is still a recommendation rather than a requirement .
WDYT?

c-knowles:
My personal preference is try to use the same install methods where we can. We might choose kubeadm for the core components and Helm install for anything third party? kube2iam is a good case of a third party addition I think. If we used the Chart, we'd benefit from that community, extra features and testing that goes into the Helm and the Charts plus we'd reduce duplicate effort. In this case, everything I stated is dealt with in the Helm Chart but only because some of the arguments are generic passthroughs. This is the custom config I pass to Helm right now:

```yaml
# Custom values for kube2iam
extraArgs:
  backoff-max-elapsed-time: 5s
  backoff-max-interval: 2s
  auto-discover-base-arn: true
  auto-discover-default-role: true

host:
  iptables: true
  # needed for flannel network
  interface: cni0

rollingUpdate: true

podAnnotations:
  scheduler.alpha.kubernetes.io/critical-pod: ''

resources:
  limits:
    cpu: 10m
    memory: 32Mi
  requests:
    cpu: 10m
    memory: 32Mi
```

mumoshu:
@c-knowles Thanks 👍 

> My personal preference is try to use the same install methods where we can. We might choose kubeadm for the core components and Helm install for anything third party?

Yes, I agree with you completely.
My point is just that it isn't a requirement but a recommendation(we could say it a "strong" recommendation if it works better in communication).
Helm is great. Charts are great. I'd like to use them everywhere I can.
However, personally, I don't like to sacrifice anything by using charts.
The use of helm isn't our purpose but a process.
We use charts to better cooperate with the community "plus" achieving our goal.

Anyways, according to your information, the kube2iam chart seems very nice. So, there's no reason
 we won't use it? 😃 
Then, I'd be happy if @camilb updates this PR with helm, or someone else submits an another PR to switch from the k8s manifest to a helm chart for deploying kube2iam.

c-knowles:
I can't think of any reason to not use it if the Tiller stuff we have is ready for use, can you?

mumoshu:
@c-knowles So am I 👍 

Perhaps it is just that we're talking about different aspects of this.

If you're asking if this PR should be updated immediately to use the chart before merging, I'd say no.
I just don't want to force @camilb to update this PR to use the chart.

I suppose for many people kube-aws is a side-project and therefore a PR author isn't always available to afford time to jump at once towards our final goal (=use helm everywhere in this case).
We are the community and we can gradually improve things towards what we want.

For this concrete case, we "can" merge this PR as-is and submit an another PR later to use the chart instead. Even in such case, this PR acts as a base-line which we shouldn't break while the migration from the k8s manifest to the chart.

core/controlplane/config/config.go
```diff
@@ -1482,3 +1482,9 @@ func withHostedZoneIDPrefix(id string) string {
 	}
 	return id
 }
+
+func (c Config) AWSAccountID() string {
+	re := regexp.MustCompile("(?:[0-9]{12})")
```
redbaron: I guess templating is done before kube-aws contacts AWS, so it is too early to receive real AWS account name?

Reason this might not work in 0.00001% of cases is if KMS key is managed by security teams across accounts, but I think having kube2iam out of the box is more important :)

c-knowles:
@mumoshu I totally agree about merging just wondering where we draw the line and didn't mean to bring up so much discussion in this PR. Perhaps we can set a time to meet on slack for a few of us to discuss?

camilb:
@c-knowles As I understand, at this moment, using the chart adds an extra manual step, am I right? In order to avoid additional steps when a user enables `kube2IamSupport`  in cluster.yaml I think we should wait for #791.

redbaron:
Problem with Helm/Tiller is that it is very buggy and don't do basic things right. For instance failed deploy in Helm causes havok, making subsequent deploys do not do what you expect from them, unless failed deploy is rolled back immediately after failure happened.  

So we need to be very careful not to ignore helm errors, otherwise CF rollback will not happen and failed release will go into that limbo state. With user's ability to edit YAML freely it will cause lots of headache.

Another point to consider is the fact that if we have Helm releases installed once cluster is up, people will see it using their helm tools, which kinda encourage them to do updates/tweaks to them using helm, bypassing kube-aws, which in turn leads to errors on subsequent cluster rolls and/or ASG node restarts, where userdata-driven script are going to undo changes.  Given that RBAC controls is non-existent with stock Tiller (correct me if I am wrong, things might change), we can't prevent it.

c-knowles:
@camilb my understanding is @mumoshu was already doing some of the work to allow us to automate chart installation. Definitely I'm against manual steps for built in features.

@redbaron I believe RBAC in Tiller is coming along. I experienced a few issues but they have been resolved and it's working well right now so not sure what you thought was buggy but probably a longer conversation 😆  Given that I can Helm install apps that also sit alongside kube-aws functions now and can drastically change functionality, is there much difference in risk profile for operators? For example installing kube2iam and intercepting AWS metadata calls. If the lack of RBAC is a concern for an operator, I suspect they are not allowing wide access to Tiller already. If we are to get this going we should probably be outputting helm chart values files during `kube-aws render` and recommend any out of kube-aws upgrades are done using `--reuse-values` flag only.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2689,6 +2697,105 @@ write_files:
...
+            {{- if .Experimental.Plugins.Rbac.Enabled}}
+            serviceAccountName: kube2iam
+            {{- end}}
+            hostNetwork: true
+            containers:
+              - image: jtblin/kube2iam:latest
+                name: kube2iam
+                args:
+                  - "--app-port=8282"
+                  - "--base-role-arn=arn:aws:iam::{{.AWSAccountID}}:role/"
```
Fsero: i would recommend that these flag is removed and people annotate their pods with the full ARN, it works better if you need to assume roles in several AWS accounts and not only the same account of the AWS cluster

Fsero:
Hi all, we have a kube-aws clusters with several teams that uses kube2iam extensively (almost every application running in the cluster contact AWS resources and needs credentials injected through kube2iam). We install kube2iam like @c-knowles is proposing, (except its a custom version including this PR which is never merged https://github.com/jtblin/kube2iam/pull/59). 

It works well the only issue we had found is if clients using the cluster wants to use helm to deploy their own apps they need to use the same version of helm, or you will be upgrading / downgrading tiller in every deploy. 

To avoid that you have two choices:

- You use a global Tiller with privileges to deploy in any namespace, this could be an issue from a security point of view but suits better if the source of your apps is in a CI/CD system controlled by only one team.

- Every client deploys a tiller in their namespace that will have only permissions through rbac to modify their own namespace, it implies that with every helm invocation you need to instruct Helm to use the right namespace. The tradeoff is that you have several Tillers running, which is far from ideal from a resource efficiency point of view but it has a better security model, an user belongs to a group, a group has privileges to modify some namespaces and can deploy using a namespaced Tiller and the desired Helm version.

If kube-aws recommends the second model some of the issues pointed by @redbaron are mitigated, you will have a central tiller for kube-aws and others tillers for user applications, kube-aws tiller would have privileges to deploy on kube-system and default namespace whereas user's tillers will have the privileges defined in the default serviceaccount of the namespace (more info here https://github.com/kubernetes/helm/issues/2224), however i also agree with @redbaron that i would avoid Helm for core applications needed to bootstrap kube-aws, and for me kube2iam is likely a core application since it's hard to imagine that someone wants to deploy k8s in AWS without the capability of using AWS services through IAM credentials.

core/controlplane/config/templates/cloud-config-controller
camilb: Thanks for suggestion. Actually helped me realise that by using `--auto-discover-base-arn` and `--auto-discover-default-role` we can use both, role name and role arn. Apart from this we don't have to inject the `AWSAccountID` in template and `--auto-discover-default-role` also provide the necessary role for kube-resources-autosave as fallback fixing #805.

camilb:
@mumoshu @redbaron @c-knowles I made some small changes by replacing the `--base-role-arn` flag with `--auto-discover-base-arn` and `--auto-discover-default-role`.
This solves the fallowing :

 * injecting the AWSAccountID in kube2iam definition is not necessary anymore. Closes #806.
 * kube-resources-autosave works correctly by using the `IAMRoleWorker` as fallback. Closes #805.
 * use both role names and role arn to assume roles in various accounts. 

On the other hand, I think this method is simple enough and covers various use cases and using helm to install it adds more complexity.

mumoshu:
Thank you everyone for your supports 🍺 

> @camilb my understanding is @mumoshu was already doing some of the work to allow us to automate chart installation. Definitely I'm against manual steps for built in features.

Ah, sorry for that! It would be addressed partly via #791. Basically, I have implemented helm releases injectable through cloud-config's `write_files` and a loop over "release files" which installs releases one by one. After that, it would be a matter of providing additional helm release(s) to be installed.

The things @redbaron pointed out should be considered though. I'm currently inclined to the second option @Fsero suggested. Regarding resource usage, I suppose it is just dozens of megabytes per a tiller deployment AFAIK, which doesn't look very problematic to me.

mumoshu:
@camilb Wow, thanks for the improvements! Really impressed to how kube2iam nicely supports edge-cases(?) like those and the outcome of the collaboration here 😊 

LGTM.

mumoshu:
Let me also note again.

> My personal preference is try to use the same install methods where we can. 

This is where we should eventually go, towards simplicity and lesser divergence of the community. This should eventually be possible without sacrificing anything.
Thank you very much for the suggestion @c-knowle 👍

c-knowles:
After updating to latest master, I noticed a potential problem for existing users of the experimental flag. It's experimental after all so just leaving a note here for any other users of the existing flag - you may have 2 x kube2iam daemon sets or your existing kube2iam daemon set config may be wiped by kube-aws update. If you have 2, it's likely you'll need to remove your existing one.

* We already have `experimental.kube2IamSupport.enabled` true in configuration as we need the assume role permission from that (there are other ways to gain this, just kube-aws provided this way so we used it)
* kube-aws now uses the same flag to deploy kube2iam, in our case the deployment name is not identical so we'd end up with 2 x kube2iam daemon sets

PR #818, opened 2017-08-04
Put kube2iam update strategy in the correct place

c-knowles:
Introduced in https://github.com/kubernetes-incubator/kube-aws/pull/809.


camilb:
Hi @c-knowles thanks for spotting this. It's weird since I had it this way locally but committed before saving. I think it's safe to be merged.

c-knowles:
@camilb controllers and cluster will still function if you performed a `kube-aws update` with this change (mine did). However, you'd see the bootstrap install failing in the controller logs. I'm not sure if `kube-aws up` on a new cluster would fail outright or not.

PR #819, opened 2017-08-04
Allow OIDC config without Dex

c-knowles:
If using a single OIDC provider, Dex is not needed.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -89,6 +89,14 @@ func NewDefaultCluster() *Cluster {
 				Enabled: false,
 			},
 		},
+		OIDC: model.OIDC{
+			Enabled:       false,
+			IssuerURL:     "https://accounts.google.com",
+			ClientID:      "YOUR_CLIENT_ID.apps.googleusercontent.com",
+			UsernameClaim: "email",
+			GroupsClaim:   "",
+			SelfSignedCa:  false,
+		},
 		Dex: model.Dex{
 			Enabled:         false,
 			Url:             "https://dex.example.com",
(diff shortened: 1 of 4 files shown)


c-knowles:
Work in progress. Trying to figure out the best way to configure this. It's just OIDC without the Dex part. One idea is to put the `dex` config under the new `oidc` section and make dex specific config optional there.

camilb:
@c-knowles We discussed this a while ago and I'm still thinking that dex can be deployed separately . Keeping the OIDC is also enough for dex to work after the cluster is up.

camilb:
I'm also thinking renaming the dex certificates as oidc certificates.

camilb:
I think we can modify the existing dex code since already contains the OIDC part. What do you think?

c-knowles:
@camilb maybe that's ok, I'm not sure of the cert requirements as never touched that bit. Happy to combine efforts, my only requirement is ability to turn on the API server OIDC flags without deploying anything Dex related since it turned out it wasn't needed if a user just wants to enable login via single OIDC provider.

camilb:
@c-knowles the certs are only necessary for other providers, like dex and only if they use a custom CA. If you use google or github it works fine.

camilb:
I'll prepare a PR these days, that will allow using only OIDC or Dex. I'm planning to replace the Dex config from the cluster.yaml with a much simpler OIDC config and move Dex to "contrib" directory with instructions on how to deploy it.

c-knowles:
Cool. So probably this PR can be replaced by yours. I tried to align our parameter names to the API server flags in this one but it doesn't matter a lot. Also, I'd be happy to start a contrib section in our new documentation with Dex as first example if we think that's appropriate (with file download links etc).

camilb:
@c-knowles At this stage, #820 works fine using only one OIDC provider or Dex. Still thinking if makes sense to keep the certificates for custom providers.

mumoshu:
Closing this in favor of #820. Thanks for your collaborative work @c-knowles and @camilb 🎉

PR #820, opened 2017-08-04
Simplify configuration for OIDC Authenticator.

camilb:


contrib/dex/README.md
```diff
@@ -1,5 +1,20 @@
...
+ 
+ 2. Create a secret containing your [GitHub OAuth2 client credentials](https://github.com/settings/applications/new)
+ 
+     kubectl create secret \
+         generic github-client \
+         --from-literal=client-id=$GITHUB_CLIENT_ID \
+         --from-literal=client-secret=$GITHUB_CLIENT_SECRET
+         
+ 3. Deploy Dex: `kubectl apply -f contrib/dex/dex.de.yaml`
+
```
mumoshu: Probably you need to update cluster.yaml accordingly if you wanted the k8s cluster to use the dex just deployed as the auth provider? Then, an example snippet of cluster.yaml here would be helpful.

camilb: You're right, already updated.

test/integration/maincluster_test.go
```diff
@@ -1167,32 +1164,13 @@ experimental:
...
       - sg-12345678
-  dex:
+  oidc:
     enabled: true
-    url: "https://dex.example.com"
-    clientId: "example-app"
-    username: "email"
+    url: "https://accounts.google.com"
+    clientId: "kubernetes"
+    username: "sub"
```
c-knowles: If we use google as an example then I think username claim is `email` so I think fine to keep the way it was. At least, we have it confirmed with `email` and it works.

camilb: Sure, no problem, didn't use Google yet. Got the config from the kubernetes's documentation. This will work for dex also.

mumoshu:
> Other than that my preference is to align to the API server param names in the hope it's easier to understand where those go

Great suggestion 👍
I agree that we should make setting keys consistent with corresponding names of apiserver params.
@camilb WDYT?

@c-knowles Would you mind submitting an another PR if we agreed on it, after this PR gets merged?

camilb:
@c-knowles I made the necessary adjustments to align with the API server param names. Also removed the self signed certificate generation for OIDC. If anyone use a custom provider that has a self signed certificate, most probably they are not using the ones generated by kube-aws. As for Dex, the users have several [options](https://github.com/camilb/kube-aws/blob/3ac34d943c1253f4b650e835eb02bd061daa5346/contrib/dex/README.md#exposing-dex-service) to  expose the service using a trusted certificate.

camilb:
> I guess the original intention behind the current key names was to be consistent with the dex config keys?

@mumoshu Yes, they were initially aligned with Dex parameter names. Agree using the names from API parameters and already made the changes. I also removed the generation of self signed certificates for OIDC as Dex can be easily  exposed  using a trusted certificate.

mumoshu:
@camilb Thanks for your efforts 👍 I understood about the changes regarding self-signed cert generation, too.

PR #824, opened 2017-08-07
Fix managed role name validation

adyromantika:
When creating iam roles for the controller or node pools, validation checks for the whole string consisting of `<cluster_name>-<nested_stack_name>-<12_characters_random_alphanumeric_characters>-<region>-<iam_role_name>`. The validation works as intended if `iam.role.name` is not specified and auto-generated.

However when `iam.role.name` is specified, kube-aws is actually creating `<region_name>-<managed_role_name>` so for the following config it'll actually create a role named `<region>-myrole` so a new validation is needed.

```
      iam:
        role:
          name: "myrole"
```

This allows for a little bit more leeway to have a more meaningful name for the fixed role.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -1023,8 +1023,14 @@ func (c Cluster) validate() error {
 		fmt.Println(`WARNING: instance types "t2.nano" and "t2.micro" are not recommended. See https://github.com/kubernetes-incubator/kube-aws/issues/258 for more information`)
 	}
 
-	if e := cfnresource.ValidateRoleNameLength(c.ClusterName, c.NestedStackName(), c.Controller.IAMConfig.Role.Name, c.Region.String()); e != nil {
-		return e
+	if len(c.Controller.IAMConfig.Role.Name) > 0 {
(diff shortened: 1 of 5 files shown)


core/controlplane/config/config.go
```diff
@@ -1023,8 +1023,14 @@ func (c Cluster) validate() error {
 		fmt.Println(`WARNING: instance types "t2.nano" and "t2.micro" are not recommended. See https://github.com/kubernetes-incubator/kube-aws/issues/258 for more information`)
 	}
 
-	if e := cfnresource.ValidateRoleNameLength(c.ClusterName, c.NestedStackName(), c.Controller.IAMConfig.Role.Name, c.Region.String()); e != nil {
-		return e
+	if len(c.Controller.IAMConfig.Role.Name) > 0 {
+		if e := cfnresource.ValidateManagedRoleNameLength(c.Controller.IAMConfig.Role.Name, c.Region.String()); e != nil {
```
mumoshu: I guess we'd better name it `ValidateStableRoleNameLength` whereas the former `ValidateRoleNameLength` renamed to `ValidateUnstableRoleNameLength` so that we can better make distinction between two?

adyromantika:
Thank you @mumoshu I have made changes as suggested.

mumoshu:
@adyromantika LGTM. Thanks for the contribution 👍

PR #826, opened 2017-08-07
Updating instructions for MFA token

sathiyas:
When using MFA for the AWS Credentials, additional Env variable needs to be exported


mumoshu:
@sathiyas LGTM. Thanks for the contribution 👍

mumoshu:
@sathiyas BTW, how do you obtain a set of an access key id and secret key, session token?
I tend to use `aws sts assume-role` command for that - if you do it too, how about adding to the doc about a full series of commands required to obtain credentials when using MFA?

sathiyas:
Thanks for the acceptance - I will add that documentation

PR #827, opened 2017-08-08
Fixed typo in the error message

sathiyas:
Fixed the typo in the Error message in the code


mumoshu:
@sathiyas Thanks for the good catch 👍

PR #829, opened 2017-08-08
Add AWS_PROFILE to FAQ

### LINKED ISSUE #716: Possibility to use different AWS profile than default
Is there any solution how to pass AWS profile to kube-aws? We have 2 separate clusters in two separate AWS accounts and it's not so easy to do something with these to cluster because every time i need to check which account i currently have set as default.

Vrtak-CZ:
closes #716


mumoshu:
@Vrtak-CZ  LGTM. Thanks a lot for taking your time on this 👍

Issue #799, opened 2017-07-26
Update Kubernetes to 1.7.3


mumoshu:

Vrtak-CZ:
1.7.3 released - https://github.com/kubernetes/kubernetes/releases/tag/v1.7.3

c-knowles:
@Vrtak-CZ hopefully CoreOS publishes it soon - https://quay.io/repository/coreos/hyperkube?tag=latest&tab=tags.

mumoshu:
Resolved via #817.
Thanks for the contribution 👍

Vrtak-CZ:
CoreOS released 1.7.3 hyperkube https://quay.io/repository/coreos/hyperkube?tab=tags we should open new issue and close this one.

c-knowles:
Let's get it updated.

redbaron:
Fixed by #830

PR #830, opened 2017-08-08
Bump default k8s to 1.7.3

c-knowles:
For #799


mumoshu:
@c-knowles Thanks for your support ☺️

PR #832, opened 2017-08-08
Update Calico to v2.4.1

tmjd:
network plugin: Update Calico to v2.4.1

One thing that needs to be mentioned is that this update includes a change in the default deny behavior for network policy, Calico has switched this behavior to match the move of Kubernetes NetworkPolicy to v1.  You can see the release notes for Calico at https://github.com/projectcalico/calico/releases/tag/v2.4.0 (see the changes under k8s-policy).
I am looking for guidance on how the kube-aws project would like to handle this behavior change or at least know that it is expected that users should understand the changes of the components they are using.

Just to point out how this new behavior works:
- DefaultDeny is enabled for a pod if there is any policy that selects the pod
  - That means the new way to enable DefaultDeny is create a policy that selects all pods (in a namespace)
  - This also means that if DefaultDeny is not desired then all policy needs to be removed that targets a pod.
- The old default deny annotation has no effect with the new Calico version


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -869,6 +869,8 @@ write_files:
                         key: etcd_endpoints
                   - name: CALICO_NETWORKING_BACKEND
                     value: "none"
+                  - name: CLUSTER_TYPE
+                    value: "kubeaws,canal"
```
mumoshu: Just for my education, are there any conventions for deciding what the value should be?
For instance, [kops seems to have `kops,canal` for the value](https://github.com/kubernetes/kops/pull/3162/files), which looks consistent with this one for kube-aws.

mumoshu:
@redbaron @camilb @danielfm @c-knowles Are you using calico on your clusters? Anyways, FYI, there are some manual steps required to maintain your network policy behave the same as before while upgrading k8s to v1.7.
The new calico v2.4.1 introduced by this PR seems to correctly implement the semantics of networking/v1 api and therefore the change in k8s v1.7 becomes effective after merging this PR.

mumoshu:
@tmjd Thank you very much for the contribution with the detailed description 👍 

> it is expected that users should understand the changes of the components they are using.

Yes, they should. I'm going to add a note about the changes in the release note for the next kube-aws release 👍

c-knowles:
@mumoshu, thanks. Not using it here.

danielfm:
+1, I enabled it, but I'm not using network policies so far.

camilb:
Same for me, enabled but with no policy configured yet.

PR #833, opened 2017-08-09
Fix potential YAML parsing error

sathiyas:
Fixed the name subnet name parameter in nodePools section so that YAML can parse. When you uncomment the name, a - is needed for successfully parsing YAML

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -239,7 +239,7 @@ worker:
 #      # If omitted, public subnets are created by kube-aws and used for worker nodes
 #      subnets:
 #      - # References subnets defined under the top-level `subnets` key by their names
-#        name: ManagedPublicSubnet1
+#        - name: ManagedPublicSubnet1
 #
 #      # Existing "glue" security groups attached to worker nodes which are typically used to allow
 #      # access from worker nodes to services running on an existing infrastructure


core/controlplane/config/templates/cluster.yaml
```diff
@@ -239,7 +239,7 @@ worker:
 #      # If omitted, public subnets are created by kube-aws and used for worker nodes
 #      subnets:
 #      - # References subnets defined under the top-level `subnets` key by their names
```
redbaron: this `-` is the one you want for successful parsing

sathiyas: yes - when uncommenting the line, we need the " - ". Pls check another location like line 65 for comparison

mumoshu: @sathiyas My assumption was given this block of yaml:

```yaml
  nodePools:
    - # Name of this node pool. Must be unique among all the node pools in this cluster
      name: nodepool1
#      subnets:
#      - # References subnets defined under the top-level `subnets` key by their names
#        name: ManagedPublicSubnet1
```

One would uncomment it like:

```yaml
  nodePools:
    - # Name of this node pool. Must be unique among all the node pools in this cluster
      name: nodepool1
      # If omitted, public subnets are created by kube-aws and used for worker nodes
      subnets:
      - # References subnets defined under the top-level `subnets` key by their names
        name: ManagedPublicSubnet1
```

which is parsed into:

```json
{
  "nodePools": [
    {
      "subnets": [
        {
          "name": "ManagedPublicSubnet1"
        }
      ], 
      "name": "nodepool1"
    }
  ]
}
```

mumoshu: @sathiyas So perhaps your assumption is that one could forget to uncomment:

```yaml
#      - # References subnets defined under the top-level `subnets` key by their names
```

in

```
#      subnets:
#      - # References subnets defined under the top-level `subnets` key by their names
#        name: ManagedPublicSubnet1
```

which results in an invalid yaml like:

```yaml
      subnets:
#      - # References subnets defined under the top-level `subnets` key by their names
        name: ManagedPublicSubnet1
```

Am I following you correctly?

mumoshu: @sathiyas If that's the case, I'd suggest moving the yaml comment outside of the array item:

```yaml
#      subnets:
#      # References subnets defined under the top-level `subnets` key by their names
#      - name: ManagedPublicSubnet1
```

so that one never fails like that.
What do you think?

mumoshu: @sathiyas Thanks for the confirmation 👍
Did you have a time to work on it?

mumoshu:
Closing this in favor of #888.
Thanks anyway @sathiyas!

PR #835, opened 2017-08-10
cfn-signal depends on install-kube-system

dvdthms:
These changes add the functionality so that cfn-signal fires if and only if the install-kube-system has finished successfully.

We add StartLimitInterval of 5m20s as a timeout value, so that the service does not cycle indefinitely on "ExecStartPre=/usr/bin/systemctl is-active install-kube-system.service"

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -316,10 +316,12 @@ coreos:
         StartLimitInterval=0
         RestartSec=10
         Restart=on-failure
+        RemainAfterExit=true
         ExecStartPre=/usr/bin/systemctl is-active kubelet.service
         ExecStartPre=/usr/bin/systemctl is-active docker.service
         ExecStartPre=/usr/bin/curl -s -f http://127.0.0.1:8080/version
-        ExecStart=/opt/bin/install-kube-system
+        ExecStartPre=/opt/bin/install-kube-system
+        ExecStart=/bin/true
 
 {{ if $.ElasticFileSystemID }}
     - name: rpc-statd.service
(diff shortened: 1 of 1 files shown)


c-knowles:
Nice! Had been wondering why faults in kube system install were introduced a few times recently.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -345,11 +347,16 @@ coreos:
...
-        Wants=kubelet.service docker.service
-        After=kubelet.service
+        Wants=kubelet.service docker.service install-kube-system.service
+        After=kubelet.service install-kube-system.service
 
         [Service]
-        Type=oneshot
+        Type=simple
+        Restart=on-failure
+        RestartSec=60
```
mumoshu: I'm wondering why we don't need `RemainAfterExit=true` for this even though we have one for `install-kube-system.service`.
Perhaps because there are no units checking activeness of cfn-signal by running `systemctl is-active cfn-signal`, right?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -345,11 +347,16 @@ coreos:
...
       content: |
         [Unit]
-        Wants=kubelet.service docker.service
-        After=kubelet.service
+        Wants=kubelet.service docker.service install-kube-system.service
+        After=kubelet.service install-kube-system.service
 
         [Service]
-        Type=oneshot
+        Type=simple
```
mumoshu: Yes, `type=simple` is better!
I'm now wondering where we actually need `Type=oneshot`. Nowhere?
Can we just transform every `Type=oneshot` service into `Type=simple` service with all the meaningful pieces in `ExecStartPre`s and `ExecStart` being `/bin/true`?

mumoshu:
@dvdthms Thanks for the contribution 👍 
LGTM overall. I've left two minor comments though.

core/controlplane/config/templates/cloud-config-controller
dvdthms: Indeed, cfn-signal is at the end of the services chain so has no units dependent on it.

core/controlplane/config/templates/cloud-config-controller
dvdthms: I can't speak for the rest of the processes, but here simple was preferred because oneshot would exit if the check for install-kube-system failed at all. Changing to simple allowed the adding of the functionality of the status checking. Might be worth checking into changing others if it makes the whole sequence more robust, but I think it would be better suited for another PR.

dvdthms:
Any more thoughts? @mumoshu

mumoshu:
@dvdthms Sorry for the long silence due to my vacation - no, this LGTM 👍 Thanks for the contribution.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -345,11 +347,16 @@ coreos:
...
+        Restart=on-failure
+        RestartSec=60
+        StartLimitInterval=320
```
dvdthms: Might need to bump RestartSec up too, potentially. There's nothing dependant on these values being as they are, so they are completely adjustable to user needs.

mumoshu: @dvdthms Thanks for the suggestion 👍 I'll take it into account.

Anyways, for this concrete case, I'm considering to change cfn-signal and install-kube-system to `Type=oneshot` with retrying via shell scripting so that we can control execution order without spamming journal like now.

PR #845, opened 2017-08-16
update default version of ETCd to 3.2.5

Vrtak-CZ:


redbaron:
@Vrtak-CZ , there was an important change "transport: use reverse lookup to match wildcard DNS SAN", did you test it on your clusters that certs generated by kube-aws are still valid and accepted by Etcd?

Vrtak-CZ:
@redbaron yes i'm test it with `kube-aws version v0.9.7` with ETCd `3.2.5` and it works ok.

PR #850, opened 2017-08-22
Feature/add new docs

c-knowles:
Plan is to host them on github pages but still generate using gitbook CLI.

@mumoshu I tried this out locally and working fine. Now we just need to connect to `gh-pages` branch we per our discussion.


docs/book.json
```diff
@@ -0,0 +1,19 @@
...
+    }
+  },
+  "links": {
+    "sidebar": {
+    }
+  },
+  "styles": {
+    "website": "styles/site.css"
+  }
+}
```
mumoshu: AFAIK, the gitbook convention is to place book.json in the project root like I did in https://github.com/mumoshu/kube-aws/blob/master/book.json

c-knowles: Can do, it also means moving things like the .gitignore for node_modules into the root as well and the `_book` will be generated in the root. In this PR it's all self contained in the `docs` folder. So if you want to combine them it's possible, are you ok that it will impact these other parts?

mumoshu: Good points - Thanks for the confirmation! Yes, I'm ok with those.

mumoshu:
@c-knowles Thanks for the contribution 👍 I'll try to publish this to gh-pages manually, and the try to automate the process.

docs/book.json
c-knowles: Did you still want book.json in the root?

mumoshu: Yes but honestly it's a low priority thing because the reason I want it is just for gitbook compatibility.
We can defer it until gitbook fixes the issue with repos under github orgs.

I'd appreciate it if you could address it though 😉

PR #851, opened 2017-08-22
Add support for Node Authorizer + Node Restriction admission controller

danielfm:
After https://github.com/kubernetes/kubernetes/pull/48707, we can now add support for the node authorizer + node restriction admission controller to kube-aws without breaking our node drainer implementation.

In order to use this, Kubelets must use certificates that conform with the username/group restrictions documented [here](https://kubernetes.io/docs/admin/authorization/node/), which makes TLS bootstrapping a prerequisite.

Also, as noted in `cluster.yaml`, this feature can only be used in Kubernetes 1.7.4+. (See #853)

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1450,6 +1450,7 @@ write_files:
             - daemonsets
             verbs:
             - get
+          # Can be removed if node authorizer is enabled
           - apiGroups: [""]
             resources:
             - nodes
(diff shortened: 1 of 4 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1762,15 +1763,16 @@ write_files:
           - --audit-log-maxbackup=1
           {{ end }}
           {{if .Experimental.Plugins.Rbac.Enabled}}
-          - --authorization-mode=RBAC
-          - --authorization-rbac-super-user=kube-admin
```
mumoshu: At first glance I couldn't see why this was removed.
I now suppose it was due to that since k8s 1.6.0 (1) `--authorization-rbac-super-user` has been deprecated and (2) the `system:masters` org provided to our admin certs is enough to make `kube-admin`(the common name of the cert) able to create roles and role bindings. If that's correct, this LGTM 👍

danielfm: Yes, that's correct.

mumoshu:
/lgtm

Thanks for the contribution @danielfm!

PR #859, opened 2017-08-23
Automate publishing of the dedicated kube-aws documentation site

mumoshu:
* Print node.js version to debug when the script failed due to an incompatible node.js version
  * `nvm` for version-locking node.js isn't available in Travis CI's golang image

* Read the repo slug(`user/reponame`) from env vars so that we do not need to hard-code it

* Improve `make publish-docs` to support various cases
  * travis builds and local runs against a forked repo


mumoshu:
This is already verified to work on my fork of kube-aws.
Let's see it works here, too.

PR #861, opened 2017-08-23
"Edit This Page" links on the doc is linking to wrong places

mumoshu:
They should be linked to https://github.com/kubernetes-incubator/kube-aws/tree/master/docs/* rather tha https://github.com/kube-aws/docs/tree/master/*

### DIFF
--- a/book.json
+++ b/book.json
@@ -6,7 +6,7 @@
   "plugins": ["edit-link", "prism", "-highlight", "github", "anchorjs"],
   "pluginsConfig": {
     "edit-link": {
-      "base": "https://github.com/kube-aws/docs/tree/master",
+      "base": "https://github.com/kubernetes-incubator/kube-aws/tree/master/docs",
       "label": "Edit This Page"
     },
     "github": {


mumoshu:
@c-knowles I suggest that the link url would better be under `edit/` rather than `tree/`(alias of `blob`) so that a link url becomes:

https://github.com/kubernetes-incubator/kube-aws/edit/master/docs/README.md

which will allow the user to immediately start editing the page for contribution.
I suppose the user may expect this behavior from the text "Edit This Page".

c-knowles:
/lgtm

@mumoshu one thing to consider is the only people who get an edit page is those with write access. I'd like it if we enabled branch protection on `master` to avoid direct editing without PR. Anyone without access will be greeted by the `You need to fork this repository to propose changes` message and a button to fork so that seems fine. But then not sure if it directs them back to the document in question later.

mumoshu:
Thanks!

I've enabled branch protection with push restriction on master.

PR #865, opened 2017-08-23
Fix an invalid link in the doc

mumoshu:
The documentation site is now hosted at https://kubernetes-incubator.github.io/kube-aws/ rather than gitbook.


cknowles (approved):
/lgtm

PR #867, opened 2017-08-23
Do not overwrite the tokens.csv file when re-rendering credentials

### LINKED ISSUE #810: tokens.csv overwritten by render credentials in 0.9.7
Hello!

As of v0.9.7 of kube-aws, `tokens.csv` is being overwritten with blank contents by `kube-aws render credentials`. 

Using a test token file on v0.9.7:

```sh
$ cat credentials/tokens.csv
fake_token,username,username

$ file credentials/tokens.csv
tokens.csv: ASCII text

$ kube-aws version
kube-aws version v0.9.7

$ kube-aws render credentials
Generating credentials...
-> Parsing existing TLS CA
-> Generating new assets

$ file credentials/tokens.csv
tokens.csv: empty
```

with v0.9.6:

```sh
$ cat credentials/tokens.csv
fake_token,username,username

$ file credentials/tokens.csv
tokens.csv: ASCII text

$ kube-aws-0.9.6 version
kube-aws-0.9.6 version

$ kube-aws-0.9.6 render credentials
Generating credentials...
-> Parsing existing TLS CA
-> Generating new assets

$ file credentials/tokens.csv
credentials/tokens.csv: ASCII text
```

Let me know if I can provide any more details that would help in investigating this. Thanks!

danielfm:
Closes #810.

### DIFF
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -7,6 +7,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net"
+	"os"
 	"time"
 
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


mumoshu:
/lgtm

Thanks for the improvement 👍

redbaron:
Why tokens.csv is special? Re-rendering ca.pem is even more dangerous, as it causes etcd rotation when replaced nodes are not trusted by not yet replaced ones causing all sorts of trouble.

danielfm:
The `tokens.csv` file is not special in that sense. The issue was that `kube-aws render credentials` was truncating this file, so any information put there by the user was inadvertently lost, causing tokens previously specified there to be "revoked" in case the user updates the cluster afterwards.

danielfm:
Also, `kube-aws render credentials` do not replace the CA unless you specify `--generate-ca`, so running it and then running `kube-aws update` will not (supposedly) mess a running cluster.

PR #882, opened 2017-08-28
Do not encrypt certificate files, dont' load fingerprints if no raw file available

redbaron:
1. Certificates are not secrets, they can be kept unencrypted
 2. If only .enc file is provided and no unencrypted version is found
    then don't try to load fingerprint files. This change allows .enc files
    to be commited to Git and then used as-is

### DIFF
--- a/core/controlplane/config/encrypted_assets_test.go
+++ b/core/controlplane/config/encrypted_assets_test.go
@@ -7,7 +7,6 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/pem"
-	"fmt"
 	"github.com/kubernetes-incubator/kube-aws/model"
 	"github.com/kubernetes-incubator/kube-aws/test/helper"
 	"os"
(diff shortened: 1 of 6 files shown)


core/controlplane/config/encrypted_assets_test.go
```diff
@@ -168,56 +169,34 @@ func TestReadOrCreateCompactAssets(t *testing.T) {
...
+				{"AdminCert", original.AdminCert, regenerated.AdminCert},
+				{"CACert", original.CACert, regenerated.CACert},
+				{"WorkerCert", original.WorkerCert, regenerated.WorkerCert},
+				{"APIServerCert", original.APIServerCert, regenerated.APIServerCert},
+				{"EtcdClientCert", original.EtcdClientCert, regenerated.EtcdClientCert},
+				{"EtcdCert", original.EtcdCert, regenerated.EtcdCert},
+			} {
+				if v[1] != v[2] {
+					t.Errorf("%s must NOT change but it did : original = %v, regenrated = %v ", v[0], v[1], v[2])
+				}
```
mumoshu: Thanks for the improvement. It is much more readable now 😆

PR #884, opened 2017-08-28
Do not allow enabling Node authorizer without RBAC

### LINKED ISSUE #874: kubelet fails to start when RBAC isn't enabled but Node authorizer
I'm experiencing this on my brand-new cluster created with v0.9.8-rc.2 with a cluster.yaml which includes the following snippet. Please note that I had not enabled RBAC in this case.

```yaml
experimental:
  nodeAuthorizer:
    enabled: true
  tlsBootstrap:
    enabled: true
```

```
Aug 25 01:56:19 ip-10-0-0-12.ap-northeast-1.compute.internal kubelet-wrapper[28241]: Error: failed to run Kubelet: cannot create certificate signing request: User "kubelet-bootstrap" cannot create certificatesigningrequests.certificates.k8s.io at the cluster scope. (post certificatesigningrequests.certificates.k8s.io)
Aug 25 01:56:19 ip-10-0-0-12.ap-northeast-1.compute.internal kubelet-wrapper[28241]: Error: failed to run Kubelet: cannot create certificate signing request: User "kubelet-bootstrap" cannot create certificatesigningrequests.certificates.k8s.io at the cluster scope. (post certificatesigningrequests.certificates.k8s.io)
Aug 25 01:56:19 ip-10-0-0-12.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Main process exited, code=exited, status=1/FAILURE
Aug 25 01:56:19 ip-10-0-0-12.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Unit entered failed state.
Aug 25 01:56:19 ip-10-0-0-12.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Failed with result 'exit-code'.
```

Any roles and role bindings are missing because they are not created due to the fact that `experimental.plugins.rbac.enabled` is kept `false` hence RBAC is disabled at the kube-aws layer. kube-aws tries to add `RBAC` authorizer to apiserver and create roles and bindings only when RBAC is enabled.

```
kubectl get -a clusterrole,clusterrolebinding
No resources found.
```

The message:

> User "kubelet-bootstrap" cannot create certificatesigningrequests.certificates.k8s.io at the cluster scope

does seem like RBAC had rejected the requests due to missing role/binding.

If RBAC had been disabled at the kube-aws layer but somehow enabled on K8S, I can understand that this is an expected behavior.

However, what I have for apiserver is `--authorization-mode=Node`, not `--authorization-mode=Node,RBAC` as it is how kube-aws configures it when rbac is disabled but node authorizer is enabled.
I had expected RBAC to be disabled hence no such error could happen because I didn't have `RBAC` in the authorization mode list.

Relevant part of kube-aws: https://github.com/kubernetes-incubator/kube-aws/blob/d080ea00ff9f5a1a7cc6d7471ecdc0ba636b7e33/core/controlplane/config/templates/cloud-config-controller#L1765-L1769

So, can we add an additional validation to fail-fast when a configuration of this kind is specified via cluster.yaml, or anything else to prevent this issue?

cc @danielfm 

Update:

I've verified that an another cluster with the following cluster.yaml snippet did work:

```yaml
experimental:
  nodeAuthorizer:
    enabled: true
  tlsBootstrap:
    enabled: true
  plugins:
    rbac:
      enabled: true
```

danielfm:
Fixes #874.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -1090,8 +1090,14 @@ func (c Cluster) validate() error {
 		}
 	}
 
-	if c.Experimental.NodeAuthorizer.Enabled && !c.Experimental.TLSBootstrap.Enabled {
-		return fmt.Errorf("TLS bootstrap is required in order to enable the node authorizer")
+	if c.Experimental.NodeAuthorizer.Enabled {
+		if !c.Experimental.TLSBootstrap.Enabled {
+			return fmt.Errorf("TLS bootstrap is required in order to enable the node authorizer")
+		}
+
+		if !c.Experimental.Plugins.Rbac.Enabled {
+			return fmt.Errorf("RBAC is required in order to enable the node authorizer")
+		}
 	}
 
 	return nil
--- a/core/controlplane/config/config_test.go
+++ b/core/controlplane/config/config_test.go
@@ -819,7 +819,6 @@ func TestNodeAuthorizerConfig(t *testing.T) {
 			},
 		},
 		{
-
 			conf: `
 experimental:
   nodeAuthorizer:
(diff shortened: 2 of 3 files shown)


redbaron:
Thanks for these early validation warnings!

PR #877, opened 2017-08-25
Fix the existing TLS CA support

mumoshu:
It seems to have broken at some point.

This is verified to fix the problem by creating a cluster with a configuration which requires ca-key.pem to be deployed on controller nodes:

```
cd e2e && EXISTING_CA=1 RBAC=1 NODE_AUTHORIZER=1 KUBE_AWS_USE_CALICO=1 ETCD_DISASTER_RECOVERY_AUTOMATED=1 ETCD_SNAPSHOT_AUTOMATED=1 ETCD_COUNT=1 CONTROLLER_COUNT=1 KUBE_AWS_CLUSTER_NAME=clustername bash -c './run all'
```

and an another one which doesn't require ca-key.pem on controller nodes:

```
cd e2e && EXISTING_CA=1 ETCD_COUNT=1 CONTROLLER_COUNT=1 KUBE_AWS_CLUSTER_NAME=clustername bash -c './run all'
```

Fixes #839

cc @danielfm @kamatama41

PR #885, opened 2017-08-28
Dedicated worker CA and Etcd trusted CA bundle

redbaron:
TLSBootstap CA is now loaded from worker-ca.pem
Etcd trusted CA bundle is now loaded from etcd-trusted-ca.pem

Both files by default are symlinks to a well known ca.pem

This change enables advanced, more secure setup, where certificates
stored on worker nodes are trusted only by APIserver and NOT
Etcd server. This kind of setup is not yet supported by
"kube-aws render credentials" command out of the box, and
therefore currently requires users to be generating and
encrypting certs on their own to benefit from it.

Depends on  https://github.com/kubernetes-incubator/kube-aws/pull/882

### DIFF
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -451,7 +451,6 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error 
 	}
 
 	for _, sl := range symlinks {
-		from := filepath.Join(dirname, sl.from)
 		to := filepath.Join(dirname, sl.to)
 
 		if _, err := os.Lstat(to); err == nil {
@@ -460,7 +459,7 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error 
 			}
 		}
 
-		if err := os.Symlink(from, to); err != nil {
+		if err := os.Symlink(sl.from, to); err != nil {
 			return err
 		}
 	}
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -317,43 +317,52 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, encryptor Cach
 	r := new(EncryptedAssetsOnDisk)
 
 	type entry struct {
-		name         string
-		data         *EncryptedCredentialOnDisk
-		defaultValue *string
+		name          string
+		data          *EncryptedCredentialOnDisk
+		defaultValue  *string
+		readEncrypted bool
 	}
 
 	files := []entry{
-		{"tokens.csv", &r.AuthTokens, &defaultTokensFile},
-		{"kubelet-tls-bootstrap-token", &r.TLSBootstrapToken, &defaultTLSBootstrapToken},
+		{"tokens.csv", &r.AuthTokens, &defaultTokensFile, true},
+		{"kubelet-tls-bootstrap-token", &r.TLSBootstrapToken, &defaultTLSBootstrapToken, true},
 	}
 
 	if manageCertificates {
 		files = append(files, []entry{
-			{"ca.pem", &r.CACert, nil},
-			{"ca-key.pem", &r.CAKey, nil},
-			{"apiserver.pem", &r.APIServerCert, nil},
-			{"apiserver-key.pem", &r.APIServerKey, nil},
-			{"worker.pem", &r.WorkerCert, nil},
-			{"worker-key.pem", &r.WorkerKey, nil},
-			{"admin.pem", &r.AdminCert, nil},
-			{"admin-key.pem", &r.AdminKey, nil},
-			{"etcd.pem", &r.EtcdCert, nil},
-			{"etcd-key.pem", &r.EtcdKey, nil},
-			{"etcd-client.pem", &r.EtcdClientCert, nil},
-			{"etcd-client-key.pem", &r.EtcdClientKey, nil},
+			{"ca.pem", &r.CACert, nil, false},
+			{"ca-key.pem", &r.CAKey, nil, true},
+			{"apiserver.pem", &r.APIServerCert, nil, false},
+			{"apiserver-key.pem", &r.APIServerKey, nil, true},
+			{"worker.pem", &r.WorkerCert, nil, false},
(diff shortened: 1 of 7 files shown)


mumoshu:
@redbaron Thanks for your efforts. I like the motivation behind this change 👍 

/lgtm

core/controlplane/config/encrypted_assets.go
```diff
@@ -378,31 +406,67 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error
...
 		// the file already exists
 		{"tokens.csv", r.AuthTokens, false},
 	}
+
+	if includeCAKey {
+		assets = append(assets, struct {
+			name      string
+			data      []byte
+			overwrite bool
+		}{"ca-key.pem", r.CACert, true})
```
mumoshu: `r.CACert` here should be `r.CAKey`

core/controlplane/config/encrypted_assets.go
```diff
@@ -378,31 +406,67 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error
...
+	for _, sl := range symlinks {
+		to := filepath.Join(dirname, sl.to)
+
+		if _, err := os.Lstat(to); err == nil {
+			if err := os.Remove(to); err != nil {
+				return err
+			}
+		}
+
+		if err := os.Symlink(sl.from, to); err != nil {
```
mumoshu: `to` should be just `sl.to` while calling chdir before/after creating the symlink. Otherwise we get mysterious error when tried to read it:

```
# The file is there
$ ls assets/k8s88/credentials/worker-ca-key.pem
assets/k8s88/credentials/worker-ca-key.pem

# However, actually trying to read it results in a no-such-file error
$ cat assets/k8s88/credentials/worker-ca-key.pem
cat: assets/k8s88/credentials/worker-ca-key.pem: No such file or directory
```

```
$ ls -lah
total 136
drwxr-xr-x  20 mumoshu  staff   680B  9  1 15:29 .
drwxr-xr-x   7 mumoshu  staff   238B  9  1 15:29 ..
-rw-r--r--   1 mumoshu  staff     1B  9  1 15:29 .gitignore
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 admin-key.pem
-rw-------   1 mumoshu  staff   1.1K  9  1 15:29 admin.pem
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 apiserver-key.pem
-rw-------   1 mumoshu  staff   1.2K  9  1 15:29 apiserver.pem
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 ca-key.pem
-rw-------   1 mumoshu  staff   1.0K  9  1 15:29 ca.pem
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 etcd-client-key.pem
-rw-------   1 mumoshu  staff   1.1K  9  1 15:29 etcd-client.pem
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 etcd-key.pem
lrwxr-xr-x   1 mumoshu  staff    18B  9  1 15:29 etcd-trusted-ca.pem -> credentials/ca.pem
-rw-------   1 mumoshu  staff   1.1K  9  1 15:29 etcd.pem
-rw-------   1 mumoshu  staff   344B  9  1 15:29 kubelet-tls-bootstrap-token
-rw-------   1 mumoshu  staff     0B  9  1 15:29 tokens.csv
lrwxr-xr-x   1 mumoshu  staff    22B  9  1 15:29 worker-ca-key.pem -> credentials/ca-key.pem
lrwxr-xr-x   1 mumoshu  staff    18B  9  1 15:29 worker-ca.pem -> credentials/ca.pem
-rw-------   1 mumoshu  staff   1.6K  9  1 15:29 worker-key.pem
-rw-------   1 mumoshu  staff   1.1K  9  1 15:29 worker.pem
```

mumoshu: Here's my work-around:

```
diff --git a/core/controlplane/config/encrypted_assets.go b/core/controlplane/config/encrypted_assets.go
index b56dc51..3203582 100644
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -509,8 +509,13 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error
                {"ca-key.pem", "worker-ca-key.pem"},
        }

+       if err := os.Chdir("credentials"); err != nil {
+               return err
+       }
+
        for _, sl := range symlinks {
-               to := filepath.Join(dirname, sl.to)
+               from := sl.from
+               to := sl.to

                if _, err := os.Lstat(to); err == nil {
                        if err := os.Remove(to); err != nil {
@@ -518,10 +523,15 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error
                        }
                }

-               if err := os.Symlink(sl.from, to); err != nil {
+               if err := os.Symlink(from, to); err != nil {
                        return err
                }
        }
+
+       if err := os.Chdir("../"); err != nil {
+               return err
+       }
+
        return nil
 }
```

core/controlplane/config/encrypted_assets.go
```diff
@@ -378,31 +406,67 @@ func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error
...
+	// these can be separate CAs to ensure that worker nodes have no certs which would let them
+	// access etcd directly. If worker-ca.pem != ca.pem, then ca.pem should include worker-ca.pem
+	// to let TLS bootstrapped workers acces APIServer.
+	symlinks := []struct {
+		from string
+		to   string
+	}{
+		{"ca.pem", "worker-ca.pem"},
+		{"ca.pem", "etcd-trusted-ca.pem"},
+		{"ca-key.pem", "worker-ca-key.pem"},
```
mumoshu: `worker-ca-key.pem` becomes a symlink to an inexistent file when `ca-key.pem` was not written because the user provided `--ca-key-path` while the TLS bootstrapping is disabled.
We should not create the `worker-ca-key.pem` symlink or try to read it afterwards when `includeCAKey` is `false`, the same as what we do for `ca-key.pem` in #877

PR #886, opened 2017-08-28
core: add kube2iam image repo and tag

jpb:
The kube2iam image and tag used for the kube-aws provided kube2iam installation
can be specified in cluster.yaml.


redbaron:
sweet!

PR #889, opened 2017-08-29
Enable RBAC by default

### LINKED ISSUE #655: Enabling RBAC by default
Thanks to [huge efforts](https://github.com/kubernetes-incubator/kube-aws/pulls?q=is%3Apr+rbac+is%3Aclosed) from multiple kube-aws contributors @gianrubio @danielfm @camilb, kube-aws supports RBAC, which is still disabled by default.

On the other hand, people seem to start noticing the default behavior of kubernetes/kube-aws **"
deploy the token with the cluster-admin privilege to every pod"** is a security hole/a major attack surface:
Ref: https://raesene.github.io/blog/2017/04/02/Kubernetes-Service-Tokens/

So - is it time enable RBAC by default?
Can we just enable it since v0.9.7(the next release), or hopefully v0.9.8?

mumoshu:
Resolves #655

**Release note**:

* If you had never enabled RBAC before, you might have not configured appropriate cluster role/role binding/service account for your apps. It is strongly recommended to do so before upgrading your cluster or otherwise your apps may fail due to missing RBAC permissions.
* Although it isn't recommended for security, you can also retain the existing behavior that all the service accounts have all the permissions by running the following command to create permissive RBAC permissions:
  ```
  kubectl create clusterrolebinding permissive-binding \
    --clusterrole=cluster-admin \
    --user=admin \
    --user=kubelet \
    --group=system:serviceaccounts
  ```

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -93,7 +93,7 @@ func NewDefaultCluster() *Cluster {
 		},
 		Plugins: Plugins{
 			Rbac: Rbac{
-				Enabled: false,
+				Enabled: true,
 			},
 		},
 		Oidc: model.Oidc{
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -91,11 +91,7 @@ func NewDefaultCluster() *Cluster {
 			Enabled:      false,
 			DrainTimeout: 5,
 		},
-		Plugins: Plugins{
-			Rbac: Rbac{
-				Enabled: true,
-			},
-		},
+		Plugins: Plugins{},
 		Oidc: model.Oidc{
 			Enabled:       false,
 			IssuerUrl:     "https://accounts.google.com",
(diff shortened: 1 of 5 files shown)


redbaron:
Shouldn't we completely remove it as an option? Reason is , that unlikely any features are going to be tested without RBAC enabled, therefore it is soon going to be pretty much broken

mumoshu:
@redbaron Thanks for the great suggestion  :+1:
I agree with you. I'll address it before merging.

redbaron:
you are quick to act :) It was a proposal, what do others think? All side project we are using, recently got support for RBAC (nginx-ingress controller, external-dns, prometheus-operator). So this change is "free" for us, but it might not be the same for others. Seeing ~80 lines of `if` switches makes me happy though

mumoshu:
@redbaron I was afraid if this would result in a blocker for k8s apps missing RBAC support.

However, I think we can suggest users to work-around by manually provide the default service account all the permissions to retain previous behavior. Although it isn't recommended for security, it doesn't make things worse than now 😄
We can also write about the work-around in our release note.
IMHO, it still isn't free but cheap enough for everyone.

// Sry I was too lazy to write about this in my prev comment. It was written on my iPhone on which I'm not happy to write lengthy comment 😉

mumoshu:
Release notes for this PR is included in the description.

PR #900, opened 2017-08-31
feat: Encrypting Secret Data at Rest

jpb:
Add `EncryptionConfig` as a managed credential. The default `EncryptionConfig` encrypts secrets while at rest in etcd.

https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1782,6 +1782,9 @@ write_files:
           - --oidc-groups-claim={{.Experimental.Oidc.GroupsClaim}}
           {{ end -}}
           {{ end -}}
+          {{if .Experimental.EncryptionAtRest.Enabled}}
```
mumoshu: We have started to keep the `experimental` key at distance so I'm now rather ok with something like:

```yaml
kubernetes:
  encryptionAtRest:
    enabled: true
```

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1182,7 +1182,7 @@ write_files:
           -ec \
           'echo decrypting assets
            shopt -s nullglob
-           for encKey in /etc/kubernetes/{ssl,{{ if or (.AssetsConfig.HasAuthTokens) ( and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken) }}auth{{end}}}/*.enc; do
+           for encKey in /etc/kubernetes/{ssl,additional-configs,{{ if or (.AssetsConfig.HasAuthTokens) ( and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken) }}auth{{end}}}/*.enc; do
```
mumoshu: I'd rather prefer `encryption` over `additional-configs` like other directories at the same level, so that we can be more explicit about what are stored under the `encryption` directory - an encryption config.

mumoshu:
Just wondering how this feature handles the key rotation #1013

mumoshu:
FYI: I'll try to resolve conflicts and adapt this towards v0.9.10-rc.1

core/controlplane/config/templates/cloud-config-controller
mumoshu: Addressed in a follow-up commit

mumoshu: Oops. No, we already have this additional-configs directory for another feature. So Im happy to just rely on it today.

core/controlplane/config/templates/cloud-config-controller
mumoshu: Addressed in a follow-up commit

mumoshu:
I did a quick test and this shouldn't be breaking anything! And I've addressed my comments on my own. I'm happy to merge this now.

Thank you very much for your contribution @jpb !

PR #904, opened 2017-09-01
Fix too permissive IAM policy for CA associated to controller nodes

### LINKED ISSUE #903: Too permissive IAM policy for CA associated to controller nodes
The IAM policy associated to controller nodes is unnecessarily permissive for running CA.

```
{
                "Effect": "Allow",
                  "Action": [
                    "autoscaling:DescribeAutoScalingInstances",
                    "autoscaling:DescribeTags",
                  ],
                  "Resource": "*"
                },
```

What we actually want instead would look exactly same as the one for worker nodes:

```
{
                  "Action": [
                    "autoscaling:DescribeAutoScalingInstances",
                    "autoscaling:DescribeTags",
                  ],
                  "Effect": "Allow",
                  "Resource": "*"
                },
                {
                  "Action": [
                  ],
                  "Condition": {
                    "Null": { "autoscaling:ResourceTag/kubernetes.io/cluster/{{.ClusterName}}": "false" }
                  },
                  "Effect": "Allow",
                  "Resource": "*"
                },
```

The latter is better because it prevents any app running on the nodes from terminating EC2 instances outside of the K8S cluster. For instance, it would have prevented a misbehaving CA to terminate EC2 instances outside of the K8S cluster like we experienced in #800.

mumoshu:
Fix #903

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -235,14 +235,23 @@
                 {{end}}
                 {{if .Addons.ClusterAutoscaler.Enabled}}
                 {
-                "Effect": "Allow",
                   "Action": [
                     "autoscaling:DescribeAutoScalingGroups",
                     "autoscaling:DescribeAutoScalingInstances",
-                    "autoscaling:DescribeTags",
+                    "autoscaling:DescribeTags"
+                  ],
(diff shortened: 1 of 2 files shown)


redbaron:
Sweet, least permissive policies are the best

Issue #800, opened 2017-07-26
Deploying cluster autoscaler but not enabling any node pools results in confusing logs and access to non-kube ASGs


c-knowles:
As of 61e411229d5b6fa6c0fbbec1c0e3e91952a4c053, turning on the cluster autoscaler with RBAC enabled gets logs similar to the following:

```W  Failed to find readiness information for k8s-preprod-v7-Controlplane-1VFZCIM8O6ZUO-Controllers
W  Failed to find readiness information for k8s-dev-v8-PoolC-1JBFUB0G59YY7-Workers 
W  Failed to find readiness information for k8s-dev-v8-Controlplane-1ERY9ADUQZRDI-Etcd0
```

I've only deployed the autoscaler to a single cluster (dev) but there are also the same logs for another totally different cluster in the same AWS account.

mumoshu:
@c-knowles Thanks for the feedback!

I can't recall whether the messages had been there or not when I first developed the feature on CA.

For the purpose of gathering more information on investigation, would you mind confirming:
* Did you enable autoscaling by `worker.nodePools[].autoscaling.clusterAutoscaler.enabled`?
* Does autoscaling(both scaling-up and down) work or not?

c-knowles:
@mumoshu I can confirm that the autoscaler app is deployed. I have not enabled it on any pools yet however I did this as a first step to see it deployed and thought it should be doing nothing rather than logging warnings about unrelated ASGs. I'm also getting logs about ASGs unrelated to k8s/kube-aws which is a bit concerning. Maybe the autoscaler needs to filter prior to logging about unrelated ASGs.

c-knowles:
Just confirmed if I enable on a single node pool that all the warnings disappear. I suppose this is mostly unexpected behaviour then when it's not activated but why is the autoscaler even concerned about those other ASGs at all?

mumoshu:
@c-knowles Thanks for the confirmation 👍 

AFAIK, CA should [try to get readiness information only when there are one or more node groups](https://github.com/kubernetes/autoscaler/blob/591ced9c7c61ec3ccdedded46f75e1be1856016d/cluster-autoscaler/clusterstate/clusterstate.go#L237).
Node groups in CA are backed by Spot Fleets or ASGs in kube-aws. Spot Fleets are not supported by CA though).

Perhaps there is a bug in CA that CA tries to get readiness information on an unknown node group when there are no node groups configured as scaling targets?

[...]

PR #905, opened 2017-09-01
Fix the scheduling and permissions issue of CA

### LINKED ISSUE #897: Cluster-autoscaler pod is launched on worker instances, but permissions for scaling are granted to controller instances only

mumoshu:
by fixing the conditional in stack-template.json for worker and controller stacks, while making configuration easier.

Resolves #897

## Context

The intended way of enabling cluster-autoscaler on your cluster has been to:

1. Set `addons.clusterAutoscaler.enabled` to `true`
2. Set `experimental.clusterAutoscalerSupport.enabled` to `true` to allow CA to be run on controller nodes, while providing appropriate IAM permissions to the controller nodes
3. Set `worker.nodePools[].clusterAutoscalerSupport.enabled` to `true` to allow CA to be run on worker nodes, while providing appropriate IAM permissions to the worker nodes
4. Finally, set `worker.nodePools[].autoscaling.clusterAutoscaler.enabled` to `true` for registering the node pool(s) as the scaling targets

Note that:

* 1. is required in order to deploy CA on your cluster.
* 2. and/or 3. is required to actually allow scheduling CA to any node
* 4. is required to instruct CA to start auto-scaling. Btw, missing this step had been resulted in #800 before.

## The fix

`worker.nodePools[].clusterAutoscalerSupport.enabled` has been configured CA to run on worker nodes, certainly. However, it was not providing required IAM permissions to the worker nodes.
More concretely, you had to set `worker.nodePools[].addons.clusterAutoscaler.enabled` instead and that's not an expected way of configuring things. This is now fixed.

#897 is considered to be affected by this.

## The easier configuration

To make this process easier, `experimental.clusterAutoscalerSupport.enabled` is now set to `true` by default.
This way, 2. and 3. can be completely omitted when you don't mind on which node CA runs.

A validation to prevent a misconfiguration is also added.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -64,7 +64,7 @@ func NewDefaultCluster() *Cluster {
 			Enabled: false,
 		},
 		ClusterAutoscalerSupport: model.ClusterAutoscalerSupport{
-			Enabled: false,
+			Enabled: true,
 		},
 		TLSBootstrap: TLSBootstrap{
 			Enabled: false,
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -233,7 +233,7 @@
                   "Resource": [ "*" ]
                 },
                 {{end}}
-                {{if .Addons.ClusterAutoscaler.Enabled}}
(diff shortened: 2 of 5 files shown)


mumoshu:
Get the ball rolling 🍙

PR #916, opened 2017-09-04
Bump default k8s to 1.7.5

c-knowles:


c-knowles:
Couple places missed during https://github.com/kubernetes-incubator/kube-aws/pull/853.

camilb:
Seems to be a bug in this version: [The Pod-To-Pod network does not work when deploy flanneld with DaemonSet in kubernetes 1.7.5](https://github.com/kubernetes/kubernetes/issues/51881).
I'll test it later to confirm.

redbaron:
@camilb should I revert then?

camilb:
@redbaron I still don't know if this affects us also or it's something related with a specific use case.  I'll start a new cluster based on 1.7.5 to make sure. I'll let you know as soon as I have the cluster up.

camilb:
@redbaron Just finished the tests with 1.7.5 and can not replicate the [issue](https://github.com/kubernetes/kubernetes/issues/51881). I think is related with his setup, not kubernetes version.

c-knowles:
Seems like it's not present in my dev cluster or maybe just didn't hit the problem yet.

PR #934, opened 2017-09-08
Remove `mapPublicIPs` and `routeTableId` in the top-level of cluster.yaml

### LINKED ISSUE #929: The combination of `mapPublicIPs` and `routeTableId` in cluster.yaml is considered hard to configure
because it is very confusing and seems hard to configure correctly as described in https://github.com/kubernetes-incubator/kube-aws/pull/284#issuecomment-275998862.
@ketzacoatl seems to have been blocked by this in #913.

Currently:

* The inverse of `mapPublicIPs` is copied to `subnets[].private`
* `routeTableId` is copied to `subnets[].routeTable.id`

Plus

1. `mapPublicIPs: false` + non-empty `routeTableId` requires one or more private subnets under the `subnets:` section of cluster.yaml, for setting up controller and etcd nodes 
2. `mapPublicIPs: true` + non-empty `routeTableId` requires one or more public subnets to exist under the `subnets:` section of cluster.yaml, for setting up controller and etcd nodes 

I believe that the latter two are not very intuitive to reason about.

Under the hood, `mapPublicIPs` and `routeTableId` are magically translated to `subnets[].private` and `subnets[].routeTable.id` as of today and therefore you can just use those two keys under `subnets:` to gain the same result.

So, I suggest to remove those two keys and make the combination of `subnets[].private` and `subnets[].routeTable.id` to be only way to configure subnets for top-level, controller and etcd nodes.

mumoshu:
Resolves #929

**Release note**:

* `mapPublicIPs` is removed in favor of `apiEndpoints[].loadBalancer.private`
  * `mapPublicIPs: true` should be migrated to `private: false` and vice versa
* `routeTableId` is removed in favor of `subnets[].routeTable.id` in order to reduce gotchas and implications in configuring controller and etcd subnets

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -109,7 +109,6 @@ func NewDefaultCluster() *Cluster {
 			ContainerRuntime:   "docker",
 			Subnets:            []model.Subnet{},
 			EIPAllocationIDs:   []string{},
-			MapPublicIPs:       true,
 			Experimental:       experimental,
 			ManageCertificates: true,
 			AmazonSsmAgent: AmazonSsmAgent{
@@ -246,7 +245,9 @@ func (c *Cluster) Load() error {
 		return fmt.Errorf("invalid cluster: %v", err)
 	}
 
-	c.SetDefaults()
+	if err := c.SetDefaults(); err != nil {
+		return fmt.Errorf("invalid cluster: %v", err)
+	}
 
 	if c.ExternalDNSName != "" {
 		// TODO: Deprecate externalDNSName?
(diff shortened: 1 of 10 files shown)


mumoshu:
@redbaron @camilb Would you mind reviewing this?

core/controlplane/config/config.go
```diff
@@ -345,15 +317,17 @@ func (c *Cluster) SetDefaults() {
 	}
 
 	if len(c.Controller.Subnets) == 0 {
-		if privateTopologyImplied {
-			c.Controller.Subnets = c.PrivateSubnets()
-		} else {
-			c.Controller.Subnets = c.PublicSubnets()
+		c.Controller.Subnets = c.PublicSubnets()
```
redbaron: I was always lost in this part of code. Why `PublicSubnets` is better default than `PrivateSubnets()`?

mumoshu: That's because you have no private subnets by default.
Anyway, I guess we can discuss it for
alternatively defaulting to either PrivateSubnets or just Subnets in a separate issue!

PR #936, opened 2017-09-11
Remove `createRecordSet` to help newcomers bring up their first clusters

### LINKED ISSUE #928: Is it hard for a first-time user to correctly configure `createRecordSet` + `hostedZone.id`?
I have been occasionally asked to debug why one's cluster failed while creating worker nodes.

The cause tend to be `createRecordSet: false` in cluster.yaml - without suprises - which is the default when one run `kube-aws init` without `--hosted-zone-id`. Someone misses to set `apiEndpoints[].loadBalancer.hostedZoneId` altogether while someone does not, which seems hard to notice the difference for a first-time user.

How could we improve the situation?

My suggestion is:

1. Completely remove `apiEndpoints[].loadBalancer.createRecordSet` and make kube-aws creates a record set for you whenever `apiEndpoints[].loadBalancer.hostedZone.id` is specified
   * I can't remember why we had needed `createRecordSet` in the first place 😉
   * What could `hostedZone.id` be used for when `createRecordSet: false`. Similarly, what could `createRecordSet: true` do when `hostedZone.id` is not present? If nothing, shouldn't we just create the record set when the hosted zone id is present?

2. Make `kube-aws init` require `--hosted-zone-id` by default and allow it to be omitted when and only when a flag like `--bring-your-own-dns` is explicitly provided by the user.

@c-knowles @redbaron @danielfm @camilb WDYT?

mumoshu:
by preventing the Route53 misconfiguration.

The `--hosted-zone-id` flag and the corresponding `hostedZone.id` key in cluster.yaml are now required by default. You can explicitly omit them by setting `--no-record-set` or `recordSetManaged: false`. This way, one can't omit hosted zone id without knowing the implication of omitting the hosted zone id.

Resolves #928

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/kubernetes-incubator/kube-aws/core/controlplane/config"
@@ -17,7 +18,7 @@ var (
 		SilenceUsage: true,
 	}
 
-	initOpts = config.Config{}
+	initOpts = config.InitialConfig{}
 )
 
 func init() {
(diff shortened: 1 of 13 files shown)


redbaron:
looks like it is renaming `createRecordSet` to `recordSetManaged`, do I miss something?

mumoshu:
@redbaron Yes and no  :+1:

It turned out that we need a config key to allow one to force creation of LB even though one didn't specify any key other than that. CreateRecordSet has also 
been used for that purpose but it isn't equivalent to RecordSetManaged.

Basically, RecordSetManaged has fewer meanings and a better default than CreateRecordSet.

mumoshu:
@redbaron Thanks for reviewing :beers:

nick4fake:
https://github.com/kubernetes-incubator/kube-aws/pull/936/commits/5cdb74ef7e24d01bf2d1aa946538c9428c7fb1a0#diff-971c12e3b00cffb1d084e6e1f5942103R99

There should not be a check for nil in e.RecordSetManaged

jorge07:
I could be a good idea add this change to `Actions required` in the release section.

I found that issue digging into issues after look at release notes and it should be the opposite.

Anyway its works ;)

PR #938, opened 2017-09-12
Handle worker-ca correctly when TLSBootstarpping is enabled

redbaron:


redbaron:
Looks reasonable enough, self-merging :)

PR #944, opened 2017-09-14
Bump default k8s to 1.7.6

Vrtak-CZ:


mumoshu:
@Vrtak-CZ LGTM. Thanks for the contribution!

Issue #598, opened 2017-04-26
ELB IP changes can bring the cluster down


danielfm:
I ran into https://github.com/kubernetes/kubernetes/issues/41916 twice in the last 3 days in my production cluster, with almost 50% of worker nodes transitioning to `NotReady` state almost simultaneously in both days, causing a brief downtime in critical services due to Kubernetes default (and agressive) eviction policy for failing nodes.

I just contacted AWS support to validate the hypothesis of the ELB changing IPs at the time of both incidents, and the answer was _yes_.

My configuration (multi-node control plane with ELB) matches exactly the one in that issue, and probably most kube-aws users are subject to this.

Have anyone else ran into this at some point?

mumoshu:
@danielfm Thanks for the report.

I think I have never encountered the issue myself for my production cluster but anyway, I believe this is a BIG issue.

IMHO, ideally, kubelet should detect any dead long-polling connection and then re-discover ELB instances via DNS and reconnect to one of them. However I'm happy to work-around the issue in kube-aws.

## Possible work-arounds

Perhaps:
* periodically monitoring the list of ips an route53 alias record for an ELB and then
* restarting kubelet whenever the monitor detects the list to be changed
so that kubelet can reconnect to one of active ELB ips before it marks the k8s node `NotReady`?

If we go that way, the monitor should be executed periodically with an interval of `(<node-monitor-grace-period> - <amount of time required for kubelet to start> - <amount of time required for each monitor run>) / <a heuristic value: N> - <amount of time required to report node status after kubelet started>` at least, so that we can provide kubelet at most N chances to successfully report its status to possibly available ELB.

Or we can implement a DNS round robin with a health-checking mechanism for serving k8s API endpoints like suggested and described in #281 #373

mumoshu:
This hash changes only when backend IPs of an ELB has been changed.

```
$ dig my.k8s.cluster.example.com +noall +answer +short | sort | sha256sum | awk '{print $1}'
96ff696e9a0845dd56b1764dd3cfaa1426cdfbd7510bab64983e97268f8e0bc4

$ dig my.k8s.cluster.example.com +noall +answer +short | sort
52.192.165.96
52.199.129.34
<elb id>.ap-northeast-1.elb.amazonaws.com.
```

mumoshu:
[...]

PR #945, opened 2017-09-18
Add initial support for NLB API endpoints

### LINKED ISSUE #937: Switch to NLB instead of ELB
Main benefit: stable IP addresses within VPC, therefore it is expected to be much less broken TCP connections when AWS roll underneath infrastructure

danielfm:
This adds support for an optional `type` attribute to `apiEndpoints[].loadBalancer`, which supports the following values:

- _undefined_ (defaults to Classic ELBs, just like before)
- `type: classic` (same as above)
- `type: network` (provisions a Network Load Balancer instead of a Classic ELB)

The idea is to use NLB-backed API endpoints for worker nodes in an attempt to mitigate issues like #598.

The first version implemented in this PR has a few limitations:

- It can only be used if load balancer is managed by kube-aws, that is, we cannot set `type: network` when specifying the load balancer `id`
- It allows inbound connections to TCP port 443 from the VPC CIDR to the controller security group if at least one NLB-backed API endpoint is defined (see below)

A few considerations regarding NLBs at their present form:

- It's not possible to override the check interval (which defaults to 10)
- It's not possible to override the check timeout (which defaults to 10)
- As of now, it's not possible to specify a different value for healthy and unhealthy thresholds, so I set them both to 3
- It's not possible to attach a security group to a NLB; this means controller nodes must accept direct inbound connections to TCP-443 from the VPC CIDR, according to the [docs](http://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html#target-security-groups)
 
TODO

- [x] Document new flag in cluster.yaml
- [x] Fix `kube-aws info` by using the ELBV2 API when trying to fetch NLBs DNS name
- [x] Test with different combinations of public/private controllers/workers
- [x] Add inbound rule to port 443 (TCP) from the VPC CIDR to controller node SG if there are any NLB-backed API endpoints
- [x] Add validation to forbid creation of NLB-backed API endpoints if deploying to the China region, which do not support NLB
- [x] Test upgrade path from ELB API endpoint to a NLB API endpoint for worker nodes
- [x] Fix/add tests

Closes #937.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1088,7 +1088,7 @@
             "IpProtocol": "tcp",
             "ToPort": 443
           },
-          {{ range $_, $r := $.APIAccessAllowedSourceCIDRs -}}
+          {{ range $_, $r := $.APIAccessAllowedSourceCIDRsForControllerSG -}}
           {
             "CidrIp": "{{$r}}",
             "FromPort": 443,
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -49,7 +49,13 @@
           {{end}}
         ],
         "LoadBalancerNames" : [
-          {{range $i, $ref := .APIEndpoints.ELBRefs -}}
+          {{range $i, $ref := .APIEndpoints.ELBClassicRefs -}}
+          {{- if gt $i 0}},{{end}}
+          {{ $ref }}
+          {{- end}}
+        ],
+        "TargetGroupARNs": [
+          {{range $i, $ref := .APIEndpoints.ELBV2TargetGroupRefs -}}
           {{- if gt $i 0}},{{end}}
           {{ $ref }}
           {{- end}}
@@ -895,6 +901,50 @@
       }
     },
     {{ end -}}
+    {{ if .LoadBalancer.LoadBalancerV2 }}
+    "{{.LoadBalancer.LogicalName}}TargetGroup": {
+      "Type": "AWS::ElasticLoadBalancingV2::TargetGroup",
+      "Properties": {
+        "HealthCheckIntervalSeconds": "10",
+        "HealthyThresholdCount": "3",
+        "UnhealthyThresholdCount": "3",
+        "Port": "443",
+        "VpcId": {{$.VPCRef}},
+        "Protocol": "TCP"
+      }
+    },
+    "{{.LoadBalancer.LogicalName}}Listener": {
+      "Type": "AWS::ElasticLoadBalancingV2::Listener",
+      "Properties": {
+        "DefaultActions": [
+          {
+            "TargetGroupArn": {"Ref": "{{.LoadBalancer.LogicalName}}TargetGroup"},
+            "Type": "forward"
+          }
+        ],
+        "LoadBalancerArn": {"Ref": "{{.LoadBalancer.LogicalName}}"},
+        "Port": "443",
+        "Protocol": "TCP"
+      }
+    },
+    "{{.LoadBalancer.LogicalName}}" : {
+      "Type" : "AWS::ElasticLoadBalancingV2::LoadBalancer",
+      "Properties" : {
+        "Type": "network",
+        "Subnets" : [
+          {{range $index, $subnet := .LoadBalancer.Subnets}}
+          {{if gt $index 0}},{{end}}
+          {{$subnet.Ref}}
+          {{end}}
+        ],
+        {{if .LoadBalancer.Private}}
+        "Scheme": "internal"
+        {{else}}
+        "Scheme": "internet-facing"
+        {{end}}
+      }
+    },
+    {{ else }}
     "{{.LoadBalancer.LogicalName}}" : {
       "Type" : "AWS::ElasticLoadBalancing::LoadBalancer",
       "Properties" : {
@@ -936,6 +986,7 @@
         ]
       }
     },
+    {{ end }}
     {{if .LoadBalancer.ManageSecurityGroup -}}
     "{{.LoadBalancer.SecurityGroupLogicalName}}" : {
       "Properties": {
@@ -1029,6 +1080,14 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ if $.KubeClusterSettings.APIEndpointConfigs.HasNetworkLoadBalancers }}
+          {
+            "CidrIp": "{{.VPCCIDR}}",
+            "FromPort": 443,
+            "IpProtocol": "tcp",
+            "ToPort": 443
+          },
+          {{ end }}
           {
             "SourceSecurityGroupId" : { "Ref" : "SecurityGroupElbAPIServer" },
             "FromPort": 443,
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -901,7 +901,7 @@
       }
     },
     {{ end -}}
-    {{ if .LoadBalancer.LoadBalancerV2 }}
+    {{ if .LoadBalancer.NetworkLoadBalancer }}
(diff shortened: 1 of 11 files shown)


redbaron:
> In summary, maybe we could only allow type: network for private load balancers. Does this make sense?

China region doesn't have them yet

danielfm:
Yes, we could add some validation code to address this.

danielfm:
Any feedback on this @redbaron @mumoshu @c-knowles?

model/api_endpoint_lb.go
```diff
@@ -56,7 +58,22 @@ func (e APIEndpointLB) ManageELB() bool {
...
+}
+
+// LoadBalancerV2 returns true if the load balancer is a ELBV2 load balancer (only network load balancer is supported for now)
+func (e APIEndpointLB) LoadBalancerV2() bool {
+	return e.Type != nil && *e.Type != "classic"
+}
+
+// NetworkLoadBalancer returns true if the load balancer is a ELBV2 network load balancer
+func (e APIEndpointLB) NetworkLoadBalancer() bool {
+	return e.Type != nil && *e.Type != "classic"
```
redbaron: typo? maybe `*e.Type == "network"`?

danielfm: 👍  Good catch

model/api_endpoint_lb.go
```diff
@@ -84,6 +102,10 @@ func (e APIEndpointLB) Validate() error {
 			return errors.New("hostedZone.id should not be specified when an API endpoint LB is not managed by kube-aws")
 		}
 
+		if e.Type != nil && len(*e.Type) > 0 {
```
redbaron: maybe also validate possible values here? accept `classic` or `network`, but nothing else

core/controlplane/config/templates/stack-template.json
```diff
@@ -895,6 +901,50 @@
       }
     },
     {{ end -}}
+    {{ if .LoadBalancer.LoadBalancerV2 }}
```
redbaron: should be `.LoadBalancer.NetworkLoadBalancer`

danielfm: Agreed.

core/controlplane/config/templates/stack-template.json
```diff
@@ -895,6 +901,50 @@
...
+    "{{.LoadBalancer.LogicalName}}TargetGroup": {
+      "Type": "AWS::ElasticLoadBalancingV2::TargetGroup",
+      "Properties": {
+        "HealthCheckIntervalSeconds": "10",
+        "HealthyThresholdCount": "3",
```
redbaron: it seems to be spamming APIserver with non-SSL healthchecks which produce annoying entry in logs, does `HealthCheckProtocol: HTTPS` work?

danielfm: TCP is the only supported protocol for NLB health checks at this point.

danielfm: es, I just checked the API server logs, and there are a lot of TLS handshake errors from these health checks. Do you have a suggestion on how can we address this?

model/api_endpoint_lb.go
danielfm: If `id` is specified, we only validate that no `type` attribute is specified _at all_, which is the correct thing to do in case the load balancer `id` is specified (since we are reusing a ELB, there's no need to specify a type for it).

The check to verify the validity of `type` itself is performed down this same function:

https://github.com/kubernetes-incubator/kube-aws/pull/945/files/6af87179b66f5fd5da35cfabbd2fa7c9cb2db7d6#diff-971c12e3b00cffb1d084e6e1f5942103R136

core/controlplane/config/templates/stack-template.json
redbaron: it seems there is no way out of it without patching apiserver to not to complain about it on info leveld.

model/api_endpoint_lb.go
```diff
@@ -107,10 +129,32 @@ func (e APIEndpointLB) Validate() error {
...
+		if len(e.SecurityGroupIds) > 0 {
+			return errors.New("cannot specify security group IDs for a network load balancer")
+		}
+
+		if len(e.APIAccessAllowedSourceCIDRs) != len(defaultRanges) {
+			return errors.New("cannot override apiAccessAllowedSourceCIDRs for a network load balancer")
+		}
+
+		for i, r := range defaultRanges {
+			if r != e.APIAccessAllowedSourceCIDRs[i] {
```
mumoshu: I guess we should add an controller sg ingress corresponds to apiAccessAllowedSourceCIDRs also for an NLB.

Oherwise a public API endpoint backed by an NLB isn't accessible from the Internet?

If we go that way, allowing to customize CIDRs here would be nicer for higher security.

danielfm: @mumoshu It's not possible to attach Security Groups to NLBs; one way to think of NLB is that it's an Elastic IP address that happens to go to multiple instances, instead of just one. Everything else stays the same.

What we can do here is restrict the usage of NLB-backed API endpoints to private NLBs only. This way, we still get to use NLB-backed API endpoints in worker nodes (in the hopes of mitigating issues like #598, but prevent users from creating a publicly-accessible NLB.

mumoshu: @danielfm Thx! I think I understand that. Therefore, I suggest to alter "controller" sg ingress to allow apiAccessAllowedSourceCIDRs for publicly-accessible NLBs, instead of trying to attach sgs to an NLB which isn't possible.

danielfm: Not sure I'm following. What you're suggesting is to append the `apiAccessAllowedSourceCIDRs` from each NLB-backed API endpoint to the controller SG ingress rules?

mumoshu: Exactly. It would look like similarly to the controller sg ingress for a vpc cidr you've added, which I believe the intention is to allow worker nodes to acess k8s api via NLB.

redbaron: @danielfm , suggestion by @mumoshu will work because NLB preserves source IP address, therefore SG would be able to filter traffic

danielfm: Done.

danielfm:
I'll still reviewing these changes myself, so please do not merge this just yet! I'll run a few more tests in the next few hours.

mumoshu:
LGTM. Thanks for your efforts @danielfm and @redbaron :+1:

core/controlplane/config/templates/stack-template.json
micahhausler: Just FYI, there are HTTP and HTTPS health checks on NLB ([docs](http://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html))

danielfm: I don't see this here:

image

As far as I know, HTTP(S) health checks are only available for application load balancers (ALB, L7); NLBs operate on the TCP layer.

micahhausler: Once you create an NLB listener and HealthCheck, the AWS console doesn't let you change it to HTTP/HTTPS (unfortunately). Try (manually) making a new Target Group with an HTTPS health check on an alternate port. You can then attach it to the NLB. I'm doing it right here: https://github.com/micahhausler/kubernetes/blob/662f4eee0d11d07bc2d0ca64198f2b0c051f684e/pkg/cloudprovider/providers/aws/aws_loadbalancer.go#L499-L514

micahhausler: To be clear, the listener protocol is TCP, but the _target group_ health check can be HTTPS

danielfm: Great, thanks for letting us know! That's great news.

PR #963, opened 2017-10-03
Support deployment to AWS GovCloud

amdonov:
I found that the IAM partition and S3 endpoint used for the us-gov-west-1 region were incorrect.


mumoshu:
@amdonov LGTM. Thanks for the contribution!

PR #966, opened 2017-10-04
Convert kube-proxy to a DaemonSet

### LINKED ISSUE #793: Update kube-proxy to match the upstream way of deployment
Our method of deploying kube-proxy is somewhat outdated. As far as I know, these are the things we must do to address that:

- Convert kube-proxy to a DaemonSet (pay attention to the required tolerations)
- Use the `kube-proxy` service account to authorize requests from kube-proxy pods
- Use the `--kubeconfig` flag to specify the API server location + service account credentials

Reference: [kubeadm](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/addons/manifests.go)

danielfm:
This PR changes the way of deploying kube-proxy; instead of deploying it via a Pod manifest on every node, it creates a `DaemonSet` that tolerates every possible taint.

The way this was implemented does not break any kube-aws feature, most notably the ability to use different API endpoints in different node pools. (This is why the kube-proxy kubeconfig file was not deployed as a `ConfigMap`, [as is done by kubeadm](https://github.com/kubernetes/kubernetes/blob/9d725da4c38ddbce5f8aec6b881df1e2f310c457/cmd/kubeadm/app/phases/addons/proxy/manifests.go#L22-L48))

Closes #793.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -2737,6 +2737,7 @@ write_files:
     content: {{.AssetsConfig.EtcdTrustedCA}}
 {{ end }}
 
+  # File needed on every node (used by the kube-proxy DaemonSet), including controllers
   - path: /etc/kubernetes/kubeconfig/kube-proxy.yaml
     content: |
         apiVersion: v1
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1524,6 +1524,9 @@ write_files:
           - kind: ServiceAccount
             name: kube-proxy
             namespace: kube-system
+          # Not needed after migrating to DaemonSet-based kube-proxy
+          - kind: Group
+            name: system:nodes
         roleRef:
           kind: ClusterRole
           name: system:node-proxier
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -250,6 +250,7 @@ coreos:
         ExecStartPre=/usr/bin/mkdir -p /var/lib/cni
         ExecStartPre=/usr/bin/mkdir -p /var/log/containers
         ExecStartPre=/usr/bin/mkdir -p /opt/cni/bin
+        ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/manifests
         ExecStartPre=/usr/bin/etcdctl \
                        --ca-file /etc/kubernetes/ssl/etcd-trusted-ca.pem \
                        --key-file /etc/kubernetes/ssl/etcd-client-key.pem \
@@ -758,7 +759,7 @@ write_files:
       kubectl apply -f "${mfdir}/kube-dns-cm.yaml"
 
       # Service Accounts
-      for manifest in {kube-dns,heapster}; do
+      for manifest in {kube-dns,heapster,kube-proxy}; do
           kubectl apply -f "${mfdir}/$manifest-sa.yaml"
       done
 
@@ -773,6 +774,11 @@ write_files:
       # Deployments
       for manifest in {kube-dns,kube-dns-autoscaler,kube-dashboard,{{ if .Addons.ClusterAutoscaler.Enabled }}cluster-autoscaler,{{ end }}heapster{{ if .KubeResourcesAutosave.Enabled }},kube-resources-autosave{{ end }}}; do
           kubectl apply -f "${mfdir}/$manifest-de.yaml"
+      done
+
+      # Daemonsets
+      for manifest in {kube-proxy,}; do
+          kubectl apply -f "${mfdir}/$manifest-ds.yaml"
       done
 
       # Services
@@ -1504,10 +1510,8 @@ write_files:
           name: cluster-admin
           apiGroup: rbac.authorization.k8s.io
 
-  # Allows both `kube-worker` user and members of the `system:nodes` group
-  # to perform actions needed by the `kube-proxy` component. Once kube-proxy
-  # is migrated to DaemonSets, we could set up a ServiceAccount for it and
-  # associate this role to it instead.
+  # Also allows `kube-worker` user to perform actions needed by the
+  # `kube-proxy` component.
   - path: /srv/kubernetes/rbac/cluster-role-bindings/node-proxier.yaml
     content: |
         kind: ClusterRoleBinding
@@ -1517,8 +1521,9 @@ write_files:
         subjects:
           - kind: User
             name: kube-worker
-          - kind: Group
-            name: system:nodes
+          - kind: ServiceAccount
+            name: kube-proxy
+            namespace: kube-system
         roleRef:
           kind: ClusterRole
           name: system:node-proxier
@@ -1692,10 +1697,10 @@ write_files:
           apiGroup: rbac.authorization.k8s.io
 {{ end }}
 
-  - path: /etc/kubernetes/manifests/kube-proxy.yaml
+  - path: /srv/kubernetes/manifests/kube-proxy-ds.yaml
     content: |
-        apiVersion: v1
-        kind: Pod
+        apiVersion: extensions/v1beta1
+        kind: DaemonSet
         metadata:
           name: kube-proxy
           namespace: kube-system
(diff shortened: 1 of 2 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,14 +2743,31 @@ write_files:
     content: {{.AssetsConfig.EtcdTrustedCA}}
 {{ end }}
 
+  - path: /etc/kubernetes/proxy-kubeconfig.yaml
```
redbaron: doesn't it go by default to localhost:8080 by default?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1692,43 +1697,60 @@ write_files:
...
+                - proxy
+                - --kubeconfig=/etc/kubernetes/proxy-kubeconfig.yaml
+                - --cluster-cidr={{.PodCIDR}}
+                securityContext:
+                  privileged: true
+                volumeMounts:
+                - mountPath: /etc/kubernetes
+                  name: kubeconfig
+                  readOnly: true
+                - mountPath: /var/run/dbus
```
redbaron: it needs dbus for firewalld on Fedora servers, shouldn't be needed on coreos

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,14 +2743,31 @@ write_files:
...
+
   - path: /etc/kubernetes/controller-kubeconfig.yaml
     content: |
         apiVersion: v1
         kind: Config
         clusters:
         - name: local
           cluster:
-            server: http://localhost:8080
+            server: http://127.0.0.1:8080
```
redbaron: ipv6 issues?

core/controlplane/config/templates/cloud-config-controller
redbaron: ignore me, it is a daemonset now, and you need `--kubeconfig` file on every pod even on controllers

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1692,43 +1697,60 @@ write_files:
...
+              containers:
+              - name: kube-proxy
+                image: {{.HyperkubeImage.RepoWithTag}}
+                command:
+                - /hyperkube
+                - proxy
+                - --kubeconfig=/etc/kubernetes/proxy-kubeconfig.yaml
+                - --cluster-cidr={{.PodCIDR}}
+                securityContext:
+                  privileged: true
```
redbaron: it might be enough to have `NET_ADMIN` capability only

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,14 +2743,31 @@ write_files:
...
+  - path: /etc/kubernetes/proxy-kubeconfig.yaml
+    content: |
+        apiVersion: v1
+        kind: Config
+        clusters:
+        - name: local
+          cluster:
+            server: http://127.0.0.1:8080
+        users:
+        - name: kubelet
```
redbaron: I think it should use service acccount token to have same permissions granted by RBAC

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1692,43 +1697,60 @@ write_files:
...
+              - operator: Exists
+                key: CriticalAddonsOnly
+              hostNetwork: true
+              containers:
+              - name: kube-proxy
+                image: {{.HyperkubeImage.RepoWithTag}}
+                command:
+                - /hyperkube
+                - proxy
+                - --kubeconfig=/etc/kubernetes/proxy-kubeconfig.yaml
```
redbaron: can we avoid mounting /etc/kubernetes into pod? There are ssl certs which this pod really dont need

core/controlplane/config/templates/cloud-config-controller
danielfm: kube-proxy sets a few sysctl params at bootstrap, so I think it won't work. Besides, this is how kubeadm is doing.

core/controlplane/config/templates/cloud-config-controller
danielfm: I don't know about that, only that there are several other places using `127.0.0.1` instead of `localhost`.

What do you propose? To change everything to `localhost`?

core/controlplane/config/templates/cloud-config-controller
danielfm: Can we access the API server on the controller nodes via `localhost` _and_ use the service account cert at the same time? I don't think the certificate includes `localhost` (or `127.0.0.1`). (I'll give it a try nevertheless)

core/controlplane/config/templates/cloud-config-controller
redbaron: Just asking why this change. Maybe you hit some problem with `localhost` resolved to ipv6 ` ::1` .

core/controlplane/config/templates/cloud-config-controller
redbaron: you can volunteer to auth requests when talking to insecure APIServer port. That is what kube-controller-manager is doing: runs using serviceaccount, but uses insecure port to communicate with apiserver

core/controlplane/config/templates/cloud-config-controller
danielfm: With this change, the only place where `localhost` is being used is in the `kube-node-label.service` unit, but I personally don't have a preference. Will undo this change and deal with the 127.0.0.1 vs localhost issue in another PR, if needed.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,7 +2737,28 @@ write_files:
...
-  - path: /etc/kubernetes/controller-kubeconfig.yaml
+  # File needed on every node (used by the kube-proxy DaemonSet), including controllers
+  - path: /etc/kubernetes/kubeconfig/kube-proxy.yaml
+    content: |
+        apiVersion: v1
+        kind: Config
+        clusters:
+        - name: local
+          cluster:
+            certificate-authority: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
redbaron: 8080 is http, not https, so no need for this cert

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,7 +2737,28 @@ write_files:
...
+            server: http://localhost:8080
+        users:
+        - name: kubelet
```
redbaron: maybe call it  kube-proxy after a service this kubeconfig is for

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2711,7 +2737,28 @@ write_files:
...
+          user:
+            tokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+        contexts:
+        - context:
+            cluster: local
+            user: kubelet
```
redbaron: same, kube-proxy.

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -303,12 +304,12 @@ coreos:
         --cloud-provider=aws \
         --cert-dir=/etc/kubernetes/ssl \
         {{- if and .Experimental.TLSBootstrap.Enabled .AssetsConfig.HasTLSBootstrapToken }}
-        --experimental-bootstrap-kubeconfig=/etc/kubernetes/worker-bootstrap-kubeconfig.yaml \
+        --experimental-bootstrap-kubeconfig=/etc/kubernetes/kubeconfig/worker-bootstrap.yaml \
```
redbaron: worker bootstrap actually need SSL certs, at least ca.pem to verify  connection

core/controlplane/config/templates/cloud-config-controller
redbaron: maybe call it `default` like in worker's kubeconfig for kube-proxy

core/controlplane/config/templates/cloud-config-worker
danielfm: What is your suggestion to fix this? Re-introduce the volume mount to `/etc/kubernetes/ssl` to kube-proxy `DaemonSet`, or somehow putting the `ca.pem` file in another directory in order to mount just this file?

Either way, I think we should deprecate the old way of bootstrapping worker nodes ASAP in favor of the TLS bootstrap feature.

core/controlplane/config/templates/cloud-config-controller
redbaron: just to be clear, I am fine with either, was curious to ask why that change, but not verbose enough in my question :)

danielfm:
Done.

Conformance test results: https://scanner.heptio.com/f0c6b47099acbe4f21ee4ebe8c647588/diagnostics/

redbaron:
/lgtm

did you run cluster update with this change?

danielfm:
@redbaron Yes, I ran the update in a test cluster.

In order to allow for a smooth update, you'd need to remove the change in which we remove the `system:nodes` group from the `kube-aws:node-proxier` cluster role binding before applying the update.

By doing this, manifest-based kube-proxy pods in old nodes will continue to work as expected even though the `DaemonSet`-based kube-proxy pod will crash-loop due to the missing kubeconfig file. Then, as new nodes come up, these crash-looping kube-proxy pods will be replaced by working kube-proxy pods.

danielfm:
So, in other words, we could make this change more update-friendly by not removing the `system:nodes` group from the `kube-aws:node-proxier` cluster role binding. @redbaron So, do you think it's a good idea to do that or leave this up to the operator?

redbaron:
There is a `system:kube-proxy` user with appropriate role binding already. Maybe remove kube-aws:node-proxier all together?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1692,43 +1697,54 @@ write_files:
           apiGroup: rbac.authorization.k8s.io
 {{ end }}
 
-  - path: /etc/kubernetes/manifests/kube-proxy.yaml
+  - path: /srv/kubernetes/manifests/kube-proxy-ds.yaml
     content: |
-        apiVersion: v1
-        kind: Pod
+        apiVersion: extensions/v1beta1
```
redbaron: apps/v1beta1 :)

redbaron:
So existing one usses `/etc/kubernetes/ssl/worker.pem` therefore existing kube-proxy is `system:nodes` which we dont touch in this PR, so they continue to work fine on update, no? :)

danielfm:
I assume you are asking this considering for clusters with TLS bootstrap enabled; for other clusters, they will authenticate as the `kube-worker` user, which continue to be associated with the `kube-aws:node-proxier` role binding, so everything will work.

Now, for clusters with TLS bootstrap enabled.

Since this PR removes `system:nodes` group from the `kube-aws:node-proxier` cluster role binding, I'd expect requests coming from "legacy" kube-proxy to fail due to missing permissions (ie. list/watch endpoints).

core/controlplane/config/templates/cloud-config-controller
danielfm: Actually, according to the API reference, the `apiVersion` for the `DaemonSet` object is `extensions/v1beta1` for Kubernetes v1.7. When we upgrade to v1.8, then we'll need to change this to `apps/v1beta2` (if I read the documentation correctly).

redbaron:
> Now, for clusters with TLS bootstrap enabled.

I don't understand how TLS boostrap case used to work :(   We have https://github.com/kubernetes-incubator/kube-aws/blob/ccfd6e2dba305ed4cc5fd4411bb41a288224301a/core/controlplane/config/templates/cloud-config-worker#L927 which references https://github.com/kubernetes-incubator/kube-aws/blob/ccfd6e2dba305ed4cc5fd4411bb41a288224301a/core/controlplane/config/templates/cloud-config-worker#L971 which is created only when TLS bootstrap disabled.

So what proxy uses when TLS bootstrap is enabled then?

danielfm:
@redbaron if you specify both `--experimental-bootstrap-kubeconfig` and `--kubeconfig` (and the file in `--kubeconfig` does not exist), the kubelet sends a CSR request using the bootstrap kubeconfig (the one with the bootstrap token), and generates the `--kubeconfig` file with the bootstrapped certificates, which is the kubeconfig file used from this point forward when talking to the apiserver.

redbaron:
This is awesome, thanks for your patience :)

Issue #952, opened 2017-09-22
Dashboard 1.7.0


Vrtak-CZ:
https://github.com/kubernetes/dashboard/releases/tag/v1.7.0

Vrtak-CZ:
ref: https://github.com/kubernetes-incubator/kube-aws/pull/962

camilb:
Closed in #971

PR #971, opened 2017-10-10
Kubernetes Dashboard 1.7 with RBAC and authentication.

camilb:
Ref. #952


redbaron:
OK, continuing Dashboard discussion. 

Dashboard change default mode of operation where it starts with limited permissions and lets user to paste token or full kubeconfig to act in users behalf.

It is quite a difference from previous use of Dashboard where it wsa running with admin privileges. My suggestion is to give dashboard account full privileges as it had before and run updated dashboard. So we merge this PR.


Next, if @camilb feels like running Dashboard with limited privileges is a feature worth having, we can introduce flag in cluster.yaml where we'd allow users to choose which mode to run dasbhoard mode: admin or minimal. That flag will control content of RBAC policy installed for Dashboard. How does it sound to you?

camilb:
Hi @redbaron updated the PR with default admin privileges, that can be changed in  `cluster.yaml` by setting the `kubeDashboardAdminPrivileges` flag to false.

mumoshu:
Thanks as always for your contribution 👍

c-knowles:
Does anyone remember why `resources` were removed here?

Issue #957, opened 2017-09-29
Start deploying metric-server for 1.8 Kubernetes 


redbaron:
From 1.8 release notes:

> metrics-server is now the recommended way to provide the resource metrics API. Deploy metrics-server as an add-on in the same way that you deploy Heapster.

mumoshu:
Addressed in #973 #1035 #1036

PR #973, opened 2017-10-10
Metrics Server addon.

camilb:
More info here:
https://kubernetes.io/docs/tasks/debug-application-cluster/core-metrics-pipeline/

Ref. #957


danielfm:
What about heapster? Can this add-on run alongside heapster in such a way we can migrate to it in a safe manner?

camilb:
Didn't have time to investigate it further, made a quick test in parallel with heapster in k8s 1.8 and works fine. `kubectl top` is one of the components that should use the new metrics-server but it didn't work without heapster.

mumoshu:
@camilb Thanks for the contribution 👍 
This is conflicting with your latest work which has just been merged 😢 
Otherwise, LGTM.

mumoshu:
@camilb Thanks for contributing!

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2578,6 +2654,89 @@ write_files:
...
+          template:
+            metadata:
+              name: metrics-server
+              labels:
+                k8s-app: metrics-server
+            spec:
+              serviceAccountName: metrics-server
+              containers:
+              - name: metrics-server
+                image: gcr.io/google_containers/metrics-server-amd64:v0.2.0
```
redbaron: shouldn't image be templated from  a value from cluster.yaml?

camilb: yes, fixed in #1035

PR #975, opened 2017-10-11
Rollback to 1.7.7 hyperkube image, as 1.7.8 is actually 1.9.0 alpha

redbaron:
See for details: https://github.com/coreos/kubernetes/issues/171


redbaron:
Emergency merge

redbaron:
It happens. What we  can learn is that cfn-signal should check for kube-proxy health

PR #985, opened 2017-10-17
GPU Build Fix + Adding G3 and P3 Instances

kylegato:
This PR fixes the build process for building NVIDIA drivers. It also adds support for P3 and G3 instances.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -1040,7 +1040,7 @@ write_files:
       set -e
       function is_gpu_enabled(){
         local instance_type=$1
-        [[ -n $instance_type ]] && ([[ $instance_type == p2* ]] || [[ $instance_type ==  g2* ]])
+        [[ -n $instance_type ]] && ([[ $instance_type == p2* ]] || [[ $instance_type ==  g2* ]] || [[ $instance_type == g3* ]])
       }
 
       INSTANCE_TYPE=$(curl -s http://169.254.169.254/latest/meta-data/instance-type)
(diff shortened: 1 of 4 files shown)


core/controlplane/config/templates/cloud-config-worker
```diff
@@ -1099,14 +1099,27 @@ write_files:
...
   - path: /opt/nvidia-build/_container_build.sh
     owner: root:root
     permissions: 0755
     content: |
       #!/bin/sh
 
       # Default: use binary packages instead of building everything from source
-      EMERGE_SOURCE_FLAGS=gK
+      # Default: build everything from source.
+      #EMERGE_SOURCE_FLAGS=gK
```
mumoshu: What pros and cons would there be by building everything from source instead of using binary packages?
Probably it now takes a lot more time for a GPU-enabled worker node to be up and running, at least?

mumoshu:
@kylegato Thanks for your contribution 👍 
I've left a question but this looks good to me overall.

core/controlplane/config/templates/cloud-config-worker
kylegato: I was getting 403s when trying to grab the binary packages from the repos. At least with building from source it's consistent for me, in my testing.

mumoshu: Ah, thanks for investigating it. Then, this change LGTM 👍 

Btw, would you mind sharing me the full log messages of the installation process which helps me which files are getting 403s, perhaps in a gist?
So that I can probably try to contact someone maintains the files to see what's going on.

kylegato: @mumoshu Sure, here you are: https://gist.github.com/kylegato/689a366b62f744702b634e710cd4cc76

mumoshu: @kylegato I've reported the 403s issue to https://github.com/coreos/bugs/issues/2207 and it seems to have fixed now. In nutshell, it wasn't a planned or expected change.

Sorry for the back and forth but would you be willing to revert only the build fix part of this PR so that we can still benefit from prebuilt binaries for faster node startup?

Also, thank you very much for locating the issue 👍

kylegato:
Given the referenced issues, this PR looks like it may not be needed anymore @mumoshu

mumoshu:
@kylegato The half part of this PR - G3 instances support - would still remain effective though!

kylegato:
@mumoshu Updated PR, and added support for P3 instances.

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -1391,6 +1391,8 @@ write_files:
 
       mkdir -p /etc/ld.so.conf.d/ 2>/dev/null
       echo "/opt/nvidia/current/lib64" > /etc/ld.so.conf.d/nvidia.conf
+      rm /opt/nvidia/current/lib64/libEGL.so.1
```
everpeace: Hi! Thanks for supporting new GPU instance types!! (I'm the original contributor of this feature.)

This `rm` and `ln` would be some workaround for [this bug](https://bugs.launchpad.net/ubuntu/+source/nvidia-graphics-drivers-375/+bug/1662860).

I think referring the bug in comments might help to future contributors because the bug will be fixed in the near future.

What do you think?

mumoshu: @everpeace Thanks for the suggestion 👍 Seems like so to me - I will add a comment anyway.
@kylegato Please correct me if I was wrong!

kylegato: Sounds good to me, nice find on that bug @everpeace

PR #986, opened 2017-10-18
Fixes #976 - kube-proxy no longer supports command line flags.

### LINKED ISSUE #976: kube-proxy no longer supports *any command line options -- support kube 1.8
We need to move --cluster-cidr into a config file for kube-proxy.  With kube 1.8, the command line options are fully depricated and the error is observed:
```
kube-proxy-wzccp kube-proxy W1012 17:28:13.637563       1 server.go:191] WARNING: all flags other than --config, --write-config-to, and --cleanup are deprecated. Please begin using a config file ASAP.
```

current command line is:
```
/hyperkube proxy --kubeconfig=/etc/kubernetes/kubeconfig/kube-proxy.yaml --cluster-cidr=10.2.0.0/16
```

Reference issue: https://github.com/kubernetes/kubernetes/issues/50041  I'm still looking for an example config, we can output one with kube proxy --write-config-to ...

jeremyd:


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -760,6 +760,7 @@ write_files:
 
       # Configmaps
       kubectl apply -f "${mfdir}/kube-dns-cm.yaml"
+      kubectl apply -f "${mfdir}/kube-proxy-config.yaml"
```
danielfm: Can you use a style more consistent with the other objects, like this:

```bash
# ConfigMaps
for manifest in {kube-dns,kube-proxy}; do
    kubectl apply -f "${mfdir}/$manifest-cm.yaml"
done
```
.. and use the `-cm` suffix in the ConfigMap file, instead of `-config`?

danielfm:
Also, it would be nice if you could run the conformance tests in a cluster that contains this change; you could use [Heptio Sonobuoy](https://scanner.heptio.com) for that.

core/controlplane/config/templates/cloud-config-controller
jeremyd: Good idea, changed.

jeremyd:
https://scanner.heptio.com/0224690663e4b33302e3771e9252101b/  test results from sonobuoy for this PR.  What a neat tool!  thanks for recommending that, another test I can use on my clusters :+1:

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1799,18 +1816,23 @@ write_files:
                 command:
                 - /hyperkube
                 - proxy
-                - --kubeconfig=/etc/kubernetes/kubeconfig/kube-proxy.yaml
-                - --cluster-cidr={{.PodCIDR}}
+                - --config=/etc/kubernetes/kube-proxy/kube-proxy-config.yaml
```
mumoshu: Very nit but I'd prefer `kube-proxy/config.yaml` for less repeating.
If you could agree, I'd appreciate it if you could address it when you work on the config afterwards.. probably for the ipvs support? 😉

mumoshu:
@jeremyd Thanks as always for your contribution 👍

PR #988, opened 2017-10-18
Add setting for worker nodePools to specify --max-pods.

jeremyd:
I found this option useful in a testing cluster.  Normally there is a hard limit of 100 pods per node.  If you're using very small resource limits you can easily fit more pods than 100 on a node.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-worker
+++ b/core/controlplane/config/templates/cloud-config-worker
@@ -296,6 +296,9 @@ coreos:
         {{end}}--register-node=true \
         {{if .Taints}}--register-with-taints={{.Taints.String}}\
         {{end}}--allow-privileged=true \
+        {{- if .MaxPods }}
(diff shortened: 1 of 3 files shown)


mumoshu:
@jeremyd Thanks for your contribution 👍 
This looks nice. Would you also mind adding a corresponding comment in the cluster.yaml template so that our users can see this feature exists.

jeremyd:
Thanks for the feedback @mumoshu.  I added the setting to the cluster.yaml and also changed its type to int32 in the cluster.yaml to make it more user-friendly.

redbaron:
This is *really* unpractical to have a flag for each possible kubelet flag. we already have $KUBELET_OPTS there, which oyu can populate from cluster.yaml

jeremyd:
Oh, I did not realize that.  Ok thanks!  We don't need this then..

PR #993, opened 2017-11-02
Fix max iops typo

Vrtak-CZ:
Max IOPS is `20000` not `2000` according to https://github.com/kubernetes-incubator/kube-aws/blob/master/vendor/github.com/aws/aws-sdk-go/models/apis/autoscaling/2011-01-01/api-2.json#L970 and https://aws.amazon.com/ebs/details/


mumoshu:
@Vrtak-CZ Thanks as always for your contribution!

PR #995, opened 2017-11-03
Only take snapshot if current member is a leader

martinssipenko:


mumoshu:
@martinssipenko Thanks for your contribution 👍
Just confirming but how did you test this change?
Would you like me to test this in specific cases (if any)?

mumoshu:
I'm going to merge anyway beucase the code LGTM.

PR #997, opened 2017-11-03
Load member .env file that is generated by etcdadm.

martinssipenko:


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -269,6 +269,7 @@ coreos:
 
             [Service]
             EnvironmentFile=-/etc/etcd-environment
+            EnvironmentFile=-/var/run/coreos/etcdadm/etcd-member.env
```
mumoshu: @martinssipenko Hi, thanks for contributing!
Would you mind sharing me your intention to add this line?
Seemed like this env file is read from nowhere.

martinssipenko: This file is generated by etcdadm

mumoshu: Yes! However, I thought we already feed it to etcd somewhere else?
Where is the env file used from when written here?

martinssipenko:
Its mever being fed to etcd, at least i could not find a place where it does. There was an old open pull request that roughly did the same, but it seems to be abandoned, i took few parts from the old PR.

Martins

mumoshu:
@martinssipenko Thanks for clarifying! Yes, you are correct. I was missing the fact that the newly added EnvironmentFile statement was inside the drop-in for the etcd-member.service. So it might work, of course!

mumoshu:
Thanks for contributing :bow:

PR #1002, opened 2017-11-07
Experimental addons can all be uncommented

c-knowles:
They all have `enabled` flags so uncommenting them means:
- Highlighting in IDE/text editor works right away and we can easily spot validation problems (current indentation is actually wrong in our comments)
- Easier to diff between newly generate cluster.yaml and existing ones in source control as not hundreds of different lines changes, just some boolean values. Changes to default values are more obvious.
- We don't have comments in comments

We could extend this pattern to other properties. Empty strings could mean off/disabled and other defaults could be uncommented rather than spread here and throughout the code base.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1172,78 +1172,88 @@ addons:
     enabled: false
 
 # Experimental features will change in backward-incompatible ways
-# experimental:
-#   # Enable admission controllers
-#   admission:
-#     podSecurityPolicy:
-#       enabled: true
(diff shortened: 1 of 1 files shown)


mumoshu:
I think this would provide more meaningful diffs between customized vs default cluster.yaml when upgrading kube-aws when combined with the git workflow described in #462 
cc @redbaron @danielfm @camilb

c-knowles:
@mumoshu at least a few of us are in favour of this, shall we merge it?

mumoshu:
@c-knowles Yes, sorry for the silence!

PR #1004, opened 2017-11-07
Add kube-proxy health check

dvdthms:
We had noticed that there wasn't a check for the kube-proxy healthz endpoint in the cfn-signal service.

We hit an issue where the cfn-signal was firing, but the proxy was not healthy (it could not talk to the API-server). We've added this check to make sure the proxy is healthy before allowing the cfn-signal to fire.

Out of interest -- was there a reason this check wasn't included in the command before?

The port used for the kube-proxy healthz can be seen here: https://github.com/kubernetes/kubernetes/blob/v1.7.0/pkg/master/ports/ports.go#L41-L43


mumoshu:
> was there a reason this check wasn't included in the command before?

No. Thanks for the good catch!

PR #1008, opened 2017-11-14
Add some additional cleanup to rkt containers

dvdthms:
Most of the services that use rkt containers had a command to perform some cleanup.

I've added the corresponding command to a service that was missing it, and also added some additional syntax to pre-existing commands in order to keep them consistent.


mumoshu:
@dvdthms Thanks for your contribution as always!

PR #1009, opened 2017-11-15
Adding AlwaysPullImages admission controller option

kylegato:
This adds an experimental option to enable the "AlwaysPullImages" admission plugin.

Per https://kubernetes.io/docs/admin/admission-controllers/#alwayspullimages as of now:

> This plug-in modifies every new Pod to force the image pull policy to Always. This is useful in a multitenant cluster so that users can be assured that their private images can only be used by those who have the credentials to pull them. Without this plug-in, once an image has been pulled to a node, any pod from any user can use it simply by knowing the image’s name (assuming the Pod is scheduled onto the right node), without any authorization check against the image. When this plug-in is enabled, images are always pulled prior to starting containers, which means valid credentials are required.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -41,6 +41,9 @@ func NewDefaultCluster() *Cluster {
 			PodSecurityPolicy{
 				Enabled: false,
 			},
+			AlwaysPullImages{
+				Enabled: false,
+			},
 			DenyEscalatingExec{
 				Enabled: false,
 			},
@@ -520,8 +523,13 @@ type Experimental struct {
 
 type Admission struct {
 	PodSecurityPolicy  PodSecurityPolicy  `yaml:"podSecurityPolicy"`
+	AlwaysPullImages   AlwaysPullImages   `yaml:"alwaysPullImages"`
 	DenyEscalatingExec DenyEscalatingExec `yaml:"denyEscalatingExec"`
 	Initializers       Initializers       `yaml:"initializers"`
+}
+
+type AlwaysPullImages struct {
+	Enabled bool `yaml:"enabled"`
 }
 
 type PodSecurityPolicy struct {
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1177,6 +1177,10 @@ experimental:
   admission:
     podSecurityPolicy:
       enabled: false
+    # alwaysPullImages Note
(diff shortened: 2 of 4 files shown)


core/controlplane/config/config.go
```diff
@@ -520,10 +523,15 @@ type Experimental struct {
 
 type Admission struct {
 	PodSecurityPolicy  PodSecurityPolicy  `yaml:"podSecurityPolicy"`
+	AlwaysPullImages   AlwaysPullImages   `yaml:"alwaysPullImages"`
 	DenyEscalatingExec DenyEscalatingExec `yaml:"denyEscalatingExec"`
 	Initializers       Initializers       `yaml:"initializers"`
 }
 
+type AlwaysPullImages struct {
+	Enabled bool `yaml:"enabled"`
```
camilb: I suggest leaving this disabled by default. Had one issue when the docker registry could not be accessed and some health checks failed, making it impossible to restart the pods. Also, when a health check fails, it takes more to restart.

kylegato: @camilb Enabled is a true/false field and believe that struct is simply defining Enabled as a bool to be based off of the yaml value, I don't believe I'm telling it to enable it by default. If you look at the other files, and the cluster.yaml template config, I'm specifying it as false by default.

camilb: Sorry, I was looking on the wrong field.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -1177,6 +1177,8 @@ experimental:
   admission:
     podSecurityPolicy:
       enabled: false
+    alwaysPullImages:
```
mumoshu: Would you mind adding a comment to suggest users to ensure that docker registries are reliable enough, and warn about otherwise the issue like @camilb [noted](https://github.com/kubernetes-incubator/kube-aws/pull/1009#discussion_r151197787) could happen?

What I'd expect would be:

```
Recommended to turn this on when and only when docker registries are reliable enough. Otherwise automatic pod restarts can be delayed due to temporally docker registry outages.
Please see https://github.com/kubernetes-incubator/kube-aws/pull/1009#discussion_r151197787 for more info.
```

// Please feel free to correct/improve any aspect of the sentences. I'm not a good writer 😄

kylegato: Sounds good to me. Added.

mumoshu:
LGTM. Thanks for contributing!

Issue #1013, opened 2017-11-20
Cluster TLS certs/keys rotation


mumoshu:
@redbaron @c-knowles @danielfm I'm still scratching my head on how to achieve this safely and automatically.

For now, I'm recreating clusters until the certs expire but that's painful!

In general, we'd want to replace all the certs/keys targeted for rotation across all the nodes very fast, without replacing the node one by one just for cert/key update, right?

// Please see https://github.com/kubernetes/kubernetes/issues/25379#issuecomment-281533623 for more context.

Otherwise, delay in update process could result in downtime. For example, when apiservers w/ old key/cert on their nodes got restarted due to some reasons after etcd-server key/cert rotation on etcd nodes but before etcd-client key rotation on controller nodes, each restarted apiserver will fail connecting etcd with old key.

Related topics:

- For kubelets we can rely on TLS bootstraping. See #922.
- For etcd-ca, #907

mumoshu:
I'm not really wanting to introduce an another moving-part to kube-aws but would it be feasible to utilize aws-ssm-agent or somethig for fast key/cert rotation?

c-knowles:
I think for rotation the minimum requirement is for the target system to support two public keys as [noted here](https://github.com/kubernetes/kubernetes/issues/25379#issuecomment-281534343). We could use that facility and introduce a new kube-aws command or deal with it automatically in each `update` call?

cheungpat:
I found myself needing to rotate credentials today which is really painful.

The `--service-account-key-file` option accepts public and private key files, perhaps we can keep previously generated apiserver cert files and use them for this option? Currently this option is set to `apiserver-key.pem` which is recreated each time `kube-aws` render credentials.

I think we can do this as part of `kube-aws render credentials`, but I think it is also okay to do this in a new command.

mumoshu:
@cheungpat Thanks a lot for reminding that flag! I have been forgotten about that since https://github.com/kubernetes-incubator/kube-aws/issues/907#issuecomment-345665123

Yes, it would work.

Regarding the command, something like `kube-aws rotate service-account-key` would be cool but i'm unsure whats a realistic option from implementors perspective, as you need multiple steps to complete a rotation, even with the flag.

1. Add a new private and public key. The public key is appended to the pem file passed to service-account-key-file.
2. Ensure that old service account tokens are deleted and regenerated(any way to automate this process?
3. Remove the old pub key from the pem file.

[...]

PR #1014, opened 2017-11-20
Fix the render command to not wipe ca.pem

### LINKED ISSUE #1003: render credentials sometimes wipes ca.pem on v0.9.9-rc.2
Steps:
* `init` a new cluster render
* Copy `ca.pem` and `ca-key.pem` into the credentials folder for this new cluster
* Run `render credentials`, all creds generated as expected
* Run `render credentials` again, for example after changing some LB config
* `ca.pem` is now empty

mumoshu:
Fixes #1003 

@c-knowles Would this work for you?

### DIFF
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -440,39 +440,46 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, caKeyRequiredO
 }
 
 func (r *RawAssetsOnMemory) WriteToDir(dirname string, includeCAKey bool) error {
+	workerCAKeyDefaultSymlinkTo := ""
+	if includeCAKey {
+		workerCAKeyDefaultSymlinkTo = "ca-key.pem"
+	}
+
 	assets := []struct {
-		name      string
-		data      []byte
-		overwrite bool
+		name             string
+		data             []byte
+		overwrite        bool
+		ifEmptySymlinkTo string
 	}{
-		{"ca.pem", r.CACert, true},
-		{"worker-ca.pem", r.WorkerCACert, true},
-		{"worker-ca-key.pem", r.WorkerCAKey, true},
-		{"apiserver.pem", r.APIServerCert, true},
-		{"apiserver-key.pem", r.APIServerKey, true},
-		{"worker.pem", r.WorkerCert, true},
-		{"worker-key.pem", r.WorkerKey, true},
-		{"admin.pem", r.AdminCert, true},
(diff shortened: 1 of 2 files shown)


mumoshu:
Squashed a fix for an easy mistake. Please reload to see the latest commit id if anyone has already seen it.

core/controlplane/config/encrypted_assets.go
```diff
@@ -440,39 +440,46 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, caKeyRequiredO
...
+	}
+
 	assets := []struct {
-		name      string
-		data      []byte
-		overwrite bool
+		name             string
+		data             []byte
+		overwrite        bool
+		ifEmptySymlinkTo string
```
c-knowles: Does this mean if populated, create a symlink?

mumoshu:
Squashed again. Seems to work so far.

core/controlplane/config/encrypted_assets.go
mumoshu: If we had nothing to write, default to create a symlink to denoted by the `ifEmptySymlinkTo`.
Basically, we have nothing to write for `worker-ca.pem` as long as the user explicitly overwrite or put the file before running `kube-aws render credentials`. So in that case, create a symlink to `ca.pem`(=the same as the previous behavior).

mumoshu:
I've manually verified this to work by running `kube-aws render` to regenerate credentials and then updating my cluster with `kube-aws update`.
The cfn stack update was successful while kube-dashboard went into CrashLookBackOff probably due to the rotation of the service account signing key.

```
$ k logs kubernetes-dashboard-7f46664f8f-n7gm4
2017/11/21 13:50:43 Starting overwatch
2017/11/21 13:50:43 Using in-cluster config to connect to apiserver
2017/11/21 13:50:43 Using service account token for csrf signing
2017/11/21 13:50:43 No request provided. Skipping authorization
2017/11/21 13:50:43 Error while initializing connection to Kubernetes apiserver. This most likely means that the cluster is misconfigured (e.g., it has invalid apiserver certificates or service accounts configuration) or the --apiserver-host param points to a server that does not exist. Reason: the server has asked for the client to provide credentials
Refer to the troubleshooting guide for more information: https://github.com/kubernetes/dashboard/blob/master/docs/user-guide/troubleshooting.md
```

However, this issue is not yet resolved in kube-aws - see #1013 for more info.

mumoshu:
I'll merge this once someone put LGTM on it

PR #996, opened 2017-11-03
Bump kubernetes version to 1.8.2.

camilb:
Tests from yesterday: https://scanner.heptio.com/4024f60c6b1535881b81328e952b4404/diagnostics/

PR #991, opened 2017-10-24
Check to ensure new API servers can receive communication from workers

dvdthms:
An issue was encountered where RBAC was turned on in cluster and the new controllers started rejecting all communication from the 'old' workers.

The controllers have no check for problems like this, so their cfn-signal returned 'true' to cloud-formation, which triggered the update of the worker nodepools.

The controllers should not be considered healthy/deployed-successfully if a new apiserver instance cannot receive communication from the existing workers.

PR #1015, opened 2017-11-21
 Fix kubelet-to-apiserver connection checks on controller nodes not to fail in certain cases

mumoshu:
- check-worker-connection was failing when audit logging is disabled
- In addition to that, it was failing since K8S v1.8 due to the change in K8S that no audits are logged when missing an audit policy file

Follow-up for #991 and #996 
Fix for https://github.com/kubernetes-incubator/kube-aws/pull/991#commitcomment-25459898
Depends on #1014

### DIFF
--- a/core/controlplane/config/credential.go
+++ b/core/controlplane/config/credential.go
@@ -80,7 +80,10 @@ func EncryptedCredentialCacheFromRawCredential(raw *RawCredentialOnDisk, bytesEn
 }
 
 func RawCredentialFileFromPath(filePath string, defaultValue *string) (*RawCredentialOnDisk, error) {
-	if _, err := os.Stat(filePath); os.IsNotExist(err) && defaultValue != nil {
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		if defaultValue == nil {
+			return nil, fmt.Errorf("%s must exist. Please confirm that you have not deleted the file manually", filePath)
+		}
 		if err := ioutil.WriteFile(filePath, []byte(*defaultValue), 0644); err != nil {
 			return nil, err
 		}
@@ -102,6 +105,9 @@ func (c *RawCredentialOnDisk) Fingerprint() string {
 }
 
 func (c *RawCredentialOnDisk) Persist() error {
+	if len(c.content) == 0 {
+		return fmt.Errorf("%s is going to be empty. Maybe a bug", c.filePath)
+	}
 	if err := ioutil.WriteFile(c.filePath, c.content, 0600); err != nil {
 		return err
 	}
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -390,7 +390,9 @@ coreos:
         {{ if .UseCalico }}
         ExecStartPre=/usr/bin/bash -c "until /usr/bin/docker run --net=host --pid=host --rm {{ .CalicoCtlImage.RepoWithTag }} node status > /dev/null; do sleep 3; done && echo Calico running"
         {{ end }}
+        {{if .Experimental.AuditLog.Enabled -}}
         ExecStartPre=/opt/bin/check-worker-communication
+        {{end -}}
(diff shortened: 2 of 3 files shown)


core/controlplane/config/credential.go
```diff
@@ -80,7 +80,10 @@ func EncryptedCredentialCacheFromRawCredential(raw *RawCredentialOnDisk, bytesEn
 }
 
 func RawCredentialFileFromPath(filePath string, defaultValue *string) (*RawCredentialOnDisk, error) {
-	if _, err := os.Stat(filePath); os.IsNotExist(err) && defaultValue != nil {
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		if defaultValue == nil {
+			return nil, fmt.Errorf("%s must exist. Please confirm that you have not deleted the file manually", filePath)
+		}
```
danielfm: It seems like the contents of #1014 got mixed in here as well.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2942,6 +2951,109 @@ write_files:
...
+        # so only log at the Metadata level.
+        - level: Metadata
+          resources:
+            - group: "" # core
+              resources: ["secrets", "configmaps"]
+            - group: authentication.k8s.io
+              resources: ["tokenreviews"]
+          omitStages:
+            - "RequestReceived"
+        # Get repsonses can be large; skip them.
```
danielfm: Typo in `repsonses`.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2942,6 +2951,109 @@ write_files:
...
+        - level: Request
+          verbs: ["get", "list", "watch"]
+          resources: ${known_apis}
```
danielfm: We are not substituting the `known_apis` variable, like it's done [here](https://github.com/kubernetes/kubernetes/blob/v1.8.3/cluster/gce/gci/configure-helper.sh#L497).

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -3092,20 +3204,32 @@ write_files:
...
 
       queryserver() {
         echo "Checking to see if workers are communicating with API server."
-        docker logs --tail 10 ${DOCKERIMAGE} |& grep -v 127.0.0.1 | awk 'BEGIN {C=1} /kubelet\// && $8 == "200" {C=0} END {exit C}'
+        auditlogs | grep -v 127.0.0.1 | awk 'BEGIN {C=1} /kubelet\// && $8 == "200" {C=0} END {exit C}'
+      }
+
+      auditlogs() {
+        if [ "$AUDIT_LOG_PATH" == "/dev/stdout" ]; then
+          docker logs --since 11s ${DOCKERIMAGE} |& cat
```
danielfm: Is there a reason behind this magic number here (`11s`)?

mumoshu: `queryserver` is called every 10+ seconds below. I wanted to gather logs for periods slightly longer than that not to drop any lines(hopefully)

core/controlplane/config/templates/cloud-config-controller
mumoshu: Thanks! Manually substituted them anyway 😄

core/controlplane/config/templates/cloud-config-controller
mumoshu: Good catch! Fixed.

core/controlplane/config/credential.go
mumoshu: Thanks for pointing that out!
Sorry but I had no time to resolving possible conflicts and testing them in isolations.
Marking this to depend on #1014 anyway.

mumoshu:
I've manually verified this to work by updating my cluster few times:

- From no auditing to auditing enabled while setting logpath of /dev/stdout
- And then to /var/log/apiserver-audit.log

mumoshu:
@danielfm Thanks for reviewing!

PR #1017, opened 2017-11-21
fixes #990 disable sshd password authentication

### LINKED ISSUE #990: Disable sshd password authentication by default
https://github.com/coreos/bugs/issues/2114

jeremyd:
Tested and working.  This is the same defaults for sshd_config taken from the upstream CoreOS alpha with the addition of:

```
PasswordAuthentication no
ChallengeResponseAuthentication no
```


mumoshu:
@jeremyd Thanks for contributing as always!

PR #1021, opened 2017-11-22
Make kube-node-label more reliable

mumoshu:
kube-node-label now retries until at most 10 seconds pass.
It is 7 seconds longer for controllers and 10 seconds longer for workers.

Resolves #866

cc @danielfm This is a fix for the issue you've reported before.

PR #1023, opened 2017-11-22
Add support for S3 buckets with the default encryption enabled

### LINKED ISSUE #831: Provide support for encrypted State buckets
Currently kube-aws gets a 403 when using a bucket with encryption enforced. Should be able to support buckets with encryption policy (KMS/AES256)

mumoshu:
Resolves #831
Depends on #1021 

Please see [the second commit](https://github.com/kubernetes-incubator/kube-aws/pull/1023/commits/88dd6511a1d81cc4583c7c0e5ef6587102599109) for reviewing changes independently from the PR this is dependent on.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -5,14 +5,21 @@
 export COREOS_PRIVATE_IPV4 COREOS_PRIVATE_IPV6 COREOS_PUBLIC_IPV4 COREOS_PUBLIC_IPV6
 REGION=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r '.region')
 USERDATA_FILE=userdata-controller
-while ! /usr/bin/rkt run \
-   --net=host \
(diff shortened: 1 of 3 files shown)


mumoshu:
I've verified this to work by updating my cluster with `kube-aws update --s3-uri s3://my-bucket-with-awskms-as-default-encryption` and then `kube-aws update --s3-uri s3://my-bucket-with-aes256-as-default-encryption`.

FYI, S3 buckets can be configured with the default encryption of `awskms` by e.g. checking this option on the console:

image

The "AES-256" option on the console corresponds to `AES256`.

Even though kube-aws doesn't explicitly set any value for the `x-amz-server-side-encryption` header, the S3 objects uploaded by kube-aws were encrypted as expected:

image

danielfm:
Will unencrypted buckets continue to work after this change? Did you test that scenario as well?

mumoshu:
@danielfm Thanks for pointing that out! I just tested that scenario and everything went well.
Basically, we can switch unencrypted/encrypted buckets anytime by running `kube-aws update` with different `s3-uri`s.

mumoshu:
@danielfm Thanks for reviewing!

PR #1024, opened 2017-11-22
Fix the default FleetIamRole

### LINKED ISSUE #1022: The default IamFleetRole changed on AWS-side?
kube-aws: v0.9.9-rc.3, but any version would be affected

I've just added a spot-fleet based node pool to my cluster and it failed like this:

image

The value for the `IamFleetRole` property was:

```
"IamFleetRole":{"Fn::Join":["",["arn:aws:iam::",{"Ref":"AWS::AccountId"},":role/aws-ec2-spot-fleet-role"]]}
```

which had certainly worked before.

As far as I remember, `aws-ec2-spot-fleet-role` was taken from the "Request Spot Instances" on AWS console because it was the default role at that time.
However, interestingly the default shown today is `aws-ec2-spot-fleet-tagging-role`.

image

Can we just change the default to the new role?
Does your AWS account shows the new role as the default one once you've browsed to the "Request Spot Instances" pane?

mumoshu:
Closes #1022

Depends on #1023
Please see [the third commit](https://github.com/kubernetes-incubator/kube-aws/pull/1024/commits/e15af51297f2edad5736d014e56a95920ac7ec3a) for reviewing changes independently from the PR this is dependent on.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -5,14 +5,21 @@
 export COREOS_PRIVATE_IPV4 COREOS_PRIVATE_IPV6 COREOS_PUBLIC_IPV4 COREOS_PUBLIC_IPV6
 REGION=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r '.region')
 USERDATA_FILE=userdata-controller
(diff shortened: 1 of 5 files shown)


mumoshu:
I've manually verified this to work by adding a spot-fleet based node pool to my cluster.

PR #1027, opened 2017-11-22
Support EC2 instance tags per node role

### LINKED ISSUE #1026: Tagging controller, etcd and worker nodes differently
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/901#issuecomment-346352317

It will be handy if your monitoring tools discovers EC2 instances and then groups resource metrics with EC2 instance tags.

cc @c-knowles

mumoshu:
@c-knowles Would this be what you had needed?

This feature will be handy when e.g. your monitoring tools discovers EC2 instances and then groups resource metrics with EC2 instance tags.

Changes:
- Add support for stackTags to SpotFleet-based node pool
- Add a new configuration key `instanceTags` to cluster.yaml per `worker.nodePools[]`, `controller` and `etcd`.

I recomment to use `stackTags` for adding cluster-wide metadata to all the resources managed by Cfn, whereas `instanceTags` are by definition used for node-group-wide metadata to EC2 instances(node pool name, node role, and anything configurable per node pool for example).

A typical cluster.yaml using both `stackTags` and `instanceTags` would look like:

```yaml
worker:
  nodePools:
  - name: pool1
     instanceTags:
       myrole: worker
       type: ondemand
     # Propagated to (hopefully, it's up to cfn) all the stack resources for the pool1 stack
     stackTags:
       env: prod
  - name: pool2
     spotFleet:
       targetCapacity: 2
     # Propagated to EC2 instances managed by this spot fleet (via tag-spot-instance.service)
     instanceTags:
       myrole: worker
       type: spot
     # Propagated to (hopefully, it's up to cfn) all the stack resources for the pool2 stack
     stackTags:
       env: prod

controller:
  instanceTags:
    myrole: controller
    type: ondemand
  # invalid as controller nodes don't have a dedicated cfn stack
  # stackTags:

etcd:
  instanceTags:
    myrole: etcd
    type: ondemand
  # invalid as etcd nodes don't have a dedicated cfn stack
  # stackTags:

stackTags:
  env: prod
```

Resolves #1026
Depends on #1024 in order to test in combination with spot fleet support

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -164,6 +164,10 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #  # CAUTION: Don't use t2.micro or the cluster won't work. See https://github.com/kubernetes/kubernetes/issues/18975
 #  instanceType: t2.medium
 #
+#  # EC2 instance tags for controller nodes
+#  instanceTags:
+#    instanceRole: controller
+#
 #  rootVolume:
 #    # Disk size (GiB) for controller node
 #    size: 30
@@ -342,6 +346,10 @@ worker:
 #      # Instance type for worker nodes
 #      # CAUTION: Don't use t2.micro or the cluster won't work. See https://github.com/kubernetes/kubernetes/issues/16122
 #      instanceType: t2.medium
+#
+#      # EC2 instance tags for worker nodes
+#      instanceTags:
+#        instanceRole: worker
 #
 #      rootVolume:
 #        # Disk size (GiB) for worker nodes
@@ -576,6 +584,10 @@ worker:
 #
 #  # Instance type for etcd node
 #  instanceType: t2.medium
+#
+#  # EC2 instance tags for etcd nodes
+#  instanceTags:
+#    instanceRole: etcd
 #
 #  rootVolume:
 #    # Root volume size (GiB) for etcd node
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -397,7 +397,7 @@ worker:
 #        # IAM role to grant the Spot fleet permission to bid on, launch, and terminate instances on your behalf
 #        # See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-fleet-requests.html#spot-fleet-prerequisites
 #        #
-#        # Defaults to "arn:aws:iam::youraccountid:role/aws-ec2-spot-fleet-role" assuming you've arrived "Spot Requests" in EC2 Dashboard
+#        # Defaults to "arn:aws:iam::youraccountid:role/aws-ec2-spot-fleet-tagging-role" assuming you've arrived "Spot Requests" in EC2 Dashboard
 #        # hence the role is automatically created for you
 #        iamFleetRoleArn: "arn:aws:iam::youraccountid:role/kube-aws-doesnt-create-this-for-you"
 #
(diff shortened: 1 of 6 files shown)


mumoshu:
An example capture of spot instance tags:

image

```yaml
worker:
  nodePools:
  - name: spot
     spotFleet:
       targetCapacity: ...
     stackTags:
       Environment: Production
     instanceTags:
       myrole: worker
       mytype: spot
```

c-knowles:
@mumoshu wow, this is great! Let's merge it asap. Maybe it's more than most people will need but the flexibility will be good.

What's the use case for `worker[*].stackTags` when there is also a root `stackTags` applicable to everything plus `worker[*].instanceTags` applicable to each node pool?

I was looking through the code to see why we hadn't added much about stack tags here and I found some `Tags` methods [here](https://github.com/kubernetes-incubator/kube-aws/blob/0dc583e21ea22dffc42ddc8d1c306ae3499d6305/core/root/template_params.go#L57) and [here](https://github.com/kubernetes-incubator/kube-aws/blob/0dc583e21ea22dffc42ddc8d1c306ae3499d6305/core/root/template_params.go#L91). Could those be reused or maybe they are now deprecated?

core/controlplane/config/templates/cluster.yaml
```diff
@@ -397,7 +405,7 @@ worker:
 #        # IAM role to grant the Spot fleet permission to bid on, launch, and terminate instances on your behalf
 #        # See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-fleet-requests.html#spot-fleet-prerequisites
 #        #
-#        # Defaults to "arn:aws:iam::youraccountid:role/aws-ec2-spot-fleet-role" assuming you've arrived "Spot Requests" in EC2 Dashboard
+#        # Defaults to "arn:aws:iam::youraccountid:role/aws-ec2-spot-fleet-tagging-role" assuming you've arrived "Spot Requests" in EC2 Dashboard
```
c-knowles: Seems like part of https://github.com/kubernetes-incubator/kube-aws/pull/1024/ slipped in here?

mumoshu:
@c-knowles Thanks for reviewing!
> What's the use case for `worker[*].stackTags` when there is also a root `stackTags` applicable to everything plus `worker[*].instanceTags` applicable to each node pool?
Just wanted to sync up with you but have you verified this behavior?
When I've tested it while developing this feature, the root `stackTags` had not propagated to workers nodes, at least.

c-knowles:
@mumoshu nope, I've not run that case. OK then, I expected it to work like that as we've described `stackTags` as recommended for cluster-wide metadata.

c-knowles:
Perhaps we could simplify and remove `stackTags` from node pools?

core/controlplane/config/templates/cluster.yaml
mumoshu: Yes, you are correct. Sorry but I had no time to test this in isolation, as noted in `Depends on #1024` in the PR description!

mumoshu:
@c-knowles Thx for the suggestion!
For what it's worth, I don't have specific use-cases, but in case you'd like to tag the cfn stack created per node pool, `nodePools[].stackTags` are the only way to go right now.

mumoshu:
@c-knowles I've opened #1038 for discussing that

c-knowles:
@mumoshu I've just been trying this, I could not get `instanceTags` to work across any of controller/etcd/worker. Just trying to debug now.

After updating an existing cluster, I also have a root stack has a tag that does not exist any more in the root `stackTags`. I think this is an AWS problem I saw before though - the tags are not updated properly using the SDK.

PR #1032, opened 2017-11-23
Add prometheus metrics ports

kevtaylor:
Based on the coreos prometheus helm charts in:
https://github.com/coreos/prometheus-operator/tree/master/helm

There are a number of metrics ports that get exposed for scraping such as kubelets, controller manager etc. which need to be visible across the nodes for the scrape targets to appear 'up' and healthy.

This PR contains those port security groups

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1348,6 +1348,118 @@
       },
       "Type": "AWS::EC2::SecurityGroupIngress"
     },
+    "SecurityGroupWorkerIngressFromWorkerToControllerKubelet": {
+      "Properties": {
+        "FromPort": 10250,
+        "GroupId": {
+          "Ref": "SecurityGroupController"
+        },
+        "IpProtocol": "tcp",
+        "SourceSecurityGroupId": {
+          "Ref": "SecurityGroupWorker"
+        },
+        "ToPort": 10250
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
+    "SecurityGroupWorkerIngressFromWorkerToWorkerKubelet": {
+      "Properties": {
+        "FromPort": 10250,
+        "GroupId": {
+          "Ref": "SecurityGroupWorker"
+        },
+        "IpProtocol": "tcp",
+        "SourceSecurityGroupId": {
+          "Ref": "SecurityGroupWorker"
+        },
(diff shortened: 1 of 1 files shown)


danielfm:
Have you tried the [example](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L53-L129) provided in Prometheus repository?

In that example, it shows how to use the `/proxy` endpoint from the apiserver to access in-cluster services - such as cAdvisor and node-exporter - without having to open any additional ports.

kevtaylor:
@danielfm Thanks but not sure how that helps us as we aren't using prometheus in that way - we are using the prometheus operator and servicemonitors

core/controlplane/config/templates/stack-template.json
```diff
@@ -1348,6 +1348,118 @@
...
+          "Ref": "SecurityGroupController"
+        },
+        "IpProtocol": "tcp",
+        "SourceSecurityGroupId": {
+          "Ref": "SecurityGroupWorker"
+        },
+        "ToPort": 10251
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
```
mumoshu: I'd like this being an optional feature security-wise, given theoretically it is possible that
a push-based monitoring agent understands the prometheus format can live without exposing all those ports.
Ref: https://github.com/DataDog/dd-agent/issues/3119

Would you mind adding a new flag in cluster.yaml for this?

kevtaylor: @mumoshu Sure. I'll look at that

mumoshu: @kevtaylor No hurry, just checking in, but any chance you work on it? 😄

zach-dunton-sf: I ran into the same issue the other day and added the flag in cluster.yaml

https://github.com/kubernetes-incubator/kube-aws/pull/1092

mumoshu:
Closing in favor of #1092. Thanks a lot for starting the work!

VinceMD:
Sorry if already documented Does that include:
* port 2379 from workers to etcd nodes
* port  9100 and 10250 from worker to controller nodes
* port   9100 and 10250 from worker to worker

I found this is what I need to make prometheus operator work fully (including etcd servicemonitoring)

PR #1033, opened 2017-11-23
Add rkt container cleanup to journald-cloudwatch-logs service

dvdthms:
A case was encountered where the journald-cloudwatch-logs service was failing, and this caused an issue where the rkt gc cannot unmount the containers and then the number of mountpoints causes the kernel to be unresponsive.

This change adds a cleanup step, so that the container is removed in the case that the service functionality has been interrupted.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -106,7 +106,9 @@ coreos:
                   --mount volume=journal,target=/var/log/journal \
                   --volume machine-id,kind=host,source=/etc/machine-id,readOnly=true \
                   --mount volume=machine-id,target=/etc/machine-id \
+                  --uuid-file-save=/var/journald-cloudwatch-logs/journald-cloudwatch-logs.uuid \
                   {{ .JournaldCloudWatchLogsImage.RktRepo }} -- {{.ClusterName}}
+        ExecStopPost=/usr/bin/rkt rm --uuid-file=/var/journald-cloudwatch-logs/journald-cloudwatch-logs.uuid
         Restart=always
         RestartSec=60s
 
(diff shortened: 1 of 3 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -106,7 +106,9 @@ coreos:
                   --mount volume=journal,target=/var/log/journal \
                   --volume machine-id,kind=host,source=/etc/machine-id,readOnly=true \
                   --mount volume=machine-id,target=/etc/machine-id \
+                  --uuid-file-save=/var/journald-cloudwatch-logs/journald-cloudwatch-logs.uuid \
```
mumoshu: Can we use `/var/run/coreos/` instead of `/var/journald-cloudwatch-logs` for consistency?

mumoshu: Perhaps we could replace all the `/var/run/coreos` to `/var/run/<svc>/` or `/var/run/kube-aws` as they aren't really tied to coreos but that would be an another github issue though!

mumoshu:
@dvdthms Thanks for your contribution as always!
Just wanted to know but how did you test this change? 😃

core/controlplane/config/templates/cloud-config-controller
dvdthms: Whichever you prefer! I noticed there was an existing journald-cloudwatch-logs directory so put the file in there, but there's no reason why it can't be moved elsewhere.

dvdthms:
First I introduced a failing condition into the service, so that I could replicate the issue, and saw unused containers by using `rkt list`. The default rkt gc is set at 24hrs I believe, which is too long a window if this issue is present on a node.

I then added the removal command, with the fail condition still present, and saw the old containers being removed after the service was restarted. Lastly, I removed the fail condition, and saw no interruption in the container function. 

Let me know if there's any problems 👍

core/controlplane/config/templates/cloud-config-controller
mumoshu: Thanks for confirming!
Yea, it seems like the directory is used for storing the state file according to [@jollinshead's setup](https://github.com/jollinshead/journald-cloudwatch-logs/blob/master/docker/init.sh#L6).
I don't have a strong opinion here so ;et's go ahead with the current directory.

mumoshu:
@dvdthms Sounds great! Thanks for clarifying 👍

PR #1034, opened 2017-11-23
Quick start and high availability guides

c-knowles:
WIP since I ran through this guide and keep hitting `Invalid arn` messages during TLS bootstrap.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,6 +1,7 @@
 # Summary
 
 * [Home](README.md)
+* [Quick Start](tutorials/quick-start.md)
 * [Getting Started](getting-started/README.md)
   * [Prerequisites](getting-started/prerequisites.md)
   * [Step 1: Configure](getting-started/step-1-configure.md)
@@ -26,4 +27,3 @@
 * [Troubleshooting](troubleshooting/README.md)
   * [Known Limitations](troubleshooting/known-limitations.md)
   * [Common Problems](troubleshooting/common-problems.md)
-* [Quick Start \(WIP\)](tutorials/quick-start.md)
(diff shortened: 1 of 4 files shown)


c-knowles:
@mumoshu by any chance have you seen this error before during cluster validate?

```
Error: Failed to initialize cluster driver: failed to read/create encrypted assets: Error encrypting credentials/kubelet-tls-bootstrap-token: NotFoundException: Invalid arn
	status code: 400, request id: 6396d065-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

I checked a few times and my KMS arn seems to be fine. I even tried recreating the `credentials` folder from scratch.

mumoshu:
> hitting Invalid arn messages during TLS bootstrap.

@c-knowles Would you mind elaborating a bit more about this?
// Probably in an another issue!

I have never encountered that myself and am interested.

c-knowles:
@mumoshu running through the basic quick start will get you there, but right now I've started to hit another problem in https://github.com/kubernetes-incubator/kube-aws/issues/1003#issuecomment-347084904 so harder to figure out what's going on.

I think these are related to the new code introduced around TLS bootstrap as I didn't have any issues with initialisation until then. I have a script which does the init and renders for an entirely new cluster each time so I can compare the latest release changes against our clusters. But it's just this:
```
rm -rf render/compare/
mkdir -p render/compare/credentials/
cp render/dev/credentials/ca*.pem render/compare/credentials/
cd render/compare
kube-aws init <MANDATORY_PARAMS>
kube-aws render credentials
kube-aws render stack
```

c-knowles:
I filed https://github.com/kubernetes-incubator/kube-aws/issues/1041 to cover the problem.

c-knowles:
I've found the cause of https://github.com/kubernetes-incubator/kube-aws/issues/1041 which we can either add validation for or update the KMS key handling.

@mumoshu this is good for merge assuming someone reviews.

mumoshu:
/lgtm

Thank you very much for the great write-up!

PR #1035, opened 2017-11-24
Use templated image for metrics-server.

camilb:
Use templated image for metrics-server
Fix a bug in `install-kube-system.service` introduced in #973


camilb:
Merging this because the bug introduced in #973 prevents controllers to start.

PR #1040, opened 2017-11-27
Improve the configuration for Kubernetes Dashboard.

camilb:


docs/advanced-topics/kubernetes-dashboard.md
```diff
@@ -0,0 +1,134 @@
+# Kubernetes Dashboard Access and Authentication
+
+## Default Setup
+
+    kubernetesDashboard:
```
c-knowles: nit, ` ```yaml`

docs/advanced-topics/kubernetes-dashboard.md
```diff
@@ -0,0 +1,134 @@
...
+In the default setup the Dashboard is configured using the `--auto-generate-certificates` flag and has Admin privileges
+
+###### Access the dashboard  using `kubectl proxy` command:
+
+    http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/
+
+
+## Expose the Dashboard using a ELB with self-signed certificates
+
+kubernetesDashboard:
```
c-knowles: Same as above, would like to put this in a code block.

docs/advanced-topics/kubernetes-dashboard.md
```diff
@@ -0,0 +1,134 @@
...
+    kind: Service
+    metadata:
+      annotations:
+        external-dns.alpha.kubernetes.io/hostname: "kubedash.example.com"
+      name: kubernetes-dashboard
+      labels:
+        run: kubernetes-dashboard
+      namespace: kube-system
+    spec:
+      type: LoadBalancer
```
c-knowles: Should we add this as a config option? I guess I still don't understand why the dashboard can no longer be access via the kubeapi ELB which already exists, that's already SSL end to end and uses the certs we generate.

docs/advanced-topics/kubernetes-dashboard.md
```diff
@@ -0,0 +1,134 @@
...
+  adminPrivileges: false
+  insecureLogin: false
+
+Ex.
+
+    apiVersion: v1
+    kind: Service
+    metadata:
+      annotations:
+        external-dns.alpha.kubernetes.io/hostname: "kubedash.example.com"
```
c-knowles: I guess we are assuming `external-dns` is the default now but might be worth excluding from this example?

docs/advanced-topics/kubernetes-dashboard.md
```diff
@@ -0,0 +1,134 @@
...
+      ports:
+      - port: 443
+        targetPort: 9090
+        protocol: TCP
+      selector:
+        k8s-app: kubernetes-dashboard  
+
+###### Access the dashboard  using `kubectl proxy` command:
+
+    http://localhost:8001/api/v1/namespaces/kube-system/services/http:kubernetes-dashboard:/proxy/
```
c-knowles: Is `proxy` relevant when exposed via ELB?

camilb:
> So far I've not understood why the existing use case of dashboard via kubeapi ELB is not possible 

As I understand, the only method to access the dashboard via kubeapi ELB is by using basic auth for API authentication.  In theory, this should work fine `https://<master-ip>:<apiserver-port>/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/`, but how do you authenticate against the API in browser?

docs/advanced-topics/kubernetes-dashboard.md
camilb: Sometimes can be useful for debugging. Just wanted to make it clear, but I suppose we can specify it only once, separately.

docs/advanced-topics/kubernetes-dashboard.md
camilb: Are you using `basic-auth` to access the kubernetes API? How do you authenticate against the API in browser when using a client certificate?

docs/advanced-topics/kubernetes-dashboard.md
camilb: Yes, I guess it can be excluded. I let it because it won't cause any error if it's specified.

docs/advanced-topics/kubernetes-dashboard.md
camilb: I'll update it

docs/advanced-topics/kubernetes-dashboard.md
c-knowles: Sure, seems fine either way just an opportunity to simplify.

docs/advanced-topics/kubernetes-dashboard.md
c-knowles: Just the original kube-aws mechanism using client SSL certs. i.e. add the CA to the approved ones and the admin key to the relevant keychain for client SSL certs. Then when browsing to the dashboard via kubeapi ELB one receives a client SSL cert prompt to select which admin key.

c-knowles:
@camilb maybe good if we could have a quick chat on k8s Slack?

docs/advanced-topics/kubernetes-dashboard.md
camilb: Ok, thanks for the info. Never used it like this. The good news is that I tested and it works fine with the default configuration and it's something like: `https://kubeapi.example.com/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy`

camilb:
@c-knowles Yes, I would like to. I'll be online in ~6h.  I'm on GMT+2 and it's 4 AM here right now. See ya!

docs/advanced-topics/kubernetes-dashboard.md
c-knowles: @camilb great. I'd like to try to enable token auth via this mechanism without needing to dish out the client SSL cert (as that allows super admin access).

c-knowles:
@camilb 👍  didn't realise it was 4am there! Take a rest!

camilb:
Hi @mumoshu, had a chat on slack with @c-knowles and we consider this PR can be merged at this point, if you also considering so.

mumoshu:
@camilb @c-knowles Thanks for your efforts!

PR #1043, opened 2017-11-28
Various fixes to 0.9.9 rc.3

mumoshu:
@c-knowles @danielfm This is intended to fix #1042 and #1039, which is already working for my clusters.
Would you mind building & testing this?

PR #1060, opened 2017-12-06
Fix install-kube-system when node drainer is enabled

mumoshu:
Follow-up for #1043


c-knowles:
Tested this and seems to have fixed kube system deploys.

/lgtm

PR #1066, opened 2017-12-07
Added Priority to the admissioncontrol list

zonzamas:
In order to properly support Pod Priorities and Preemption (kubernetes 1.8 https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) we need the Priority admission controler

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -50,6 +50,9 @@ func NewDefaultCluster() *Cluster {
 			Initializers{
 				Enabled: false,
 			},
+			Priority{
+				Enabled: false,
+			},
 		},
 		AuditLog: AuditLog{
 			Enabled: false,
@@ -529,6 +532,7 @@ type Admission struct {
 	AlwaysPullImages   AlwaysPullImages   `yaml:"alwaysPullImages"`
 	DenyEscalatingExec DenyEscalatingExec `yaml:"denyEscalatingExec"`
 	Initializers       Initializers       `yaml:"initializers"`
+	Priority           Priority           `yaml:"priority"`
 }
 
 type AlwaysPullImages struct {
@@ -544,6 +548,10 @@ type DenyEscalatingExec struct {
 }
 
 type Initializers struct {
+	Enabled bool `yaml:"enabled"`
+}
+
+type Priority struct {
 	Enabled bool `yaml:"enabled"`
 }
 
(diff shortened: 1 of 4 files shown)


zonzamas:
Already signed the CLA

camilb:
Hi @zonzamas can you please add the necessary configuration in `core/config/templates/cluster.yaml` ? 
```yaml
experimental:
  admission:
    priority:
      enabled: false
```

zonzamas:
I guess you mean `core/controlplane/config/templates/cluster.yaml` right? the other file does not exists

camilb:
@zonzamas Yes, updated.

zonzamas:
[Here](https://github.com/zonzamas/kube-aws/blob/31dee1d9c6f32006e965383b8f4165fc381f4723/core/controlplane/config/templates/cluster.yaml#L1213) ... waiting for :green_apple: from travis

mumoshu:
@zonzamas Hi, thanks a lot for contributing!

This looks good but just for clarification - probably you also want `--feature-gates=PodPriority=true` for apiservers and kubelets to actually enable the PodPriority feature?
Are you willing to address it in a future PR(s) or you just missed it in this PR?

zonzamas:
Thank you for kube-aws. Right now I'm enabling pod priority via plugins. I did this one because it seemed like a low hanging fruit. I'm not (yet) familiar with the code base but ill see what i can do. Under which configuration section do you think it should go? Experimental?

mumoshu:
@zonzamas Thank you for the response, and trying kube-aws 👍 

Oh, you meant a kube-aws plugin?

Btw, reading some code reminded me that there is already `worker.nodePools[].featureGates` which can be used to add feature gate(s) to kubelet, and `controller.featureGates` for its controller variant! So, there seem like no need to update the code base.

However, would you mind adding some comment to `admissions.priority` in cluster.yaml so that you can notice users so that they will also update `featureGates`?

zonzamas:
I will look into that @mumoshu

zonzamas:
I have changed things a bit, there are a lot of required flags to enable for Pod Priority to properly work. I have used priority admission control for that... the rest using featureGates.

mumoshu:
> here are a lot of required flags to enable for Pod Priority to properly work

Yes, everyone would thank you for your efforts 👍 

> the rest using featureGates.

What was the intention behind that? I guess you'd wanted to enable PodPriority in a subset of many node pools so that you can gradually roll-out PodPriority enabled worker nodes, right? (But not sure if things are really designed to work like that

Issue #1058, opened 2017-12-05
Integration with kube-aws-ingress-controller + skipper/istio-ingress/nginx-ingress


mumoshu:
After surveying about coreos/alb-ingress-controller, nginx-ingress-controller and so on, kube-aws-ingress-controller seems to be an optimal solution when you'd like to achieve the followings requirements as whole:

* Let an ingress controller create an ALB w/ appropriate ACM cert attached
* Appropriate worker nodes are added to target group(s) of the ALB
* DNS records are created on the ALB by external-dns
- In combination with an other ingress controller, path/cookie/weight/other rule based routing which isn't possible just with ALB but with combination of two ingress controllers

As kube-aws-ingress-controller is designed to be used in combination with an another ingress controller, we may choose skipper or istio-ingress, nginx-ingress, or else in addition to kube-aws-ingress-controller.

* skipper is well known and battle-tested in combination with kube-aws-ingress-controller by zalando. supports annotations for easy customizations. Even weight-based routing is configurable via annotations.
* istio-ingress would be a natural choice once we address #978. It works but I don't know how mature it is. I also don't know if it works well w/ kube-aws-ingress-controller though. Just have not tried yet.
* nginx-ingress-controller seems like the defacto of ingress controllers. Don't know if it works well w/ kube-aws-ingress-controller. Supports many annotations for easy customizations. Supports nginx.conf injection as the last resort.

mumoshu:
Just finished experimenting with kube-ingress-aws-controller+skipper - it worked great!

Firstly, you'd need these IAM permissions for all the ASG-backed worker nodes:

```yaml
worker:
  nodePools:
  - iam:
      policy:
        statements:
        - effect: Allow
          actions:
          - "autoscaling:DescribeAutoScalingGroups"
          - "autoscaling:AttachLoadBalancers"
          - "autoscaling:DetachLoadBalancers"
          - "autoscaling:DetachLoadBalancerTargetGroup"
          - "autoscaling:AttachLoadBalancerTargetGroups"
          - "elasticloadbalancing:AddTags"
          - "elasticloadbalancing:DescribeLoadBalancers"
          - "elasticloadbalancing:CreateLoadBalancer"
          - "elasticloadbalancing:DeleteLoadBalancer"
          - "elasticloadbalancing:DescribeListeners"
          - "elasticloadbalancing:CreateListener"
          - "elasticloadbalancing:DeleteListener"
[...]

PR #1067, opened 2017-12-08
Correct values for the `kubernetes.io/cluster/<Cluster ID>` tags

### LINKED ISSUE #1025: `kubernetes.io/cluster/<Cluster ID>` tag should have a valid value
It should be either `owned` or `shared` but is now `true`.

Ref: https://github.com/kubernetes/kubernetes/blob/v1.8.3/pkg/cloudprovider/providers/aws/tags.go#L44-L51

mumoshu:
Resolves #1025

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -554,7 +554,7 @@
         }],
         "HostedZoneTags" : [{
           "Key": "kubernetes.io/cluster/{{$.ClusterName}}",
-          "Value": "true"
+          "Value": "owned"
         }]
       }
     },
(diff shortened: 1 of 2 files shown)


mumoshu:
See my comment at https://github.com/kubernetes-incubator/kube-aws/issues/1025#issuecomment-350218190 for why I'm fixing this now.

mumoshu:
Merging this as I've already tested this while experimenting on #1058

mumoshu:
Sonobuoy Scan Result: https://scanner.heptio.com/6b16bb7c0e7f460375ac2f2cd95f2c52/diagnostics/

Issue #1000, opened 2017-11-07
Confusing validation error and cluster.yaml for route 53 setup


c-knowles:
I found some small issues from https://github.com/kubernetes-incubator/kube-aws/pull/936.

[Validation message](https://github.com/kubernetes-incubator/kube-aws/blame/ec0f9b33109fd16320e3f5e0da6a5acb4d4ca031/model/api_endpoint_lb.go#L122) says `missing hostedZone.id: hostedZone.id is required when `recordSetManaged` is set to true. If you do want to configure DNS yourself, set it to true`. I think it should recommend to set to `false` if you want to configure DNS yourself.

Default `cluster.yaml` has examples with [`managed`](https://github.com/kubernetes-incubator/kube-aws/blob/ec0f9b33109fd16320e3f5e0da6a5acb4d4ca031/core/controlplane/config/templates/cluster.yaml#L127) and [`recordSetManaged`](https://github.com/kubernetes-incubator/kube-aws/blob/ec0f9b33109fd16320e3f5e0da6a5acb4d4ca031/core/controlplane/config/templates/cluster.yaml#L136). I think "managed" has become a little confusing as one is the LB itself and one if the record set. Wondering if there is a way to clarify it. I also keep forgetting if "managed" means kube-aws manages it or it's self managed, perhaps we could clarify that while we are at it in the default cluster.yaml.

nick4fake:
Well, it looks like there is a problem in the code.

In api_endpoint_lb.go, ELB is created if RecordSetManaged is not nil, and in the same time there is a strict check:
````
if e.RecordSetManaged == nil || *e.RecordSetManaged {
			return errors.New("missing hostedZone.id: hostedZone.id is required when `recordSetManaged` is set to true. If you do want to configure DNS yourself, set it to true")
		}
````

mumoshu:
> I also keep forgetting if "managed" means kube-aws manages it or it's self managed, perhaps we could clarify that while we are at it in the default cluster.yaml.

I began to believe that we should name keys more explicitly and concisely: `kubeAwsManaged...`/`recordSetKubeAwsManaged` vs `userManaged...`.

mumoshu:
>  I think it should recommend to set to false if you want to configure DNS yourself.

Are, true! I had not noticed that until now. Indeed it is confusion 😄

PR #1069, opened 2017-12-11
Fix for #1000: Confusing validation error and cluster.yaml for route 53 setup

nick4fake:


nick4fake:
Sign check.

mumoshu:
@nick4fake Seems like we need to fix tests. I'll take a look soon.

PR #1086, opened 2017-12-19
controlplane config: Rotate Certificates on workers kubelet

zonzamas:
Added a flag to enable automatic cert rotation when TLS Bootstraping
is used.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1309,6 +1309,13 @@ experimental:
   # Must be N times more than kubelet's nodeStatusUpdateFrequency (default 10s).
 #   nodeMonitorGracePeriod: "40s"
 
+kubelet:
+  # Tell Kubelet to renew its certificate as its expiration time is approaching
+  # Requires FeatureGate RotateKubeletClientCertificate=true on worker
+  # Also requires Experimental.tlsBootstrap to be enabled
+  RotateCerts:
+    enabled: true
+
 # AWS Tags for cloudformation stack resources
 #stackTags:
 #  Name: "Kubernetes"
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -36,6 +36,11 @@ const (
 )
 
 func NewDefaultCluster() *Cluster {
+	kubelet := Kubelet{
+		RotateCerts: RotateCerts{
+			Enabled: false,
+		},
+	}
 	experimental := Experimental{
 		Admission: Admission{
 			PodSecurityPolicy{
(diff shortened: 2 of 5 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1243,6 +1243,11 @@ experimental:
   tlsBootstrap:
     enabled: false
 
+  # Tell Kubelet to renew its certificate as its expiration time is approaching
+  # Requires FeatureGate RotateKubeletClientCertificate=true on worker
+  RotateCerts:
```
mumoshu: Could you move this to `kubelet.rotateCertificates`?

Sorry for taking your time but we're migrating away from the experimental settings because not only experimental but everything could change while we're in pre v1.0 😉

zonzamas: ack, no problem

zonzamas:
I think this is what you meant @mumoshu

core/controlplane/config/templates/cluster.yaml
```diff
@@ -1309,6 +1309,13 @@ experimental:
   # Must be N times more than kubelet's nodeStatusUpdateFrequency (default 10s).
 #   nodeMonitorGracePeriod: "40s"
 
+kubelet:
+  # Tell Kubelet to renew its certificate as its expiration time is approaching
+  # Requires FeatureGate RotateKubeletClientCertificate=true on worker
```
danielfm: You say this feature gate is required, but why not adding that feature gate automatically if the certificate rotation is enabled?

zonzamas: It could be done, Is there any convention about that?

danielfm: I don't believe there's a convention for this, only that we (ideally) should avoid forcing the user to change two places (in this case, `cluster.yaml` and  `userdata/cloud-config-worker`) in order to enable any given feature.

zonzamas: FeatureGates can be added to the cluster.yaml (and this is what i meant in the comment)

```
worker:
  featureGate:
     RotateKubeletClientCertificate: true
```

I'll check how to do it automatically though

danielfm: Ah, I did not know about this flag in `cluster.yaml`. 😅 

Sorry for the confusion.

zonzamas: After reviewing the code, I've seen that automatic FeatureGate enabling is being done for another component:

See https://github.com/kubernetes-incubator/kube-aws/blob/master/core/nodepool/config/config.go#L291

mumoshu: Yep it definitely makes sense to me!
Would you mind doing the same for this feature, too?

mumoshu:
@zonzamas Thank you very much! Yes, that's exactly what I meant.

mumoshu:
@zonzamas Sorry for the delayed response. I had been
off for several days due to regional holidays.

I'm merging this not to slow things down by potential conflicts.

Would you mind addressing https://github.com/kubernetes-incubator/kube-aws/pull/1086#discussion_r159886263 in an another PR?

zonzamas:
No worries, we all have taken some time off these days.

I will address your comment in another PR

Thanks

PR #1090, opened 2017-12-28
Set V4 signatures to enable kube-resources-autosave to work in all regions

whereisaaron:
The coreos awscli container image is ancient and unmaintained and defaults to the old V2 signatures. We need to explicitly set V4 signatures to enable this feature to function in all regions, e.g. us-east-2 where only V4 is supported.


mumoshu:
@whereisaaron Sorry I had been off due to regional holidays. Thanks for the fix!

/lgtm

whereisaaron:
No problem @mumoshu, I expected you'd be away. Just this time of year is when people stop hassling me and I can get the fun stuff done :-)

PR #1092, opened 2017-12-28
Allow exposing several ports so that external Prometheus can scrape K8S system components' metrics

zach-dunton-sf:
Continuation of https://github.com/kubernetes-incubator/kube-aws/pull/1032

Added the yaml option and rebased onto the current master.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1214,6 +1214,11 @@ addons:
   metricsServer:
     enabled: false
 
+  # When set to true this configures security groups for prometheus between nodes.
+  # This includes the following ports: 10252, 10251, 10250, 9100, and 4194
+  prometheus: 
+    securityGroupsEnabled: false
+
 # Experimental features will change in backward-incompatible ways
 experimental:
   # Enable admission controllers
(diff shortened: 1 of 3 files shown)


kevtaylor:
@zach-dunton-sf Thanks for covering this as I hadn't had a chance to do it

core/controlplane/config/templates/cluster.yaml
```diff
@@ -1214,6 +1214,11 @@ addons:
   metricsServer:
     enabled: false
 
+  # When set to true this configures security groups for prometheus between nodes.
+  # This includes the following ports: 10252, 10251, 10250, 9100, and 4194
+  prometheus:
```
mumoshu: Would you mind if I changed this to something like `node.exposeKubernetesSystemCompomentsPortsForPrometheus` so that it won't look like a general configuration section for Prometheus itself?

mumoshu:
@zach-dunton-sf Thanks for your contribution!

Only a minor comment regarding ux https://github.com/kubernetes-incubator/kube-aws/pull/1092#discussion_r159891140 

/lgtm

mumoshu:
Changed the PR title according to my understanding of the use-case behind this.
Am I correct?

core/controlplane/config/templates/cluster.yaml
zach-dunton-sf: Perfect. I wasn't sure exactly how to structure the config flag.

zach-dunton-sf:
The title change, makes the purpose much cleare.

mumoshu:
@zach-dunton-sf Thanks for your contribution 👍

PR #1095, opened 2017-12-29
fix restore.sh

skloss:


skloss:
CLA signature registration done

contrib/cluster-backup/restore.sh
```diff
@@ -141,7 +141,7 @@ main() {
     for ns in ${NAMESPACES}; do
         if [ -e "${TEMP_DIR}/${ns}" ] && [ -d "${TEMP_DIR}/${ns}" ]; then
         echo "Restoring resources for '${ns}' Namespace..."
-        kubectl create -f <(echo $(cat ${TEMP_DIR}/namespaces.json | jq -r --arg NS "$ns" 'select( .items[].metadata.name | contains($NS))'))
+        kubectl create -f <(echo $(cat ${TEMP_DIR}/namespaces.json | jq -r --arg NS "$ns" '.items[] | select(.metadata.name | contains($NS))'))
```
mumoshu: @skloss Wow! So, it was (mistakenly) restoring all the namespaces at once regardless of what was said to be restored?

skloss: @mumoshu  Yes,exactly and also it wasn't creating any ressource inside the namespace

skloss: tested it also on two cluster's

skloss:
fixed bug: if the namespace was containing the namespace of other the script failed at line 142, thats why i replaced the contains with equals.

mumoshu:
@skloss Good catch! Thanks a lot for the fix 🎉

PR #1097, opened 2018-01-03
Update documentation that hostPort issue fixed since Kubernetes 1.7.0

whereisaaron:
Update that fixed since Kubernetes 1.7.0 (as noted in the upstream issue)
I've been testing with 1.8 and not see this problem there either.


mumoshu:
@whereisaaron Good news! Thanks for confirming.

/jgtm

PR #1098, opened 2018-01-03
Add missing admission controllers

danielfm:
According to the [docs](https://kubernetes.io/docs/admin/admission-controllers/), it's strongly recommended to also add the `PersistentVolumeLabel` and `DefaultTolerationSeconds` admission controllers for Kubernetes versions >= 1.6.0.


mumoshu:
Good catch. Thanks as always for your support!

PR #1101, opened 2018-01-04
Fix kubelet.service startup.

### LINKED ISSUE #1100: kubelet.service fails to start on builds from current master
Using the current master branch, the `kubelet.service` fails to start on master nodes.

The error `/run/systemd/system/kubelet.service:23: Missing '='.`  is caused by a double backslash in https://github.com/kubernetes-incubator/kube-aws/blob/d15e5cf95d43733b6a0b062d4e545d2103a8347f/core/controlplane/config/templates/cloud-config-controller#L297

camilb:
Fixes #1100

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -292,9 +292,10 @@ coreos:
         {{ if .KubeDns.NodeLocalResolver }}--cluster-dns=${COREOS_PRIVATE_IPV4} \
         {{ else }}--cluster-dns={{.DNSServiceIP}} \
         {{ end }}--cluster-domain=cluster.local \
-        --cloud-provider=aws \{{if .Experimental.Admission.Priority.Enabled}}
+        --cloud-provider=aws \
+        {{if .Experimental.Admission.Priority.Enabled}}
         --feature-gates=PodPriority=true \
-        {{end}}\
+        {{end}}
         $KUBELET_OPTS
         Restart=always
         RestartSec=10
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -295,7 +295,7 @@ coreos:
         --cloud-provider=aws \
         {{if .Experimental.Admission.Priority.Enabled}}
         --feature-gates=PodPriority=true \
-        {{end}}
+        {{end}}\
         $KUBELET_OPTS
         Restart=always
         RestartSec=10


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -292,9 +292,10 @@ coreos:
         {{ if .KubeDns.NodeLocalResolver }}--cluster-dns=${COREOS_PRIVATE_IPV4} \
         {{ else }}--cluster-dns={{.DNSServiceIP}} \
         {{ end }}--cluster-domain=cluster.local \
-        --cloud-provider=aws \{{if .Experimental.Admission.Priority.Enabled}}
+        --cloud-provider=aws \
+        {{if .Experimental.Admission.Priority.Enabled}}
         --feature-gates=PodPriority=true \
-        {{end}}\
+        {{end}}
```
danielfm: Doesn't this line also has to end with `\`, or otherwise anything passed into `$KUBELET_OPTS` will be considered to be part of the systemd configuration itself?

camilb: @danielfm maybe when `feature-gates` it's enabled works correct, but when it's disabled, this is how it looks:

```bash
--register-with-taints=node.alpha.kubernetes.io/role=master:NoSchedule \
--allow-privileged=true \
--pod-manifest-path=/etc/kubernetes/manifests \
--cluster-dns=10.3.0.10 \
--cluster-domain=cluster.local \
--cloud-provider=aws \\
$KUBELET_OPTS
Restart=always
RestartSec=10
```

camilb: The double backslash prevent `kubelet.service` to start

danielfm: Since you dropped the `{{ if }}` to the next line, wouldn't it look like this then (assuming you put the `\` after the `{{ end }}`?

```
        ...
        --cloud-provider=aws \
        \
        $KUBELET_OPTS
```

...which I believe is correct.

Anyway, we need to make sure this change doesn't break things for people relying on `$KUBELET_OPTS`.

camilb: I'll test it again with and without `--feature-gates`

camilb: > Since you dropped the {{ if }} to the next line, wouldn't it look like this then (assuming you put the \ after the {{ end }}?

Yes, by dropping the `{{ if }}` to the next line, it's also dropping the `\` to the next line.

camilb:
```yaml
    priority:
      enabled: false
```
```bash
....
        --pod-manifest-path=/etc/kubernetes/manifests \
        --cluster-dns=10.3.0.10 \
        --cluster-domain=cluster.local \
        --cloud-provider=aws \
        
        $KUBELET_OPTS
.....
```

```yaml
    priority:
      enabled: true
```
```bash
....
        --pod-manifest-path=/etc/kubernetes/manifests \
        --cluster-dns=10.3.0.10 \
        --cluster-domain=cluster.local \
        --cloud-provider=aws \
        
        --feature-gates=PodPriority=true \
        
        $KUBELET_OPTS
....
```

camilb: I think that dropping the `{{if}}`  should be enough

camilb: After the cade6e91b4893787e16fed55324a3527f6d80fc1 looks like this:
```yaml
    priority:
      enabled: false
```
```bash
....
        --pod-manifest-path=/etc/kubernetes/manifests \
        --cluster-dns=10.3.0.10 \
        --cluster-domain=cluster.local \
        --cloud-provider=aws \
        \
        $KUBELET_OPTS
.....
```

```yaml
    priority:
      enabled: true
```
```bash
....
        --pod-manifest-path=/etc/kubernetes/manifests \
        --cluster-dns=10.3.0.10 \
        --cluster-domain=cluster.local \
        --cloud-provider=aws \
        
        --feature-gates=PodPriority=true \
        \
        $KUBELET_OPTS
....
```

camilb: Tested and `kubelet.service` starts fine now.

mumoshu:
Thanks for your efforts!
Templating YAML is really killing us 😢

PR #1102, opened 2018-01-04
Fix flanneld.service when using custom flannel image.

camilb:


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -200,6 +200,8 @@ coreos:
             TimeoutStartSec=120
 
 {{if .FlannelImage.RktPullDocker}}
+    - name: flanneld.service
+      drop-ins:
```
mumoshu: Why do we need this even though there is already L177-178 with the equivalent code?

camilb: You are right, got confused. For some reason it didn't create the drop-in file. I'll check it again. Thanks!

PR #1103, opened 2018-01-04
Update various components

camilb:


c-knowles:
Original vanity domain issue is https://github.com/kubernetes/kubernetes/pull/54174.

Issue #1106, opened 2018-01-06
'kube-aws init' should insert a fixed 'amiId' in cluster.yaml and/or warn users


whereisaaron:
I just read some past scary issues where people had unexpected complete cluster rebuilds because the `amiId` setting in `cluster.yaml` was blank. I don't remember reading any warning about this and the impact on an innocent `kube-aws update`. Others have be caught by surprise in recent months (#825 #1006). I'm glad I discovered it before it bit me!

I wanted to raise it as its own issue so it doesn't get lost.

- [x] We should document this fishhook in the `kube-aws init` `cluster.yaml` and in the install documentation.
- [ ] Also we should document how to find the possible/current 'amiId' values and set them manually.
- [ ] Ideally we could follow [@danielfm's suggestion](https://github.com/kubernetes-incubator/kube-aws/issues/825#issuecomment-320760362) to have `kube-aws init` set this field when generating the `cluster.yaml` rather than leave it blank.

whereisaaron:
Made a start with a one-line warning in `cluster.yaml` in PR #1107.

```
# The AMI ID of CoreOS.
# If omitted, the latest AMI for the releaseChannel is used.
# This happens every time 'kube-aws update' is run, potentially replacing all instances with a newer version
```

Issue #436, opened 2017-03-22
Fix locksmith etcd-lock reboot strategy


Thermi:
Right now, the reboot-strategy "etcd-lock", which is the only suitable one for a HA cluster, doesn't work and is disabled by default (reboot-strategy: "off").
To make a production ready cluster, automatic updates which don't impair or disrupt the cluster's operation have to be possible.

mumoshu:
@Thermi Thanks for creating the issue 👍 
I personally prefer to turn it off for my production clusters so that I have a complete control over when to automatically upgrade Container Linux for nodes.
Would it be enough to allow selecting a reboot strategy per sub-clusters including etcd, controller, each worker node pool?

mumoshu:
Related: https://github.com/kubernetes-incubator/kube-aws/issues/240

Thermi:
I think selecting a reboot strategy per sub-cluster would be fine.
I like clusters to take care of themselves as much as possible to keep down OPEX and possibility for human error.

mumoshu:
Thanks for the confirmation 👍 
I generally agree to you.
JFYI, probably you may already know but to be clear; I believe we need an extra care to have enough replicas and/or extra nodes to allow your services to be kept up and runnig while upgrading Container Linux over all the nodes. Please see the relevant discussion at https://github.com/kubernetes-incubator/kube-aws/issues/240#issuecomment-272134598

Thermi:
Yes, sure. That's the point of the etcd-lock: To ensure only one node reboots at a time.

c-knowles:
Could we replace this one and https://github.com/kubernetes-incubator/kube-aws/issues/240 with an issue to support https://github.com/coreos/container-linux-update-operator?

mumoshu:
@c-knowles I believe so 👍

Issue #240, opened 2017-01-12
Allow CoreOS updates by adding nodes to etcd as proxy


ArchiFleKs:
Currently, updates strategy is off on Worker node and controllers.

I dont know if that would be the right approach but maybe we could add the CoreOS instances to ETCD, maybe as proxy to take advantages of the etcd-lock reboot strategy ?

mumoshu:
Hi, thanks as always!
I'm personally not going to use the automatic update feature of CoreOS in production without an imaginary feature to postpone a reboot until all the pods running on the node moves to existing node(s). I'm open to suggestions though!

redbaron:
doesn't node drain feature help doing it?

mumoshu:
Not really. AFAIK it waits for pods on a node to be deleted but not for new pods on another nodes to be running.

redbaron:
So *wait-before-reboot* logic would be:
 - mark node so that no _new_ pods are scheduled to it (how?)
 - get list of running pods from a node
 - drain a node
 - poll k8s controllers waiting until all pods from (2) are running elsewhere.
 - reboot.

Right?

c-knowles:
Short term we could taint an existing node for now? Longer term I think we should wait for https://github.com/coreos/bugs/issues/1274.

Side note, as a user of kube-aws I had switched a dev cluster over to the `etcd-lock` strategy to try it out. We should add something to the documentation if it's not advised to use it right now.

mumoshu:
@redbaron I wasn't aware of a concrete solution to the problem but your idea sounds great 👍 

Let me also add that we'll need to make enough space in the cluster for pods evicted from the drained node to be run, probably having at least one extra node for that or enable `cluster-autoscaler` to automatically add a node for those pods.

mumoshu:
@c-knowles Thanks for looking into this!

For what part of the process you intended to use taints?

I wasn't aware of it until now but am really looking forward for `klocksmithd`!

c-knowles:
@mumoshu for the taint I meant for the first point of @redbaron's proposal. Perhaps node draining and waiting is enough though?

mumoshu:
@c-knowles Ah, I guessed that `kubectl cordon` would be enough for that!

c-knowles:
@mumoshu, yep makes sense

c-knowles:
[...]

PR #1107, opened 2018-01-06
Warn that 'kube-aws update' can replace all if 'amiId' is blank

whereisaaron:
Following on from #1106, add a small warning that a blank 'amiId' can lead to all instances being rolled during any 'kube-aws update'.


c-knowles:
/lgtm

Should we add to this comment that currently the container linux automatic update is turned off so this is the only way to update? (see https://github.com/kubernetes-incubator/kube-aws/issues/436 and https://github.com/kubernetes-incubator/kube-aws/issues/240)

mumoshu:
@whereisaaron Thanks for your contribution as always 🍻

PR #1108, opened 2018-01-08
Add RotateKubeletClientCertificate feature gate automatically when rotating certificates

zonzamas:
This PR implements what was discussed in 
https://github.com/kubernetes-incubator/kube-aws/pull/1086#discussion_r159886263

In short: When we enable `rotateCerts` apart from the required flag, the feature gate `RotateKubeletClientCertificate` should be also added automatically.


zonzamas:
Ok. Done, I have squashed everything into a single commit.

mumoshu:
@zonzamas Great! Thanks for your support.

PR #1111, opened 2018-01-13
CA key still required for non-bootstrapped TLS

### LINKED ISSUE #960: Error reading credential file with custom CA
Following the guide here:

https://kubernetes-incubator.github.io/kube-aws/getting-started/step-2-render.html

I have previously set up and run test clusters using a generated CA, but wanted to switch to using my existing one.  So following that, I did:

    kube-aws render credentials --ca-cert-path=/root/ca/pki/ca.crt --ca-key-path=/root/ca/pki/private/ca.key

As far as I can tell, by comparing outputs from different command varieties, this seems to generate all the important files (though it doesn't generate `.pem.enc` or `.pem.fingerprint` files) except one - `ca-key.pem`.  Looking at `ca.pem`, which is generated, this must relate to the provided CA.  Honestly I wouldn't even want a copy of my CA's private key in there, so the fact that it isn't is good.  Except for two issues:

1) I don't know how not having it there will affect the running of the cluster.  Have the generated keys been signed?

2) The crux of this issue - I get the following error:

```
Generating credentials...
-> Parsing existing TLS CA
-> Generating new assets
Error: Error reading credential file credentials/ca-key.pem: open credentials/ca-key.pem: no such file or directory
```

As I said, all the files seem to be generated, but I can't be sure in what state - I've not had a chance to run the cluster yet.

jlk:
`kube-aws up` will fail if credentials/ca-key.pem is not present. The
file is needed whether kube-aws bootstraps the TLS certificates or no.

This passes `make test` and cluster builds for me, but would appreciate testing from others to make sure I'm not breaking something.

Fixes #960

### DIFF
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -179,7 +179,7 @@ func (c *Cluster) NewAssetsOnDisk(dir string, renderCredentialsOpts CredentialsO
 
 	tlsBootstrappingEnabled := c.Experimental.TLSBootstrap.Enabled
 	certsManagedByKubeAws := c.ManageCertificates
-	caKeyRequiredOnController := certsManagedByKubeAws && tlsBootstrappingEnabled
+	caKeyRequiredOnController := certsManagedByKubeAws
 
 	fmt.Printf("--> Summarizing the configuration\n    Kubelet TLS bootstrapping enabled=%v, TLS certificates managed by kube-aws=%v, CA key required on controller nodes=%v\n", tlsBootstrappingEnabled, certsManagedByKubeAws, caKeyRequiredOnController)
 


mad01:
i have a cluster that is created with TLSBootstrap set to false. Is this potentially also fixing the issue that `worker-ca.pem` is required when TLSBootstrap is false ? 


this is my error output i get 
```
Success! Stack rendered to ./stack-templates.

Next steps:
1. (Optional) Validate your changes to cluster.yaml with "kube-aws validate"
2. (Optional) Further customize the cluster by modifying templates in ./stack-templates or cloud-configs in ./userdata.
3. Start the cluster with "kube-aws up".
Error: Failed to initialize cluster driver: failed to read/create encrypted assets: Error reading credential file credentials/worker-ca.pem: credentials/worker-ca.pem must exist. Please confirm that you have not deleted the file manually
```

mumoshu:
@jlk Thanks for your effort!

Just for sync up, I guess we shouldn't require ca-key to be there when TLS bootstrapping is disabled, in theory and in the first place. So, probably an alternative fix for #960 
would be to fix kube-aws to NOT to unnecessary read ca-key.pem when TLS bootrapping is disabled.

jlk:
@mumoshu That makes sense on the surface - will take a look and update the code.

PR #1114, opened 2018-01-17
Migrate to golang/dep

### LINKED ISSUE #983: Migration to dep package manager
As [glide is now officially recommending dep](https://github.com/Masterminds/glide/releases/tag/v0.13.0), I'm looking forward to start using dep also in kube-aws.

It is a completely different project than kube-aws but please let me introducing https://github.com/crewjam/saml/pull/114 - it looks like a very nicely authored example pull request for introducing dep into an existing golang project 😉

mumoshu:
In nutshell, run `dep init` and then modify `Gopkg.toml` to include a specific `ignored` statement to prevent dep from scanning deps of `hack/relnote.go`, then run `dep ensure`.
Your are now ready to go.

Resolves #983


mumoshu:
Note that running `dep prune` would remove bunch of unnecessarily vendored files from the tree.
However, I have refrained from doing it in this PR to complicate reviews.

mumoshu:
Just run `make build` and `./containerized-build-release-binaries` - both worked fine, of course!

PR #1116, opened 2018-01-17
conditionally deploy kube-dns to controllers

zonzamas:
I have an scenario where kube-dns needs to be deployed on controllers instead of K8s workers. I have added a flag to allow this configuration.


mumoshu:
@zonzamas Thanks for your contribution!

Interesting idea, but please be aware of scaling kube dns accordingly to your workloads and num of worker nodes.

Perhaps you'd also want to enable kubeDns.nodeLocalResolver for better scalability?

PR #1117, opened 2018-01-18
Add priorityClassName to cloud-config-controller

sergi:
This adds a baseline priority of `system-node-critical` to system critical pods, instead of the default value of `0`.

[`system-node-critical`](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/priority/admission.go#L48) is a System Priority class, which is used by system critical pods that should not be preempted by workload pods.

`priorityClassName` is only set if the priority feature is enabled.


sergi:
I just signed the CLA

mumoshu:
@sergi Sounds great. Thanks for your contribution 👍

PR #1118, opened 2018-01-23
Tidy up markdown

whereisaaron:
Tidy up markdown, especially for pre-indented / fixed-width content, and a heading.

Thanks for this contribution @camilb! I didn't even know it was here. @mumoshu we should link to these contrib folders somewhere in the docs, like just a bullet-list on the 'add-ons' page or an extra 'Advanced Topics' page?

### DIFF
--- a/contrib/dex/README.md
+++ b/contrib/dex/README.md
@@ -1,14 +1,14 @@
 ## Configure Dex as a custom provider in cluster.yaml
 
 Example: 
-
+```
        oidc:
          enabled: true
          issuerUrl: "https://dex.example.com"
          clientId: "example-app"
          usernameClaim: "email"
          groupsClaim: "groups"
-
+```
 
 ## Deploy Dex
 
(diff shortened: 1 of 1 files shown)


c-knowles:
I think we should include a link in the advanced topics for now or perhaps we could just combine these contrib docs into the main docs? Having a single place for documentation would makes things simpler.

whereisaaron:
My suggestion is while they are 'contrib' status @c-knowles , the files and docs probably live or languish in the 'contrib' folder, with just a link from the main docs: 'You might be interested in...'. Otherwise we should bring it into the main code base as an official ('experimental') add-on and then we would definitely merge the docs into the main add-on docs.

mumoshu:
@whereisaaron Thanks for your contribution! LGTM.

I'm rather inclined to combine contrib docs to the main one, as long as the reader is able to notice that it is actually from contrib.

Issue #1105, opened 2018-01-06
'kube-node-drainer' does not work 'kube2iam' in 0.9.9


whereisaaron:
I have `kube-aws` 0.9.9 clusters (us-east-2) with `kube-node-drainer` enabled and running, but it doesn't actually appear to be working. I don't see any sign of pods draining during ASG termination. The controller doesn't reschedule any pods until well after the instance has terminated and the `nodeMonitorGracePeriod` has elapsed (which I set to 90s to test). All the pods then get rescheduled at the same instant.

The ASG instances spend 5-10 minutes in `Terminating: Pending` state before transitioning to `Terminating: Proceed`. At that point all connection to the instance drops. But only a couple minutes later do any pods get rescheduled.

A log from a `kube-node-drainer` on a node being terminated is below. It gets a 404 response from the AWS API, right up to the point when the connection drops.

The clusters have `kube2iam` installed, does that interfer or block the AWS API access for `kube-node-drainer`?

Output of `kubectl -n kube-system logs kube-node-drainer-ds-289hr --follow`
```
<=== At this point the instance is "Terminating: Wait" in ASG console
+ curl -o /dev/null -w %{http_code} -sL http://169.254.169.254/latest/meta-data/spot/termination-time
+ http_status=404
+ [ 404 -eq 200 ]
+ [ -e /etc/kube-node-drainer/asg ]
+ grep -q i-08fc5fb751b30d164 /etc/kube-node-drainer/asg
+ sleep 10
+ curl -o /dev/null -w %{http_code} -sL http://169.254.169.254/latest/meta-data/spot/termination-time
+ http_status=404
+ [ 404 -eq 200 ]
+ [ -e /etc/kube-node-drainer/asg ]
+ grep -q i-08fc5fb751b30d164 /etc/kube-node-drainer/asg
+ sleep 10
+ curl -o /dev/null -w %{http_code} -sL http://169.254.169.254/latest/meta-data/spot/termination-time
+ http_status=404
+ [ 404 -eq 200 ]
+ [ -e /etc/kube-node-drainer/asg ]
+ grep -q i-08fc5fb751b30d164 /etc/kube-node-drainer/asg
+ sleep 10
+ curl -o /dev/null -w %{http_code} -sL http://169.254.169.254/latest/meta-data/spot/termination-time
+ http_status=404
+ [ 404 -eq 200 ]
+ [ -e /etc/kube-node-drainer/asg ]
[...]

PR #1123, opened 2018-01-30
Add [optional] explicit IAM role specification to NodeDrainer

ivanilves:
Add [optional] explicit IAM role specification to NodeDrainer.

If you are happy **kube2iam** user depending on your setup you may want **no AWS credentials** being available on pods having no IAM role specifications. At least I believe default access is an evil thing to be avoided.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -104,6 +104,7 @@ func NewDefaultCluster() *Cluster {
 		NodeDrainer: model.NodeDrainer{
 			Enabled:      false,
 			DrainTimeout: 5,
+			IAMRole:      model.IAMRole{},
 		},
 		Oidc: model.Oidc{
 			Enabled:       false,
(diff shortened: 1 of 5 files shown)


ivanilves:
Maybe, not sure, but specifying explicit IAM role could help with https://github.com/kubernetes-incubator/kube-aws/issues/1105

mumoshu:
@ivanilves Thanks for the fix. Yes, I began to believe we'd need to do this everywhere!

Btw, would you mind naming it `iamRole.arn`? So that we could enhance it to understand e.g. `iamRole.create: true` to let kube-aws create the role for you in the future.

ivanilves:
Hey @mumoshu I've followed your advise.

Could you please re-review and comment, maybe something more needs to be changed?

mumoshu:
@ivanilves Thanks a lot for taking your time! LGTM.

ivanilves:
Thank you! :relaxed: :clap:

Issue #1125, opened 2018-02-06
wget in node drainer skip has stopped working


c-knowles:
Extracted from https://github.com/kubernetes-incubator/kube-aws/issues/1105#issuecomment-363316491.

The current node drainer scripting uses wget but it seems to have some problems with 502s:
```
wget -O - -q http://169.254.169.254/2016-09-02/meta-data/instance-id
wget: server returned error: HTTP/1.1 502 Bad Gateway 
```

On a `Running` node drainer pod, I've exec-ed to the pod and done this:
```
/ # wget -O - -q http://169.254.169.254/latest/meta-data/
wget: server returned error: HTTP/1.1 502 Bad Gateway

/ # curl http://169.254.169.254/latest/meta-data/
ami-id
ami-launch-index
ami-manifest-path
block-device-mapping/
hostname
[...]
```

Probably related issue in wget in busybox: http://svn.dd-wrt.com/ticket/5771

mumoshu:
Thanks for reporting!
Not sure this is affecting everyone's cluster, but let's improve node drainer to use `curl` instead and see if it works, according to our observation in the referenced issue.

vaibhavsingh97:
@c-knowles @mumoshu How can I solve this? I am new to K8s 😄

camilb:
@c-knowles @mumoshu This happens on new nodes or nodes that recently pulled the hyperkube container image. The hyperkube doesn't contain the `wget`or `curl` anymore.

camilb:
@c-knowles @mumoshu This happens for me on new nodes or nodes that recently pulled the aws-cli container image. 

```bash
bash-4.3# wget -O - -q http://169.254.169.254/2016-09-02/meta-data/instance-id
wget: error getting response
bash-4.3# apk update
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
v3.6.2-248-g3f8eeb3ea1 [http://dl-cdn.alpinelinux.org/alpine/v3.6/main]
v3.6.2-250-g51a3714b5e [http://dl-cdn.alpinelinux.org/alpine/v3.6/community]
OK: 8437 distinct packages available
bash-4.3# apk add wget
(1/1) Installing wget (1.19.2-r0)
Executing busybox-1.26.2-r9.trigger
[...]

PR #1127, opened 2018-02-08
Swap out wget for curl in node drainer scripts

c-knowles:
Resolves https://github.com/kubernetes-incubator/kube-aws/issues/1125.

Also both curl and wget used within these scripts, better to use just one of them.

I went with the option to fail fast if the curl fails.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1362,7 +1362,7 @@ write_files:
                   - /bin/sh
                   - -xec
                   - |
-                    metadata() { curl -s -f http://169.254.169.254/2016-09-02/"$1"; }
+                    metadata() { curl -s -S -f http://169.254.169.254/2016-09-02/"$1"; }
                     asg()      { aws --region="${REGION}" autoscaling "$@"; }
 
                     # Hyperkube binary is not statically linked, so we need to use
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1362,7 +1362,7 @@ write_files:
                   - /bin/sh
                   - -xec
                   - |
-                    metadata() { wget -O - -q http://169.254.169.254/2016-09-02/"$1"; }
+                    metadata() { curl -s -f http://169.254.169.254/2016-09-02/"$1"; }
```
mumoshu: `-s` also suppresses any error message, which seems undesirable.
How about `--silent --show-error` instead so that it still prints error messages?

```
$ curl --fail --silent http://httpstat.us/502
$ echo $?
22
```

```
$ curl --fail --silent --show-error http://httpstat.us/502
curl: (22) The requested URL returned error: 502 Bad Gateway
$ echo $?
22
```

c-knowles: @mumoshu sure I've done that. I've kept the short flags to be consistent with the rest of the file.

There are many other curls in this file that silence but do not have the show error flag, probably we should change them all. There's many without the specific fail flag too which is concerning (some valid when we want to check HTTP response code).

mumoshu: Thanks! Yeah, short flags would be better for consistency.

> There are many other curls in this file that silence but do not have the show error flag, probably we should change them all.

Can't agree more.
Would you mind fixing these issues too, in an another PR? 🙇

mumoshu:
Thanks as always for the fix!

PR #1129, opened 2018-02-08
Template kube-dns-autoscaler parameters

SomeoneWeird:
We need to customise these values in our cluster. Seeing as it's always deployed (and not optional) I have just uncommented the `kubeDns` section in `cluster.yaml` and added an `autoscaling` key with the values in it. If you would rather I can add default values in the template so we can keep the config commented out.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -2380,7 +2380,7 @@ write_files:
                   - --namespace=kube-system
                   - --configmap=kube-dns-autoscaler
                   - --target=Deployment/kube-dns
-                  - --default-params={"linear":{"coresPerReplica":256,"nodesPerReplica":16,"min":2}}
+                  - --default-params={"linear":{"coresPerReplica":{{ .KubeDns.Autoscaler.CoresPerReplica }},"nodesPerReplica":{{ .KubeDns.Autoscaler.NodesPerReplica }},"min":{{ .KubeDns.Autoscaler.Min}}}}
                   - --logtostderr=true
                   - --v=2
 
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -154,6 +154,11 @@ func NewDefaultCluster() *Cluster {
 			KubeDns: KubeDns{
 				NodeLocalResolver:   false,
 				DeployToControllers: false,
+				Autoscaler: KubeDnsAutoscaler{
+					CoresPerReplica: 256,
+					NodesPerReplica: 16,
+					Min:             2,
+				},
 			},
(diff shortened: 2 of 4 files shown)


SomeoneWeird:
@k8s-ci-robot signed

SomeoneWeird:
Not sure I've done this correctly or not, [the docs](https://kubernetes-incubator.github.io/kube-aws/guides/developer-guide.html#modifying-templates) say to run `make build` to rebuild the templates, but it doesn't seem to be doing anything for me.

mumoshu:
Do you see `core/controlplane/config/templates.go` generated in your work-tree?
But don't worry - we don't managed that in git!

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -2380,7 +2380,7 @@ write_files:
                   - --namespace=kube-system
                   - --configmap=kube-dns-autoscaler
                   - --target=Deployment/kube-dns
-                  - --default-params={"linear":{"coresPerReplica":256,"nodesPerReplica":16,"min":2}}
+                  - --default-params={"linear":{"coresPerReplica":{{ .kubeDns.autoscaling.coresPerReplica }},"nodesPerReplica":{{ .kubeDns.autoscaling.nodesPerReplica }},"min":{{ .kubeDns.autoscaling.min}}}}
```
mumoshu: I guess `kubeDns` should be capitalized so that it is `KubeDns`.

That would be the cause of failures seen in the travis build:

```
--- FAIL: TestRenderStackTemplate (0.11s)
	Error Trace:	cluster_test.go:768
			helper.go:100
			helper.go:22
			cluster_test.go:772
	Error:      	Received unexpected error:
	            	failed to render controller cloud config: template: ../config/templates/cloud-config-controller:2383:77: executing "s3" at <.kubeDns.autoscaling...>: can't evaluate field kubeDns in type *config.Config
	Messages:   	Unable to initialize Cluster
```
https://travis-ci.org/kubernetes-incubator/kube-aws/builds/339246310

SomeoneWeird:
@mumoshu I've updated the template with a combination of capitalisation but can't seem to get it to pass these tests.

mumoshu:
Thanks for taking your time!
According to the build output, it seems like being killed in the middle of execution by the 20-min timeout set for go test.
Maybe due to temporal slowdown of travis?

Will rerun it to see if it works.

SomeoneWeird:
Looks like it's still failing unfortunately - any idea?

SomeoneWeird:
@mumoshu Looks like this finally passed!

mumoshu:
LGTM. Thank you very much for your patience and efforts, @SomeoneWeird!

PR #1134, opened 2018-02-14
feat(integration): WIP: kiam support

### LINKED ISSUE #1055: Integration with kiam
kiam seems to succeed kube2iam on several points:

- Built on a master/agent architecture so that we need IAM permissions only on controller nodes
- Restriction of allowed IAM roles per namespace
- Reliability fix

mumoshu:
Resolves #1055


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -3569,6 +3592,253 @@ write_files:
...
+            tolerations:
+            - operator: Exists
+              effect: NoSchedule
+            - operator: Exists
+              effect: NoExecute
+            - operator: Exists
+              key: CriticalAddonsOnly
+            serviceAccountName: kiam-server
+            nodeSelector:
+              kubernetes.io/role: master
```
mumoshu: Ha, just realized that kiam's default template spec relies on `kubernetes.io/role: master`.
kube-aws has been using `node-role.kubernetes.io/master: ""`.
Both seems valid according to: https://github.com/kubernetes/kubernetes/issues/50010 and https://github.com/kubernetes/kubernetes/pull/51173

mumoshu:
```
$ k get no
NAME                                            STATUS    ROLES     AGE       VERSION
ip-10-0-1-13.ap-northeast-1.compute.internal    Ready     node      2m        v1.9.1
ip-10-0-2-170.ap-northeast-1.compute.internal   Ready     master    8m        v1.9.1
ip-10-0-3-77.ap-northeast-1.compute.internal    Ready     node      2m        v1.9.1
kuoka-yusuke-2:k8s3 kuoka-yusuke$ k get po
NAME                                                                    READY     STATUS    RESTARTS   AGE
calico-kube-controllers-7bcff664bc-trm42                                1/1       Running   0          8m
calico-node-224cr                                                       1/1       Running   0          8m
calico-node-jjvxc                                                       1/1       Running   0          2m
calico-node-wl99g                                                       1/1       Running   0          2m
cluster-autoscaler-59c6556b99-zmbbs                                     1/1       Running   5          8m
dnsmasq-node-7hxhr                                                      2/2       Running   0          2m
dnsmasq-node-dgnp4                                                      2/2       Running   0          8m
dnsmasq-node-xq5vq                                                      2/2       Running   0          2m
heapster-7f977f75f7-vsps5                                               2/2       Running   0          1m
kiam-agent-wjzts                                                        1/1       Running   0          2m
kiam-agent-zd5lq                                                        1/1       Running   0          2m
kiam-server-w654r                                                       0/1       Running   7          8m
kube-apiserver-ip-10-0-2-170.ap-northeast-1.compute.internal            1/1       Running   0          7m
kube-controller-manager-ip-10-0-2-170.ap-northeast-1.compute.internal   1/1       Running   0          7m
kube-dns-autoscaler-c895d6bbd-jhgrp                                     1/1       Running   0          8m
kube-dns-d87fdd65-7jpqr                                                 3/3       Running   0          8m
kube-dns-d87fdd65-hrt64                                                 3/3       Running   0          1m
kube-proxy-cz2lw                                                        1/1       Running   0          8m
kube-proxy-dsrt8                                                        1/1       Running   0          2m
kube-proxy-r88wh                                                        1/1       Running   0          2m
kube-rescheduler-7b75d6b8db-m72hn                                       1/1       Running   0          8m
kube-scheduler-ip-10-0-2-170.ap-northeast-1.compute.internal            1/1       Running   0          7m
kubernetes-dashboard-6ff488bcdd-zh7dd                                   1/1       Running   0          8m
tiller-deploy-8698c78577-6vvpn                                          1/1       Running   0          8m
```

Seems like I'm almost there, but not sure why readinessProbe of the kiam-server never pass

mumoshu:
exec'ed into the pod and manually run `/health ...`:

```
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no address available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no connection available
WARN[0001] error checking health: rpc error: code = Unavailable desc = there is no connection available
```

mumoshu:
Turns out something is going wrong with kube-aws generated certs & key.

It worked with the tls assets generated by the procedure described in https://github.com/uswitch/kiam/blob/master/docs/TLS.md

mumoshu:
Ah, I see that [kiam-server uses TLS mutual auth](https://github.com/uswitch/kiam/blob/master/pkg/server/server.go#L180-L182). I'd need specifically created cert/key pair for that!

mumoshu:
This was the key point: `ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},`.
You need both. kube-aws generated certs had been having only either of these.

pingles:
@mumoshu exciting ;)

Let me know if there's improvements we can make to the deployment manifest examples, docs, readme etc. to make it easier to get deployed- know that there's a little bit of jumping around that users have to do currently.

mumoshu:
@pingles Thanks for your kind words!

Maybe some explanation in your doc about for which purpose each cert/key pair is used would help.
It seems like kiam-server cert/key are used for both server auth and client auth, whereas agent seems to use it only for client auth(implied by its cert misses CN, right?).
It took some time for me to figure that out.

whereisaaron:
Super great @mumoshu. I was thinking to use `cert-manager` to issue and renew the certs for `kiam`. Since you can create your own root CA, and then automatically issue and renew client/server certs.

mumoshu:
@whereisaaron Sweet! Please, please submit a pull request for doc and/or feature if you managed to implement it 👍

pingles:
I’d love integration with automated cert management.

One thing to watch though- currently Kiam doesn’t check the identity of the
client, just that they have a cert that can be verified. In the docs at the
moment that’s because it uses a standalone CA only used for server and
agent.

Integrating with other CAs we’d need to add something to ensure the server
confirms the identity of the agent so that other processes that might be
able to generate certs from the same CA would still be unauthorised.

Not sure if I’ve explained that super well? ;)

I’m totally keen to do it- either to integrate certs with cert manager,
vault or k8s’ cert api.

If someone would be up for helping, let me know and let’s get something on
the Kiam issues list!

whereisaaron:
@pingles `cert-manager` 'Issuer' and 'Certificate' resources are namespaced. So if you create the 'Issuer' for the `kiam` CA in a `kiam` namespace or the `kube-system` namespace (and use RBAC), then only people/Pods/processes who can create 'Certificate' resources in that namespace are able to issue certificates signed by that CA. The Secrets are created in that namespace too, so only `kiam` Pods/processes can use them.

`cert-manager` is a cluster service that doesn't just provide just one authority, it supports any number of external or internal CA's, with each either namespaced (single namespace), or with cluster-wide scope. So you can easily have a private CA per namespaced application, and that application can issue automatically renewing certificates that are private to that application.

There are instructions ['Creating a simple CA based issuer'](https://github.com/jetstack/cert-manager/blob/master/docs/user-guides/ca-based-issuer.md). Just note that the instructions currently forget to tell you to [add the `CA:true` attribute to your self-signed root CA](https://github.com/jetstack/cert-manager/issues/279).

There is a plan to use `cert-manager` for [issuing Istio certificates](https://github.com/jetstack/cert-manager/issues/304) also.

My idea was to have a `helm` chart that deploys `kiam` server and agents to its own namespace with appropriate RBAC and  `cert-manager` Issuer and Certificate resources.

@mumoshu heaps of your patch is about issuing certificates for `kiam` which we then have no (easy) way to maintain/renew after cluster launch. So something like `cert-manager` could reduce maintenance and certificate-expiring outages. Better we roll `cert-manager` into `kube-aws` launches and then use it to issue the Certificates we need for other in-cluster services (probably not an option for controller/etcs/node certs though).

(The [k8s cluster certificate service](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/) is not currently an alternative for this, as it is very limited. Only one cluster-wide signing authority, you have to create your own CSR's and download the certficate to a Secret yourself! And you have to either manually approve them, or write your own approval plugin.)

pingles:
Awesome- that sounds perfect then! Just wanted to raise it in case it
didn't- was an assumption we made but may not be immediately obvious.

mumoshu:
Sounds great 👍 Would you mind creating a dedicated github issue for tracking the possible cert-manager support?

mumoshu:
Verified that this is working, by running something like:

```
#!/usr/bin/env bash

pod=iamtest-$(date +%s)
role=k8s-service-role

kubectl run -it --restart=Never --namespace kube-system \
  --image=kubeaws/awscli:0.9.0 $pod \
  --overrides '{"metadata":{"annotations":{"iam.amazonaws.com/role":"'$role'"}}}' -- bas
```

And executing `aws` commands in the shell.

mumoshu:
We may want `--assume-role-arn` of kiam to be configurable as per https://github.com/uswitch/kiam/blob/master/docs/IAM.md#server-iam-role, maybe.
It would be useful especially when we have many IAM roles assumed by kiam, while manually updating every assumed role to trust the kube-aws managed role seemed not sustainable.

This isn't kiam specific but also applies to kube2iam and even AWS IAM in general, but also note that you can trust every role by doing [the suggested setting](https://forums.aws.amazon.com/message.jspa?messageID=770658#770658).

mumoshu:
@whereisaaron 

> kiam which we then have no (easy) way to maintain/renew after cluster launch. So something like cert-manager could reduce maintenance and certificate-expiring outages. Better we roll cert-manager into kube-aws launches and then use it to issue the Certificates we need for other in-cluster services (probably not an option for controller/etcs/node certs though).

I can't agree more :) So, integrating something like cert-manager would be the way to go in the longer term.

However, you may already know, but just using cert-manager doesn't solve the problem as a whole.

More concretely, how would you trigger rolling-update of pods in a proper order to avoid downtime, in case of key/certificate update? cert-manager seems to just update k8s secrets containing keys and certs, but it doesn't coordinate restarts of its consumers of course, right? Then, we need to implement that part ourselves.

With current implementation, you can manually update part of `credentials/*.pem` and run `kube-aws update` which will rolling-update necessary nodes and hopefully affected secrets and pods to achieve. In that sense, it covers the whole lifecycle of key/cert. But the u/x is not that good until we somehow address issues like #1147 #1146.

So, assuming you're saying we should adopt cert-manager TODAY, my answer would be no.
My best bet as of today is start a dedicated, umbrella issue to track the overall progress towards the future integration of cert-manager or alike.

But please feel free to correct my I'm missing something and you think we can integrate it into kube-aws without hassle today!

WDYT?

whereisaaron:
No, I agree.  I see `cert-manager` an alternative way to deploy `kiam` such that it can manage and rotate its own certificates, with no dependence on `kube-aws`. So just helm chart deployed after cluster creation. You are right, unless `kiam` automatically reloads its Secrets occasionally, then something would need to trigger reload.

PR #1141, opened 2018-02-20
Add Mutating/ValidatingAdmissionWebhook to Admission Control List

shraykay:
In order to support Istio automatic sidecar injection in Kubernetes 1.9, these two experimental admission controllers are required.

https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use

https://istio.io/docs/setup/kubernetes/sidecar-injection.html#automatic-sidecar-injection

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -58,6 +58,12 @@ func NewDefaultCluster() *Cluster {
 			Priority{
 				Enabled: false,
 			},
+			MutatingAdmissionWebhook{
+				Enabled: false,
+			},
+			ValidatingAdmissionWebhook{
+				Enabled: false,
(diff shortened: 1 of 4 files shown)


mumoshu:
Note from @shraykay:

You need two additional flags for apiserver to actually use istio sidecar-injection:

```
--proxy-client-cert-file=/etc/kubernetes/ssl/apiserver.pem
--proxy-client-key-file=/etc/kubernetes/ssl/apiserver-key.pem
```

You can add it by manually editing userdata/cloud-config-controller.

mumoshu:
@shraykay LGTM! Thanks for your contribution 👍

Issue #464, opened 2017-03-26
Suggestion for consolidating our efforts


mumoshu:
To be short, I'd like to find a way to merge our efforts in the area of kubernetes deployment tools.
Would "you" mind suggesting how's and/or express agreement(s) to one of the suggestions possibly in this thread?

Please read [the kubernetes-dev thread](https://groups.google.com/forum/#!topic/kubernetes-dev/diN0KIOpLro) for more context.

Merging our efforts would enable:

* Every kubernetes user easily get to one or two appropriate tools according to one's requirements
* We, tool developers, won't need to redo works already done for another tools

The list of tools includes:
* kube-aws
* kops
* kargo
* kubernetes-anywhere
* bootkube
* kubeadm
* etc.

---

As of today, kube-aws already plans to [integrate with bootkube in the future](https://github.com/kubernetes-incubator/kube-aws/blob/master/ROADMAP.md#v09x).

However I want more; just for example, how about merging kube-aws into kops? More concretely, I imagine about kube-aws being the cloudformation provider for kops.

Of course, being one of providers for kops doesn't mean kube-aws can lose the following important features:

* Manageable with SCM like Git to support customizations of clusters (#238)
* Based on CloudFormation 😃 
* Integrated node pools
* H/A everywhere
* Flexibility to allow customizing via cluster.yaml, stack template templaets, stack templates, userdata
* Bring your own AWS resources to be used (where necessary, where possible)
* Various work-arounds to go beyond AWS limits(16KB in userdata, 512KB in cfn stack template's size)
* etc.

I'll keep every aspect of kube-aws for our current users, as much as possible. However, I don't believe that is something prevents us from merging our efforts.

Personally, [I suggested in sig-aws about merging kube-aws into kops](https://groups.google.com/forum/#!topic/kubernetes-sig-aws/L1E2F2Easbw).
Would someone agrees and/or wants to take a leadership to make it a reality?

redbaron:
I appreciate effort to avoid duplication, but merging with kops would be a major disappointment for us, for following few, but very important reasons:
 1. kops codebase is huge, last time I checked it was ~40k LOC vs ~4k LOC in kube-aws (excluding tests and vendored libs). Kops tries to reimplement what TF/CF do well. Relative ease of adding custom features to  a tool influenced our choice significantly.
[...]

PR #1154, opened 2018-02-22
Update ROADMAP.md

mumoshu:
Improving our roadmap according to the states of other projects and also use-cases and feedbacks I've received since last release

cc @camilb @c-knowles @danielfm @redbaron and everyone relying on/following this project

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -78,16 +78,25 @@ Please file an issue to make suggestions on this roadmap!
 
 ## v0.9.10
 
-  * kubeadm support to simplify k8s components configuration (#654)
-    * Reduces the amount of code required in kube-aws
-    * To better follow upstream improvements on how k8s components are deployed
-  * [More manageable Calico + Flannel](https://github.com/kubernetes-incubator/kube-aws/pull/675#issuecomment-303669142) (@redbaron)
-  * istio integration
-    * Probably after k8s supported injecting init containers from PodPreset
-      * [Upstream issue](https://github.com/kubernetes/kubernetes/issues/43874)
+  * Kubernetes 1.9.x
+  * Security+Usability improvements
+    * [kiam](https://github.com/uswitch/kiam/) integration (#1055)
+    * [authenticator](https://github.com/heptio/authenticator) integration (#1153)
+    * Support for pregenerating IAM roles used by kube2iam/kiam (#1145, #1150)
+  * Operatability improvements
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -78,16 +78,25 @@ Please file an issue to make suggestions on this roadmap!
...
     * To better follow upstream improvements on how k8s components are deployed
-  * [More manageable Calico + Flannel](https://github.com/kubernetes-incubator/kube-aws/pull/675#issuecomment-303669142) (@redbaron)
-  * istio integration
+  * (After scalability/reliability/upgradability cleared) istio integration
     * Probably after k8s supported injecting init containers from PodPreset
       * [Upstream issue](https://github.com/kubernetes/kubernetes/issues/43874)
-
-## v0.9.11
-
   * Migrate from coreos-cloudinit to ignition for node bootstrapping (@redbaron)
```
mumoshu: Honestly, I don't have specific use-cases require ignition.

Can't we just use bash rather than coreos-cloud-init or ignition for most of node configuration steps? :)
The goal of node configuration is just reliably fetching assets from S3 and configuring systemd units.

Note that even with that direction, we could make ignition rather than cloud-init to be the entrypoint and just call the kube-aws's bash script for node configuration. So that users can still modify ignition config directly when necessary.

I'm just saying that we don't necessarily migrate all the things we've been doing in cloud-config to ignition.

c-knowles:
Also no specific need for ignition for us. There's chatter in various issues about performing operations more and more outside of CloudFormation, does it mean kube-aws functionality will more closely align with kops?

mumoshu:
Thx!

> does it mean kube-aws functionality will more closely align with kops?

Does it? I really don't know how kops works under the hood!

c-knowles:
@mumoshu hmmm, I think perhaps. kops has always used a state sync model which seems to be implied by some of our issues although I'm not sure what the plan is for those. I think since we last chatted about it kops has added HA master, declarative YAML and CloudFormation support.

mumoshu:
@c-knowles Yep! kops is progressing in a surprising pace!

JFYI, when I've revisited about kops last time, CloudFormation support seemed to have lacked supports for the 16K userdata limit and the CFN stack template size limit. I remember that I've commented about that somewhere in the kops issues.

You may know but kube-aws had addressed the two in its early days. So, I guess there may be still many gotchas when someone tried to switch between kops and kube-aws. I'm not happy about that yet.

And I have not yet given up on #464!

PR #1171, opened 2018-03-07
Fix validation failure on gp2 volume type for node pools

luck02:
Improved tests and code to respect changes in EC2 validate volume API. We can no longer submit IOPS 0 as being synomous with unset field / null.

In `core/nodepool/cluster.go`, the function: `validateWorkerRootVolume` is changed to call an another function to generate EC2 volume parameters. Also tweaked tests around that.

### DIFF
--- a/core/nodepool/cluster/cluster.go
+++ b/core/nodepool/cluster/cluster.go
@@ -197,13 +197,7 @@ func (c *ClusterRef) validateKeyPair(ec2Svc ec2DescribeKeyPairsService) error {
 func (c *ClusterRef) validateWorkerRootVolume(ec2Svc ec2CreateVolumeService) error {
 
 	//Send a dry-run request to validate the worker root volume parameters
-	workerRootVolume := &ec2.CreateVolumeInput{
-		DryRun:           aws.Bool(true),
-		AvailabilityZone: aws.String(c.Subnets[0].AvailabilityZone),
-		Iops:             aws.Int64(int64(c.RootVolume.IOPS)),
-		Size:             aws.Int64(int64(c.RootVolume.Size)),
-		VolumeType:       aws.String(c.RootVolume.Type),
-	}
+	workerRootVolume := c.getWorkerRootVolumeConfig()
 
 	if _, err := ec2Svc.CreateVolume(workerRootVolume); err != nil {
 		operr, ok := err.(awserr.Error)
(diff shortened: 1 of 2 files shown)


core/nodepool/cluster/cluster.go
```diff
@@ -216,6 +210,44 @@ func (c *ClusterRef) validateWorkerRootVolume(ec2Svc ec2CreateVolumeService) err
 	return nil
 }
 
+func (c *ClusterRef) getWorkerRootVolumeConfig() *ec2.CreateVolumeInput {
+	var workerRootVolume = &ec2.CreateVolumeInput{}
+
+	switch c.RootVolume.Type {
+	case "standard":
```
mumoshu: Almost nit but `case "standard", "gp2":` would be a bit more concise!

core/nodepool/cluster/cluster_test.go
```diff
@@ -40,9 +40,14 @@ func (svc dummyEC2CreateVolumeService) CreateVolume(input *ec2.CreateVolumeInput
 		)
 	}
 
-	if aws.Int64Value(input.Iops) != aws.Int64Value(expected.Iops) {
+	if (input.Iops == nil && expected.Iops != nil) ||
```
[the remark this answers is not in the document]
mumoshu: Makes sense!

core/nodepool/cluster/cluster.go
luck02: Certainly

luck02:
@mumoshu I've pushed a fix, as previous I ran the unit tests locally, not e2e tests. 

If I were to push a hotfix branch (Including this and https://github.com/kubernetes-incubator/kube-aws/pull/1172) to 0.9.8 (IE a 0.9.8+hotfix) etc(which we're on now) is there any chance we could get that released?  We plan on upgrading to the newest version of kube-aws soonish but we're partially blocked right now.  

Thanks!

mumoshu:
@luck02 Thanks! LGTM.

Regarding the hotfix release, honestly saying, I don't have bandwidth to do so. Would you mind trying to build it yourself? Probably there might be no conflict which will complicate building your own fork of kube-aws.

But please feel free to ask me anything if you had troubles while building your own kube-aws binary.

Thanks for your support!

luck02:
@mumoshu we've built it, that's not the issue, more along the lines of distributing from a known upstream.

We can get along, I wasn't aware that doing a hotfix was a major effort.  If possible i'd be happy to help there.  

I mean we can get along without it, however I doubt we're the only people who will / have run into this issue.  

Thanks for the assistance!

Issue #1169, opened 2018-03-06
kube-aws / autoscaler add permissions "autoscaling:DescribeLaunchConfigurations"


luck02:
This autoscaler doc's state that to scale up from 0 instances the permission: `autoscaling:DescribeLaunchConfigurations` is required.

https://github.com/kubernetes/autoscaler/blob/5302c37d6d6e89a2eeac0c74b4afe6caadfa3340/cluster-autoscaler/cloudprovider/aws/README.md#scaling-a-node-group-to-0

How / where do we specify this in kube-aws or is that something we'd do orthogonally?

Thanks!

mumoshu:
@luck02 Thanks for the report!

Adding the action to https://github.com/kubernetes-incubator/kube-aws/blob/9df90636b0ad3df8453cfad6b4d7f9bcedacf228/core/controlplane/config/templates/stack-template.json#L252 would do the job.

You can do it for your `./stack-templates/control-plane.json.tmpl` generated by `kube-aws render stack`, or even better submit a PR to add to the line I've noted above.

luck02:
@mumoshu pushed a PR above.

whereisaaron:
Thanks for identifying this @luck02!

luck02:
So sorry, I :copy-pastad: the wrong Cloud Formation permission :(

New PR: https://github.com/kubernetes-incubator/kube-aws/pull/1181

luck02:
@mumoshu sorry, I'm unable to re-open issue for PR: https://github.com/kubernetes-incubator/kube-aws/pull/1181  

I can open a new issue if that's better for your workflow.

PR #1172, opened 2018-03-08
added autoscaling:DescribeAutoScalingGroups to stack-template

luck02:
core/controleplane/config/templates

When scaling from 0 -> n nodes we require DescribeAutoscalingGroups in order to scale up based on docs: https://github.com/kubernetes/autoscaler/blob/5302c37d6d6e89a2eeac0c74b4afe6caadfa3340/cluster-autoscaler/cloudprovider/aws/README.md#scaling-a-node-group-to-0

fixes: https://github.com/kubernetes-incubator/kube-aws/issues/1169


mumoshu:
@luck02 LGTM! Thanks a lot for your support 🍻

PR #1188, opened 2018-03-15
closes #1186 - removed go generate

### LINKED ISSUE #1186: remove `go generate` code
we would like to use the `core` of the project with our custom commands. Basically we need to replace `init` and `render credentials`, because we are using custom templates to generate `cluster.yaml` and vault to generate our credentials. The idea is to split this project into 'core' and 'cmd', so core could be used in other projects.

We would like to import this project as dependency and we would only use the core (not the commands). But what makes this hard is `go:generate` code. If the project is imported the paths are incorrect, plus `dep` tool refuses to import non-go code, so we cannot get the templates.

My suggestion is to move templates into go code (basically run build, print byte arrays and copy and paste these and add to git). Then we could delete go generate statements, codegen package and all the json/tmpl files.

This would make it easier to write tests (no interaction with filesystem, no need to generate templates - just run go test), easier to update templates (no need to run go generate, just go build), safer ...

I've started the work, but it looks like there's a lot of changes, so just wanted to know if you are happy with me making this change.

pete911:
this is initial attempt, let me know what you think, if you are happy with the change, I will update the templates that changed recently


pete911:
ok, up to date with master, all tests are passing

build
```diff
@@ -18,10 +18,6 @@ fi
 
 echo Building kube-aws ${VERSION}
 
-go generate ./core/controlplane/config
-go generate ./core/nodepool/config
-go generate ./core/root/config
```
mumoshu: Hi! Thank you very much for your interest in kube-aws and the efforts.

Basically, I appreciate it if we could keep our existing workflow of just editing separate template files rather than battling against multiline string vars.

But on the other hand, I do believe that your use-case is reasonable. So, I'd like to discuss with you to find a sweetspot that works for both of us!

That being said, my suggestion is: how about replacing these `go generates` with something like:

```bash
# Instead of `go generate ./core/controlplane/config
pushd core/controlplane/config/templates
go run ../../../codegen/templates_gen.go CloudConfigController=cloud-config-controller CloudConfigWorker=cloud-config-worker CloudConfigEtcd=cloud-config-etcd DefaultClusterConfig=cluster.yaml KubeConfigTemplate=kubeconfig.tmpl StackTemplateTemplate=stack-template.json
gofmt -w templates.go
go run ../../../codegen/files_gen.go Etcdadm=../../../etcdadm/etcdadm
gofmt -w files.go
popd

# And repeat the above pattern for `./core/nodepool/config` and `./core/root/config` respectively
```

and DO commit generated files to git?

pete911: yeah, there is a disadvantage in having templates as content of variable in go (formatting, escaping, no native editor support - there's no way for editor to know if the string content of the variable is yaml or json or bash ...)

so yeah, I'm happy with your solution 👍

pete911: I'm happy to do it (remove from .gitignore, update make, create pull request) on Monday

PR #1189, opened 2018-03-16
closes #1186 - commit generated templates so project can be imported

### LINKED ISSUE #1186: remove `go generate` code

pete911:


mumoshu:
LGTM. Let's go forward!

PR #1194, opened 2018-03-21
Networking Daemonsets

davidmccormick:
Deploy Calico and Flannel networking as a daemonset with kubernetes API as backing store. Removes the need for nodes connecting to etcd and frees up node podCIDR leases faster -addressing cluster role issue: coreos/flannel#954.

Experimental feature, disabled by default.
Kubernetes Controllers become responsible for allocating node cidrs.
Switch between calico+flannel (canal) or flannel.
Fast roll out into existing clusters with minimal disruption.
Optional calico typha service for easing load on apiservers in large clusters.


davidmccormick:
Let me clean this up

Issue #704, opened 2017-06-15
Support `hostPort` with CNI


c-knowles:
According to https://github.com/kubernetes/kubernetes/issues/23920#issuecomment-307981726 we will need to update versions of Calico and configuration to enable this on k8s 1.7. I'm not sure what if anything needs to change for flannel to also work.

mumoshu:
@c-knowles Thanks for the info!
Perhaps I'll verify it to work until v0.9.8-rc.1 and then remove the note in [the limitations doc](https://github.com/kubernetes-incubator/kube-aws/blob/master/Documentation/kubernetes-on-aws-limitations.md).

c-knowles:
Sure, let's verify that I can try on our clusters. FYI I hit this from datadog's latest install notes which use the host port functionality.

andrejvanderzee:
Ha! This just cost me a couple of hours :-)
+1 for hostPort with CNI.

chen-anders:
I'm having issues getting CNI working with the proper conflist config for flannel on k8s 1.7.3.

This is along the lines of what I have: 
```
{
    "name": "cbr0",
    "cniVersion": "0.3.0",
    "plugins": [
      {
        "type": "flannel",
        "delegate": {
          "isDefaultGateway": true
        }
      },
      {
        "type": "portmap",
        "capabilities": {"portMappings": true}
      }
    ]
}
```

I've searched over the web and there only seems to be instructions for Calico, but not flannel.

klausenbusk:
@chen-anders That is the file extension? I have [read](https://github.com/kubernetes/kubernetes/issues/23920#issuecomment-308254308) it should be `.conflist`..

chen-anders:
@klausenbusk - `.conflist` is correct. (Edited the original post for clarity)

Update: I was able to get it working with the v0.6.0-rc2 release downloaded from here: https://github.com/containernetworking/plugins/releases

drywheat:
I guess I'm still a little confused on how one would configure `hostPort`. Like @c-knowles mentioned earlier, I'm seeing this problem when trying to install Datadog into our cluster.

FYI – Running kube-aws version **v0.9.9-rc.2**

mumoshu:
@drywheat Hi! If you have not tried yet - Perhaps you need to include the portmap plugin in the cni config?
Ref: https://github.com/kubernetes/kubernetes/issues/23920#issuecomment-315119176

c-knowles:
[...]

PR #1195, opened 2018-03-21
Add networking-daemonsets feature

davidmccormick:
Deploy Calico and Flannel networking as a daemonset with kubernetes API as backing store. Removes the need for nodes connecting to etcd and frees up node podCIDR leases faster -addressing cluster role issue: coreos/flannel#954.

Experimental feature, disabled by default.
Kubernetes Controllers become responsible for allocating node cidrs.
Switch between calico+flannel (canal) or flannel.
Fast roll out into existing clusters with minimal disruption.
Optional calico typha service for easing load on apiservers in large clusters.


davidmccormick:
I need to update the tests with the expectation of the NetworkingDaemonSets type which gets set up with default values.

core/controlplane/config/config.go
```diff
@@ -680,6 +694,16 @@ type LocalStreaming struct {
 	interval int    `yaml:"interval"`
 }
 
+type NetworkingDaemonSets struct {
+	Enabled         bool        `yaml:"enabled"`
+	Typha           bool        `yaml:"typha"`
+	CalicoNodeImage model.Image `yaml:"calico-node-image"`
```
mumoshu: Super nit but would you mind changing yaml keys to lowerCamelCase for consistency with other existing keys?

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -236,27 +257,38 @@ coreos:
             Environment="FLANNEL_IMAGE={{.FlannelImage.RktRepo}}"
             Environment="RKT_RUN_ARGS={{.FlannelImage.Options}} --uuid-file-save=/var/lib/coreos/flannel-wrapper2.uuid"
 {{end}}
+    {{- end}}
     - name: kubelet.service
       command: start
       runtime: true
       content: |
         [Unit]
+        {{ if not .Experimental.NetworkingDaemonSets.Enabled -}}
```
mumoshu: - How about renaming `NetworkingDaemonSets` to e.g. [`SelfHostedCanal`](https://github.com/projectcalico/canal/blob/master/k8s-install/README.md)? After reading the canal doc, I now believe we can safely call it so
- Can we move the new configuration(NetworkingDaemonSets before the my suggested rename) out of `.Experimental`? Sorry for the confusion but we're basically avoiding to use the `.Experimental` key anymore. We're in pre-1.0 and there is no strict rule about what's considered experimental or when it graduates from experimental and so on.
  - However, a note about the experimental feature in cluster.yaml comments would be welcomed!

core/controlplane/config/config.go
davidmccormick: Sure - no problem

core/controlplane/config/templates/cloud-config-worker
davidmccormick: I didn't want to name it Canal as you can have selfhosted flannel if you don't supply UseCalico: true.  Perhaps another name better captures the hosted-in-kubernetes, removed-from-coreos-supplied-flannel-ness of the change?  Shall we call it KubeHostedNetworking, or SelfHostedNetworkFabric maybe?

I'll remove it from Experimental and make it top level?

mumoshu: Thanks for pointing it out! Made sense a lot.

How about `Kubernetes.Networking.SelfHosting` then?

Probably we can also consider migrating other network-related keys like `useCalico` under that for more cohesion, in the future.

davidmccormick: Ok, I'll make those changes and re-push

davidmccormick: Err, did you want me to create a Kubernetes stanza?  If so should I move kubeDns, kubeProxy, kubernetesDashboard, + lots of others into it? (Feels quite a big change).

Or create it just for this settings?
Or could I get away with KubernetesNetworkingSelfHosting?

mumoshu:
> Or create it just for this settings?

This is what I meant!
I'll create separate issues for migrating others.

davidmccormick:
Hi, I've updated the settings to kubernetes.networking.selfhosting, squashed a couple of bugs, and merged the latest changes.  I have new versions of the template files and these seem to be conflicting and not sure how to clear them.

Is their a way to select my versions?  Can you do that?

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -270,6 +270,7 @@ coreos:
             {{end -}}
 
             [Service]
+            Type=simple
```
davidmccormick: etcd rollout stalls at the first etcd server in a new cluster when the etcdadm-check (etcd disaster recovery enabled) service is enabled.  This service, etcd-member, is type 'notify' by default - but etcd won't send the notify until the cluster is healthy (and so systemd reports etcd as starting-not started even though etcd is running ok).  Etcadm-check won't start until etcd-member has started and so no cfn-signal can be sent to start the subsequent etcd servers.  This means that your roll out gets stuck at the first etcd server.  By changing the type of etcd-member to 'simple' we allow etcdadm-check start and the cloud-formation signal can go out.  There is no side effect from changing the type so that as soon as etcd has been started everything can continue.

core/controlplane/config/templates/stack-template.json
```diff
@@ -1295,6 +1295,34 @@
...
+          "Ref": "SecurityGroupWorker"
+        },
+        "IpProtocol": "tcp",
+        "SourceSecurityGroupId": {
+          "Ref": "SecurityGroupWorker"
+        },
+        "ToPort": 5473
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
```
mumoshu: Note to self: This rule should be enabled only when typha is enabled.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -1096,7 +1096,52 @@ worker:
...
+# From Legacy Calico to self-hosted Flannel - OK - ISSUES1
+# From self-hosted Flannel to self-hosted Canal - GOOD
+# From self-hosted Canal to self-hosted Flannel - OK - ISSUES1
+# From Legacy Flannel to self-hosted Canal or Flannel - ISSUES2
+# From self-hosted Canal or Flannel to legacy Flannel - ISSUES3
+# 
+# ISSUES1 - calico iptables rules remain on the nodes, recycle nodes to remove.
+# ISSUES2 - Existing pod ip addresses are not tracked and can cause clashing ips - please cordon all nodes whilst
+# performing the upgrade until all workers/nodes have been recycyled - then it will be ok.
+# ISSUES3 - Rolling back to old networking will require all nodes are cycled/rolled and the network is disrupted throughout.
```
mumoshu: I really appreciate the comprehensive and detailed documentation here!

core/controlplane/config/templates/cloud-config-worker
```diff
@@ -181,19 +183,61 @@ coreos:
...
+      enable: true
+      command: start
+      content: |
+        [Path]
+        PathExists=/etc/kubernetes/cni/net.d/net-migration
+    - name: net-migration-helper.service
+      enable: false
+      content: |
+        [Unit]
+        Description=Perform actions which help when migrating from legacy to selfhosted networking.
```
mumoshu: Shouldn't this be run when `.Kubernetes.Networking.SelfHosting.Enabled` is `true`?
The proposed implementation seems to be doing the opposite of that, or perhaps I'm missing something?

mumoshu: @davidmccormick I think I started to understand the good of taking care of migration from old to new networking for minimal downtime.

But on the other hand I'm still unable to understand why this implementation helps that. To me, this systemd unit seems to run "whenever a node w/ legacy networking starts", but not "before a node w/ new networking comes up".

Also, how does a node w/ legacy networking starts successfully if you disable non-self-hosted flanneld this way?

Probably I'm still missing something...

davidmccormick: Hi, sorry - I haven't explained it all that well and it is a little complicated.  The net-migration-helper is deployed to all nodes when the self-hosting option has been set to off/disabled - ie. all legacy style nodes.  On its own it wont do anything - you see it will only trigger if someone writes a file /etc/kubernetes/cni/net.d/net-migration which I'm using as a kind of semaphore/trigger file.  This file gets written when either of the self-hosting canal or flannel daemonsets are deployed (effectively signalling all legacy nodes that a migration is happening so they can shutdown their local flannel).  Migrated nodes don't need to worry about this file though - they don't have a local flannel and so this service isn't deployed to them.

The whole goal of this is to minimise the disruption when rolling a legacy cluster to the new networking (if every node has to be rolled before the network is good this would cause lengthy disruption on large clusters).  This way the cluster can reconfigure its networking with only a minute or less of disruption as it rolls out (in fact in some rolls there is no disruption at all).

As soon as the user selects self-hosting in their cluster.yaml and rolls their cluster, and once the first new controller comes up and runs install-kube-system,  it will deploy the canal or flannel daemonsets to ALL nodes - so for a while we have legacy nodes which have their previously running flannel on them AND the new flannel deployed by daemonset.  

I saw some strange behaviour in testing that led me to believe that the safest way to deploy into an existing cluster is to roll the cluster so the net-migration helper goes out first on legacy nodes.  I'm not convinced that it is needed in all cases, but it does give me a but more security for performing a roll out into a large cluster.

davidmccormick: Do you need any more docs/testing?

c-knowles:
Awesome! I would heart this PR more if I could. I believe this will resolve https://github.com/kubernetes-incubator/kube-aws/issues/704 as well which I would need to work on soon. Might be worth commented it as experimental in cluster.yaml?

core/controlplane/config/templates/cloud-config-worker
mumoshu: Totally understood! Thank you so much for the clarification.

For anyone interested, several points you should know to understand about how this is implemented:

- `net-migration-helper.path` is implicitly paired to `net-migration-helper.service`, so that once the file is created on the path, the `.service` counterpart is started by systemd.
- A `canal` daemonset deployed via the first controller node rolled via `kube-aws update` to enable canal writes the file to all the controller and worker nodes, so that `net-migration-helper.path` is triggered on every legacy node for faster migration. Implementation: https://github.com/kubernetes-incubator/kube-aws/pull/1195/files#diff-ef25536c536667a40b993d4d24ab7567R1241

core/controlplane/config/templates/cloud-config-etcd
mumoshu: @davidmccormick Good point!

But I take this change to actually result in a side-effect of we no longer postpone the rolling update of etcd nodes in case etcd-member failed to join the existing cluster. I think this is better explained in https://github.com/kubernetes-incubator/kube-aws/issues/1206#issuecomment-378857233.

Not saying this change is no good. However, at least, this chance should be discussed and made separately!
That being said, would you mind reverting this?

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -284,8 +317,7 @@ coreos:
         --container-runtime={{.ContainerRuntime}} \
         --rkt-path=/usr/bin/rkt \
         --rkt-stage1-image=coreos.com/rkt/stage1-coreos \
-        --node-labels node-role.kubernetes.io/master{{if .NodeLabels.Enabled}},{{.NodeLabels.String}} \
-        {{end}} \
+        --node-labels=node-role.kubernetes.io/master="",kubernetes.io/role=master,{{if .NodeLabels.Enabled}},{{.NodeLabels.String}}{{end}} \
```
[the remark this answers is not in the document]
davidmccormick: Ah yes let me send a small correction

mumoshu:
@davidmccormick Thank you so much for all the your awesome works. I've reviewed this again, and am happy to get this merged 👍

jorge07:
Did someone test this?

davidmccormick:
I did lots of testing of it - are you having a particular issue?

jorge07:
We're having issues creating a cluster from master and looks related with canal. I'll try to collect as much logs as possible and post it here asap

rastut:
Hey i am a colleague from @jorge07 . Yesterday we were doing some testing with the master branch and we have found that the cluster was not  able to run containers correctly. We are running the cluster without self-hosting the network .We have been looking logs around, and the problem was that kubelet was not able to locate cni binaries.

```
Apr 11 10:42:10 ip-10-0-2-59.us-east-2.compute.internal kubelet-wrapper[1410]: E0411 10:42:10.546050    1410 cni.go:259] Error adding network: failed to find plugin "loopback" in path [/opt/loopback/bin /opt/cni/bin]
Apr 11 10:42:10 ip-10-0-2-59.us-east-2.compute.internal kubelet-wrapper[1410]: E0411 10:42:10.546706    1410 cni.go:220] Error while adding to cni lo network: failed to find plugin "loopback" in path [/opt/loopback/bin /opt/cni/bin]
Apr 11 10:42:10 ip-10-0-2-59.us-east-2.compute.internal kubelet-wrapper[1410]: E0411 10:42:10.547302    1410 cni.go:259] Error adding network: failed to find plugin "loopback" in path [/opt/loopback/bin /opt/cni/bin]
Apr 11 10:42:10 ip-10-0-2-59.us-east-2.compute.internal kubelet-wrapper[1410]: E0411 10:42:10.551897    1410 cni.go:220] Error while adding to cni lo network: failed to find plugin "loopback" in path [/opt/loopback/bin /opt/cni/bin]
```

After taking a look basically i have found that the kubelet system unit basically is mounting the host path `/opt/cni/bin` inside the kubelet container, this is causing that the path of the container is override with the host path and the kubelet is not able to found the cni binaries.

https://github.com/kubernetes-incubator/kube-aws/blob/3f116956373f6fa817618b257b4479a8cb464b45/core/controlplane/config/templates/cloud-config-worker#L329-L330

This is the output of one of the nodes that currently are not working correctly:

```
core@ip-10-0-2-59 /opt/cni/bin $ rkt list                                                                                                  
UUID		APP		IMAGE NAME				STATE		CREATED		STARTED		NETWORKS
9d5945e6	awscli		quay.io/coreos/awscli:master		exited garbage	1 hour ago	1 hour ago	
9f0dfd23	flannel		quay.io/coreos/flannel:v0.10.0		exited		1 hour ago	1 hour ago	
eadd1353	flannel		quay.io/coreos/flannel:v0.10.0		running		1 hour ago	1 hour ago	
f1dcc1fc	hyperkube-amd64	k8s.gcr.io/hyperkube-amd64:v1.9.1	running		59 minutes ago	59 minutes ago	
core@ip-10-0-2-59 /opt/cni/bin $ sudo rkt enter f1dcc1fc                                                                                   
enter: no command specified, assuming "/bin/bash"
root@ip-10-0-2-59:/# ls /opt/cni/bin
root@ip-10-0-2-59:/# 
```

If we delete the argument from the unit then the containers began to run correctly and on the kubelet container we can see the binaries:

```
core@ip-10-0-1-182 /usr/lib64/systemd $ sudo rkt enter 4cd0af93                                                                  
enter: no command specified, assuming "/bin/bash"
root@ip-10-0-1-182:/# ls /opt/cni/bin
bridge	dhcp  flannel  host-local  ipvlan  loopback  macvlan  portmap  ptp  sample  tuning  vlan
root@ip-10-0-1-182:/#
```

davidmccormick:
Hi thanks for the info - I will have a look at making the mount dependent on the networking selected.  I'm guessing that you are not running with calico?  as this would have installed the cni binaries as part of its bootstrap.

ktateish:
Hi all,
I've opened an issue #1232 without noticing this discussion.   (Thanks @kevtaylor for notifying me)

I'll keep #1232 open for users who search issues with `ContainerCreating`, so please include `Close #1232` in your commit message when you create a fix commit for this problem in the future.

I'll also keep investigating it.

rastut:
Yep not using calico. I have just try some changes, modifying the template, but then i am getting and error in relation but then i am getting some strange failure. If i go through the templates and i modify the mount, then i am getting an error in relation  to the signature verification of the image.

```Apr 11 13:05:51 ip-10-0-2-229.us-east-2.compute.internal kubelet-wrapper[2258]: + exec /usr/bin/rkt run --volume coreos-etc-kubernetes,kind=host,source=/etc/kubernetes,readOnly=false --volume coreos-etc-ssl-certs,kind=host,source=/etc/ssl/certs,readOnly=true --volume coreos-usr-share-certs,kind=host,source=/usr/share/ca-certificates,readOnly=true --volume coreos-var-lib-docker,kind=host,source=/var/lib/docker,readOnly=false --volume coreos-var-lib-kubelet,kind=host,source=/var/lib/kubelet,readOnly=false,recursive=true --volume coreos-var-log,kind=host,source=/var/log,readOnly=false --volume coreos-os-release,kind=host,source=/usr/lib/os-release,readOnly=true --volume coreos-run,kind=host,source=/run,readOnly=false --volume coreos-lib-modules,kind=host,source=/lib/modules,readOnly=true --mount volume=coreos-etc-kubernetes,target=/etc/kubernetes --mount volume=coreos-etc-ssl-certs,target=/etc/ssl/certs --mount volume=coreos-usr-share-certs,target=/usr/share/ca-certificates --mount volume=coreos-var-lib-docker,target=/var/lib/docker --mount volume=coreos-var-lib-kubelet,target=/var/lib/kubelet --mount volume=coreos-var-log,target=/var/log --mount volume=coreos-os-release,target=/etc/os-release --mount volume=coreos-run,target=/run --mount volume=coreos-lib-modules,target=/lib/modules --hosts-entry host --stage1-from-dir=stage1-fly.aci docker://k8s.gcr.io/hyperkube-amd64:v1.9.1 --exec=/kubelet -- --cni-conf-dir=/etc/kubernetes/cni/net.d --cni-bin-dir=/opt/cni/bin --network-plugin=cni --container-runtime=docker --rkt-path=/usr/bin/rkt --rkt-stage1-image=coreos.com/rkt/stage1-coreos --node-labels=kubernetes.io/role=node,k8s.paysites/env=ci,k8s.paysites/tier=ci,kube-aws.coreos.com/cluster-autoscaler-supported=true,kube-aws.coreos.com/role=worker --register-node=true --allow-privileged=true --pod-manifest-path=/etc/kubernetes/manifests --cluster-dns=10.3.0.10 --cluster-domain=cluster.local --cloud-provider=aws --cert-dir=/etc/kubernetes/ssl --tls-cert-file=/etc/kubernetes/ssl/worker.pem --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem --kubeconfig=/etc/kubernetes/kubeconfig/worker.yaml --require-kubeconfig
Apr 11 13:05:52 ip-10-0-2-229.us-east-2.compute.internal kubelet-wrapper[2258]: run: signature verification for docker images is not supported (try --insecure-options=image)

```

rastut:
Yep not using calico. I have just try some changes, modifying the template, but then i am getting and error in relation but then i am getting some strange failure. If i go through the templates and i modify the mount, then i am getting an error in relation  to the signature verification of the image.

```Apr 11 13:05:51 ip-10-0-2-229.us-east-2.compute.internal kubelet-wrapper[2258]: + exec /usr/bin/rkt run --volume coreos-etc-kubernetes,kind=host,source=/etc/kubernetes,readOnly=false --volume coreos-etc-ssl-certs,kind=host,source=/etc/ssl/certs,readOnly=true --volume coreos-usr-share-certs,kind=host,source=/usr/share/ca-certificates,readOnly=true --volume coreos-var-lib-docker,kind=host,source=/var/lib/docker,readOnly=false --volume coreos-var-lib-kubelet,kind=host,source=/var/lib/kubelet,readOnly=false,recursive=true --volume coreos-var-log,kind=host,source=/var/log,readOnly=false --volume coreos-os-release,kind=host,source=/usr/lib/os-release,readOnly=true --volume coreos-run,kind=host,source=/run,readOnly=false --volume coreos-lib-modules,kind=host,source=/lib/modules,readOnly=true --mount volume=coreos-etc-kubernetes,target=/etc/kubernetes --mount volume=coreos-etc-ssl-certs,target=/etc/ssl/certs --mount volume=coreos-usr-share-certs,target=/usr/share/ca-certificates --mount volume=coreos-var-lib-docker,target=/var/lib/docker --mount volume=coreos-var-lib-kubelet,target=/var/lib/kubelet --mount volume=coreos-var-log,target=/var/log --mount volume=coreos-os-release,target=/etc/os-release --mount volume=coreos-run,target=/run --mount volume=coreos-lib-modules,target=/lib/modules --hosts-entry host --stage1-from-dir=stage1-fly.aci docker://k8s.gcr.io/hyperkube-amd64:v1.9.1 --exec=/kubelet -- --cni-conf-dir=/etc/kubernetes/cni/net.d --cni-bin-dir=/opt/cni/bin --network-plugin=cni --container-runtime=docker --rkt-path=/usr/bin/rkt --rkt-stage1-image=coreos.com/rkt/stage1-coreos --node-labels=kubernetes.io/role=node,k8s.paysites/env=ci,k8s.paysites/tier=ci,kube-aws.coreos.com/cluster-autoscaler-supported=true,kube-aws.coreos.com/role=worker --register-node=true --allow-privileged=true --pod-manifest-path=/etc/kubernetes/manifests --cluster-dns=10.3.0.10 --cluster-domain=cluster.local --cloud-provider=aws --cert-dir=/etc/kubernetes/ssl --tls-cert-file=/etc/kubernetes/ssl/worker.pem --tls-private-key-file=/etc/kubernetes/ssl/worker-key.pem --kubeconfig=/etc/kubernetes/kubeconfig/worker.yaml --require-kubeconfig
Apr 11 13:05:52 ip-10-0-2-229.us-east-2.compute.internal kubelet-wrapper[2258]: run: signature verification for docker images is not supported (try --insecure-options=image)

```

davidmccormick:
Thanks, it looks like I introduced this bug with cni after my flannel testing.  I'm testing an update to the cloud-config that copies the binaries into the host /opt/cni/bin directory.  I need to keep this mounted so that users can switch from legacy flannel/calico mode to the self-hosting daemonsets with minimal cluster disruption.

PR #1198, opened 2018-03-23
Doc: notice about service token invalidation after credentials update.

hartym:


hartym:
CLA signed.

mumoshu:
@hartym Awesome! Thank you very much for the improvement 👍

PR #1200, opened 2018-03-23
Dedicated Service account / Cluster Role for Tiller Deploy

### LINKED ISSUE #878: Cluster role and binding, service account for tiller
We'd need those until RBAC is enabled by default.

https://github.com/kubernetes/helm/blob/3f336517135233ae4a65e4ea931e200adf6f7f27/docs/service_accounts.md

kylehodgetts:
Currently Tiller Deploy is running as a cluster admin through the default service account.
This PR gives the tiller deploy its own service account binded to a cluster role with only the correct permission scope (taken from [here](https://github.com/kubernetes/helm/blob/3f336517135233ae4a65e4ea931e200adf6f7f27/docs/service_accounts.md)).

Resolves #878 

For the future, it would be worth checking if anything else is running with the default service account binded to a role with more permissions than it needs.


mumoshu:
@kylehodgetts Thank you very much for your contribution!
Would you mind signing CLA?

kylehodgetts:
@mumoshu Doing that now :)

kylehodgetts:
@mumoshu This is good to go now, assuming travis-ci comes back positively.

jorge07:
this branch needs a rebase

kylehodgetts:
I think the reason for all these commits is because the first commit I made for this PR was with a different email, as I hadn't updated my git to use my current work one.

Therefore, I had to go back and change the commit with that bad email in it, as the rest of the commits were okay. It has ended up pulling in a whole bunch of other commits too. I don't know why.

Maybe I should close this PR and try again?

PR #1202, opened 2018-03-27
Dedicated Service account / Cluster Role for Tiller Deploy

### LINKED ISSUE #878: Cluster role and binding, service account for tiller

kylehodgetts:
Currently Tiller Deploy is running as a cluster admin through the default service account.
This PR gives the tiller deploy its own service account binded to a cluster role with only the correct permission scope (taken from here).

Resolves #878

For the future, it would be worth checking if anything else is running with the default service account binded to a role with more permissions than it needs.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -870,6 +870,9 @@ write_files:
       # Service Accounts
       applyall "${mfdir}"/{kube-dns,heapster,kube-proxy,kubernetes-dashboard}"-sa.yaml"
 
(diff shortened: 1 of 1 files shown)


kylehodgetts:
@mumoshu 

The old PR's were a mess due to the wrong email being assigned to a commit.
I've fixed the issue and opened a fresh PR. Apologies for the messing around

PR #1203, opened 2018-03-27
Remove UnknownKeys field and Fail methods

### LINKED ISSUE #1143: Use go-yaml's UnmarshalStrict to emit errors for user typos in cluster.yaml
UnmarshalStrict has been introduced to go-yaml last year: https://github.com/go-yaml/yaml/pull/262

kube-aws has been implementing our own logic to emit errors in such way with [`UnknownKeys`
](https://github.com/kubernetes-incubator/kube-aws/search?utf8=%E2%9C%93&q=unknownkeys&type=) as [suggested originally in go-yaml](https://github.com/go-yaml/yaml/issues/136#issuecomment-213658588).

But honestly, it isn't/never have been sustainable 😃
Basically we should have UnknownKeys in every struct we use, which isn't achieved yet.

Just use `UnmarshalStrict` so that we never need to deal with this problem.

kylehodgetts:
Replace `yaml.Unmarshal` calls with `yaml.UnmarshalStrict`

Resolves #1143 

Note:

The error messages returned by `UnmarshalStrict` are in this format: 
`line x: field f not found in struct s` Where line x is always relative to the innermost struct, not the whole structure that could have other structs as fields.

also struct s will be prepended by any package name that is needed in order to reference the struct.
Not a huge deal but definitely worth thinking about


kylehodgetts:
Will investigate why tests are failing

mumoshu:
@kylehodgetts Thanks for the PR! The build is failing due to code format error. You can run `make format` to format code for you!

kylehodgetts:
The tests are failing now because of test yaml strings containing fields `createRecordSet`, `controllerRootVolumeType` and `controllerRootVolumeSize` and they are not defined in the Cluster struct. This is the correct behavior of `yaml.UnmarshalStrict`.

The question is now, do you want me to add these fields into the Cluster struct or remove these fields from the tests? @mumoshu

mumoshu:
@kylehodgetts Thx. Removing the fields from tests would be much appreciated. If a test case isn't meant for verifying kube-aws to correctly catch invalid config keys, those invalid keys are bugs in test data.

kylehodgetts:
Handing off to @Matei207

PR #1213, opened 2018-04-03
Fix Tiller permissions

### LINKED ISSUE #1212: Tiller permissions too restrictive
## Background
Introduced in #1202. Originally tiller ran with the default service account bound to cluster-admin which is quite bad when you consider that everything that is created without a specified service account will be assigned the default and thus be able to do ANYTHING with the cluster. That's no longer the case.

## New problem.
I ran into an issue when attempting to install the prometheus-operator chart with these new permissions. These permissions are too restrictive and thus don't allow for the creation of service-accounts and cluster role bindings, which it will need when installing these kind of charts.

The easy fix is to bind the new tiller service account added in #1202 to cluster-admin. This is better than what we had before because now a specified service account has cluster-admin permissions for tiller and not just the default.

Ideally, I'd like this to prompt a discussion about what exactly tiller should have permission to do by default.

For instance, cluster-admins are able to create namespaces. Does tiller really need to do this?

kylehodgetts:
## Background
This is a modification of #1202. Originally tiller ran with the default service account bound to cluster-admin which is quite bad when you consider that everything that is created without a specified service account will be assigned the default and thus be able to do ANYTHING with the cluster. That's no longer the case.

## New problem.
I ran into an issue when attempting to install the prometheus-operator chart with these new permissions. These permissions are too restrictive and thus don't allow for the creation of service-accounts and cluster role bindings, which it will need when installing these kind of charts.

The easy fix is to bind the new tiller service account added in #1202 to cluster-admin. Ideally, I'd like this to prompt a discussion about what exactly tiller should have permission to do by default.

For instance, cluster-admins are able to create namespaces. Does tiller really need to do this?

Resolves #1212


amoskyler:
I am currently dealing with this issue. Is this fix not being pursued?

PR #1219, opened 2018-04-04
Allow new etcd clusters to boot with recovery enabled

davidmccormick:
When I tried to launch new clusters with the following etcd config: 

```
etcd:
  snapshot:
    automated: true
  disasterRecovery:
    automated: true
```

The rollout stalls at etcd0 because cfn-signal service doesn't start.  It doesn't start because the etcdadm-check service hasn't started and the etcdadm-check wont start until the etcd server, i.e. the etcd-member service signals that it has started up via the systemd notify mechanism.

The issue seems to be that etcd-member will not indicate that it has started unless all of the members are healthy - which is a problem, we have only started the first one.  So, unless we take intervention then the etcd0 will time out after 15mins and we don't get our cluster.

This behaviour can be changed by changing the etcd-member service from type notify to type simple instead.  Then out etcdadm-check service can start even though it will report that the cluster is unhealthy and subsequently cfn-signal can run and so we move on to launching etcd1.  By changing the /etc/systemd/system/etcd-member.service.d/20-aws-cluster.conf which we are already creating we can override the systemd type as required.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -270,6 +270,7 @@ coreos:
             {{end -}}
 
             [Service]
+            Type=simple
             EnvironmentFile=-/etc/etcd-environment
             EnvironmentFile=-/var/run/coreos/etcdadm/etcd-member.env
 


mumoshu:
@davidmccormick Thanks for your contribution as always!

We do conditionally set systemd unit type to either `simple` or `notify` according to how many etcd nodes are remaining or required until a quorum is met.

So, as long as `etcdadm` is correctly recognizing current/remaining num of etcd nodes until the quorum, it should just work. Perhaps anything else is going wrong so that `cluster_num_running_nodes` in `etcdadm` is returning wrong results?

mumoshu:
Updated the prev comment with a link to relevant code.

mumoshu:
@davidmccormick Does https://github.com/kubernetes-incubator/kube-aws/issues/1206#issuecomment-378785704 look relevant to your issue?

davidmccormick:
Apologies, the problem is not with etcdadm-check - it gets stuck along with cfn-signal because they both have ExecStartPre checks that etcd-member is active.  Neither cfn-signal nor etcdadm are able to start until the etcd cluster reports ready - and it will never do so until it has quorum.  Another etcd node needs to start for quorum to be achieved but it is stopped doing do because the cloud-formation stack deploys each etcd one at a time and expects a cfn-signal from each before moving on to the next.

davidmccormick:
Closing this is it is too simple a solution to this problem.

Issue #1218, opened 2018-04-04
kube-aws does not always roll nodes as expected


davidmccormick:
In testing out a new kube-aws feature I found that changes made to my cluster.yaml doesn't always result in my cluster nodes rolling even though changes to my choices in the cluster.yaml should have resulted in differences to my resultant node cloud-configs.  On such example, could be toggling kubernetes.networking.selfHosting on and off via the settings in cluster.yaml.

This looks to be because the user-data and cloud config have been modelled as two separate parts; the 'instance' part is a script that loads the second 's3' part (which contains the actual cloud-config configuration).  The 'instance' script will change (causing nodes to roll) if the template changes because the script references the cloud-config file with an added 'fingerprint' (sha256 hash) of the template as part of the filename.  I think the nodes will only roll when there are differences in the template.

If there are no changes to the template - then the nodes will not roll, regardless of any changes that I have made to the settings within my cluster.yaml.  I believe that the nodes should roll as a result of any changes **either in the template or the values** which results in a difference to the **rendered** cloud-config.  i.e. I believe that the actual user-data - the 'instance' part needs to be tied to the rendered s3 part, and not to template itself.

To this end, I would like to propose that we add a fingerprint of the rendered 's3' userdata component to the script in the 'instance' component as a non-executed line (i.e. comment) so that when there is a change to the rendered 's3' userdata component then the 'instance' component will also change, causing the nodes to roll next time we run update.

I will submit a pull request to add this functionality.

regards

Dave

mumoshu:
@davidmccormick Thanks for the detailed feedback!

> the rendered s3 part, and not to template itself.

I was under the impression that we do calculate a fingerprint from a rendered template according to:

- https://github.com/kubernetes-incubator/kube-aws/blob/5e15ac9059a95c051600c378cec204621c3a722f/core/nodepool/cluster/cluster.go#L121
- https://github.com/kubernetes-incubator/kube-aws/blob/548bb386dd14dbee7d9c9fc012e2b3a4fe5c6b75/cfnstack/assets.go#L85
- https://github.com/kubernetes-incubator/kube-aws/blob/0dc583e21ea22dffc42ddc8d1c306ae3499d6305/model/userdata.go#L123

However then, why your node didn't roll at that time? I'm not sure yet.

davidmccormick:
[...]

PR #1220, opened 2018-04-04
Make sure that servers roll when there are data changes

davidmccormick:
Please see issue https://github.com/kubernetes-incubator/kube-aws/issues/1218 for a description of the problem - which is that nodes are not rolling when purely cluster.yaml changes are being made, only when there are changes to the templates.

This pull request shows that it is possible to insert a fingerprint into the 'instance' userdata which will trigger rolls where the templates haven't changed but selected features have changed in the cluster.yaml. 

As part of this change, I've made the nodes, controllers and etcd servers all act the same in how they load their data and call out for cloud-formation stack details.

### DIFF
--- a/docs/getting-started/step-4-update.md
+++ b/docs/getting-started/step-4-update.md
@@ -37,6 +37,9 @@ More concretely, steps should be taken in order to rotate your certs on nodes ar
   kube-aws update
   ```
 
+There are cases where the service account tokens used by the system pods become invalid after credentials update, and
+some of your system pods will break (especially `kube-dns`). Deleting the said secrets will solve the issue (see https://github.com/kubernetes-incubator/kube-aws/issues/1057).
+
 ## The etcd caveat
 
 There is no solution for hosting an etcd cluster in a way that is easily updateable in this fashion- so updates are automatically masked for the etcd instances. This means that, after the cluster is created, nothing about the etcd ec2 instances is allowed to be updated.


core/controlplane/config/templates/cloud-config-worker
```diff
@@ -24,14 +25,6 @@ INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
 exec /usr/bin/coreos-cloudinit --from-file /var/run/coreos/$USERDATA_FILE
 {{ end }}
 
-{{ define "instance" -}}
-{ "Fn::Base64": { "Fn::Join" : ["", [
-  "#!/bin/bash -xe\n",
-  {"Fn::Join":["",[ "echo '{{.StackNameEnvVarName}}=", { "Ref": "AWS::StackName" }, "' >> {{.StackNameEnvFileName}}\n" ]]},
-  {{ (execTemplate "instance-script" .) | toJSON  }}
```
mumoshu: Just trying to understand the assumption made hre, but probably you could alternatively add the fingerprint here:
```json
{"Fn::Join":["",[  "# s3-part-fingerprint: ", {{ (execTemplate "s3" .) | fingeprint }} ]]},
{"Fn::Join":["",[ "echo '{{.StackNameEnvVarName}}=", { "Ref": "AWS::StackName" }, "' >> {{.StackNameEnvFileName}}\n" ]]},
{{ (execTemplate "instance-script" .) | toJSON  }}
```

so that the brand-new `RenderTemplate` func becomes unnecessary?

davidmccormick: Hmm, I'm not super expert at templating but if fingerprint is available in the Func map then I don't see why that wouldn't work.
A question why have a different method of generating the node templates from etcd and controller?  One of my thoughts was that they should work the same?

mumoshu: Ah, yes, good point. I'm a bit confused 😉 

We definitely should render all the node templates in the same way, for consistency. Your change to introduce `RenderInstance` and the improvement to make worker nodes use that func for userdata generation is definitely a good thing.

Back to current implementation, I believe that the worker template is slightly more advanced than other templates because it achieves the same goal of `cfn-signal` and `cfn-init` with reduced AWS API calls. Hopefully we could keep that.

core/nodepool/config/templates/stack-template.json
```diff
@@ -101,7 +101,7 @@
               {{end}}
             ],
             "SubnetId": {{$workerSubnet.Ref}},
-            "UserData": {{ $.UserDataWorker.Parts.instance.Template }}
+            "UserData": {{ $.UserDataWorker.RenderInstance true | checkSizeLessThan 16384 | quote }}
```
mumoshu: Suppose we now embed the rendered "instance" template which [contains the S3URI](https://github.com/kubernetes-incubator/kube-aws/pull/1220/files#diff-a14eb9921ff28b29aa4b3394378d27d6R21) which contains the hash of the rendered "s3" template, can we safely omit the fingerprint inclusion done in the new `RenderInstance` func?

core/controlplane/config/templates/cloud-config-worker
mumoshu: Maybe append this part of the `instance-script` to every "instance" template for all the node types, via `RenderInstance` func?

```
{ "Fn::Base64": { "Fn::Join" : ["", [
"#!/bin/bash -xe\n",
{"Fn::Join":["",[ "echo '{{.StackNameEnvVarName}}=", { "Ref": "AWS::StackName" }, "' >> {{.StackNameEnvFileName}}\n" ]]},
```

davidmccormick: Hmm - that's an interesting idea!  Let me go away and have a look!

core/nodepool/config/templates/stack-template.json
mumoshu: Ah, sorry again and thanks for pointing it out!
I somehow misunderstood that the problem of the inappropriate fingerprint affected worker nodes only but that's not true, indeed.
That being said, I now believe that inserting a "correct" fingerprint through `RenderInstance` is completely valid.

Regarding the API call minimization thing, I still wonder if what [I have suggested above](https://github.com/kubernetes-incubator/kube-aws/pull/1220/files#r179750922) is possible. If it doesn't work, yeah, it would be a hard journey... In that case, I'm open to drop the API call minimization trick for worker nodes for now and discuss about bring it back in an another issue.

core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -270,6 +270,7 @@ coreos:
             {{end -}}
 
             [Service]
+            Type=simple
```
mumoshu: I appreciate it if you could revert this again, as we've done for the prev PR.

davidmccormick: Yes, no problem - I keep on adding it for my testing as my clusters fail to spin up properly without manual intervention. :)

PR #1222, opened 2018-04-04
Add GPU support for kubernetes 1.9+ using device plugins

Lemmons:
This also greatly simplifies gpu driver installation, putting it into a
daemonset which runs on all nodes with the `kube-aws.coreos.com/gpu: nvidia` node label

It also installs a daemonset for `NVIDIA GPU device plugin used by GKE/GCE` which allows
close to native access to gpus in kubernetes, without having to install
nvidia-docker.

This adds a new experimental config, `gpuSupport` which has three
fields: `enabled`, `version` and `installedImage`. `installedImage` is
optional, and only needed if you want to self-host the image used to
install the nvidia drivers. It is the docker image repo:tag to be used
to as the nvidia install image.  `version` is the nvidia driver version to
install.  `375.66` has worked pretty well for me, though it should
support anything your current coreos kernel supports. `enabled` should
be pretty self-explanatory.

To use this, there are two things required:
1. add the `kube-aws.coreos.com/gpu: nvidia` node label to the gpu node
pools you want it installed on
2. add the `gpuSupport` section to the `experimental` section of your
cluster.yaml, as described above.
3. *DO NOT* add the `gpu.nvidia` section to the gpu nodepool.  At best
this will greatly increase node startup time, at worst it will severely
conflict with the new way of installing the drivers.


mumoshu:
@Lemmons Wow, I greatly appreciate this contribution! Eager to merge this 👍 

Regarding the failing travis build, it looks like that you need to add `shelmangroup/coreos-nvidia-driver-installer:latest` as the expected experimental settings parsed from test data, so that the bunch of failures of `maincluster_test.go:157: experimental settings didn't match` goes away.

mumoshu:
@Lemmons Please Let me know if you're short on time and want me to fix tests!

Lemmons:
I have a few more changes to add, mainly due to the `WARNING: Nvidia GPU driver intallation is disabled although instance type %v does support GPU.  You have to install Nvidia GPU driver by yourself to schedule gpu resource.` message still showing up when using this new GpuSupport.  I'll add those today or tomorrow, and will try to fix the tests while I'm at it.

Lemmons:
@mumoshu Alright. Fixed the tests and added the checks I wanted.  It could use more tests for the extra code paths, but unfortunately, I don't have time to write those right now.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -4873,6 +4878,118 @@ write_files:
...
+      spec:
+        template:
+          metadata:
+            labels:
+              name: nvidia-driver-installer
+            annotations:
+              scheduler.alpha.kubernetes.io/critical-pod: ''
+          spec:
+            nodeSelector:
+              kube-aws.coreos.com/gpu: nvidia
```
mumoshu: Completed untested but just my guess - We still need to tolerate all the taints here. Otherwise, the driver won't be installed on tainted nodes and therefore we can't make gpu nodes dedicated to some workloads.

PR #1225, opened 2018-04-05
Fix synchronization on starting etcdadm-reconfigure

### LINKED ISSUE #1206: Unable to create cluster with more than 1 etcd
Hi I've been trying for a few hours to create a cluster with 3 etcd instances but always got a timeout. It looks like the ASG for Etcd0 is created first and the instance keeps trying to connect to the other two Etcd instances but they do not yet exist and the initialisation times out. If the Etcd1 and Etcd2 ASGs were created in parallel it would probably work as the instances would start up simultaneously and could connect to each other. 

I had the same results both with .etcd.memberIdentityProvider == eip and with eni - in both cases etcd0 tried to connect to the other not-yet-existing nodes, either over EIP or over ENI. In either case it timed out.

I'm using pre-existing VPC with existing subnets - 3x Private with NAT and 3x DMZ with public IP enabled by default. I tried to put the etcd nodes both in Private and in DMZ and both failed when requested more than 1 node.

ktateish:
Resolves #1206

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -156,6 +156,7 @@ coreos:
         RestartSec=5
         EnvironmentFile=-/etc/etcd-environment
         EnvironmentFile=-/var/run/coreos/etcdadm-environment
+        ExecStartPre=/usr/bin/systemctl is-active format-etcd2-volume.service
         ExecStartPre=/usr/bin/systemctl is-active cfn-etcd-environment.service
         ExecStartPre=/usr/bin/mkdir -p /var/run/coreos/etcdadm/snapshots
         ExecStart=/opt/bin/etcdadm reconfigure


davidmccormick:
This looks rather complex as a means of saying start the etcd servers together?  Why do we need a service to adjust the type of the service at all?

mumoshu:
@ktateish Would you mind signing the CLA? Thx!

mumoshu:
@ktateish Seems like bot finally recognized your sign. Thx 😄 

> BTW we need more discussion for #1206

Certainly. However, I'm eager to merge this as I believe that we can assume it as two separate things - fixing a timing-issue which causes race-condition of etcd nodes(this PR) vs simplifying implementation(@davidmccormick's suggestion, which is great).

mumoshu:
> I believe that we can assume it as two separate things

Does this sentence make sense to you two?

Issue #1226, opened 2018-04-06
Support for user provided service-account signing keys


davidmccormick:
At hotels.com we are trying to get back from using a modified versions of kube-aws to using the vanilla kube-aws release.  

We have numerous clusters provisioned with an alternative service-account signing key process that I want to remove from our build.  Our code creates new service-account signing keys and partially rotates them on a schedule.  It's overkill and I want to remove it but I need to do it in a way where we continue to support our previous keys - i.e. existing services accounts created with our custom keys are still valid.  Kubernetes allows you to have 

I would like to propose a change to kube-aws core that will allow a user to specify their own service-account signing keys, but will use the apiserver-key when rolling into existing clusters when the user has not specifically provided a key.  New clusters will have a service-account-key generated which is separate from the apiserver key.

This should not affect existing users but will add functionality that allows users to support their own processes for provisioning service-account-keys.  I will submit a pull request that adds this functionality.

PR #1227, opened 2018-04-06
Enable user provided service-account-signing-keys

davidmccormick:
kube-aws presently assumes that we will always use the apiservers key for signing service-accounts.  This change separates the service-account-signing-key into its own asset.

Behaviour:-
* New clusters get separately generated apiserver-key.pem and service-account-key.pem keys.
* Where deploying to an existing cluster the service-account-key.pem will be copied from the apiserver-key.pem file.
* Where users specify their own service-account-key.pem file, that will be used.

Related issue: https://github.com/kubernetes-incubator/kube-aws/issues/1226

### DIFF
--- a/core/controlplane/config/credential.go
+++ b/core/controlplane/config/credential.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 )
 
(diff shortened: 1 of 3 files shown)


mumoshu:
LGTM. Great work, especially on implementing the migration plan 🎉

PR #1228, opened 2018-04-06
Another implementation of user-data fingerprinting

davidmccormick:
Replaces https://github.com/kubernetes-incubator/kube-aws/pull/1220
Resolves issue https://github.com/kubernetes-incubator/kube-aws/issues/1218

Implemented as a template function and stack-name harmonised across etcd, apiserver and nodes.

This is good to go!


davidmccormick:
Hi this is ready... the change does the following..

* Take SHA256 checksum of s3 cloud config part and insert into the instance user data as a comment.  So that changes to cluster.yaml will trigger a change to instance user-data.
* SHA256 sum implemented as a template function - fingerprint.
* Etcd, Controller and Nodes made the same! :)
* Removed aws calls from etcd/controllers/nodes for looking up AWS::StackName from tags - instead it is passed through cloud-formation when user-data is written.

:)

mumoshu:
@davidmccormick I've squashed and merged the commits with reorganizing the commit message based on PR descriptions of this and #1220. Thank you so much for the awesome work!

PR #1229, opened 2018-04-08
Fix issue with tiller having to restrictive permissions

### LINKED ISSUE #1212: Tiller permissions too restrictive

kylehodgetts:
## Background
This is a modification of #1202. Originally tiller ran with the default service account bound to cluster-admin which is quite bad when you consider that everything that is created without a specified service account will be assigned the default and thus be able to do ANYTHING with the cluster. That's no longer the case.

## New problem.
I ran into an issue when attempting to install the prometheus-operator chart with these new permissions. These permissions are too restrictive and thus don't allow for the creation of service-accounts and cluster role bindings, which it will need when installing these kind of charts.

The easy fix is to bind the new tiller service account added in #1202 to cluster-admin. Ideally, I'd like this to prompt a discussion about what exactly tiller should have permission to do by default.

For instance, cluster-admins are able to create namespaces. Does tiller really need to do this?

Resolves #1212

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -3191,16 +3191,6 @@ write_files:
           loadBalancer: {}
 
   - path: /srv/kubernetes/manifests/tiller-rbac.yaml
-    content: |
-        apiVersion: rbac.authorization.k8s.io/v1
-        kind: ClusterRole
-        metadata:
-          name: tiller
-        rules:
-          - apiGroups: ["", "extensions", "apps"]
-            resources: ["deployments", "replicasets", "pods", "configmaps", "secrets", "namespaces"]
-            verbs: ["*"]
-        ---
         apiVersion: v1
         kind: ServiceAccount
         metadata:
@@ -3217,7 +3207,7 @@ write_files:
           namespace: kube-system
         roleRef:
           kind: ClusterRole
-          name: tiller
+          name: cluster-admin
           apiGroup: rbac.authorization.k8s.io
 
   - path: {{.KubernetesManifestPlugin.ManifestListFile.Path}}
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -3191,6 +3191,7 @@ write_files:
           loadBalancer: {}
 
   - path: /srv/kubernetes/manifests/tiller-rbac.yaml
+    content: |
         apiVersion: v1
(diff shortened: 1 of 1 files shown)


Fsero:
I don't understand really the issue here, so i could be totally wrong about this.

However, in our case we consider this tiller the cluster-level tiller used by cluster maintainers to bootstrap cluster wide applications (kube2iam, autoscaler, ValidatingWebhooks et al, cluster monitoring and cluster logging, ingress.. ) where tenants can have a namespaced tiller with lesser privileges just to deploy applications.

This is a waste in the sense we run multiple copies of tiller, but it locks down better what a tiller can do in an application namespace, and is up to the application owner or cluster-maintainer to set that up.

So if you want to lock-down tiller permissions in a given namespace you should create a tiller  locked-down serviceaccount  in that namespace and deploy tiller in that namespace using that locked service account.

Then as a helm client you need to instruct helm to use the namespaced tiller, also ideally your user will be locked down via RBAC to not have access to the cluster-wide tiller.

Does that make sense for you?

kylehodgetts:
@Fsero 

I completely agree.

However this PR is just to fix an issue introduced in a previous contribution, causing errors due to not being able to create service accounts and such, which Tiller was able to do BEFORE I made that contribution.

I believe we should be moving towards what you described in the future. For now, at least, I want to make it so `cluster-admin` cluster role is not bound to the `default` service account.

That is all this PR is trying to fix.

mumoshu:
LGTM. Thank you very much for your efforts @kylehodgetts and @Fsero 🙏

c-knowles:
This might all go away as the Helm v3 proposal appears to be agreed almost and it moves to a client only model. So the permissions will be RBAC based and Tiller will disappear from what I’ve read.

Issue #1112, opened 2018-01-15
Moving ETCD and controllers outside CloudFormation Nested Stacks 


camilb:
Hi @mumoshu, I want to discuss some possible changes in CloudFormation especially regarding the NestedStacks and  DesiredCapacity.

Had all the clusters upgraded to 1.9.1 last week. All went fine, except on the last cluster, the last NestedStack (a nodepool) failed to upgrade and started a RollBack.
The upgrade failed because a new instance was added and didn't respond by any form (no ssh, no ping, nothing in AWS console "Get System Logs"). Since CF is expecting 
a signal from the launched instances, the quickest solution was to terminate the instance from console but it didn't launched a new one, so I increased the ASG capacity 
to be able to receive the signal before a timeout but right after that I received a error in CloudFormation: 
`New SetDesiredCapacity value 20 is below min value 21 for the AutoScalingGroup.`
And the RollBack started, rolling back all the node pools, controllers and ETCD for ~2h. It's not the first time when something like this happens, once the CloudFormation displayed a message that a new instance was added but didn't show in console so I had to change the ASG manually to avoid signal timeout.
 
On our CF stacks we never use `DesiredCapacity` for ASGs because they can be resized due to  a traffic increase and CF will rollback the update if the size differs.
Then, regarding the NestedStacks, I think ETCD and controllers should be in separated stacks. If the update fails on workers, we should not rollback the ETCD and controllers that were successfully upgraded. Sometimes a rollback will not work for ETCD or controllers with specific versions. For example on test clusters was unable to revert from ETCD 3.2.X to ETCD 3.0.X or from Kubernetes 1.9.X to 1.7.X and this require repeating the upgrade. Workers are quite safe to rollback. 

Also, the RollBack process for NestedStacks takes a very long time for large clusters with multiple node pools. I know that using Nested Stacks it's much cleaner, but rolling back everything when we have updates in ETCD or Kubernetes version requires repeating the upgrade to recover the cluster, causing a very long downtime.

I' don't know how we can remove the `DesiredCapacity` from CF right now, but at least maybe we can move out ETCD and Controllers from Nested Stacks. What do you think?

Fsero:
@camilb i've been maintaining and testing several k8s clusters using kube-aws the last year and i do agree with your idea, it will be good that we can pin specific k8s version and amiID for the control plane and etcd and another one for workers. That will tremendously help upgrades and updates

redbaron:
[...]

PR #1233, opened 2018-04-11
feat: Fine-grained stack update

mumoshu:
USAGE:

Update all the stacks including root, control-plane, worker:

```console
kube-aws update --force
```

Update the control-plane stack only:

```console
kube-aws update --targets control-plane --force
```

Note that, for the root stack, just the `Resources.Controlplane.Properties.TemplateURL` is updated in this case.

Update the worker node pool stack(s) only:

```console
kube-aws update --targets worker --force
```

NOTES:

- The `root` stack is updated implicitly when you update all the stacks including `worker`, `control-plane`,  `etcd` and `network`

TODO:

- [x] The separation and extraction of the etcd stack and the network stack from the control-plane stack
- [ ] Add a new option in cluster.yaml `rollbackOnFailure` for control-plane and etcd stack only

Everything else should be there

Ref #1112


core/root/cluster.go
```diff
@@ -237,51 +249,103 @@ func (c clusterImpl) Info() (*Info, error) {
 	return describer.Info()
 }
 
-func (c clusterImpl) prepareTemplateWithAssets() (string, error) {
-	assets, err := c.Assets()
+func (c clusterImpl) generateAssets(targets OperationTargets) (cfnstack.Assets, error) {
```
mumoshu: This is where the magic happens.

mumoshu:
This is triggering a considerable amount of refactorings... 😉

mumoshu:
Whoa. Can't believe this is actually working 🤣 

```
$ KUBECONFIG=kubeconfig kubectl get no
NAME                                            STATUS    ROLES     AGE       VERSION
ip-10-0-0-180.ap-northeast-1.compute.internal   Ready     node      7m        v1.9.1
ip-10-0-0-20.ap-northeast-1.compute.internal    Ready     master    15m       v1.9.1
```

image

mumoshu:
Go ahead!

c-knowles:
Is this intended to work with just `--targets etcd`? I did an update like that but it also updated controllers and started on workers too. The controllers appeared to be mostly blank container linux instances, they didn't even have a kubelet service! To fix it I cancelled the update and the cluster has a functioning control plane again.

PR #1234, opened 2018-04-11
handle kubectl apply failures by trying delete+create

davidmccormick:
We have seen an issue where an attempted update to tiller roll bindings is causing a cluster roll/update to fail.

There are some objects and fields in kubernetes that you can not update with a kubectl apply, for example ClusterRoleBindings.  I've added a little logic to handle the failure of kubectl apply by attempting a delete followed by a create when an apply fails.  If the create also fails then the unit will go into a failed state which will prevent cfn-signal from firing and therefore a roll-back.

This change add resilience to the install-kube-system script and will prevent cluster update failures.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -819,6 +819,7 @@ write_files:
 
       # Try to batch as many files as possible to reduce the total amount of delay due to wilderness in the API aggregation
       # See https://github.com/kubernetes-incubator/kube-aws/issues/1039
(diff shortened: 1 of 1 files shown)


redbaron:
do you think   `--force` flag can help here? From kubectl apply docs:

```
 --force=false: Delete and re-create the specified resource, when PATCH encounters conflict and has retried for 5 times.
```

davidmccormick:
Oops, yes that sounds cleaner! :)

mumoshu:
LGTM. Thanks as always for your contribution 👍

PR #1237, opened 2018-04-12
Make etcdadm-reconfigure.service retryable

Sriharivignesh:
Along with etcdadm-update-status.service, etcdadm-check.service, etcdadm-save.service

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -153,7 +153,7 @@ coreos:
         After=network.target
 
         [Service]
-        Type=oneshot
+        Restart=on-failure
         RemainAfterExit=yes
         RestartSec=5
         EnvironmentFile=-/etc/etcd-environment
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -153,7 +153,7 @@ coreos:
         After=network.target
 
         [Service]
-        Type=oneshot
+        Restart=on-failure
```
mumoshu: Good point! But just changing this to `Type=simple`(as you removed `Type=oneshot`) results in systemd to not wait until `etcdadm-reconfigure` finishes.

If you really wanted `etcadm-reconfigure` to be retried automatically, would you mind achieving it like `/opt/bin/retry 5 /opt/bin/etcdadm reconfigure` instead?

FYI, you can see similar implementations are done in `cloud-config-controller` for `install-kube-system` and `apply-kube-aws-plugins`.

PR #1239, opened 2018-04-16
KIAMImage should affect server as well as client.

davidmccormick:
We want to change the version of the KIAMServer but found that the image is hardcoded whereas the client isn't.  Both the client and server should use the same image.


davidmccormick:
Associated issue: https://github.com/kubernetes-incubator/kube-aws/issues/1245

PR #1241, opened 2018-04-16
Disable coreOS auto-updates

### LINKED ISSUE #1240: [Proposal] Disable CoreOS update-engine
We're having some issues due to CoreOS update-engine. Some machines has been upgraded to a newer CoreOS version. Immutability is important and this is causing upgrades to docker `17.09.1-ce` that has serious issues like: #1135

jorge07:
Fixes #1240

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -13,6 +13,9 @@ s3URI: {{.S3URI}}
 # The AMI ID of CoreOS.
 amiId: "{{.AmiId}}"
 
+# Container Linux has automatic updates https://coreos.com/os/docs/latest/update-strategies.html. This can be a risk in certain situations and this is why is disabled by default and you can enable it by setting this param to false.
+disableContainerLinuxAutomaticUpdates: true
+
 # The ID of hosted zone to add the externalDNSName to.
 # Either specify hostedZoneId or hostedZone, but not both
 #hostedZoneId: ""
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -454,18 +454,19 @@ type ComputedDeploymentSettings struct {
 // Though it is highly configurable, it's basically users' responsibility to provide `correct` values if they're going beyond the defaults.
 type DeploymentSettings struct {
 	ComputedDeploymentSettings
-	CloudFormation              model.CloudFormation  `yaml:"cloudformation,omitempty"`
-	ClusterName                 string                `yaml:"clusterName,omitempty"`
-	S3URI                       string                `yaml:"s3URI,omitempty"`
-	KeyName                     string                `yaml:"keyName,omitempty"`
-	Region                      model.Region          `yaml:",inline"`
-	AvailabilityZone            string                `yaml:"availabilityZone,omitempty"`
-	ReleaseChannel              string                `yaml:"releaseChannel,omitempty"`
-	AmiId                       string                `yaml:"amiId,omitempty"`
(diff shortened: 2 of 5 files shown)


jorge07:
Not tested yet

core/controlplane/config/templates/cluster.yaml
```diff
@@ -13,6 +13,9 @@ s3URI: {{.S3URI}}
 # The AMI ID of CoreOS.
 amiId: "{{.AmiId}}"
 
+# CoreOS has automatic updates https://coreos.com/os/docs/latest/update-strategies.html. This can be a risk in certain situations. Set this param to true to disable it in all instances.
+disableCoreOSUpdates: false
```
mumoshu: I'd prefer `disableContainerLinuxAutomaticUpdates` as it has officially renamed to Container Linux a year or so ago 😃

jorge07:
@mumoshu I updated the PR renaming this. I did not have time to test it yet, but I'll try to do it asap.

davidmccormick:
I think that, once tested, this should be on by default - not behind a feature gate. :)

mumoshu:
@davidmccormick Ah! That's definitely a good point. I agree.

jorge07:
Sorry for this waiting for a week. I've the remote team in the office and it's a super busy week. I hope find some time for rebase the branch and launch the cluster next week as I think it would be great have it ready for 0.9.10 final release

mumoshu:
@jorge07 No worry but I just want to sync up with you that I am about to cut v0.9.10 this week! Do you have any reason you want this before #1233 which is I'm going to merge soon after cutting v0.9.10?

jorge07:
I did the QA today and this is not working at all.

The `update-engine.service` is running event is `sysctemctl status` show it as masked.

If I restart the service it fails, obviously because is masked. Any suggestions?

mumoshu:
@jorge07 Thanks for your efforts. Unfortunately, it seems like we need ignition to actually disable update-engine according to https://github.com/coreos/bugs/issues/1646 and linked issues...

An alternative idea is that to automatically remove `update-engine` related systemd unit file somewhere in the filesystem so that update-engine would never start in case of a reboot.

mumoshu:
Hm, or maybe just specifying both `command: stop` and `mask: true` would do the job?

ref: https://github.com/coreos/bugs/issues/1982#issuecomment-304957249

jorge07:
I tried masking the update engine with the stop option too. Same issue, it ends as `failed`. The cluster its created and it works, but having a failed unit its not nice. Should I start considering refactor into ignition? Sounds like a huge epic

mumoshu:
@jorge07 Thanks for your efforts. If you still want to make it in v0.9.10, I'd recommend my suggestion of removing update-engine(and possibly locksmithd related units).

It would look something like introducing `disable-automatic-update.service` which would basically run:

```
for u in update-engine locksmithd; do
  find / | grep lib64/systemd | grep $u | xargs rm
  systemctl stop ${u}.service
  systemctl disable ${u}.service
fi
```

You probably need this to clear failed units:

```
systemctl daemon-reload
systemctl reset-failed
```

But beware that this is completely untested!

mumoshu:
Of course, I'd appreciate it if you could work on the migration to ignition 👍

jorge07:
I thought the same, I think that create a work around unit that can stop it and clean the failures it's the best option meanwhile migrate into ignition. I'll work on this next week. Thanks for the suggestions!

mumoshu:
@jorge07 LGTM. Thank you very much for your effort 🍻

PR #1244, opened 2018-04-16
Inherit controlplane Kubenetes-Newtworking config in node pools.

davidmccormick:
Apologies, I found another bug with the Self Hosting Networking, this one does not cause any failures, but it does prevent nodes from rolling (and the local flannel becoming de-activated).  When I moved the feature from Experimental to Kubernetes.Networking.SelfHosting, I didn't realise that the settings from the Controlplane do not automatically propagate down to the nodes.

This change corrects this by ensuring that the Kubernetes top level object gets propagated from the control-plane to node.


davidmccormick:
Associated Issue: https://github.com/kubernetes-incubator/kube-aws/issues/1246

PR #1251, opened 2018-04-23
Bump Calico/Typha Versions now Typha Issue resolved.

davidmccormick:
In the present implementation of Kubernetes.Networking.SelfHosting, I have had to work around an issue where if Typha came up before any Calico nodes that it would write a bad global cluster information record which would prevent calico/typha from subsequently correctly starting.

I raised an upstream issue, https://github.com/projectcalico/typha/issues/118, which has now been resolved with the latest releases.  Given that the issue is resolved, this change bumps Calico and Typha to the latest releases and removes my work-around for Typha bug (which was polling for the record via an init container).

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -30,11 +30,11 @@ const (
 	userDataDir    = "userdata"
 
 	// Experimental SelfHosting feature default images.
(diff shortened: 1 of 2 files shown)


mumoshu:
LGTM. Thank you for your continuous efforts 🙏

PR #1260, opened 2018-04-26
Add some limited certificate credential validation

davidmccormick:
Let's make it harder to spin or update a kube-aws cluster with either expired certs or apiserver certs with missing or incorrect DNS or IP SANS.  Deploying expired certificates or certs with the incorrect Subject Alternate names can crash it and often in ways that doesn't fail fast and are hard to debug.
This PR starts to add some more basic validation of the cert assets against the cluster configuration:-

- checks if any certificate has expired when credentials are loaded.
- once all creds and cluster configuration is loaded cross check that configured apiEndpoints.dnsName is present in the certificate dnsNames SANs.
- Also check that the resultant address of the kubernetes service (ServiceCIDR +1) is present as an IP SAN in the apiserver certificate.

Examples of failure conditions: -
1. Test enabling check on expired admin cert (admin cert check is disabled by default).
```
$ kube-aws update
Error: Failed to read cluster config: failed to read/create encrypted assets: The following certificate in file credentials/admin.pem has expired:-

Subject: CN=kube-admin,O=system:masters
Issuer: CN=davem-lab-secure Intermediate CA
Valid From: 2018-01-25 10:18:26 +0000 UTC
Expires: 2018-01-25 11:18:56 +0000 UTC
```

2. Testing  where the apiserver cert IP SAN does not match our desired ServiceCIDR (change ServiceCIDR from 192.168.128.0/17 to 192.168.20.0/17): -
```
$ kube-aws update
Error: failed to validate control plane: the apiserver cert does not contain the kubernetes service ip address 192.168.112.1, please regenerate or resolve
```
3. Testing where the apiserver cert does not match DNSName: -
```
$ kube-aws update
Error: failed to validate control plane: the apiserver cert does not contain the external dns name test123.k8s.secure.lab.us-west-2.hcom.cloud, please regenerate or resolve
```


davidmccormick:
implementation for issue https://github.com/kubernetes-incubator/kube-aws/issues/1261

gzipcompressor/gzipcompressor.go
```diff
@@ -34,3 +36,20 @@ func CompressString(str string) (string, error) {
 	bytes := []byte(str)
 	return CompressData(bytes)
 }
+
+// DecompressString given a gzipped and base64 encoded string will decode back to
+// uncompressed text
+func DecompressString(d string) (string, error) {
```
jorge07: Renaming `d` var to something like `encodedString` will make it easier to follow and the comment no longer necessary.

davidmccormick: Yes, I see your point that the code can be self-documenting but are we interested in documenting our packages with godoc?  My IDE (vscode) spends a lot of time informing me that I should comment any function/method that is exported.  What do you think?

jorge07: I prefer to avoid comments and make things explicit but just my preference

davidmccormick:
Merged in latest master

mumoshu:
@davidmccormick LGTM. Thank you so much for your continuous supports 👍

PR #1263, opened 2018-04-27
removed unused s3 flags from commands

pete911:


cmd/calculator.go
```diff
@@ -30,7 +30,7 @@ func init() {
 	cmdCalculator.Flags().BoolVar(&calculatorOpts.awsDebug, "aws-debug", false, "Log debug information from aws-sdk-go library")
 }
 
-func runCmdCalculator(cmd *cobra.Command, args []string) error {
+func runCmdCalculator(_ *cobra.Command, _ []string) error {
```
mumoshu: Oh! Good catch. And honestly, forgive my golang beginnerish... I just realized that the go compiler doesn't complain on unused parameters unlike unused variables.

pete911: this commit was to remove s3uri options from update and validate commands, but I thought that I'll just clean it up as well as one commit.

Issue #1254, opened 2018-04-24
Controller Manager memory leak


jorge07:
I'm currently fighting with this issue https://github.com/kubernetes/kubernetes/issues/52657

As there's no fix for this ATM, what do you think about add some memory limit as parameter for controller manager as workaround like it's suggested in https://github.com/kubernetes/kubernetes/issues/52657#issuecomment-371804110?

```
$ kubectl top po -n kube-system
NAME                                                    CPU(cores)   MEMORY(bytes)
cluster-autoscaler-6b89b7bfc9-mtk9n                     5m           62Mi
heapster-68db74c6dc-2tnlb                               1m           75Mi
kube-apiserver-ip-YYYYYY.ec2.internal                   17m         741Mi
kube-apiserver-ip-YYYYYY.ec2.internal                  100m         522Mi
kube-controller-manager-YYYYYY.ec2.internal             0m           17Mi
kube-controller-manager-ip-YYYYYY.ec2.internal         21m         2090Mi
```

mumoshu:
@jorge07  Hey! Thx for the detailed report and the suggestion.

Adding memory limit sounds good to me. Would you mind submitting a PR for that? Just my two cents but my quick guess is that `kubernetes.controllerManager.requests.memory` and `kubernetes.controllerManager.limits.memory` in cluster.yaml, propagated as is to controller-manager static pods, would work.

jorge07:
@mumoshu that's sounds like a very clever solution! I'll try it asap

jorge07:
Related with: https://github.com/kubernetes-incubator/kube-aws/pull/1266/files

mumoshu:
kubespray seems to define default limits/requests [here](https://github.com/kubernetes-incubator/kubespray/blob/07cc98197187535619ac8e57ee4e04ae02631a54/roles/kubernetes/master/defaults/main.yml#L28-L31).
`kube_controller_memory_requests ` of `100M` and `kube_controller_memory_limit` of `512M` sound reasonable.

PR #1266, opened 2018-04-30
Allow configure Controller Manager compute resources

jorge07:


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1094,6 +1094,14 @@ worker:
 #kubernetes:
 #  networking:
 #    selfHosting:
+#  controllerManager:
+#    resources:
+#      requests:
+#        cpu: 200m
```
c-knowles: Usually the values here are the defaults. Are there any recommended defaults from upstream?

jorge07: This is the default value righ know for cpu. That I added for limits are based on our production cluster requirements. If you know where I can find any recommendation for this, please, share

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -3323,7 +3323,19 @@ write_files:
           {{end}}
           resources:
             requests:
-              cpu: 200m
+              cpu: {{ if .Kubernetes.ControllerManager.ComputeResources.Requests.Cpu }}{{ .Kubernetes.ControllerManager.ComputeResources.Requests.Cpu }}{{ else }}200m{{ end }}
```
mumoshu: Nit: We'd better define the default of `200m` in go and `cluster.yaml` template, similarly to other settings.

mumoshu:
LGTM. Thank you so much for your contribution @jorge07!

I'd appreciate it if you could update the defaults once we find any in upstream

core/controlplane/config/templates/cluster.yaml
mumoshu: Let's continue the discussion in #1254

PR #1268, opened 2018-04-30
autoscaler: update cloud-config-controller

### LINKED ISSUE #1253: cluster autoscaler --skip-nodes-with-system-pods=false ?
Hello All,

Apologies if I'm missing something, but with a recently new cluster deployed with kube-aws v0.9.10-rc.3, I seem to be missing autoscaler functionality.  The problem is that this new cluster's AS doesn't honor the min settings in our node pool configurations as specified in the cluster.yaml and will not terminate after we've manually scaled the nodepool up, and then after we've done a kube-aws update with the new lower/min setting.

Looking at the actual AS git code/repo:
ref: [Cluster Autoscaler on AWS](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws)

> By default, cluster autoscaler will not terminate nodes running pods in the kube-system namespace. You can override this default behaviour by passing in the --skip-nodes-with-system-pods=false flag.

On that note, I received a private message / response to my post in the kube-aws IRC channel stating

> thats known problem, you need to edit user-data files and add option for cluster-autoscaler `--skip-nodes-with-system-pods=false`

Regarding kube-aws deployed clusters and the above fixes and suggestions, it's not clear how the above should be done.

Thanks

cmcconnell1:
modifies the behavior of the autoscaler to terminate worker and
controller nodes no longer needed when minSize is set manually.

Note that I had to delete the initial autoscaler pod after
deploying a new cluster with this modification in place.
After that initial autoscaler pod was deleted the second
pod functioned without errors and as expected.
I will update the issue comments with more details.

Fixes #1253

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -4049,6 +4049,7 @@ write_files:
                     - --stderrthreshold=info
                     - --cloud-provider=aws
                     - --skip-nodes-with-local-storage=false
+                    - --skip-nodes-with-system-pods=false
                     - --expander=least-waste
                     - --node-group-auto-discovery=asg:tag=k8s.io/cluster-autoscaler/enabled,kubernetes.io/cluster/{{.ClusterName}}
                   env:


mumoshu:
@cmcconnell1 Thanks for the quick response, too!

My only concern is this:

> Note that I had to delete the initial autoscaler pod after
deploying a new cluster with this modification in place.

Do you mean that, you created a brand new cluster wWITH this change, and then you had to remove the autoscaler pod by hand to make it actually work?

cmcconnell1:
Yes @mumoshu that is what I saw.  I updated the logs from the initial crashing AS pod here
[1253#](https://github.com/kubernetes-incubator/kube-aws/issues/1253#issuecomment-385568895)

cmcconnell1:
I am deploying a new fresh cluster now to validate if that happens again.
`kube-aws version v0.9.10-rc.3`

PR #1269, opened 2018-05-01
Fix etcd startup

Confushion:
After applying the fix for issue #1206, timing/race  issues were still showing when creating etcd clusters with more than 1 instance.

This patch fixes the race condition by having `etcdadm-reconfigure.service` depend on service `var-lib-etcd2.mount` instead of `format-etcd2-volume.service` (`var-lib-etcd2.mount` already depends on that service).

Also a potential circulair dependency on `cfn-etcd-environment.service` has been eliminated.


Confushion:
Closed PR, committed with wrong username. Use PR #1270 instead.

PR #1270, opened 2018-05-01
Fix etcd startup

Confushion:
After applying the fix for issue #1206, timing/race issues were still showing when creating etcd clusters with more than 1 instance.

This patch fixes the race condition by having `etcdadm-reconfigure.service` depend on service `var-lib-etcd2.mount` instead of `format-etcd2-volume.service` (`var-lib-etcd2.mount` already depends on that service).

Also a potential circulair dependency on `cfn-etcd-environment.service` has been eliminated.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -151,6 +151,7 @@ coreos:
         Wants=cfn-etcd-environment.service
         After=cfn-etcd-environment.service
         After=network.target
+        After=var-lib-etcd2.mount
 
         [Service]
         Type=oneshot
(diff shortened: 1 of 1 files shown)


Confushion:
Yes, I managed to bring up 2 new clusters with 3 etcd instances each.

Although still with setting ‘waitSignal.enabled: false’, but that seems to be a separate issue...

mumoshu:
@Confushion Thanks for the confirmation.

So, are you still seeing issues when the wait signal is enabled?
Would you mind sharing logs(even a full output from `journalctl` can be helpful), statuses or anything you think important?
I'd like to sort these out, too!

Issue #1273, opened 2018-05-03
cfn-signal service runs far after CF stack updates if controller node terminates


c-knowles:
If a controller node terminates for any reason and another is put in its place outside of a CloudFormation update, the cfn-signal.service still executes and continues to fail (assuming `waitSignal` is enabled).

An example from one of our controller nodes:
```
2018-04-28T14:28:52.000Z {
  "instanceId": "i-0cc90d2b46c92759b",
  "pid": 1,
  "uid": 0,
  "gid": 0,
  "cmdName": "systemd",
  "exe": "/usr/lib64/systemd/systemd",
  "cmdLine": "/usr/lib/systemd/systemd --switched-root --system --deserialize 16",
  "systemdUnit": "init.scope",
  "bootId": "f234d4bc209c45d1a5bbc6e5f644a510",
  "hostname": "ip-10-0-31-213.eu-west-1.compute.internal",
  "transport": "journal",
  "priority": "INFO",
  "message": "cfn-signal.service: Scheduled restart job, restart counter is at 305.",
  "messageId": "5eb03494b6584870a536b337290809b3",
  "syslog": {
    "facility": 3,
    "ident": "systemd"
  },
  "kernel": {}
}
```

The message that appears on failure is:
```
"message": "[10741.860149] awscli[7]: ValidationError: Stack arn:aws:cloudformation:eu-west-1:OBFUSCATED:stack/k8s-dev-v8-Controlplane-1ERY9ADUQZRDI/9faa62d0-2b19-11e7-82f7-503ac9e74c8d is in UPDATE_COMPLETE state and cannot be signaled",
```

Using kube-aws version 548bb386dd14dbee7d9c9fc012e2b3a4fe5c6b75.

Would a swap to Ignition help us in any way for maintenance or testing of these systemd units?

PR #1274, opened 2018-05-03
Allow cfn-signal.service to cleanly start under certain circumstances

davidmccormick:
Stops cfn-signal service from failing in circumstances where the cloud-formation stack has succeeded as in issue https://github.com/kubernetes-incubator/kube-aws/issues/1273

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -782,7 +782,7 @@ write_files:
       result_code=$?
       rkt rm --uuid-file=/var/run/coreos/cfn-signal.uuid || :
 
-      # Filter out beneign failures to signal stack
+      # Filter out benign failures to signal stack
       if [ $result_code -ne 0 ]; then
         test_for_already_sent=$(echo "$result" | grep "for resource Controllers already exists.")
         [ -n "${test_for_already_sent}" ] && exit 0
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -771,13 +771,13 @@ write_files:
         --volume=awsenv,kind=host,source=/var/run/coreos,readOnly=false \
         --mount volume=awsenv,target=/var/run/coreos \
         --uuid-file-save=/var/run/coreos/cfn-signal.uuid \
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -778,10 +778,20 @@ write_files:
...
+          ')
+      result_code=$?
       rkt rm --uuid-file=/var/run/coreos/cfn-signal.uuid || :
 
+      # Filter out beneign failures to signal stack
+      if [ $result_code -ne 0 ]; then
+        test_for_already_sent=$(echo "$result" | grep "for resource Controllers already exists.")
+        [ -n "${test_for_already_sent}" ] && exit 0
+        test_for_stack_completed=$(echo "$result" | grep "is in UPDATE_COMPLETE state and cannot be")
+              [ -n "$test_for_stack_completed" ] && exit 0
```
c-knowles: nit, align to other lines?

c-knowles:
/lgtm

Currently testing this out on our dev cluster

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -778,10 +778,23 @@ write_files:
...
-
+          ')
+      result_code=$?
+      echo $result
       rkt rm --uuid-file=/var/run/coreos/cfn-signal.uuid || :
 
+      # Filter out benign failures to signal stack
+      if [ $result_code -ne 0 ]; then
+        case $result in
+          *"for resource Controllers already exists"*)     exit 0
```
mumoshu: I can't remember what was this message and when it happens 🤔 Would you mind providing me an example?

davidmccormick: It comes up if the message has already been sent before the cloudformation update is complete, so I saw it when I was testing the script for the failure responses.  We should probably squash both of them and have the service succeed should someone restart the cfn-signal service or the node for some reason.

mumoshu: Ah, got it! Thanks for the clarification.

mumoshu:
@davidmccormick Thanks for your contribution 🎉

PR #1276, opened 2018-05-05
Disable alpha batch APIs by default

c-knowles:
Batch has been beta since k8s 1.8.

Without this it's possible to start getting this error using latest kubectl.
```
{  cronjob} matches multiple kinds [batch/v1beta1, Kind=CronJob batch/v2alpha1, Kind=CronJob]
```


mumoshu:
LGTM. Thanks for the fix 🍻

Issue #1215, opened 2018-04-04
kiam missing from default cluster.yaml


c-knowles:
Master supports kiam since https://github.com/kubernetes-incubator/kube-aws/issues/1055, we should add the options to the default generated cluster.yaml.

PR #1277, opened 2018-05-05
kiam support improvements

c-knowles:
Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1215.

Also contains some additional tests, bug fix to AssumeRole and update to kiam 2.7.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -193,7 +193,7 @@ kmsKeyArn: "{{.KMSKeyARN}}"
 #  # If you specify managedIamRoleName the role created for controller nodes will not suffix the random id at the end
 #  # Role will be created with "Ref": {"AWS::StackName"}-{"AWS::Region"}-yourManagedRole
 #  # to follow the recommendation in AWS documentation  http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html
-#  # This is also intended to be used in combination with .experimental.kube2IamSupport. See #297 for more information.
+#  # This is also intended to be used in combination with .experimental.kube2IamSupport or .experimental.kiamSupport. See #297 for more information.
 #  #
 #  # ATTENTION: Consider limiting number of characters in clusterName and iam.role.name to avoid the resulting IAM
 #  # role name's length from exceeding the AWS limit: 64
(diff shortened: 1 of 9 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -998,6 +1004,12 @@ worker:
 #  tag: 1.1.2
 #  rktPullDocker: false
 
+# kiam image repository to use.
+kiamImage:
```
c-knowles: Similar to https://github.com/kubernetes-incubator/kube-aws/pull/1002, I think having all the defaults in a single place is beneficial. 2.7 here now differs from 2.6 in the code, I'm not sure if it'd be better to remove the code level default or now.

c-knowles:
From what I can see the validation code could do with an overhaul, it seems we have multiple pieces of code in the same files performing various forms of validation. Adding to that was a lot less simple than I expected.

core/controlplane/config/templates/cluster.yaml
mumoshu: I'm ok with removing all the code level defaults, as long as missing values in cluster.yaml doesn't make kube-aws fail.
The reasoning is that probably users want to avoid including e.g. kiam related settings at all in cluster.yaml when one doesn't enable kiam.

docs/getting-started/step-6-configure-add-ons.md
```diff
@@ -38,10 +38,10 @@ The above example configuration would:
   * If there are no unschdulable, pending pod(s) that is waiting for more capacity and one or more nodes are in low utlization, CA will remove node(s), down until the min size `1`
 * The second node pool `notScaled` is scaled manually by YOU, because you had not the autoscaling on it(=missing `autoscaling.clusterAutoscaler.enabled`)
 
-## kube2iam
+## kube2iam / kiam
  
-[kube2iam](https://github.com/jtblin/kube2iam) is an add-on which provides IAM credentials for target IAM roles to pods running inside a Kubernetes cluster based on annotations.
-To allow kube2iam deployed to worker and controller nodes to assume target roles, you need the following configurations.
+[kube2iam](https://github.com/jtblin/kube2iam) and [kiam](https://github.com/uswitch/kiam) are add-ons which provides IAM credentials for target IAM roles to pods running inside a Kubernetes cluster based on annotations.
```
mumoshu: Thanks for your love on the docs ❤️

mumoshu:
> From what I can see the validation code could do with an overhaul, it seems we have multiple pieces of code in the same files performing various forms of validation. Adding to that was a lot less simple than I expected.

Are you talking about the huge maincluser_test.go?

mumoshu:
LGTM! Thanks for the improvement 👍

core/controlplane/config/templates/cluster.yaml
c-knowles: I agree some users want a minimal config file. We could always put all ththe defaults in on place in the code then generate this.

c-knowles:
@mumoshu not just the main cluster test but config validation too, some files are fairly unreadable. I’m trying to devise a plan where we could have a base test that validates and somehow separate out the concerns of specific addons or checks.

core/nodepool/config/templates/stack-template.json
```diff
@@ -349,7 +349,7 @@
                   "Resource": "arn:{{.Region.Partition}}:s3:::{{ .KubeResourcesAutosave.S3Path }}/*"
                 },
                 {{end}}
-                {{if .Kube2IamSupport.Enabled }}
+                {{if or .Kube2IamSupport.Enabled .KIAMSupport.Enabled }}
```
c-knowles: @mumoshu sorry, this bit was incorrect - it's not needed (rest is fine though). Only the kiam server needs this. I'm going to remove it in https://github.com/kubernetes-incubator/kube-aws/pull/1285 if that's ok as there are some merge conflicts there.

PR #1282, opened 2018-05-05
Fix temporary kubelet.service failures

### LINKED ISSUE #1278: kubelet.service failing due to missing key
@davidmccormick Hey! I'm seeing kubelet.service on controller nodes being failed on initial startup, due to missing etcd-client-key.pem like this:

```
May 05 12:33:07 ip-10-0-0-214.ap-northeast-1.compute.internal etcdctl[1131]: Error:  open /etc/kubernetes/ssl/etcd-client-key.pem: no such file or directory
May 05 12:33:07 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Control process exited, code=exited status=4
May 05 12:33:07 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Failed with result 'exit-code'.
May 05 12:33:07 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: Failed to start kubelet.service.
May 05 12:33:17 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Service hold-off time over, scheduling restart.
May 05 12:33:17 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: kubelet.service: Scheduled restart job, restart counter is at 1.
May 05 12:33:17 ip-10-0-0-214.ap-northeast-1.compute.internal systemd[1]: Stopped kubelet.service.
```

Note that I have not yet enabled self-hosting calico+flannel.
Looking into the code-base, it turns out we miss decrypt-tls-assets.service in dependency when self-hosting is disabled:

https://github.com/kubernetes-incubator/kube-aws/blob/15458e8469ce1f3c7bce7a11ef3a0ab405cd1950/core/controlplane/config/templates/cloud-config-controller#L274-L279
https://github.com/kubernetes-incubator/kube-aws/blob/15458e8469ce1f3c7bce7a11ef3a0ab405cd1950/core/controlplane/config/templates/cloud-config-controller#L311-L319

Do you think we can just extract this block out of the `else` block so that we can ensure `decrypt-tls-assets` is executed, as it is required regardless of self-hosting is enabled or not, anyway?

```
Wants=decrypt-assets.service 
 After=decrypt-assets.service 
```

mumoshu:
Fixes #1278

Depends on #1281, but will be rebased and squashed after merging it.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -273,10 +273,9 @@ coreos:
         After=cfn-etcd-environment.service
         {{ if not .Kubernetes.Networking.SelfHosting.Enabled -}}
         Wants=flanneld.service
-        {{ else -}}
+        {{ end -}}
         Wants=decrypt-assets.service
         After=decrypt-assets.service
-        {{- end }}
 
         [Service]
         # EnvironmentFile=/etc/environment allows the reading of COREOS_PRIVATE_IPV4
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -92,7 +92,6 @@ coreos:
         Restart=on-failure
         RemainAfterExit=true
         ExecStartPre=/opt/bin/cfn-etcd-environment
-        ExecStartPre=/usr/bin/mv -f /var/run/coreos/etcd-environment /etc/etcd-environment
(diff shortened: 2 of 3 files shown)


mumoshu:
Thanks for the review!
Verified that it fixes the specific race issue by creating a brand new cluster from this branch. Merging

Issue #1045, opened 2017-11-29
`experimental.ClusterAutoscalerSupport` is not documented


c-knowles:
The code references `experimental.ClusterAutoscalerSupport` but it's not documented in the default `cluster.yaml` or the documentation site.

PR #1284, opened 2018-05-06
Cluster autoscaler docs/tests

c-knowles:
Basic tests for cluster autoscaler config.

Bug fix also found during testing - only label the controller nodes if `experimental.clusterAutoscalerSupport.enabled` is set to true (which is the default).

Improved cluster autoscaler documentation as well which will fix https://github.com/kubernetes-incubator/kube-aws/issues/1045

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -1161,9 +1161,13 @@ func (c Cluster) APIAccessAllowedSourceCIDRsForControllerSG() []string {
 	return cidrs
 }
 
+func (c Cluster) ClusterAutoscalerSupportEnabled() bool {
+	return c.Addons.ClusterAutoscaler.Enabled && c.Experimental.ClusterAutoscalerSupport.Enabled
+}
+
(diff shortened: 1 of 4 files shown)


core/controlplane/config/config.go
```diff
@@ -1141,9 +1141,13 @@ func (c Cluster) NestedStackName() string {
...
 }
 
+func (c Cluster) ClusterAutoscalerSupportEnabled() bool {
+	return c.Addons.ClusterAutoscaler.Enabled && c.Experimental.ClusterAutoscalerSupport.Enabled
+}
+
 func (c Cluster) NodeLabels() model.NodeLabels {
 	labels := c.NodeSettings.NodeLabels
-	if c.Addons.ClusterAutoscaler.Enabled {
+	if c.ClusterAutoscalerSupportEnabled() {
```
mumoshu: Ah, good catch 👍

mumoshu:
@c-knowles Would you mind rebasing this too?

mumoshu:
LGTM. Thanks as always!

Issue #753, opened 2017-07-11
Remove deprecated keys in cluster.yaml


mumoshu:
Finally remove keys deprecated via #372 #545 

More concretely, remove:

* [ ] `externalDNSName` in favor of `apiEndpoints[].dnsName`
* [ ] `createRecordSet` in favor of `apiEndpoints[].loadBalancer.createRecordSet`
* [ ] `recordSetTTL` in favor of `apiEndpoints[].loadBalancer.recordSetTTL`
* [ ] `hostedZoneId` in favor of `apiEndpoints[].loadBalancer.hostedZone.id`
* [x] `controllerCount` in favor of `controller.count`
* [ ] `controller.loadBalancer` in favor of `apiEndpoints[].loadBalancer`
* [x] `controllerCreateTimeout` in favor of `controller.createTimeout`
* [x] `controllerInstanceType` in favor of `controller.instanceType`
* [x] `controllerRootVolumeSize` in favor of `controller.rootVolume.size`
* [x] `controllerRootVolumeType` in favor of `controller.rootVolume.type`
* [x] `controllerRootVolumeIOPS` in favor of `controller.rootVolume.iops`
* [x] `controller.managedIamRoleName` in favor of `controller.iam.role.name`
* [x] `etcdCount` in favor of `etcd.count`
* [x] `etcdInstanceType` in favor of `etcd.instanceType`
* [x] `etcdRootVolumeSize` in favor of `etcd.rootVolume.size`
* [x] `etcdRootVolumeType` in favor of `etcd.rootVolume.type`
* [x] `etcdRootVolumeIOPS` in favor of `etcd.rootVolume.iops`
* [x] `etcdDataVolumeEphemeral` in favor of `etcd.dataVolume.ephemeral`
* [x] `etcdDataVolumeSize` in favor of `etcd.dataVolume.size`
* [x] `etcdDataVolumeType` in favor of `etcd.dataVolume.type`
* [x] `etcdDataVolumeIOPS` in favor of `etcd.dataVolume.iops`
* [x] `etcdDataVolumeEncrypted` in favor of `etcd.dataVolume.encrypted`
* [x] `etcdTenancy` in favor of `etcd.tenancy`
* [x] `worker.nodePools[].rootVolumeSize` in favor of `worker.nodePools[].rootVolume.size`
* [x] `worker.nodePools[].rootVolumeType` in favor of `worker.nodePools[].rootVolume.type`
[...]

Issue #698, opened 2017-06-09
Feedback on IAM role creation


c-knowles:
Following up on @Fsero's nice work from https://github.com/kubernetes-incubator/kube-aws/pull/607. So far it's mostly been exactly what we need. I hit a couple of doc issues I think we should improve:

1. `cluster.yaml` mentions `{"AWS::StackName"}-{"AWS::Region"}-yourManagedRole` a couple of times but this was changed in https://github.com/kubernetes-incubator/kube-aws/pull/607 to not include the stack name. I think the recommendation should be to include the stack name otherwise we end up with fairly generic `eu-west-1-poolname-role` role names.
1. The validation for the character limit is still in place prefixed with stack name - [see here](https://github.com/kubernetes-incubator/kube-aws/blob/1f08f3bb7ebe12cca738ba25df42263ca077cc3a/cfnresource/naming.go#L8)
1. It was not totally obvious to me at first that the role names get exported from the node pool stacks as well as the root stack. The exports from the node pools look like `k8s-dev-PoolName-Y1WKSC7JU6JS-WorkerIAMRoleArn` but the one from the root looks like `k8s-dev-NodePoolPoolNameWorkerIAMRoleArn` which is obviously easier to reference
1. There's warnings to not use the documented settings as below but the settings recommended don't seem to work (both controller and worker settings under `iam` seem to not do anything:

```
WARN: worker.nodePools[].managedIamRoleName is deprecated and will be removed in v0.9.7. Please use worker.nodePools[].iam.managedRoleName instead
WARN: controller.managedIamRoleName is deprecated and will be removed in v0.9.7. Please use controller.iam.managedIamRoleName instead
```

Then one question for @Fsero - given that with an export we can reference the ARN of the worker pool roles, what does overriding the unique ID name of the actual role provide? I'd assume any references would use the ARN export and not reference the name directly.

c-knowles:
I think I answered my own question from the final part of the above. Even if the exported role name is consistent, AWS still produces an error if the underlying ARN changes and another stack depends on the exported name. I wonder if they will change that behaviour at some point so it can be a bit more dynamic.

Fsero:
Hi @c-knowles! Thanks for your comments :)

Let us review them one by one

[...]

PR #1285, opened 2018-05-06
Improvement/named iam roles cleanup

c-knowles:
Roles used to include the stack name but since removing that they can clash for clusters with mirrored config in the same account+region. Fixes most of #698.

Doco updates fixes comments from https://github.com/kubernetes-incubator/kube-aws/issues/753#issuecomment-372704508


c-knowles:
@Fsero might be worth you having a quick look as this will affect your use case but only in a very basic way, the names are still stable.

core/controlplane/config/templates/cluster.yaml
```diff
@@ -491,14 +485,8 @@ worker:
...
-#      # This will use SSL certificates generated during `kube-aws render credentials`, ensure this has been run with kube-aws v0.9.9+
-#      kiamSupport:
-#        enabled: true
-#
-#      # When enabled it will grant sts:assumeRole permission to the IAM roles for worker nodes in this pool.
-#      # This is intended to be used in combination with managedIamRoleName. See #297 for more information.
+#      # When enabled this will grant sts:assumeRole permission to the IAM roles for worker nodes in this pool.
+#      # This is intended to be used in combination with nodePools[].iam.role.name. See #297 for more information.
 #      kube2IamSupport:
 #        enabled: true
```
c-knowles: @mumoshu do you know why we have `kube2IamSupport` on node pools? I thought it was to allow only some node pools to have this enabled but having looked at it again I cannot see anything which prevents the daemon set from running across everything.

mumoshu: @c-knowles Wow, good catch! I can't imagine how I miss that until now, but probably we need node affinity:

```yaml
              affinity:
                nodeAffinity:
                  requiredDuringSchedulingIgnoredDuringExecution:
                    nodeSelectorTerms:
                    - matchExpressions:
                      - key: "kube-aws.coreos.com/assume-role-supported"
                        operator: "In"
                        values:
                        - "true"
```

and corresponding node labels?

core/nodepool/config/templates/stack-template.json
```diff
@@ -349,7 +349,7 @@
                   "Resource": "arn:{{.Region.Partition}}:s3:::{{ .KubeResourcesAutosave.S3Path }}/*"
                 },
                 {{end}}
-                {{if or .Kube2IamSupport.Enabled .KIAMSupport.Enabled }}
+                {{if .Kube2IamSupport.Enabled }}
```
mumoshu: Ah, yes. We have a nodeSelector to stick kiam server to controller nodes. So we don't need to consult `.KIAMSupport.Enabled` here. Good catch!

core/controlplane/config/templates/cluster.yaml
```diff
@@ -491,14 +485,8 @@ worker:
 #      ephemeralImageStorage:
 #        enabled: true
 #
-#      # When enabled this will install kiam using the repo from `kiamImage` (ref https://github.com/uswitch/kiam)
-#      # It will also grant sts:assumeRole permission to the IAM role for worker nodes in this pool.
-#      # This will use SSL certificates generated during `kube-aws render credentials`, ensure this has been run with kube-aws v0.9.9+
-#      kiamSupport:
-#        enabled: true
```
mumoshu: Ah, yes. Currently we have a nodeSelector to stick kiam server to controller nodes. So no need for kiamSupport in worker node pools. Good catch 👍

mumoshu:
@c-knowles Should we note about a breaking change for this? Probably you need to update trust-relationship of the roles assumed by kube2iam or kiam before running `kube-aws update`. (I'd recommend bringing up a new cluster for kube-aws upgrade, but worth noting ,anyway?

Fsero:
@mumoshu i think is a good idea to note the breaking change over the release notes yes.
@c-knowles thanks for the refactor, i think it makes sense and it does not affect me because i'm handling IAM outside kube-aws and instructing kube-aws to use an existing instance-profile since my IAM Roles scheme are pretty specific for our use case.

mumoshu:
@Fsero Thanks for the confirmation 👍 Will add a release note.

core/controlplane/config/templates/cluster.yaml
c-knowles: @mumoshu yeah something like that, it seems totally missing from the beginning of this feature as far as I can tell.

Issue #419, opened 2017-03-17
"write /dev/stdout: no space left on device" errors in kube-apiserver when audit logging is enabled


mumoshu:
`/dev/stdout` seems to be rotated when the experimental "auditLog" is enabled for kube-apiserver and the tmpfs mounted at `/dev` in the apiserver pod eventually get filled up.

---

My kube-apiserver pods started spamming error messages like:

```
$ docker logs -f 0d9cfab181d7
*snip*
E0317 09:15:55.998046       1 audit.go:126] Unable to write audit log: 2017-03-17T09:15:55.998018096Z AUDIT: id="ccac3ba1-4433-4fad-abbe-00fb118bce8d" ip="127.0.0.1" method="GET" user="<none>" as="<self>" asgroups="<lookup>" namespace="kube-system" uri="/api/v1/namespaces/kube-system/endpoints/kube-scheduler"
, the error is: write /dev/stdout: no space left on device
```

How `/dev/stdout` can be filled up like this?
Investigating into the primary container of one of the pods revealed that the original `/dev/stdout`(=`/proc/self/fd/1`) had been rotated over and over the "current" `/dev/stdout` is now a regular file 😆 

```
$ docker exec -it 0d9cfab181d7 sh
# df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          27G  4.4G   22G  18% /
tmpfs           1.9G  1.9G     0 100% /dev
tmpfs           1.9G     0  1.9G   0% /sys/fs/cgroup
/dev/xvda9       27G  4.4G   22G  18% /etc/hosts
/dev/xvda3      985M  588M  347M  63% /etc/ssl/certs
shm              64M     0   64M   0% /dev/shm
```

```
# ls -lah /dev/stdout
-rw-------. 1 root root 84M Mar 17 09:28 /dev/stdout
# ls -lah /dev
total 1.9G
drwxr-xr-x. 5 root root     760 Jan 12 12:36 .
drwxr-xr-x. 1 root root    4.0K Dec 27 06:36 ..
lrwxrwxrwx. 1 root root      11 Dec 27 06:36 core -> /proc/kcore
lrwxrwxrwx. 1 root root      13 Dec 27 06:36 fd -> /proc/self/fd
crw-rw-rw-. 1 root root  1,   7 Dec 27 06:36 full
crw-rw-rw-. 1 root root 10, 229 Dec 27 06:36 fuse
[...]

PR #1289, opened 2018-05-06
Normalize audit log attribute names in cluster.yaml + add support for missing attributes

danielfm:
The `auditLog.(logpath|maxage)` attribute names were normalized to `auditLog.(logPath|maxAge)`, and added support for the `maxBackup` (which was hard-coded to `1`) and `maxSize` flags.

Also, the `cluster.yaml` now uses the same default value for `auditLog.logPath` as the one defined in the struct.

### DIFF
--- a/core/controlplane/config/templates/cluster.yaml
+++ b/core/controlplane/config/templates/cluster.yaml
@@ -1311,8 +1311,10 @@ experimental:
   # Enable audit log for apiserver. Recommended when `rbac` is enabled.
   auditLog:
     enabled: false
-    maxage: 30
-    logpath: /dev/stdout
+    logPath: /dev/stdout
+    maxAge: 30
+    maxBackup: 1
+    maxSize: 100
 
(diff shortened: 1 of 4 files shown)


core/controlplane/config/templates/cluster.yaml
```diff
@@ -1311,8 +1311,10 @@ experimental:
   # Enable audit log for apiserver. Recommended when `rbac` is enabled.
   auditLog:
     enabled: false
-    maxage: 30
-    logpath: /dev/stdout
+    logPath: /var/log/kube-apiserver-audit.log
```
c-knowles: Ref https://github.com/kubernetes-incubator/kube-aws/pull/579 when this was changed in the code. Never seen stdout working well so far with regard to https://github.com/kubernetes-incubator/kube-aws/issues/419.

mumoshu:
@danielfm LGTM! Thanks again for your contribution.

Issue #1291, opened 2018-05-07
Stop relying on the apiserver insecure port for control plane components and bootstrapping


danielfm:
There are several control plane components (i.e. controller manager, scheduler, etc) and bootstrapping tasks (i.e. systemd `ExecStartPre`s, `kube-node-label`, etc) on `userdata/cloud-config-controller` still relying on apiserver's insecure port.

Aside from being a security concern, the insecure port binding in apiserver is deprecated and will soon be removed.

davidmccormick:
Hi

There are a number of systemd service scripts that use the insecure port including install-kube-system so we must be careful to give them a certificate they can use before disabling the port.  I only mention this because I have implemented a new systemd service that assumes it can use the insecure port! :)

PR #1292, opened 2018-05-08
Enable TLS for communication between control plane components and the local apiserver

danielfm:
This is a first step towards disabling the apiserver insecure port binding entirely; in this PR, we generate TLS client certificates for `kube-controller-manager` and `kube-scheduler`, and change those manifests to communicate with the apiserver via the encrypted (authenticated) port.

The `kube-proxy` pods on controller nodes were also using the apiserver insecure port, and this PR changes that as well.

Ref: #1291


danielfm:
Conformance test results:
https://scanner.heptio.com/fdfe8f4cbb8e41efcaf0fe7e836fc969/diagnostics/

mumoshu:
@danielfm Thank you so much for your continuous efforts on security 👍 
LGTM.

Issue #1255, opened 2018-04-25
v0.9.10-rc.5 Unknown lvalue '--feature-gates'


VinceMD:
Creating a cluster from v0.9.10-rc.5 

(is there already a Github ticket for release candidate related issues? let me know if that's not the right process) 

Under worker.nodepool[0]

`      featureGates:
        PodPriority: true`

And Experimental.Admission.Priority.Enabled: true

Controller failed to be created:

`"/run/systemd/system/kubelet.service:25: Unknown lvalue '--feature-gates' in section 'Service'"`

* Why is kubelet complaining?
* kube-aws v0.9.10-rc.5  uses Kubernetes 1.9.3. Do I still need a flag to enable PodPriority? Or is it included by default?  

I would like to use and experiment with Overprovisioning in autoscaler (https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler) which allows one to create a buffer thanks to low priority pods. Any tips on getting access to this feature in kube-aws?

mumoshu:
@VinceMD Thanks for the report!
Not sure exactly but it seems like a bug in kube-aws, probably due to wrong templating.

Could you share an output from `systemctl cat kubelet.service` on the failed controller node?

mumoshu:
@VinceMD Thanks! Not sure how this had been considered working until now. Anyway I believe the template should be changed:

before:

```
        --cloud-provider=aws \
        {{if .Experimental.Admission.Priority.Enabled}}
        --feature-gates=PodPriority=true \
        {{end}}\
        $KUBELET_OPTS
```

after:

```
        --cloud-provider=aws \
        {{if .Experimental.Admission.Priority.Enabled -}}
        --feature-gates=PodPriority=true \
        {{end -}}
        $KUBELET_OPTS
```

otherwise the stray empty lines would isolate `--feature-gates=PodPriority=true \` which ends up the error you're seeing...

VinceMD:
@mumoshu it didn't work as expected as discussed in https://github.com/kubernetes-incubator/kube-aws/issues/1265


```cat /etc/kubernetes/manifests/kube-apiserver.yaml
apiVersion: v1
kind: Pod
metadata:
  name: kube-apiserver
  namespace: kube-system
  labels:
    k8s-app: kube-apiserver
spec:
  hostNetwork: true
  containers:
  - name: kube-apiserver
[...]

PR #1243, opened 2018-04-16
Admission Controller for PersistentVolumeClaimResize

kevtaylor:

Issue #1204, opened 2018-03-27
Support for volumeMounts on controllers/etcd


kiich:
This is more of a question - when I try to use `volumeMounts` in controller and etcd, sure enough I get an error:

```
Error: Failed to read cluster config: file cluster.yaml: unknown keys found in controller: volumeMounts
```
And also for etcd
```
Error: Failed to read cluster config: file cluster.yaml: unknown keys found in etcd: volumeMounts
```

I can see that there is no support for this but was wondering why - I have a use case where we need to dump journald logs (filtered) to file system and I was not so keen on catering for this extra data on root file system and thought the extra ebs will do the job for me.

Let me know if there is any other way to do something like this, thanks.

mumoshu:
@kiich Good catch! It sounds like a good feature request.

If you're willing to submit a PR for this, I'd be happy to review!
In that case, the declaration of `systemd-modules-load.service` service seen in `core/controlplane/config/templates/cloud-config-worker` would be a good starting point.

kiich:
Sounds good @mumoshu - I'll take a look!

PR #1293, opened 2018-05-08
Add in ExpandPersistentVolumes as part of feature-gates on controller

kiich:
Compliments https://github.com/kubernetes-incubator/kube-aws/pull/1243 by adding `ExpandPersistentVolumes` as part of `--feature-gates` option when `PersistentVolumeClaimResize` is enabled via admission control.

Also noticed fix for https://github.com/kubernetes-incubator/kube-aws/issues/1255 so rolled the fix into `kubelet` part.

Ideally, would like to proceed further by supporting same method as workers by supporting following for `controller` block:
```
featureGates:
        ExpandPersistentVolumes: true
```


mumoshu:
@kiich Thanks! In the future, we should make it more general so that we won't need to add available feature gates one by one..

mumoshu:
@kiich Would you mind rebasing?

kiich:
@mumoshu I was actually looking into how the `featureGates` works so would be up for PR-ing similar/same logic for the rest. Will carry on after upstreaming PR for https://github.com/kubernetes-incubator/kube-aws/issues/1204 which I'm nearly done with.

kiich:
@mumoshu re: rebasing - i just fixed the conflict in github editor - was this enough? (sorry for obvious questions)

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -348,9 +348,9 @@ coreos:
         {{ else }}--cluster-dns={{.DNSServiceIP}} \
         {{ end }}--cluster-domain=cluster.local \
         --cloud-provider=aws \
-        {{if .Experimental.Admission.Priority.Enabled}}
-        --feature-gates=PodPriority=true \
-        {{end}}\
+        {{if or (.Experimental.Admission.Priority.Enabled) (.Experimental.Admission.PersistentVolumeClaimResize.Enabled) -}}
+        --feature-gates=PodPriority={{.Experimental.Admission.Priority.Enabled}},ExpandPersistentVolumes={{.Experimental.Admission.PersistentVolumeClaimResize.Enabled}} \
+        {{end -}}\
```
mumoshu: Good try! But I guess the trailing slash still breaks the commmand. Let'ts fix it if this turns out to remain problematic.

mumoshu:
@kiich LGTM. Thanks a ton for the fix!

Regarding #1255, I have a concern noted in https://github.com/kubernetes-incubator/kube-aws/pull/1293#discussion_r187495186. I appreciate it if you could confirm that.

core/controlplane/config/templates/cloud-config-controller
kiich: Hi @mumoshu 
I thought so as well but it seems to be working/not showing the same problem as #1255 
Whenever i toggle either of `Priority` or `PersistentVolumeClaimResize`, the resulting kubelet service seems to pass ok.
e.g. When I enable `Priority`, the following is shown for `kubelet.service` and Controllers seems to come up.

```
--cloud-provider=aws \
--feature-gates=PodPriority=true,ExpandPersistentVolumes=false \
\
$KUBELET_OPTS
Restart=always
RestartSec=10```

As you can see, no spaces/lines that was causing the issue in #1255 as far as I can tell.

**Comment by kiich** [comment]
Thanks @mumoshu - I can confirm the blank lines seen in #1255 does not seem to occur and controllers coming up ok as far as I can tell.

File: core/controlplane/config/templates/cluster.yaml [hunk 2]
```diff
@@ -1319,6 +1322,8 @@ experimental:
       enabled: false
     OwnerReferencesPermissionEnforcement:
       enabled: false
+    # persistentVolumeClaimResize enables PersistentVolumeClaim in the API server, scheduler and kubelet. you need to manually add the
+    # featureGate ExpandPersistentVolumes:true in the worker
```
c-knowles: @kiich @mumoshu, do you know when enabling the feature gate on node pools is necessary as per the comment? Is it only when control plane pods are scheduled onto workers?

PR #1299, opened 2018-05-11
tlsutil cleanup plus added tests

pete911:
Started with tlsutil - adding tests and cleaning up.

### DIFF
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -19,6 +19,7 @@ import (
 	"github.com/kubernetes-incubator/kube-aws/gzipcompressor"
 	"github.com/kubernetes-incubator/kube-aws/model"
 	"github.com/kubernetes-incubator/kube-aws/netutil"
+	"github.com/kubernetes-incubator/kube-aws/tlscerts"
 	"github.com/kubernetes-incubator/kube-aws/tlsutil"
 )
 
@@ -485,7 +486,7 @@ func ReadRawAssets(dirname string, manageCertificates bool, caKeyRequiredOnContr
 			return nil, fmt.Errorf("error reading credential file %s: %v", path, err)
 		}
 		if file.expiryCheck {
-			certs, err := tlsutil.ToCertificates(data.content)
+			certs, err := tlscerts.FromBytes(data.content)
 			if err != nil {
 				return nil, err
 			}
@@ -577,7 +578,7 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, caKeyRequiredO
 				return nil, fmt.Errorf("error reading credential file %s: %v", path, err)
 			}
 			if file.expiryCheck {
-				certs, err := tlsutil.ToCertificates(raw.content)
+				certs, err := tlscerts.FromBytes(raw.content)
 				if err != nil {
 					return nil, err
 				}
--- a/core/controlplane/config/encrypted_assets.go
+++ b/core/controlplane/config/encrypted_assets.go
@@ -482,11 +482,17 @@ func ReadRawAssets(dirname string, manageCertificates bool, caKeyRequiredOnContr
 		path := filepath.Join(dirname, file.name)
 		data, err := RawCredentialFileFromPath(path, file.defaultValue)
 		if err != nil {
-			return nil, fmt.Errorf("Error reading credential file %s: %v", path, err)
+			return nil, fmt.Errorf("error reading credential file %s: %v", path, err)
 		}
 		if file.expiryCheck {
-			if err := tlsutil.CheckAllCertsValid(path, data.content); err != nil {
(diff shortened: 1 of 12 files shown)


tlsutil/cert.go
```diff
@@ -68,9 +102,22 @@ func ToCertificates(data []byte) ([]Certificate, error) {
...
+				IPAddresses: c.IPAddresses,
 			},
 		)
 	}
 	return certificates, nil
 }
+
+// returns certificate that matches subject CN match regex (Subject.CommonName), if the certificate cannot be found,
+// second returned value will be false
+func GetCertificate(certificates []Certificate, subjectCNMatch string) (cert Certificate, ok bool) {
```
mumoshu: I prefer `GetCertificateBySubjectCommonNamePattern`, or even nicer `certificates.GetBySubjectCommonNamePattern` w/ `type Certificates []Certificates`!

core/controlplane/config/encrypted_assets.go
```diff
@@ -558,22 +564,28 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, caKeyRequiredO
...
 			}
 		} else {
 			raw, err := RawCredentialFileFromPath(path, file.defaultValue)
 			if err != nil {
-				return nil, fmt.Errorf("Error reading credential file %s: %v", path, err)
+				return nil, fmt.Errorf("error reading credential file %s: %v", path, err)
 			}
 			if file.expiryCheck {
-				if err := tlsutil.CheckAllCertsValid(path, raw.content); err != nil {
+				certs, err := tlsutil.ToCertificates(raw.content)
```
mumoshu: I prefer `tlsutil.CertificatesFromBytes` to make it even more explicit about in/out.
An alternative would be `certs.FromBytes` if we can safely extract a dedicated package `certs` out of the `tlsutil`. When we do many things in a single pkg named *util, I tend to split it into meaningful packages.

core/controlplane/config/encrypted_assets.go
```diff
@@ -558,22 +564,28 @@ func ReadOrEncryptAssets(dirname string, manageCertificates bool, caKeyRequiredO
...
+				if err != nil {
 					return nil, err
 				}
+				for _, cert := range certs {
+					if cert.IsExpired() {
+						return nil, fmt.Errorf("the following certificate in file %s has expired:-\n\n%s", path, cert)
```
mumoshu: Seems like we miss the useful info originally included in the error in [`CheckAllCertsValid`](https://github.com/kubernetes-incubator/kube-aws/pull/1299/files#diff-ff7418bed798ff0b9347acef6765cb23L192)?

core/controlplane/config/encrypted_assets.go
```diff
@@ -482,12 +482,18 @@ func ReadRawAssets(dirname string, manageCertificates bool, caKeyRequiredOnContr
 		path := filepath.Join(dirname, file.name)
 		data, err := RawCredentialFileFromPath(path, file.defaultValue)
 		if err != nil {
-			return nil, fmt.Errorf("Error reading credential file %s: %v", path, err)
+			return nil, fmt.Errorf("error reading credential file %s: %v", path, err)
```
mumoshu: Yep, an error message should start with a lower case character! Thx for starting to conform that.

mumoshu:
@pete911 Thank you so much for your efforts!
Just finished the first-pass review. WDYT?

tlsutil/cert.go
pete911: like the Certificates type, what do you think of `certificates.GetBySubjectCNPattern` a bit shorter

core/controlplane/config/encrypted_assets.go
pete911: certificate has `String()` method that prints all the info about certificate. This method is used in show certificates command as well

mumoshu: Ah! Awesome. I missed that. LGTM 👍

tlsutil/cert.go
mumoshu: `certificates` in this context was a variable rather than pkg name.

So overall, I am thinking about naming the package `tlscerts` and introduce:

- `type Certificates []Certificate`
- `func (c *Certificates) GetBySubjectCommonNamePattern(...) ...`
- Include these in a newly introduced pkg `tlscerts`
- According to https://github.com/kubernetes-incubator/kube-aws/pull/1299#discussion_r187797767, you use it like:

```
certs, err := tlscerts.FromBytes(raw.content)
cert, ok := certs.GetBySubjectCommonNamePattern(...)
// do something useful with the found cert
```

WDYT?

pete911: done, let me know what you think

pete911:
ok, all the comments should be done

mumoshu:
@pete911 Wow, that was quick! LGTM. I'll merge this once the build passes.

PR #1305, opened 2018-05-16
append comma when necessary to the EtcdNodes range

### LINKED ISSUE #1303: Unable to validate using latest master after etcd stack was separated out
Not sure if I am the only one seeing this issue but I've tried it using latest master with following `etcd` settings and I can always reproduce the below error when doing `kube-aws validate`:
```
Error: Failed to initialize cluster driver: failed to initialize etcd stack: Error while rendering template: invalid character '"' after object key:value pair:
json syntax error (offset=15384), in this region:
-------
}



    "Etcd1ENI": {
      "Properties": {
        "SubnetId": "subnet-xxxxxxxx",
        "GroupSet": [

-------
```

With etcd config:
```
etcd:
  count: 3
  instanceType: m4.large
  subnets:
    - name: a
    - name: b
    - name: c
  memberIdentityProvider: eni
```
The last setting doesn't seem to matter as I get similar syntax error with eip.

It looks like when rendering happens, the code that iterates through etcd's launch config is not appending necessary `,` for each LC and hence it sees the syntax error.

Rendered result looked like this:
```
    "Etcd0LC": {
      "Properties": {
        "BlockDeviceMappings": [
          {
            "DeviceName": "/dev/xvda",
            "Ebs": {
              "VolumeSize": "30",

              "VolumeType": "gp2"
            }
          }

        ],
[...]
      },
      "Type": "AWS::AutoScaling::LaunchConfiguration"
    }




    "Etcd1ENI": {
      "Properties": {
        "SubnetId": "subnet-xxxxxxxx",
        "GroupSet": [


          {"Fn::ImportValue" : {"Fn::Sub" : "${NetworkStackName}-EtcdSecurityGroup"}}

        ]
      },
      "Type": "AWS::EC2::NetworkInterface"
    },
```

As you can see, there's a missing `,` before the `Etcd1ENI` starts.
If I only have 1 etcd, the syntax error goes away (though I see another error but at least not syntax error).

I'm wondering if I've done something wrong in `cluster.yaml` (since master has had this feature for a while) but couldn't spot anything obvious.

let me know if I've missed something obvious!

kiich:
Fixes #1303

### DIFF
--- a/core/etcd/config/templates/stack-template.json
+++ b/core/etcd/config/templates/stack-template.json
@@ -222,6 +222,7 @@
     },
     {{end}}
     {{range $etcdIndex, $etcdInstance := .EtcdNodes}}
+    {{if $etcdIndex }},{{end}}
     {{if $etcdInstance.RecordSetManaged}}
     "{{$etcdInstance.RecordSetLogicalName}}" : {
       "Type" : "AWS::Route53::RecordSet",


mumoshu:
LGTM, as I commented in the original issue. Thank you so much for your effort!

Issue #1307, opened 2018-05-16
You can change kubernetes pod and service CIDRs with the kube-aws distribution!


davidmccormick:
Changing an existing kubernetes cluster's pod or service CIDR is not easy to do!

Allowing administrators to change their pod network CIDR or services network CIDR is not something that is well supported by the kubernetes community.  There isn't a way of changing the CIDRs on-the-fly in a live cluster and it is still not straight-forward even if you are willing to accept that the change will require downtime.

And yet there are still inevitably circumstances where you might find that you have no choice but to consider changing your CIDRs: -

- You've run out of room for nodes, if you picked a /16 pod network CIDR then you only have up to a 256 node cluster (because flannel takes the last 8 bits for pod addresses by default).
- Either you find that your pod network or service network clash with duplicate ranges outside your cluster that you need to be able to route to.
- Or you want to run a fancy service mesh across different clusters which can't because both clusters share the same internal ranges.

We have found that we have six clusters that we must change because we have run out of space to grow.  We don't operate multi-clusters and so it is desirable to update them in-place with some downtime (but no loss of app/services when they come back up).

The standard kubernetes response to wanting to change a CIDR is to "build another cluster" but this may not be straight-forward (especially if your cluster is already large with important customers), unless your application deployment tool-set already accommodates a multi-cluster architecture.   This is where I feel that kubernetes distributions like 'kube-aws' can step in and deliver real value for our users and administrators by assisting/facilitating such changes through automation in kube-aws.

I have been able to successfully change both pod and service cidrs in 'kube-aws' clusters by following the procedure below: -

WARNING: The procedure below WILL result in downtime!

1. Update cluster.yaml with new podCIDR, ServiceCIDR (remember to also update dnsServiceIP).
1. Remove and re-generate your apiserver.pem certificate (if you changed your ServiceCIDR).
1. Scale the AWS Autoscaling groups for your controllers and all nodes to 0 (so you are only left with etcd servers)
1. Render stack and roll cluster to bring up controllers and nodes with new CIDRs++

++note: special updated kube-aws version required that handles cluster cidr changes upon start-up required.

Unfortunately, it is not possible to simply stop all controllers and nodes and bring everything back up to change the CIDRs - this will result in some nasty errors that will leave you with a non-functional cluster.  The 'kube-controller-manager' will not start up if it finds objects in etcd with incompatible node, pod or service cidrs:-

```
[...]

PR #1308, opened 2018-05-16
Handle cluster CIDR changes

davidmccormick:
Implementation of a controller based service that is able to handle the changing of either pod, service CIDR or both.

See issue: https://github.com/kubernetes-incubator/kube-aws/issues/1307

- Only performs checks when the cidrs changed.
- Cleans up and recreates bad objects
- Implements STONITH with any previous controllers with incorrect CIDR settings
- Rolls all nodes and controllers when either cidr changes
- Manages a lock to avoid races with other controllers


core/controlplane/config/config.go
```diff
@@ -255,11 +257,6 @@ func NewDefaultCluster() *Cluster {
 		EtcdSettings: EtcdSettings{
 			Etcd: model.NewDefaultEtcd(),
 		},
-		FlannelSettings: FlannelSettings{
```
davidmccormick: I move PodCIDR and ServiceCIDR into KubeClusterSettings so that they can be accessed in the cloud-config on the nodes (so we can make them roll).  I happen to think that they are KubeClusterSettings anyway and there was little to be gained by having the FlannelSettings type at all.

davidmccormick:
I've now used this to migrate 4 clusters to different pod and service CIDRs which enabled us to increase the size that our clusters can scale to from 128 to 4096 nodes.

mumoshu:
@davidmccormick LGTM. Thank you very much for your huge effort!

PR #1309, opened 2018-05-17
Enable Network-Self-Hosting by default

davidmccormick:
The kubernetes flannel and canal daemonsets were introduced in the last release.
Let's make them the default in the next release and then remove the legacy functionality in the following release.


mumoshu:
@davidmccormick Thanks for the contribution as always. Let's try!

PR #1310, opened 2018-05-17
Move templates into their related packages

davidmccormick:
Now that we are separating out the stacks it might make sense to move some of the templates around to belong with the stack that they most closely relate: -

e.g.

cloud-config-etcd could live in the core/etcd/config package
cloud-config-worker could live in the core/nodepool/config package
cluster.yaml could live in core/root/config

What do you think?


mumoshu:
Thanks for rebasing 🎉

PR #1314, opened 2018-05-21
Feature/volume mounts for controllers and etcds

kiich:
PR for https://github.com/kubernetes-incubator/kube-aws/issues/1204
Would be grateful for feedback as I had some questions about the validation part.


core/etcd/config/templates/cloud-config-etcd
```diff
@@ -77,6 +77,33 @@ coreos:
         {{- end }}
       {{- end}}
 {{- end}}
+    - name: systemd-modules-load.service
+      command: restart
```
mumoshu: @kiich Could you clarify the reason to add this?
AFAIK, this was needed only for worker and controller nodes in order to enable ipvs-based kube-proxy in 8d872b2b9534bd0bf9d3b2e5ebf01ac9d4ae58d5
This is etcd's cloud-config so we don't need?

kiich: Hi @mumoshu - sorry, I was just copying the way it was done in the worker cloud-config without realising it fully - I'll look into it and refactor.

PR #1318, opened 2018-05-23
Default controller manager limits

jorge07:
Fixes: #1254


mumoshu:
@jorge07 Thanks for the PR!

Honestly I'm still unsure this is the safe defaults. Ideally there should be a heuristic to calculate necessary request/limit that is proportional to number of resources managed by k8s. But let's start by seeing how it works in practice.

jorge07:
I've this patch in production since 3 weeks. I'll report any issue related with this.

mumoshu:
@jorge07 I greatly appreciate your support!

Issue #1316, opened 2018-05-22
Self hosted canal network doesn't work with kube2iam/kiam


c-knowles:
Currently when self hosted network is enabled with canal, we configure kube2iam/kiam with network interface of `cni0` (since the `UseCalico` flag is ignored):

https://github.com/kubernetes-incubator/kube-aws/blob/master/core/controlplane/config/templates/cloud-config-controller#L4962-L4966
https://github.com/kubernetes-incubator/kube-aws/blob/master/core/controlplane/config/templates/cloud-config-controller#L5356-L5360

However, this doesn't appear to work. It needs to be `cali+` for canal.

Probably we need it before https://github.com/kubernetes-incubator/kube-aws/pull/1309?

davidmccormick:
Hi, we've not seen any issues with self-hosted canal + kiam in our hcom clusters. Let me have a look at one of our clusters and check which interface is being used.

PR #1320, opened 2018-05-23
1316/kiam kube2iam with kube net selfhosting

davidmccormick:
Allow migration from use of UseCalico setting to resolve issue https://github.com/kubernetes-incubator/kube-aws/issues/1316 and allow migration of Kubernetes->Networking->Selfhosting by default.


mumoshu:
@davidmccormick Thanks again for your contribution!

Issue #1175, opened 2018-03-08
Mount a filesystem backed by a Raid0 assembly of EBS volumes 


rckgrail:
I've been a long time user of kube-aws (since v0.8.0) and use it to help process genomic data
sometimes running up to 1000 nodes.

I've been using a single 2T volume mounted on /data that gets used for processing.  There's
also per worker monitoring the remaining free space in /data to grow it automatically if needed
using the AWS ModifyVolume et al APIs along with resize2fs (I use mkfs.ext4 in cloud-config-worker).

It turns out that a Raid0 striped assembly of ~10 214G EBS volumes has far faster throughput
than a single 2T EBS volume.  I build the Raid0 using lvm since it still allows me to dynamically
increase the size of the filesystem by running ModifyVolume on each of the 10 volumes, followed
by pvresize and lvextend.  (mdadm wouldn't allow these subsequent resize operations).

Therefore, I would like to see if there would be any interest in integrating into kube-aws something
similar to VolumeMounts, but for creating these Raid0 / EBS backed higher performance filesystems.

Something like in cluster.yaml  (similar to volumeMounts in nodePools)

      raid0Mounts:
      - type: "gp2"
        iops: 0
        size: 214  # size of each device
        devices:
        - /dev/xvdf
        - /dev/xvdg
        - /dev/xvdh
        - /dev/xvdi
        - /dev/xvdj
        - /dev/xvdk
        - /dev/xvdl
        - /dev/xvdm
        path: "/data"

Under the covers _devices_ would need to be checked against other raid0Mounts and volumeMounts for duplication, same thing for _path_.

Construction of the filesystem would then happen via something like:

    pvcreate /dev/xvdf /dev/xvdg /dev/xvdh /dev/xvdi /dev/xvdj /dev/xvdk /dev/xvdl /dev/xvdm
    vgcreate data_group /dev/xvdf /dev/xvdg /dev/xvdh /dev/xvdi /dev/xvdj /dev/xvdk /dev/xvdl /dev/xvdm
    lvcreate -l 100%VG --stripes 8 --stripesize 256 -n data_vol data_group
    mkfs.ext4 /dev/data_group/data_vol
    mount /dev/data_group/data_vol /data

Please let me know what you think.

rckgrail:
FYI I have an implementation of this ready to go.

PR #1322, opened 2018-05-23
Raid0 support

siddharthab:
Resolves https://github.com/kubernetes-incubator/kube-aws/issues/1175.

### DIFF
--- a/core/nodepool/config/templates/cloud-config-worker
+++ b/core/nodepool/config/templates/cloud-config-worker
@@ -95,7 +95,7 @@ coreos:
 
         [Service]
         Type=oneshot
-        ExecStart=-/usr/sbin/mkfs.xfs {{$volumeMountSpec.Device}}
+        ExecStart=-/usr/sbin/mkfs.{{$volumeMountSpec.FilesystemType}} {{$volumeMountSpec.Device}}
 
         [Install]
         WantedBy=local-fs-pre.target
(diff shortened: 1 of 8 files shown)


rckgrail:
signed CLA

core/nodepool/config/templates/cloud-config-worker
```diff
@@ -1585,5 +1658,33 @@ write_files:
...
+          fi
+        fi
+      done
+
+  - path: /opt/nvme-hack/99-aws-ebs-nvme.rules
+    owner: root:root
+    permissions: 0644
+    content: |
+      # Create the legacy device node for the volume mount and raid0 startup
+      ACTION=="add", SUBSYSTEM=="block", KERNEL=="nvme[1-26]n1", ATTRS{model}=="Amazon Elastic Block Store              ", RUN+="/opt/nvme-hack/ebs-nvme-mapping"
```
mumoshu: Oh! Thank you so much for the extra!
I believe this feature isn't the primary topic of this PR but I appreciate this anyway.
Would you mind clarifying how this implementation differs from https://github.com/coreos/bugs/issues/2399#issuecomment-379730681?
Due to my lack of expertise, I'm not sure which one is the complete solution.

rckgrail: The services seem about the same.

The one I provide tries to map all it can find (via `nvme list`), while the one in [coreos/bugs#2399 (comment)](https://github.com/coreos/bugs/issues/2399#issuecomment-379730681)
looks like it needs to be explicitly invoked for each presumed device to get the mappings.

I've also seen evidence of the `nvme id-ctrl` output both include and exclude a leading "/dev/", so my version
looks for both and maps them as discovered.

I based my effort on [github.com/oogali/ebs-automatic-nvme-mapping](https://github.com/oogali/ebs-automatic-nvme-mapping), which I had been pointed at by our AWS support people.

jotto:
Is there an example somewhere of how to use this?

PR #1323, opened 2018-05-24
Make sure that canal and flannel land on all tainted nodes

davidmccormick:
The canal and flannel daemonsets should land on ALL nodes, including those with taints.
Correction of a syntax bug when retrying kubectl commands in add-node-cidrs.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -1446,6 +1446,16 @@ write_files:
             annotations:
               scheduler.alpha.kubernetes.io/critical-pod: ''
           spec:
+            # Ensure that canal-node only targets nodes and not masters
+            affinity:
+              nodeAffinity:
+                requiredDuringSchedulingIgnoredDuringExecution:
+                  nodeSelectorTerms:
+                  - matchExpressions:
+                    - key: kubernetes.io/role
+                      operator: In
+                      values:
+                        - node
             hostNetwork: true
             serviceAccountName: canal
             tolerations:
@@ -1952,12 +1962,14 @@ write_files:
             nodeSelector:
               beta.kubernetes.io/arch: amd64
             tolerations:
-            - key: node-role.kubernetes.io/master
-              operator: Exists
-              effect: NoSchedule
-            - key: node.alpha.kubernetes.io/role
-              value: "master"
-              effect: NoSchedule
+              # Tolerate this effect so the pods will be schedulable at all times
+              - effect: NoSchedule
+                operator: Exists
+              # Mark the pod as a critical add-on for rescheduling.
+              - key: CriticalAddonsOnly
+                operator: Exists
+              - effect: NoExecute
+                operator: Exists
             serviceAccountName: flannel
             # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
             # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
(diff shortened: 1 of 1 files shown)


davidmccormick:
Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1324

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -1446,6 +1446,16 @@ write_files:
...
+            # Ensure that canal-node only targets nodes and not masters
+            affinity:
+              nodeAffinity:
+                requiredDuringSchedulingIgnoredDuringExecution:
+                  nodeSelectorTerms:
+                  - matchExpressions:
+                    - key: kubernetes.io/role
+                      operator: In
+                      values:
+                        - node
```
mumoshu: Oh, why do we need this? My understanding was that we have to deploy canal-node onto every node including controllers to retain Flannel connectivity cluseter-wide?

davidmccormick: We need to make sure it does not try to deploy to masters which have their own version of the daemonset with different settings (no Typha).

davidmccormick: Hi, I still think that we need this change - without it flannel will not properly be deployed to any node pools that contain a taint.

mumoshu: Ah, sorry for the delay! Maybe I missed that in my previous reviews. So, you chose to not enable typha on controller nodes, perhaps because we shouldn't load controller nodes to serve many calico-related agents, right?

davidmccormick: Hi, yes I don't want the controllers having to proxy their api requests via the typha service and thus create a dependency on having 'Ready' nodepools before the controllers are able to have a valid network and become 'Ready' themselves.  Avoiding one of those chicken and egg type problems! :)

mumoshu: Please let me double-check :) What happens when we enabled typha also on controllers nodes? There should be no chicken-and-egg prob. then.

davidmccormick: The act of enabling typha in kube-aws as I have implemented it is to only ever enable it on **worker nodes**, i.e. only in the daemonset **canal-node**.  The **canal-master** daemonset is always configured to **NOT** use typha regardless of the kube-aws typha option and this ds targets only the controllers. 

Typha is an apiserver cache specific to calico - and I don't want to allow the actual Typha pods to schedule and execute on the apiserver nodes themselves.  Its job is to alleviate load on the apiservers and I want to have as few services on the controllers as possible.  That's why I want Typha to be scheduled on nodes in the worker pools instead.

With that decision to have Typha only scheduled to nodes then I didn't want the masters to also depend upon it because this can cause problems spinning up a new cluster.  This way the masters can become Ready before any workers have started and Typha is running.

davidmccormick: This might also help.  See how the canal-master daemonset is already targeted at controllers:
```
            # Ensure that canal-master only targets our masters.
            affinity:
              nodeAffinity:
                requiredDuringSchedulingIgnoredDuringExecution:
                  nodeSelectorTerms:
                  - matchExpressions:
                    - key: node-role.kubernetes.io/master
                      operator: Exists
```

Now that the taints are being relaxed on the canal-node daemonset then this equivalent makes sure that canal-node only gets scheduled on workers: -

```
            # Ensure that canal-node only targets nodes and not masters
            affinity:
              nodeAffinity:
                requiredDuringSchedulingIgnoredDuringExecution:
                  nodeSelectorTerms:
                  - matchExpressions:
                    - key: kubernetes.io/role
                      operator: In
                      values:
                        - node
```

davidmccormick: The taints were already relaxed for the canal daemonsets prior to this change but the plain flannel one wasn't updated and so won't schedule on any node pools with taints.

mumoshu: @davidmccormick I appreciate your detailed explanation. I now believe that I fully understood your intention paired with the implementation 👍

PR #1326, opened 2018-05-24
added label to kube-system namespace

### LINKED ISSUE #1325: Namespaces need labels
Network Policies only allow to choose namespaces by labels.  Not all namespaces have labels.

PR to follow with labels

jasonrichardsmith:
templates: add label to kube-system ns so you can select in Network Policies

To select a namespace in network policies, you need a label on the namespace.  I have added a label to kube-system
```
name: kube-system
```

Fixes #1325


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -5167,6 +5167,9 @@ write_files:
         name: kube-system
         annotations:
           iam.amazonaws.com/permitted: ".*"
+        labels:
+          name: kube-system
```
mumoshu: Makes sense! But it would be even better if we could find a more generic way. Perhaps `kubeSystemNamespaceLabels:` in cluster.yaml?( So verbose!

WDYT?

davidmccormick:
Can’t you already use the namespace selector?

core/controlplane/config/templates/cloud-config-controller
jasonrichardsmith: @mumoshu :point_up:

mumoshu: Sorry for the delay! I'd prefer a hash containing the whole labels i.e. `kubeSystemNamespaceLabels`.

I'm not yet sure if everyone could agrees on this specific label `name: kube-system`.
I also considered something like `kube-aws/name: kube-system` that looks more specific to kube-aws. But it wouldn't be optimal. There should be one or more namespaces that you would like to add labels as well, and adding `kube-aws/name: foo` to every ns managed outside of kube-aws seems wrong.

Something like `kubeSystemNamespaceLabels` seems to provide enough options for users to consistently label namespaces across the cluster, regardless of it is managed by kube-aws or not.

WDYT?

jasonrichardsmith: Makes complete sense.  thanks for getting back to me.  I will adjust and resubmit.

jasonrichardsmith:
@mumoshu sorry for the late update, this has been updated.

mumoshu:
@jasonrichardsmith Looks great. Thanks for your contribution 👍

PR #1333, opened 2018-05-26
Tweak CloudFormation stack descriptions

c-knowles:
Matches the new network stack.


mumoshu:
Good catch. Thanks for the fix 👍

PR #1334, opened 2018-05-27
Fix flannel traffic among controllers

cheungpat:
It appears flannel traffic is not permissible among controllers. This pull
request fix this by adding a security group ingress which allows 8472/udp
from other controllers.

When kiam support is enabled, kiam-server tries to discover grpclb by querying
DNS. If kube-dns is running on controllers and there are multiple controllers,
the DNS name may not be resolved becuase kiam-server cannot contact kube-dns
running on the other controller.


mumoshu:
@cheungpat Good catch! Thank you very much for the fix.

PR #1335, opened 2018-05-27
Fix install-kube-system

c-knowles:
From https://github.com/kubernetes-incubator/kube-aws/pull/1319/files#diff-ef25536c536667a40b993d4d24ab7567R964, space causes the install-kube-system to fail:

`{kube-dns,kube-dns-autoscaler {{ if .KubernetesDashboard.Enabled }}`

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -955,10 +955,6 @@ write_files:
         "${mfdir}/metrics-server-apisvc.yaml"
       {{- end }}
 
-      {{ if .Experimental.NodeDrainer.Enabled }}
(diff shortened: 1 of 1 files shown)


c-knowles:
@mumoshu not sure if it's better for us to separate based on entity type deployment/service etc or separate based on app autoscaler/node-drainer etc. I went with entity type for now. How would you feel about storing these manifests inside the S3 bucket and then syncing them inside the UserData? We could store each app in a separate prefix, e.g. `s3://kube-aws-assets/manifests/srv/kubernetes/node-drainer/`

c-knowles:
@mumoshu FYI this fix should block any new release since install-kube-system is currently broken on master.

Issue #1209, opened 2018-04-02
doc: Suggest cluster admins to regularly update AMI ID(s) in cluster.yaml


mumoshu:
After #1201. Otherwise the cluster stucks on an old Container Linux release. Formerly, regularly running `kube-aws update` without modifying cluster.yam at all was enuogh.

c-knowles:
It would be good to have a command to update to the latest from the cluster.yaml specified channel.

cmcconnell1:
Hello,

Would like to ask how this will work with versions that hard-code/pin the CoreOS AMI image by default upon cluster.yaml file?

I.e.:  for kube-aws version v0.9.10-rc.3 we have the following 'pinned' in our cluster.yaml file by default:
amiId: "ami-07a6b767"

Would the proposed process (and doc update or commands) involve  removing the above amiID 'pinning?'
And then updates would be made once nodes were restarted?

Thanks

PR #1339, opened 2018-05-28
Instructions for updating the AMI

c-knowles:
Towards https://github.com/kubernetes-incubator/kube-aws/issues/1209.

Basic instructions for now as I think we should wire a command to the existing logic, at least to print out the ID and if not to update the cluster.yaml.

### DIFF
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -11,6 +11,9 @@ s3URI: {{.S3URI}}
 #releaseChannel: stable
 
 # The AMI ID of CoreOS.
+#
+# To update this to the latest AMI run the following command with the appropriate region and channel then place the resulting ID here
+#   REGION=eu-west-1 && CHANNEL=stable && curl -s https://coreos.com/dist/aws/aws-$CHANNEL.json | jq -r ".\"$REGION\".hvm"
 amiId: "{{.AmiId}}"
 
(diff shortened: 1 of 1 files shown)


jorge07:
Currently working in a command for this. The idea is something like:

cmd/ami_update.go
```go
...

func init() {
	RootCmd.AddCommand(cmdAMIUpdate)
}

func runUpdate(_ *cobra.Command, _ []string) error {

	cluster, err := root.ClusterFromFile(configPath, opts, false)
	if err != nil {
		return fmt.Errorf("Failed to read cluster config: %v", err)
	}

	amiID, err := amiregistry.GetAMI(cluster.DeploymentSettings.Region, cluster.DeploymentSettings.ReleaseChannel)
	initOpts.AmiId = amiID
	if err != nil {
		return fmt.Errorf("Cannot retrieve CoreOS AMI for region %s, channel %s", initOpts.Region.Name, defaultReleaseChannel)


	if err := filegen.UpdateClusterYaml(configPath, initOpts, cluster); err != nil {
		return fmt.Errorf("Error updating config template: %v", err)
	}

...
```

mumoshu:
@jorge07 Sounds interesting! Just curious but how does `UpdateClusterYaml` work? And I'm scratching my head how we should name the new command. I'd love `kube-aws update-ami` or alike to update cluster.yaml only, but it offends `kube-aws update` 😢 I can't help wishing we had named `kube-aws update` as `kube-aws apply`.

Issue #1336, opened 2018-05-27
Issue reconciling node routes with self hosted canal


c-knowles:
I've been trying to use self hosted canal enabled by default using version 9c735f9485445699d598df5c07fc73c811f682d2. k8s v1.9.7. It seems now I'm getting these logs from the controller manager although the cluster itself appears to be working still.

`
kube-controller-manager-ip-10-0-22-230.eu-west-1.compute.internal kube-controller-manager E0527 10:17:37.615993       1 route_controller.go:116] Couldn't reconcile node routes: error listing routes: found multiple matching AWS route tables for AWS cluster: kubernetes
`

The only changes made between this and previous cluster are switching self hosting on plus the separate etcd stack that comes with the latest version. It's an entirely new cluster and I've deleted the previous version of the cluster.

We do have another cluster in the same VPC but this has never been a problem before.

I had a look through every route table and subnet, all seem tagged appropriately. I was looking for potentially related issues and found https://github.com/kubernetes/kubernetes/issues/12449#issuecomment-138276913.

Anyone else having similar issues?

c-knowles:
I think I tracked this down. https://github.com/kubernetes-incubator/kube-aws/pull/1233 changed all the `kubernetes.io/cluster/{{$.ClusterName}}` tag values from `owned` to `true` but not sure why that was. @mumoshu?

c-knowles:
Ah sorry perhaps it's not that, some are `true` and some are `owned`.

c-knowles:
Just trying out the `--cluster-name` flag for controller manager, it seems we've never set it in kube-aws.

c-knowles:
@davidmccormick, I wondered if you had any ideas about this or seeing similar in your clusters? It appears the cluster still works which is odd but I'm concerned running another cluster in the same VPC will break them. The cluster name flag above did not fix it. Our other cluster in the same VPC appears to be working fine and has a similar setup. The differences are k8s 1.7.x to 1.9.x and the self hosted network enable. I'm not sure if the `--allocate-node-cidrs` and `--cluster-cidr` flags of controller manager could affect this, I'm still digging into the controller manager code.

davidmccormick:
[...]

PR #1341, opened 2018-05-28
Disable configure of cloud route when allocate-node-cidrs is enabled

c-knowles:
Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1336 by adding `--configure-cloud-routes=false` when `--allocate-node-cidrs=true`

Added `--cluster-name` to ensure it's set correctly on controller manager.

Grouped mandatory and optional flags together.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -3372,17 +3372,20 @@ write_files:
           command:
           - /hyperkube
           - controller-manager
+          {{/* mandatory flags below */}}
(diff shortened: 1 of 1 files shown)


c-knowles:
@davidmccormick for some reason I cannot assign you for review but worth you taking a look at this.

mumoshu:
@c-knowles Thanks for the fix! LGTM.

@ArchiFleKs has also encountered this only with self-hosting enabled. So your choice of adding the flag only for self-hosted configuration looks good.

Issue #1342, opened 2018-05-30
Tag volumes with names


c-knowles:
Most of the EBS volumes created by kube-aws don't have a `Name` tag and so show without a name in the console. It'd be nice to tag them where possible.

kiich:
also just to chime in that the root volumes on the nodes are not tagged neither last time i checked (and i thought i raised an issue about it but i can't seem to find it)

kiich:
https://github.com/kubernetes-incubator/kube-aws/issues/1159 is the issue i raised.

mumoshu:
@kiich @c-knowles Just had some to study about this. TD;DR; EC2 or CFN doesn't support tagging root volumes in a declarative manner 😢 

Refs: https://forums.aws.amazon.com/thread.jspa?threadID=102894
https://forums.aws.amazon.com/thread.jspa?threadID=122354

mumoshu:
A possible workaround would be to tag all the attached EBS volumes with `DeleteOnTermination` enabled, on node startup.

```bash
aws ec2 describe-instances --instance-ids $instance_id | jq -r .Reservations[].Instances[0] > instance.json
cat instance.json | jq -r '.BlockDeviceMappings[] | select(.Ebs.DeleteOnTermination == true) | .Ebs.VolumeId' > volume_ids.txt
for vol_id in $(cat volume_ids.txt); do
  add_tags $vol_id
done

# Do actual tagging for the volume-id passed as the first arg in the `add_tags` func
```

mumoshu:
I believe that the filtering with `DeleteOnTermination` is necessary in order to prevent accidental tagging of volumes not managed by kube-aws.

PR #1343, opened 2018-05-30
Tag etcd volumes with names

c-knowles:
https://github.com/kubernetes-incubator/kube-aws/issues/1342


c-knowles:
Seems like the ones in `AWS::AutoScaling::LaunchConfiguration` don't support tagging out of the box in CloudFormation.

Issue #1075, opened 2017-12-16
Optional switch for CoreDNS with k8s 1.9


c-knowles:
Placeholder based on https://github.com/kubernetes/features/issues/427. We should look to adding the same optional support for CoreDNS.

mumoshu:
Looking forward to try this!

Btw, would there be an equivalent of kube-dns-autoscaler a.k.a [dns-horizontal-autoscaler](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns-horizontal-autoscaler) for coredns?
Can we just start with paring it with kube-dns-autoscaler?

etiennetremel:
The dns-horizontal-autoscaler autoscale based on the number of cores and nodes then update the number of replica of a given deployment name. It works with both kube-dns and coredns.

More info: https://github.com/kubernetes-incubator/cluster-proportional-autoscaler/#implementation-details

mumoshu:
Good to know! Thanks for the info

PR #1345, opened 2018-05-31
Add support for CoreDNS as DNS provider

etiennetremel:
Add support for CoreDNS as DNS provider. Fix https://github.com/kubernetes-incubator/kube-aws/issues/1075

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -182,6 +182,7 @@ func NewDefaultCluster() *Cluster {
 				IPVSMode: ipvsMode,
 			},
 			KubeDns: KubeDns{
+				Provider:            "kube-dns",
 				NodeLocalResolver:   false,
 				DeployToControllers: false,
 				Autoscaler: KubeDnsAutoscaler{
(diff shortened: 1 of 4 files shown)


core/controlplane/config/config.go
```diff
@@ -817,6 +820,7 @@ type KubeDnsAutoscaler struct {
 }
 
 type KubeDns struct {
```
mumoshu: Yeah, this should retain the name for backward compatibility for now.
But how about renaming this to `ClusterLocalDns` in the **future**?
`KubeDns` sounds like literally specific to `kube-dns`, that is just one of cluster local DNS server implementations available to Kubernetes.

c-knowles:
If a user switches provider on an existing cluster, do both deployments run side by side and one has to be manually deleted? We may want to add a note about that.

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -3642,9 +3726,12 @@ write_files:
                   - /cluster-proportional-autoscaler
                   - --namespace=kube-system
                   - --configmap=kube-dns-autoscaler
+                  {{- if eq .KubeDns.Provider "coredns" }}
+                  - --target=Deployment/coredns
+                  {{- else }}
                   - --target=Deployment/kube-dns
+                  {{- end }}
                   - --default-params={"linear":{"coresPerReplica":{{ .KubeDns.Autoscaler.CoresPerReplica }},"nodesPerReplica":{{ .KubeDns.Autoscaler.NodesPerReplica }},"min":{{ .KubeDns.Autoscaler.Min}}}}
-                  - --logtostderr=true
```
c-knowles: Was `logtostderr` intentionally removed here?

mumoshu: No, at least for me! Good catch.

c-knowles:
I was trying to figure out if https://github.com/kubernetes-incubator/kube-aws/pull/792 is still really needed when using CoreDNS. There seems to be some discussion in https://github.com/kubernetes/kubernetes/issues/32749 about that, it might be that we could deploy CoreDNS as a Daemon Set and reduce the component count involved.

mumoshu:
@c-knowles Hey!

I got a chance to revisit this. And it seems like the momentum is to use CoreDNS as a node-local dns caching agent, as you've said.

See the sig-network proposal at https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/0030-nodelocal-dns-cache.md, which is implemented in the upstream at https://github.com/kubernetes/kubernetes/pull/70555

Issue #1252, opened 2018-04-23
Enable EBS-optimized flag on nodes


kiich:
We created a node pool specifically to take advantage of [EBS-optimise](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html) so they are of instance type that has it on by default.

However when we spin up a cluster, the AWS console shows the volume with no optimisation. Looking further into it, it looks like the Launch Configuration that creates the instances do not have the EBS optimised flag enabled and hence the instances created from the LC do not have it enabled neither.

If I just spin the instance via ec2 console, ebs is indeed enabled.

mumoshu:
@kiich Thanks for the good catch! I do remember the last time I have literally seen the flag, but somehow overlooked to consider it for kube-aws...

This should be addresseds .

Probably:

- Adding a helper go func named like `EbsOptimized(instanceType string) bool` to return true when and only when the instance type is greater than `t2`, `m3`, and `c3` series.
- And use the helper func inside stack-template.json for etcd, controller, worker nodes

at minimum would work.

kiich:
Sounds good @mumoshu - thanks for the heads up. I'll look into it for possible PR.

kiich:
@mumoshu regarding series that do not support EBS optimise, it seems some type in certain series that supports EBS optimise, does not - such as `i2.8xlarge` that does not support it but the rest of `i2` series do.

What do you think about catching those? Was there a reason for filtering on `t2`,`m3` and `c3` specifically? As I can see some types of `m3` such as `m3.xlarge` supprots EBS optimise.

PR #1346, opened 2018-05-31
Automatically enable EBS optimization

kiich:
Feature done by looking for instance types that do NOT support EBS optimisation from the AWS page https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html and few extra instances I found from AWS console.

PR for https://github.com/kubernetes-incubator/kube-aws/issues/1252

### DIFF
--- a/core/nodepool/config/templates/stack-template.json
+++ b/core/nodepool/config/templates/stack-template.json
@@ -266,6 +266,9 @@
         {{else}}
         "PlacementTenancy": "{{.Tenancy}}",
         {{end}}
+        {{if EbsOptimized .InstanceType }}
+        "EbsOptimized": "true",
+        {{end}}
         "UserData": {{ .UserDataWorker.Parts.instance.Template }}
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration"
--- a/filereader/texttemplate/texttemplate.go
+++ b/filereader/texttemplate/texttemplate.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
+	"strings"
 	"text/template"
 
 	"github.com/Masterminds/sprig"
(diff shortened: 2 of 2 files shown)


core/nodepool/config/templates/stack-template.json
```diff
@@ -266,6 +266,9 @@
         {{else}}
         "PlacementTenancy": "{{.Tenancy}}",
         {{end}}
+        {{if EbsOptimized .InstanceType }}
```
mumoshu: Just wanted to note that I slightly prefer `.InstanceType.EbsOptimized`, while turning `.InstanceType` to its own type and writing tests for it in go. But that's not a strong opinion.

filereader/texttemplate/texttemplate.go
```diff
@@ -42,6 +43,25 @@ var funcs2 = template.FuncMap{
 		reg := regexp.MustCompile("[^a-z0-9A-Z_.-]")
 		return reg.ReplaceAllString(data, "_")
 	},
+	"EbsOptimized": func(instanceType string) bool {
+		// Amazon instance series that do not support EBS optimize
+		nonSupportedSeries := map[string]bool{"t1": true, "t2": true}
+		// Amazon instance types that do not support EBS optimize
+		nonSupportedTypes := map[string]bool{
+			"c1.medium": true,
+			"c3.large":  true, "c3.8xlarge": true,
```
mumoshu: Interesting way to organize instance types! Good to know gofmt doesn't detect this 😄

core/nodepool/config/templates/stack-template.json
mumoshu: Please take it as I didn't want to make things too complicated and hence my comment at https://github.com/kubernetes-incubator/kube-aws/issues/1252#issuecomment-385591265 😄

mumoshu:
@kiich Edited the PR title and the commit message according to my understanding. I appreciate it if you could review it and drop comments if I'm missing something.

Thanks as always for your contribution 🍻

filereader/texttemplate/texttemplate.go
kiich: Yes I was wondering the same after deciding on the formatting - sure enough `make format` didn't complain about it. I think it looked easier on my eyes at the time.

kiich:
@mumoshu Yep, happy with the new PR title, in fact I think it better suits it so thanks for taking the time to update it.

1 question I had was regarding your comment `.InstanceType.EbsOptimized` - I am happy to update to that so let me know what your thoughts are. Initially I did have it similar to that but then saw how you do the other helper functions so went along with that approach instead.

Issue #1350, opened 2018-06-04
Kiam server health check is hardcoded


kevtaylor:
The latest vault (0.10.1) does not allow port names in the subject alternative name

Previous editions allowed this: "alt_names=kiam-server:443,localhost:443" when creating an pki issue

This has now been stopped and only valid dns names are allowed

Vault will allow localhost overriding which will create:
```
X509v3 Subject Alternative Name:
                DNS:localhost
```

...but the server health checks in the kiam-server manifest specify the port in the server address check and this is validated against the SAN in the cert

```
livenessProbe:
                  exec:
                    command:
                    - /health
                    - --cert=/etc/kiam/tls/server.pem
                    - --key=/etc/kiam/tls/server-key.pem
                    - --ca=/etc/kiam/tls/ca.pem
                    - --server-address=localhost:443
                    - --server-address-refresh=2s
                    - --timeout=5s
```

Need to either allow a configurable server-address or have some other health check

/cc @pingles

kevtaylor:
I'm going to put in a PR to make the server address configurable
/assign @kevtaylor

pingles:
We recently took a PR to address this https://github.com/uswitch/kiam/pull/86. I'm hoping we can pick up a few other changes and have a 3.0 release within a couple of weeks. The change has been integrated so there's a tagged image on Quay with the new code but there's a few other improvements I'd like to see pulled in before we move to do another release.

The TLS change is potentially breaking (depending on how certs are generated) so it is potentially worth talking about how kube-aws does that currently. If you're generating them with Vault and there's no port in the server name then you should be able to work fine.

PR #1354, opened 2018-06-06
Consolidate kiam configuration

kevtaylor:
This is a consolidation of kiam parts into the kiamSupport configuration
It adds the listeners for server and agent as configurable options
The previously KIAMImage type is now underneath kiamSupport, so this is a possible breaking change to existing cluster.yaml files

Now looks like this
```
  kiamSupport:
    enabled: true
    image:
      repo: quay.io/uswitch/kiam
      tag: v2.8
    server: localhost
    agent: kiam-server
```

The original DNS health check/server settings are preserved as defaults
The kiam image is bumped to v2.8

### DIFF
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -1037,12 +1037,6 @@ worker:
 #  tag: 1.1.2
 #  rktPullDocker: false
 
-# kiam image repository to use.
-kiamImage:
-  repo: quay.io/uswitch/kiam
-  tag: v2.7
-  rktPullDocker: false
-
 # kube2iam image repository to use.
 #kube2iamImage:
 #  repo: jtblin/kube2iam
@@ -1453,6 +1447,13 @@ experimental:
   # This is intended to be used in combination with .controller.iam.role.name. See #297 for more information.
   kiamSupport:
     enabled: false
+    image: 
+      repo: quay.io/uswitch/kiam
+      tag: v2.7
+      rktPullDocker: false
+    serverAddresses:
+      serverAddress: localhost:443
+      agentAddress: kiam-server:443
 
   # When enabled this will install the kube2iam daemon set using the repo from `kube2iamImage`
   # It will also grant sts:assumeRole permission to the IAM role for controller nodes.
--- a/test/integration/maincluster_test.go
+++ b/test/integration/maincluster_test.go
@@ -137,7 +137,9 @@ func TestMainClusterConfig(t *testing.T) {
 				Filesystem: "xfs",
 			},
 			KIAMSupport: controlplane_config.KIAMSupport{
-				Enabled: false,
+				Enabled:         false,
+				Image:           model.Image{Repo: "quay.io/uswitch/kiam", Tag: "v2.7", RktPullDocker: false},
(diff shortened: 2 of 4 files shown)


kevtaylor:
Fixes #1350

kevtaylor:
@mumoshu My original commit failed the test build. It was a very cryptic `exit 1` after a `diff` statement. After a bit of thinking I could only put it down to the `go fmt` which turned out to be a correct assumption. Kubernetes has a `verify` stage which checks for that sort of stuff which might make it clearer? What are your thoughts?

core/root/config/templates/cluster.yaml
```diff
@@ -1443,6 +1437,12 @@ experimental:
   # This is intended to be used in combination with .controller.iam.role.name. See #297 for more information.
   kiamSupport:
     enabled: false
+    image: 
+      repo: quay.io/uswitch/kiam
+      tag: v2.8
+      rktPullDocker: false
+    server: localhost:443
+    agent: kiam-server:443
```
c-knowles: Maybe worth clarifying these as something like this?
```yaml
serverAddresses:
  server: localhost:443
  agent: kiam-server:443
```

kevtaylor: Looks pretty reasonable. Will address this later today

mumoshu:
@kevtaylor Are you talking about https://travis-ci.org/kubernetes-incubator/kube-aws/builds/388804226?utm_source=github_status&utm_medium=notification specifically?

In that case, how about extracting format verification to its own make target like `make verify-fmt`? Would it be clearer for you?

kevtaylor:
@mumoshu Yes, that was the error I was referring to. Wasn't immediately obvious what the error condition was that was occurring. A verify-fmt would be a nice option I feel.

pingles:
We've had problems with additional behaviour in the client in 2.8 that waited for an address (fix is here: https://github.com/uswitch/kiam/commit/87e067ef41f4411a9de29c40881b53bf26e7a1e2#diff-56019486e31148cc73e6431f00cf320f). This caused the health check to immediately exit and fail 
 which prevented the server from successfully starting.

For now I'd suggest sticking with 2.7 until we do another release. I'm hoping this will be within the next few weeks but will likely move to 3.0 as there'll be a few significant changes.

kevtaylor:
@pingles Okay - will push a commit downgrading to 2.7

kevtaylor:
@mumoshu I have applied the `serverAddresses:` construction as suggested and downsized to 2.7
/cc @pingles 
/cc @c-knowles

kevtaylor:
I squashed the commits

test/integration/maincluster_test.go
```diff
@@ -1568,14 +1572,22 @@ worker:
 experimental:
   kiamSupport:
     enabled: true
+    image:
+      repo: quay.io/uswitch/kiam
+      tag: v2.6
```
mumoshu: FYI @c-knowles, I believe that it is better to organize the image repo/tag settings like this, so that it is easier to migrate features like this into kube-aws plugins.

mumoshu:
Edited the commit message to include your nice PR description.

PR #1357, opened 2018-06-07
Bring etcd's up in parallel on new cluster / Fail-fast to update 0.10.x clusters with newer kube-aws

davidmccormick:
Checks for the existence of the etcd stack and will bring the etcd servers up in parallel if the stack does not exist, i.e. it is a fresh cluster (or upgrade to nested stacks).  

Behavoir and benefits: -

* Performs lookup of cloudformation stacks before rendering stack templates - cusomises the resource dependencies based the on this result.
* New clusters start up quicker (servers start in parallel)
* Cluster updates still go through etcd one node at-a-time
* Removes the need to concoct a cfn-signal on the early nodes before etcd quorum and healthy cluster is achieved.

### DIFF
--- a/cfnstack/cfnstack.go
+++ b/cfnstack/cfnstack.go
@@ -61,18 +61,18 @@ func StackEventErrMsgs(events []*cloudformation.StackEvent) []string {
 }
 
 func NestedStackExists(cf CFInterrogator, parentStackName, stackName string) (bool, error) {
-	logger.Debugf("Testing whether nested stack '%s' is present in parent stack '%s'", stackName, parentStackName)
+	logger.Debugf("testing whether nested stack '%s' is present in parent stack '%s'", stackName, parentStackName)
 	parentExists, err := StackExists(cf, parentStackName)
 	if err != nil {
 		return false, err
 	}
 	if !parentExists {
-		logger.Debugf("Parent stack '%s' does not exist, so nested stack can not exist either", parentStackName)
+		logger.Debugf("parent stack '%s' does not exist, so nested stack can not exist either", parentStackName)
 		return false, nil
 	}
 
 	req := &cloudformation.ListStackResourcesInput{StackName: &parentStackName}
-	logger.Debugf("Calling AWS cloudformation ListStackResources for stack %s ->", parentStackName)
+	logger.Debugf("calling AWS cloudformation ListStackResources for stack %s ->", parentStackName)
 	out, err := cf.ListStackResources(req)
 	if err != nil {
 		return false, fmt.Errorf("Could not read cf stack %s: %v", parentStackName, err)
@@ -83,21 +83,21 @@ func NestedStackExists(cf CFInterrogator, parentStackName, stackName string) (bo
 	logger.Debugf("<- AWS responded with %d stack resources", len(out.StackResourceSummaries))
 	for _, resource := range out.StackResourceSummaries {
 		if *resource.LogicalResourceId == stackName {
-			logger.Debugf("Match! resource id '%s' exists", stackName)
+			logger.Debugf("match! resource id '%s' exists", stackName)
 			return true, nil
 		}
 	}
-	logger.Debugf("No match! resource id '%s' does not exist", stackName)
+	logger.Debugf("no match! resource id '%s' does not exist", stackName)
 	return false, nil
 }
 
 func StackExists(cf CFInterrogator, stackName string) (bool, error) {
-	logger.Debugf("Testing whether cf stack %s exits", stackName)
+	logger.Debugf("testing whether cf stack %s exits", stackName)
 	req := &cloudformation.ListStacksInput{}
-	logger.Debug("Calling AWS cloudformation ListStacks ->")
+	logger.Debug("calling AWS cloudformation ListStacks ->")
 	stacks, err := cf.ListStacks(req)
 	if err != nil {
-		return false, fmt.Errorf("Could not list cloudformation stacks: %v", err)
+		return false, fmt.Errorf("could not list cloudformation stacks: %v", err)
 	}
 	if stacks == nil {
 		return false, nil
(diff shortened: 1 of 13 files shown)


cfnstack/cfnstack.go
```diff
@@ -50,3 +59,61 @@ func StackEventErrMsgs(events []*cloudformation.StackEvent) []string {
...
+		if *resource.LogicalResourceId == stackName {
+			logger.Debugf("Match! resource id '%s' exists", stackName)
+			return true, nil
+		}
+	}
+	logger.Debugf("No match! resource id '%s' does not exist", stackName)
+	return false, nil
+}
+
+func StackExists(cf CFInterrogator, stackName string) (bool, error) {
```
mumoshu: Would you mind making this a private func `stackExists` in case this isn't called outside of the pkg?

cfnstack/cfnstack.go
```diff
@@ -50,3 +59,61 @@ func StackEventErrMsgs(events []*cloudformation.StackEvent) []string {
...
+	if err != nil {
+		return false, fmt.Errorf("Could not list cloudformation stacks: %v", err)
+	}
+	if stacks == nil {
+		return false, nil
+	}
+	logger.Debugf("<- AWS Responded with %d stacks", len(stacks.StackSummaries))
+	for _, summary := range stacks.StackSummaries {
+		if *summary.StackName == stackName {
+			logger.Debugf("Found matching stack %s: %+v", *summary.StackName, *summary)
```
mumoshu: Could you make all the log messages with their first letters in lower case?

core/etcd/config/templates/stack-template.json
```diff
@@ -451,7 +451,7 @@
       },
       "DependsOn": [
         {{if $etcdInstance.DependencyExists}}{{$etcdInstance.DependencyRef}},{{end}}
-        {{if $etcdIndex}}"{{$.Etcd.LogicalName}}{{sub $etcdIndex 1}}",{{end}}
+        {{if $.Etcd.StackExists}}{{if $etcdIndex}}"{{$.Etcd.LogicalName}}{{sub $etcdIndex 1}}",{{end}}{{end}}
```
mumoshu: Can you make it just `{{ if .IsUpdatingCluster }}` and pass a boolean value `true` for `kube-aws update`, and `false` for `kube-aws up`? It would simplify the implementation.

cfnstack/cfnstack.go
davidmccormick: Ah yes, I see what you mean - yes it isn't used now but it feels like the sort of utility function that will be used in itself very soon, e.g. we could use it to implement a kube-aws apply command (that could replace separate up and update)?

mumoshu: > utility function that will be used in itself very soon, e.g. we could use it to implement a kube-aws apply command

Yeah, that makes sense! I'd love `kube-aws apply` after `kubectl apply` and `terraform apply`.

However, even in that case what we'd need is `func (c *Provisioner) createOrUpdateStackWithTemplateURL(...)` which calls `stackExists` in order to decide it is a create or an update, so we'd better make it private within the `cfnstack` pkg? 🤔

core/etcd/config/templates/stack-template.json
davidmccormick: Yes, I was tempted to re-factor this in a similar way but in the end resisted for the following reasons: -

* We can build upon this implementation in order to build to proposed fail-fast when someone tries to update a legacy style implementation to the separate stacks implementation which will break their cluster.  The other simpler way does not allow this.
* By implementing the utility StackExists/NestedStackExists functions we can make use of these to implement other useful UI simplifications, such as creating a kube-aws 'apply' command that handles cfn create and update for us and would make implementing kube-aws into pipelines easier.

What do you think?  Is it worth sacrificing the simplicity now for these other features?  I think I could PR the fail-fast in the next day or two and the 'apply' command wouldn't take much longer?

mumoshu: Context: https://github.com/kubernetes-incubator/kube-aws/pull/1357#discussion_r194323349

mumoshu: > this implementation in order to build to proposed fail-fast when someone tries to update a legacy style implementation to the separate stacks implementation which will break their cluster

Ah, makes sense!

> Is it worth sacrificing the simplicity now for these other features?

Definitely! It turned out that this isn't sacrificing anything, given two use-cases of the `kube-aws apply` and the fail-fast feature as context.

Thank you so much for the clarifications 😄

cfnstack/cfnstack.go
davidmccormick: Please see my latest commit with a proposed fail-fast on update when the root stack or nested etcd stack doesn't exist.  The root stack does not exist is best served with the StackExists function as we want to deal with this more generic case independently before checking the missing nested stack case.

test/integration/maincluster_test.go
```diff
@@ -3475,7 +3475,7 @@ worker:
 			configBytes := validCase.configYaml
 			// TODO Allow including plugins in test data?
 			plugins := []*pluginmodel.Plugin{}
-			providedConfig, err := config.ConfigFromBytesWithEncryptService([]byte(configBytes), plugins, helper.DummyEncryptService{})
+			providedConfig, err := config.ConfigFromBytesWithMocks([]byte(configBytes), plugins, helper.DummyEncryptService{}, helper.DummyCFInterrogator{})
```
mumoshu: Nit but it sounds more like "stubs" rather than `mocks`. In my understanding stubs are these simpler alternative objects that are used to replace the complex aws client in order to test the `providedConfig` in isolation. Mocks on the other hand would be used to verify how methods are called, which doesn't seem to be what we're doing here.

core/root/cluster.go
```diff
@@ -471,6 +471,29 @@ func (c clusterImpl) tags() map[string]string {
...
+	exists, err = cfnstack.NestedStackExists(cfSvc, c.controlPlane.ClusterName, naming.FromStackToCfnResource(c.etcd.Etcd.LogicalName()))
+	if err != nil {
+		logger.Errorf("please check your AWS Credentials/Permissions")
+		return "", fmt.Errorf("can't lookup AWS CloudFormation stacks")
+	}
+	// fail fast if it looks like we are trying to update a legacy cluster.
+	if !exists {
+		logger.Errorf("the %s stack must exist in order to be able to update your cluster.", naming.FromStackToCfnResource(c.etcd.Etcd.LogicalName()))
+		logger.Error("we're sorry, but kube-aws can not presently upgrade your cluster to the new release with a separate ETCD Cloudformation stack.")
+		logger.Error("please update using the previous 0.9.x branch or consider backing up, destroying and recreating to upgrade.")
```
mumoshu: `0.10.x` instead of `0.9.x`? But anyway, I'd suggest to say this without mentioning a specific kube-aws ver. like `use a previous version of kube-aws in order to update, or consider creating an another cluster for migration.`

mumoshu:
Modified the PR title a bit to include the fail-fast feature. Thx 👍

mumoshu:
I've added some nits regarding logs and names, but this LGTM overall! I'll merge this shortly. Feel free to add commits for trivial edits, fixes, and etc. if find something.

Issue #1315, opened 2018-05-22
Duplicate controller role permissions for CloudWatch logging


c-knowles:
Small tidy up we could do, we seem to have the permissions for `CloudWatchLogging` inside `IAMManagedPolicyController` twice:
https://github.com/kubernetes-incubator/kube-aws/blob/6c509be4228791ccb3b19083277b2530643b6055/core/controlplane/config/templates/stack-template.json#L198
https://github.com/kubernetes-incubator/kube-aws/blob/6c509be4228791ccb3b19083277b2530643b6055/core/controlplane/config/templates/stack-template.json#L220

As far as I can tell it was like this from the start of this feature so no worries about a bad merge somewhere. Ref https://github.com/kubernetes-incubator/kube-aws/pull/707.

A-Hilaly:
Hello ! Can work on this ? @c-knowles

c-knowles:
@A-Hilaly sure, no rush on this one as it doesn’t block a rollout. Should be an easy fix.

mumoshu:
Closing as resolved via #1360. Thanks for your supports.

PR #1360, opened 2018-06-12
Fix duplicate controller role permissions for CloudWatch logging

A-Hilaly:
Fixing #1315


core/controlplane/config/templates/stack-template.json
```diff
@@ -207,7 +207,8 @@
                     { "Ref": "CloudWatchLogGroupARN" },
                     { "Fn::Join" : [ "", [{ "Ref": "CloudWatchLogGroupARN" }, ":log-stream:*"]] }
                   ]
-                },{{ end }}
+                },
+                {{ end }}
```
A-Hilaly: Just added a new line after the json object since it is the standard in this file

PR #1361, opened 2018-06-12
Implement 'apply' command and deprecate 'up' and 'update'

davidmccormick:
Make integration with CI/CD automation tools simpler by having a single 'apply' command that handles the correct cloud formation create or update depending on whether the cluster stacks exists or not.

This change adds an apply command and puts a deprecation message on 'up' and 'update' which then proceed to do an 'apply'.  This is to allow a period of backwards compatibility for people that have existing automations around 'up' and 'update' so they can have a notice period to change over to apply.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -75,7 +75,7 @@ func runCmdApply(_ *cobra.Command, _ []string) error {
 	}
 
 	successMsg :=
-		`Success! Your AWS resources are being applied:
+		`Success! Your AWS resources are being provisioned:
 %s
 `
 	logger.Infof(successMsg, info)
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -1,0 +1,92 @@
+package cmd
+
+import (
+	"fmt"
+
+	"bufio"
+	"os"
+	"strings"
+
+	"github.com/kubernetes-incubator/kube-aws/core/root"
+	"github.com/kubernetes-incubator/kube-aws/logger"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdApply = &cobra.Command{
+		Use:          "apply",
+		Short:        "Create or Update your cluster",
+		Long:         ``,
+		RunE:         runCmdApply,
+		SilenceUsage: true,
+	}
+
+	applyOpts = struct {
+		awsDebug, prettyPrint, skipWait, export bool
+		force                                   bool
+		targets                                 []string
+	}{}
+)
+
+func init() {
+	RootCmd.AddCommand(cmdApply)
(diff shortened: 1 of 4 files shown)


cmd/apply.go
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("error updating cluster: %v", err)
+	}
+
+	info, err := cluster.Info()
+	if err != nil {
+		return fmt.Errorf("failed fetching cluster info: %v", err)
+	}
+
+	successMsg :=
+		`Success! Your AWS resources are being applied:
```
mumoshu: How about `being provisioned`?

core/root/cluster.go
```diff
@@ -468,23 +466,34 @@ func (c clusterImpl) tags() map[string]string {
...
-		logger.Errorf("you can only 'update' clusters with a matching cloudformation stack")
-		return "", fmt.Errorf("missing cluster root stack %s", c.controlPlane.ClusterName)
+
+	if exists {
+		report, err := c.update(cfSvc, targets)
+		if err != nil {
+			return fmt.Errorf("error updating cluster: %v", err)
+		}
+		if report != "" {
+			logger.Infof("Update stack: %s\n", report)
```
mumoshu: How about `stack update started: %s`, to make it even more clear that the update isn't completed yet

mumoshu:
Awesome! I especially appreciate your effort to keep backward-compatibility.

core/root/cluster.go
davidmccormick: Hasn't update already run at this point and produced a report?

mumoshu: Oops, you are correct! Then it should be `stack updated` or maybe `stack update completed` or alike. Not yet sure what option would be most consistent with all the other log messages.

mumoshu:
After thinking twice, I'm still fine with deprecating `update` because it isn't very clear what kube-aws updates(stacks? configs?).

But how about `up`? Like `kubectl create`, I wonder if certain number of people wants a creation command that fail with no-op when there's already an existing cluster.

cc/ @c-knowles

davidmccormick:
Personally, I feel that both 'up' and 'update' should be depreciated and removed in the 0.12 release.  Then people know that a single command ('apply') will execute their changes whether they create new or update (and it works better with automation).

mumoshu:
@davidmccormick Thanks for the feedback! I can't agree more about the thought that `apply` fits better for CI&CD use-case :) It is much like I use `kubectl create` over `kubectl apply` only when I'm deploying something by hand.

mumoshu:
Thanks again for your contribution 🍻

PR #1367, opened 2018-06-15
Remove node dependency on ETCD Stack when Networking.SelfHosting is Enabled.

davidmccormick:
Clean up cross-stack dependency that is not needed when use kubernetes as backing store for flannel/calico.


mumoshu:
Ref #1370

PR #1374, opened 2018-06-20
Correct semantic version checks

### LINKED ISSUE #1373: K8sVer check is invalid for kube-proxy
https://github.com/kubernetes-incubator/kube-aws/blob/master/core/controlplane/config/templates/cloud-config-controller#L3211

```
{{if ge .K8sVer "v1.9"}}kubeproxy.config.k8s.io{{else}}componentconfig{{end}}/v1alpha1
```

The syntax here does not work for 1.10 because of the string validation

This seems to work https://github.com/mcuadros/go-version 

PR proposal on its way

kevtaylor:
Fixes #1373

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -139,10 +139,7 @@
 
 [[projects]]
   name = "github.com/stretchr/testify"
-  packages = [
-    "assert",
-    "require"
-  ]
+  packages = ["assert"]
   revision = "12b6f73e6084dad08a7c6e575284b177ecafbc71"
   version = "v1.2.1"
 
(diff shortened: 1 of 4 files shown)


c-knowles:
@kevtaylor there’s already the masterminds “semver” in use in the code checking for the reconciler, could we reuse the same to ensure dependencies are minimised?

kevtaylor:
@c-knowles I'll take a look. I didn't appreciate something was already in place.

kevtaylor:
@c-knowles @mumoshu Used semver instead

mumoshu:
@kevtaylor Thanks for the contribution! Can you also revert `Gopkg.toml` and `Gopkg.lock`? My understanding was that you don't need extra dependencies for this fix.

kevtaylor:
@mumoshu Apologies yes indeed. Submitted with reverted files

kevtaylor:
@mumoshu Seems like something occurred in the e2e tests? How do we force a retest?

PR #1379, opened 2018-06-29
0.10.x migration preparation release

davidmccormick:
This update to the 0.10.x branch needs to be run in order to allow an in place upgrade from kube-aws 0.10.x to the new 0.11.x release with separate 'network' and 'etcd' stacks.

The following changes have been made:
* When Kubernetes.Networking.SelfHosting is enabled the references to outputs from the control plane stack for etcd ip addresses are removed (they are not needed)
* A Warning is printed when you perform an upgrade without SelfHosting enabled that you won't be able to update to future releases.
* I lookup the CIDRs for the subnets containing the etcd servers and relax the incoming rule to allow connections to the etcds on port 2379 from any servers in the same subnets.  This is needed because the etcds in the new stack need to be able to connect across to the existing etcds.  Once migrated, the new migrated etcd have the same tight rules as before.
* I allow etcds to listen on all of their interfaces - which is needed so I can find them and connect to them directly without faffing around trying to work out if they have an attached ENI or not.  Again, this behaviour is only adopted for the migration and new migrated etcds will have more restrictive networking.

These changes combined break the cloudformation dependencies that prevent migration with the 0.11.x release and give access to the existing etcds so that we can migrate their contents.  A PR for the 11.x base will add the features it requires to manage the upgrade.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-etcd
+++ b/core/controlplane/config/templates/cloud-config-etcd
@@ -709,7 +709,7 @@ write_files:
 
       ETCD_INITIAL_CLUSTER_STATE=new
       ETCD_DATA_DIR=/var/lib/etcd2
-      ETCD_LISTEN_CLIENT_URLS=https://$private_ip:2379
+      ETCD_LISTEN_CLIENT_URLS=https://0.0.0.0:2379
       ETCD_ADVERTISE_CLIENT_URLS=https://$advertised_hostname:2379
       ETCD_LISTEN_PEER_URLS=https://$private_ip:2380
       ETCD_INITIAL_ADVERTISE_PEER_URLS=https://$advertised_hostname:2380" >> /var/run/coreos/etcd-environment
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1593,6 +1593,14 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ range $_, $etcdSubnet := .Etcd.Subnets -}}
+          {
+            "CidrIp": "{{ $etcdSubnet.InstanceCIDR }}",
+            "FromPort": 2379,
+            "IpProtocol": "tcp",
+            "ToPort": 2379
+          },
+          {{ end -}}
           {
             "CidrIp": "0.0.0.0/0",
             "FromPort": 3,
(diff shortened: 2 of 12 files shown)


core/controlplane/config/templates/cloud-config-etcd
```diff
@@ -709,7 +709,7 @@ write_files:
 
       ETCD_INITIAL_CLUSTER_STATE=new
       ETCD_DATA_DIR=/var/lib/etcd2
-      ETCD_LISTEN_CLIENT_URLS=https://$private_ip:2379
+      ETCD_LISTEN_CLIENT_URLS=https://0.0.0.0:2379
```
davidmccormick: This is required in order for migration process (etcdadm) running on the new etcd servers to connect to the existing cluster and export all kubernetes keys.  The migration works around this export and import between the etcd clusters.

core/controlplane/config/templates/stack-template.json
```diff
@@ -1593,6 +1593,14 @@
...
           },
           {{end -}}
+          {{ range $_, $etcdSubnet := .Etcd.Subnets -}}
+          {
+            "CidrIp": "{{ $etcdSubnet.InstanceCIDR }}",
+            "FromPort": 2379,
+            "IpProtocol": "tcp",
+            "ToPort": 2379
+          },
+          {{ end -}}
```
davidmccormick: This relaxation of the inbound rules to etcd clients allows ALL hosts within the same subnets as the etcd servers to connect to them.  In practice this should not be an issue because they are protected by TLS client/server certificates.  This is intended as a temporary step in order to faciliate the upgrade/migration

davidmccormick:
Fix for issue https://github.com/kubernetes-incubator/kube-aws/issues/1112

core/controlplane/config/templates/stack-template.json
mumoshu: Makes sense. Migrating to v0.11 results in a brand-new set of `SecurityGroupEtcd` + `SecurityGroupEtcd*Ingress`(that allows 2379 only within either the old or the new etcd cluster, but not between the two), and that's why we need this relaxation.

core/controlplane/config/templates/cloud-config-etcd
mumoshu: Awesome trick!

PR #1380, opened 2018-06-29
0.11.x migration from existing clusters without losing state

davidmccormick:
This PR is for the master 0.11.x candidate branch and is intended to allow a smoother migration for existing users with 0.10.x clusters which do not have the new separate etcd stack.  When first testing the 0.11.x code I found that the process would always fail and roll-back due to cloud-formation dependencies, but once these were cleaned up and worked around then the new etcd cluster would come up empty - effectively wiping the state of the existing cluster in the process of the upgrade.

TL/DR: Upgrades from legacy etcds by importing a copy of all their keys and then allows them to be destroyed.  ALSO you are expected to have updated to the 0.10.x migration release first otherwise the migration will fail because cloud-formation dependencies or your new etcd servers won't be able to connect to the old ones!

The approach for the upgrade is this: -
1.  Examine existing etcd cluster state to determine if we are upgrading from a legacy cluster.
2. Refuse to update unless Kubernetes.Networking.SelfHosting is enabled.
3. If migration required, lookup and construct and etcdadm connect string for the existing etcd servers by querying aws api.
4. Render the etcd cloud-formation with additonal etcd state information in the template context.
5. During a migration extra export and import systemd units are created on the new etcds.
6. A new etcd specific StackConfig has been created which contains the additional etcd state information so I can use knowledge of the existing state when rendering the etcd cfn stack json.
7. The Etcd stack creates an extra /var/run/coreos/etcdadm-environment-migration file with details for connecting to the existing etcd cluster.
8 etcdadm has been enhanced to provide extra 'cluster-is-healthy', 'member-is-leader', 'migration-export-kube-state' and 'migration-import-kube-state' commands.
9.  The etcd cluster leader will export all of the kubernetes objects under prefix '/registry', the others will write an empty file.
10.  On each of the new etcd servers if there is an export file that is not empty they will import it into the new cluster.  The others just pass through with success so that they trigger their cfn-signal but the leader will only trigger cfn-signal if the import is successful.  This means if the import fails for some reason the cluster can roll back to the previous 0.10.x version before any of the old etcds are destroyed.
11.  Node/everything loses connection to the legacy api servers during the Network phase of the update - this is an unintended yet extremely helpful side-effect of the stack change which effectivly disables the existing apiservers during the update.
12.  Once the new apiservers come up they can see and properly interact with the existing nodepools.
13. The Control plane stack will clean up all existing etcd servers and resources once everything else has been successful.

The bulk of changes are in using knowlege of existing state when templating assets, such as cloud-configs and stack templates.  I wasn't happy putting the state in the config package because these are not settings that a user can select, so I ended up working things out at the cluster package level and then looking for ways to included my extra state information into the templating contexts.  I think that perhaps some more thought and re-factoring could be applied to better model the role of config and  existing state when bringing up a cluster.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -158,6 +158,57 @@
         }
       }
     },
+    "SecurityGroupWorker": {
+      "Properties": {
+        "GroupDescription": {
+          "Ref": "AWS::StackName"
+        },
+        "SecurityGroupEgress": [
+          {
+            "CidrIp": "0.0.0.0/0",
+            "FromPort": -1,
+            "IpProtocol": "icmp",
+            "ToPort": -1
+          },
+          {
+            "CidrIp": "0.0.0.0/0",
+            "FromPort": 0,
+            "IpProtocol": "tcp",
+            "ToPort": 65535
+          },
+          {
+            "CidrIp": "0.0.0.0/0",
+            "FromPort": 0,
+            "IpProtocol": "udp",
+            "ToPort": 65535
+          }
+        ],
+        "SecurityGroupIngress": [
+          {{ range $_, $r := $.SSHAccessAllowedSourceCIDRs -}}
+          {
+            "CidrIp": "{{$r}}",
+            "FromPort": 22,
+            "IpProtocol": "tcp",
+            "ToPort": 22
+          },
+          {{end -}}
+          {
+            "CidrIp": "0.0.0.0/0",
+            "FromPort": -1,
+            "IpProtocol": "icmp",
+            "ToPort": -1
+          }
+        ],
+        "Tags": [
+          {
+            "Key": "Name",
+            "Value": "{{$.ClusterName}}-sg-worker"
+          }
+        ],
+        "VpcId": {{.VPCRef}}
+      },
+      "Type": "AWS::EC2::SecurityGroup"
+    },
     {{ if not .Controller.IAMConfig.InstanceProfile.Arn }}
     "IAMInstanceProfileController": {
       "Properties": {
@@ -617,7 +668,6 @@
     {{quote $n}}: {{toJSON $r}}
     {{end}}
   },
-
   "Outputs": {
     {{ if not .Controller.IAMConfig.InstanceProfile.Arn }}
     "ControllerIAMRoleArn": {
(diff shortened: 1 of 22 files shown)


core/controlplane/config/templates/stack-template.json
```diff
@@ -158,6 +158,57 @@
         }
       }
     },
+    "SecurityGroupWorker": {
```
davidmccormick: This security group needed returning to the control plane.  We can remove it again in later releases but without it in the updated control plane stack we can an error about it being in use by the nodepools

core/etcd/config/stack_config.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// StackConfig contains configuration parameters available when rendering CFN stack template from golang text templates
+type StackConfig struct {
+	*controlplaneconfig.Config
+	StackName string
+	controlplaneconfig.StackTemplateOptions
+	UserDataEtcd      model.UserData
+	ExtraCfnResources map[string]interface{}
+	model.EtcdExistingState
```
davidmccormick: This and func NewStackConfig are the only changes from the controlplane version.

core/etcd/config/templates/cloud-config-etcd
```diff
@@ -207,6 +207,63 @@ coreos:
         [Install]
         WantedBy=multi-user.target
 {{end}}
+    {{ if .EtcdMigrationEnabled -}}
```
davidmccormick: The migration units, path unit to trigger the import

core/root/cluster.go
```diff
@@ -504,17 +504,18 @@ func (c clusterImpl) LegacyUpdate(targets OperationTargets) (string, error) {
 
 func (c clusterImpl) update(cfSvc *cloudformation.CloudFormation, targets OperationTargets) (string, error) {
 
+	// Look at existing state of cloud formation and stacks to determine if we need to take special measures in migrating our etcd
+	// clusters from the control plane stack to their own Etcd stack.
 	exists, err := cfnstack.NestedStackExists(cfSvc, c.controlPlane.ClusterName, naming.FromStackToCfnResource(c.etcd.Etcd.LogicalName()))
 	if err != nil {
 		logger.Errorf("please check your AWS credentials/permissions")
 		return "", fmt.Errorf("can't lookup AWS CloudFormation stacks: %s", err)
 	}
```
davidmccormick: Only fail-fast if we don't have SelfHosting enabled - otherwise take this as our cue to migrate instead!

davidmccormick:
Fixes issue https://github.com/kubernetes-incubator/kube-aws/issues/1112

core/controlplane/config/templates/stack-template.json
mumoshu: Makes sense. Fine with reviving this then.

// Wish we had something like reference counting to keep AWS resources only while they're used 😆

core/controlplane/config/templates/stack-template.json
```diff
@@ -626,6 +676,11 @@
       "Export": { "Name": { "Fn::Sub": "${AWS::StackName}-ControllerIAMRoleArn" } }
     },
     {{end}}
+    "WorkerSecurityGroup" : {
+      "Description" : "The security group assigned to worker nodes",
+      "Value" :  { "Ref" : "SecurityGroupWorker" },
+      "Export" : { "Name" : {"Fn::Sub": "${AWS::StackName}-WorkerSecurityGroup" }}
+    },
```
mumoshu: This corresponds to https://github.com/kubernetes-incubator/kube-aws/pull/1380/files#r199216521

PR #1383, opened 2018-07-02
getting started: fix typo in bucket creation

kylegalbraith:
the region specified should be `us-east-1` to match the description above it.


kylegalbraith:
Signed CLA.

mumoshu:
@kylegalbraith Good catch! Thanks for your contribution.

PR #1387, opened 2018-07-06
Remove dependency on alpine:lastest docker image

davidmccormick:
use already present hyperkube image instead


kiich:
Wish I read this earlier!! :)

PR #1388, opened 2018-07-06
Fix colour causing extra newlines

davidmccormick:
This was bugging me, especially on the streamed logs from cloudwatch.  By adding the colour off character after the output string (which included the newline) the logging library Output method will add another new line: -

e.g.
logger.Infof("My log message contains a newline\n") -> "CyanOn" + "My log message contains a newline\n" + DefaultColour + "\n"

Colourize the log line and add the colour off character before any trailing whitespace/newlines.  Now the logs are in colour without the double line spacing!

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"regexp"
 )
 
 const calldepth = 3
@@ -88,7 +89,7 @@ func Debugf(format string, v ...interface{}) {
 func Log(l *log.Logger, color string, v ...interface{}) {
 	msg := fmt.Sprint(v...)
 	if Color {
-		msg = color + msg + ColorNC
+		msg = colorizeMessage(color, msg)
 	}
 	l.Output(calldepth, msg)
 }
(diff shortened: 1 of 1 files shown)


logger/logger.go
```diff
@@ -88,15 +89,23 @@ func Debugf(format string, v ...interface{}) {
...
 	msg := fmt.Sprintf(format, v...)
 	if Color {
-		msg = color + msg + ColorNC
+		msg = colorizeMessage(color, msg)
 	}
 	l.Output(calldepth, msg)
 }
+
+func colorizeMessage(color, s string) string {
+	whitespace := regexp.MustCompile(`\s*$`)
```
kylehodgetts: I wouldn't use `regexp.MustCompile` as it will panic, if for example, the regex is changed and is wrong. 

Must better to use `whitespace err := regexp.Compile(`\s*$`)` and handle the error

davidmccormick: No, I specifically want to ensure that the regex is valid so panicing should a developer break it needs to be flagged up powerfully.  If the regex was being supplied by a user I would consider using regex.Compile though. :)

kylehodgetts: I see, in that case then. it's all good

kylehodgetts: I see, in that case then. it's all good

PR #1389, opened 2018-07-06
Added in KIAM session timeout modifier

### LINKED ISSUE #1386: Kiam hardcodes session timeout
Would like the ability to dynamically alter the session duration

We note that the java AWS SDK seems to have a 15 minute fixed credential check to test whether to cache the credentials

If the session duration in kiam is set to the same, then the credentials are never cached causing an excessive overhead on calls (particularly in high transactions)

Wish to override the default by adding a configurable parameter to the kiam section so can use
`- --session-duration=30m`

Leaving the default as 15

Matei207:
Fixes #1386


Matei207:
Verified

PR #1390, opened 2018-07-06
0.11.x Extend CustomFiles to be able to render go templates

davidmccormick:
We want to remove our customisations to kube-aws from our forked branch and would like to move some of those customisations from the kube-aws code into our cluster.yaml files.  Some of these customisations need to lookup cluster configuration information and so need to be templates rather than simple strings.

This change extends to the CustomFile type to add a template field and adds a new Render function that will automatically render the template if it is present or render our the simple content as before (plus some tests of these operations).

### DIFF
--- a/model/custom_file.go
+++ b/model/custom_file.go
@@ -2,14 +2,18 @@ package model
 
 import (
 	"fmt"
+	"strings"
+	"text/template"
 
 	"github.com/kubernetes-incubator/kube-aws/gzipcompressor"
+	"github.com/kubernetes-incubator/kube-aws/logger"
 )
 
 type CustomFile struct {
 	Path        string `yaml:"path"`
 	Permissions uint   `yaml:"permissions"`
-	Content     string `yaml:"content"`
+	Content     string `yaml:"content,omitempty"`
+	Template    string `yaml:"template,omitempty"`
 	UnknownKeys `yaml:",inline"`
 }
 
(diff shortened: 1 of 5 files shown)


model/custom_file.go
```diff
@@ -2,14 +2,18 @@ package model
 
 import (
 	"fmt"
+	"strings"
+	"text/template"
 
 	"github.com/kubernetes-incubator/kube-aws/gzipcompressor"
+	"stash.hcom/run/kat/pkg/logger"
```
mumoshu: Probably you mistakenly introduced an another logger that is not in the project deps?

model/custom_file.go
```diff
@@ -25,3 +29,37 @@ func (c CustomFile) GzippedBase64Content() string {
 	}
 	return out
 }
+
+func (c CustomFile) RenderContentWithEncoding(gzb64 bool, ctx interface{}) (string, error) {
```
mumoshu: I prefer making this `RenderGzippedBase64Content(ctx interface{})` because it tells about the specific encoding of gzip+base64 than `RenderContentWithEncoding(true, ...)`. The `gzb64 bool` can be omitted because we use it from nowhere right now?

davidmccormick:
That makes sense, please see this amended version with separate RenderGzippedBase64Content function as suggested (keeping RenderContent for use where encoding is not desired).

model/custom_file.go
davidmccormick: That makes sense, please see this amended version with separate RenderGzippedBase64Content function as suggested (keeping RenderContent for use where encoding is not desired).

PR #1394, opened 2018-07-08
Bump Kubernetes to v1.10.5

### LINKED ISSUE #1375: Upgrade kubernetes to 1.10
As 1.11 is upon us, it seems sensible to raise the versions to latest kube 1.10.4

mumoshu:
Incorporate various fixes and improvements made since 1.10.4 as noted in https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#changelog-since-v1104

Resolves #1375
Relates to #1376


core/controlplane/config/config_test.go
```diff
@@ -1503,10 +1503,10 @@ func TestInvalidKubernetesVersion(t *testing.T) {
 kubernetesVersion: v1.x.3
 `,
 		`
-kubernetesVersion: v1.10.4yes
+kubernetesVersion: v1.10.5yes
```
c-knowles: Not much need to update these tests BTW - it's just checking invalid versions error out.

mumoshu: Makes sense. I'll leave it as-is because our `contrib/bump-version` script automatically catches it up. But thanks for pointing it out!

Issue #1396, opened 2018-07-08
Calico 3.1.x support


c-knowles:
Calico 3.1 supports HostEndpoint support for Kubernetes API datastore.

> Calico now supports configuration of host endpoints when using the Kubernetes API datastore. This allows you to seamlessly apply network policy to Kubernetes host machines and Kubernetes pods alike using Calico global network policies.

Ref https://docs.projectcalico.org/v3.1/releases/

It will require updates to `cloud-config-controller`. I have the changes for this locally, just checking it runs ok before submitting.

PR #1397, opened 2018-07-08
Self hosted calico v3.1.3

c-knowles:
- Bump calico images for self hosted Calico (non-self hosted is EOL)
- Include latest CRDs and RBAC for canal from https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation/flannel
- Correct the self hosted flannel images to self hosted config (previously unused)
- Mount calico dirs in canal and kubelet to fix pod init issue, ref https://github.com/projectcalico/calico/issues/1795#issuecomment-390171235

Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1396

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -35,11 +35,11 @@ const (
 	userDataDir    = "userdata"
 
 	// Experimental SelfHosting feature default images.
-	kubeNetworkingSelfHostingDefaultCalicoNodeImageTag = "v3.0.6"
-	kubeNetworkingSelfHostingDefaultCalicoCniImageTag  = "v2.0.5"
+	kubeNetworkingSelfHostingDefaultCalicoNodeImageTag = "v3.1.3"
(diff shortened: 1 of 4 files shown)


mumoshu:
JFYI, the non-self-hosted installation is not EOL yet according to the release note:

> ## Limitations
> - Offers only Kubernetes, OpenShift, and host endpoint integrations: the OpenStack, Mesos, DC/OS, rkt, and Docker orchestrators have not been tested and are not supported. (Calico still supports Docker and rkt containers.) We plan to resume support for the OpenStack, Mesos, DC/OS, and Docker orchestrators in a future release.
>
> https://docs.projectcalico.org/v3.0/releases/#v3.0.0

c-knowles:
@mumoshu to be clear I meant EOL within kube-aws. If possible we should simplify available options to avoid user confusion, especially now self hosted is enabled by default.

PR #1398, opened 2018-07-09
0.10.x Extend CustomFiles to be able to render go templates 

davidmccormick:
Please can we also include this feature in the 0.10.x?


davidmccormick:
Raised incorrectly

mumoshu:
Note: This is a equivalent to #1390, but for the 0.10.x branch.

PR #1400, opened 2018-07-09
0.10.x Add kubernetes manifests from customfiles

davidmccormick:
Please can we include this change in the 0.10.x release?  It would help us greatly if we can include these two customFile changes so that we can move some of our code out of our fork of kube-aws and closer to using the upstream version instead.

This change patches 'install-kube-system' so that it will kubectl apply any manifest files written to /srv/kubernetes/manifests/custom/*.yaml

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -979,7 +979,7 @@ write_files:
       {{ end }}
 
       # Allow installing kubernetes manifests via customFiles in the controller config - installs all manifests in mfdir/custom directory.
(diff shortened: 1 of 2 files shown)


c-knowles:
I think we want this anyway for other issues in the backlog where I imagine we eventually store the manifests into S3 then do something like `aws s3 sync --delete s3://bucket/` prior to applying.

davidmccormick:
@c-knowles yes, I have used a similiar solution previously which is not unlike the kubernetes addon-manager https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/addon-manager

This is enough for us to move out some of our patches to install-kube-system! :)

Issue #1407, opened 2018-07-15
Tweaks to network config


c-knowles:
While looking at some issues with CNI and hostPort connections mentioned in https://github.com/kubernetes-incubator/kube-aws/issues/704#issuecomment-403265649, I found some discrepancies between the kube-aws network configuration and what's expected or perhaps optimal. I've since discovered the cause of my hostPort issues was something unrelated but the items I found are likely worth including in kube-aws. I'm going to list them here first and we can always split them up.

1. We should update `cniVersion` from `0.3.0` to `0.3.1`

2. We are not setting the MTU for Calico in the CNI config. It means that for Canal the MTU of the flannel interface is 8951 but all the Calico interfaces are 1500. It still works but I believe it's not the optimal value. According to [the Calico docs](https://docs.projectcalico.org/v3.1/usage/configuration/mtu), we should set this to 8951 to match flannel:

> When using flannel for networking, the MTU for the network interfaces should match the MTU of the flannel interface. In the above table the 4th column “Calico MTU with VXLAN” is the expected MTU when using flannel configured with VXLAN.

3. We should set `externalSetMarkChain` to `KUBE-MARK-MASQ` in the CNI config to reuse existing iptables chains. It's not in the default YAMLs but I found it [here](https://github.com/containernetworking/plugins/blob/master/plugins/meta/portmap/README.md). It seems our friend redbaron [found the same](https://github.com/kubernetes-incubator/bootkube/pull/711#pullrequestreview-81553366)

> externalSetMarkChain - string, default nil. If you already have a Masquerade mark chain (e.g. Kubernetes), specify it here. This will use that instead of creating a separate chain. When this is set, markMasqBit must be unspecified.

4. We should set the `--service-cluster-ip-range` flag on controller manager as it prevents IPs being assigned in the service CIDR range should it overlap with the pod CIDR range. This doesn't seem to be documented anywhere I could find, but most bootstrap instructions set it and I just found we didn't set it. Dug into the code and found [this](https://github.com/kubernetes/kubernetes/blob/af8a1dba68f7d34c0cae45e1aaa1bc41c7299cc1/pkg/controller/nodeipam/ipam/controller.go#L139-L141).

davidmccormick:
Hi

[...]

PR #1408, opened 2018-07-15
Tweaks to network config

cknowles:
Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1407.

For now just document the `podCIDR` aspects but we should probably switch the default based on the networking setup, possibly combine this config with the `selfHosting` config into a networking section.

### DIFF
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -978,8 +978,7 @@ worker:
 # serviceCIDR: "10.3.0.0/24"
 
 # CIDR for all pod IP addresses
-# Should set to 10.244.0.0/16 for Flannel or Canal
-# Should set to 192.168.0.0/16 for Calico
+# Typically set to 10.244.0.0/16 for Flannel or Canal and 192.168.0.0/16 for Calico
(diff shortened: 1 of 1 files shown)


c-knowles:
@davidmccormick I cannot assign you for review but worth a look.

c-knowles:
@mumoshu please hold off on merging this, something is causing nodes in a couple of our clusters to shift into `SchedulingDisabled` status a bit at random and not yet sure if it's this change or the one in https://github.com/kubernetes-incubator/kube-aws/pull/1397. I believe it to be this change because we've been running the other PR in two clusters for a week without issue and it seems to only be when we rolled this change that nodes starting becoming randomly unhealthy. Note, this is even with a totally fresh stack (as we did change the pod CIDR change so wanted a fresh one).

c-knowles:
@mumoshu I believe we identified the cause of my notes issues as unrelated now. I've corrected my cluster.yaml notes according to @davidmccormick's feedback in https://github.com/kubernetes-incubator/kube-aws/issues/1407. I think this is ready to merge.

Issue #1405, opened 2018-07-13
[Proposal]


jorge07:
Add kube-aws version to the cluster yaml.

When you've many cluster with a lot of different versions is hard to identify with which version of kube-aws you launch the cluster.

What do you think?

davidmccormick:
Hi @jorge07 I like the idea of exposing the version of kube-aws that provisioned a cluster, but I'm not sure that the cluster.yaml is the right place for this.  If we embed it in the cluster.yaml what action should we take when that cluster.yaml is read by another kube-aws binary built to a different version?

I do collect the kube-aws version in the latest master branch and expose it when you login to a server. e.g: -

```
dave@mbp ~$ ssh core@10.29.26.8
Last login: Tue Jul 10 15:46:27 UTC 2018 from 172.21.133.174 on pts/1
Container Linux by CoreOS stable (1745.7.0)
Kubernetes v1.10.5
kube-aws add-kubernetes-manifests-from-customfiles/8caae2d4
Update Strategy: No Reboots
Failed Units: 1
  coreos-metadata-sshkeys@core.service
master|davem-lab-secure core@ip-10-29-26-8 ~ $
```
In this example the host was provisioned by a feature branch version of kube-aws.  I think I we could also tag aws resources with the version too - would you find that useful?

I think that cluster.yaml needs to express the desired state of the cluster and kube-aws version isn't desired state, it's more like status metadata.  I'm also unclear what action we should/can take based on the customer supplying this as an input value.

Issue #901, opened 2017-09-01
Reimplementing tagging of spot instances managed by a fleet


mumoshu:
A worker node backed by a spot instance managed by a spot fleet runs the `tag-spot-instance.service` systemd unit to tag the spot instance with:

* the cluster name tag required by k8s and
* other utility tags defined by kube-aws.

As @paalkr pointed out in #803, this can be done via cloudformation today.

https://aws.amazon.com/about-aws/whats-new/2017/07/tag-your-spot-fleet-ec2-instances/

c-knowles:
Ah, this would be good for monitoring. We are currently using a custom tag in `cluster.yaml` with the environment name and then we filter to that plus `aws:cloudformation:logical-id` == `Workers` to plot how many workers in an environment. We're using Datadog. Unfortunately right now in a spot fleet the instances receive neither of those tags so they are missing from the totals. Happy to change this to whatever we come up with, just need a way to segregate workers from other nodes and switch between environments.

c-knowles:
@mumoshu it seems this is not exposed via CF as of today? I couldn't find tags anywhere in the docs but it does seem available via the CLI.

mumoshu:
@c-knowles I've researched how the SpotRequestConfigData containing tags look like when it was generated in AWS console:

```json
{
  "IamFleetRole": "arn:aws:iam::myaccountid:role/myrole",
  "AllocationStrategy": "lowestPrice",
  "TargetCapacity": 1,
  "SpotPrice": "0.128",
  "ValidFrom": "2017-11-22T04:44:53Z",
  "ValidUntil": "2018-11-22T04:44:53Z",
  "TerminateInstancesWithExpiration": true,
  "LaunchSpecifications": [
    {
      "ImageId": "ami-2a69be4c",
      "InstanceType": "c3.large",
      "KeyName": "mykey",
      "BlockDeviceMappings": [
        {
          "DeviceName": "/dev/xvda",
          "Ebs": {
            "DeleteOnTermination": true,
            "VolumeType": "gp2",
            "VolumeSize": 8,
            "SnapshotId": "snap-123456788"
          }
        }
      ],
      "TagSpecifications": [
        {
          "ResourceType": "instance",
          "Tags": [
            {
              "Key": "Foo",
              "Value": "Bar"
            }
          ]
        }
      ]
    }
  ],
  "Type": "maintain"
}
```

[...]

PR #1409, opened 2018-07-16
Add kube-aws:version to stacktags

davidmccormick:
We don't currently tag our stack or instances with the version of kube-aws that created them.  This change adds a 'kube-aws:version' tag into each stack, including the root stack, and fixes a bug where tags are not updated when stacks are updated.  These stacktags are also inherited by instances, so all ec2 instances also recieve the tag.  This makes it very easy to use AWS API/CLI/GUI to check on the version of kube-aws used to deploy a cluster.

This idea was mentioned in https://github.com/kubernetes-incubator/kube-aws/issues/1405 where it was suggested as a good idea - giving us less reason for further requests to embed it the cluster.yaml.

### DIFF
--- a/cfnstack/provisioner.go
+++ b/cfnstack/provisioner.go
@@ -3,14 +3,15 @@ package cfnstack
 import (
 	"errors"
 	"fmt"
+	"strings"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/kubernetes-incubator/kube-aws/logger"
 	"github.com/kubernetes-incubator/kube-aws/model"
-	"strings"
-	"time"
 )
 
 type Provisioner struct {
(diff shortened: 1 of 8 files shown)


c-knowles:
Awesome! Does this also work with spot instances? They do not inherit the stack tags unfortunately and kube-aws supports some other attributes for that, although I kind of which we connected them as to the stack tags since AWS doesn't.

davidmccormick:
Hi, I haven't tested spot instances but it looks as though they are responsible for setting their own tags and the stack tags are propigated to them through their cloud-config files, so they *should* also set this new tag!

I've added the new 'kube-aws:version' tag to StackTags so it behaves like any other StackTag - and so I'm quite confident that it will be propigated down.  However, I am a little less confident that this is the behavoir that you desire (hopefully it is)!

mumoshu:
We have a set of script/unit to tag spot instances by their own! https://github.com/kubernetes-incubator/kube-aws/blob/77c43b5e16912d89a9238c4a38b573303bcdb71e/core/nodepool/config/templates/cloud-config-worker#L1090-L1136

It would be nice if we could add the support for it later.

mumoshu:
Oh, it seems like we can remove the script/unit in favor of the official cfn support for spotfleet tag specification as per #901!

davidmccormick:
Could changing the implementation of setting spot tags be another issue/pr?  I can't properly test it here.

c-knowles:
My vote goes for separate, it's a bit delicate as kube-aws also supports separate tags per node pool as well.

Issue #1417, opened 2018-07-19
Can't dep ensure from kube-aws master


davidmccormick:
When trying to pull in kube-aws using `dep ensure` I get the following error: -

```
(11)  ✗   "github.com/kubernetes-incubator/kube-aws" transitively (through 2 packages) imports
 "github.com/kubernetes-incubator/kube-aws/core/network/config", which contains malformed code: 
no buildable Go source files in /Users/dave/code/hcom/pkg/dep/sources/https---github.com-
kubernetes--incubator-kube--aws/core/network/config
```

I think that this is because network config package does not have any go code within it until it has been built and the templates.go file has been created, but we need to pull the source with dep *before* this can happen.

I propose adding a simple file to ensure that this package always contains code so that dep will be able to pull it.

davidmccormick:
This is now resolved.

PR #1418, opened 2018-07-19
Fix dep ensure empty package issue

davidmccormick:
When embedding kube-aws in another go tool we want to be able to pull it in using dep ensure.  dep refuses to pull in empty packages and so does not like the network/config package which is empty until a build is performed and the templates.go file generated.

Committing this stub file satisfies dep and allows it to successfully pull in kube-aws.

### DIFF
--- a/core/network/config/config.go
+++ b/core/network/config/config.go
@@ -1,3 +1,3 @@
 package config
 
-const DepFindsCode = true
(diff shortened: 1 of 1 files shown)


davidmccormick:
fix for https://github.com/kubernetes-incubator/kube-aws/issues/1417

Issue #1424, opened 2018-07-26
metrics server install-kube-system fix causing intermittent hang on new clusters


davidmccormick:
It looks like there has been an issue within 'install-kube-system' when the metrics service is enabled which was documented and fixed in issue:  https://github.com/kubernetes-incubator/kube-aws/issues/1039#issuecomment-348978375.   We are now finding that when we spin up some new clusters that they get stuck within 'install-kube-system' which is not completing and allowing 'cfn-signal' to fire.

We see it get stuck in the logs...
```
Jul 26 09:41:37 ip-10-4-5-95.us-west-2.compute.internal retry[2152]: Sleeping 3 seconds.
Jul 26 09:41:40 ip-10-4-5-95.us-west-2.compute.internal retry[2152]: Sleeping 3 seconds.
Jul 26 09:41:43 ip-10-4-5-95.us-west-2.compute.internal retry[2152]: Sleeping 3 seconds.
```
Looking at 'install-kube-system' it looks to be stuck in this code: -
```
# See https://github.com/kubernetes-incubator/kube-aws/issues/1039#issuecomment-348978375
if ks get apiservice v1beta1.metrics.k8s.io && ! ps ax | grep '[h]yperkube proxy'; then
  echo "apiserver is up but kube-proxy isn't up. We have likely encountered #1039."
  echo "Temporary deleting the v1beta1.metrics.k8s.io apiservice as a work-around for #1039"
  ks delete apiservice v1beta1.metrics.k8s.io

  echo Waiting until controller-manager stabilizes and it creates a kube-proxy pod.
  until ps ax | grep '[h]yperkube proxy'; do
      echo Sleeping 3 seconds.
      sleep 3
  done
  echo kube-proxy stared. apiserver should be responsive again.
fi
```
It seems that this code executes before the code that installs the kube-proxy and so gets stuck in the inner loop waiting for the proxy to come up.

PR #1426, opened 2018-07-26
Fix kube-proxy startup race condition when metric server is enabled.

omar-nahhas:
This PR is to solve the reported issue: https://github.com/kubernetes-incubator/kube-aws/issues/1424

The fix is to move the installation of kube-proxy up in the install-kube-system script, so it gets
a chance to start before the installation checks for it and goes into a loop.

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -987,6 +987,12 @@ write_files:
 
       applyall "/srv/kubernetes/manifests/kube-system-ns.yaml"
 
+     # Daemonsets
+      applyall \
+        "${mfdir}/kube-proxy-ds.yaml" \
+        {{ if .Experimental.NodeDrainer.Enabled }}"${mfdir}/kube-node-drainer-ds.yaml"{{ end }} \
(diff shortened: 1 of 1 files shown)


core/controlplane/config/templates/cloud-config-controller
```diff
@@ -987,6 +987,12 @@ write_files:
 
       applyall "/srv/kubernetes/manifests/kube-system-ns.yaml"
 
+      # Daemonsets
+      applyall \
+        "${mfdir}/kube-proxy-ds.yaml" \
+        {{ if .Experimental.NodeDrainer.Enabled }}"${mfdir}/kube-node-drainer-ds.yaml"{{ end }} \
```
mumoshu: I had been a bit nervous about moving kube-node-drainer-ds altogether, wondering if `kube-node-drainer` would depend on cluster dns. But actually we had been [deploying the drainer before dns](https://github.com/kubernetes-incubator/kube-aws/blob/c6d72bc9ad927b5509932bb55d614e69d02e19b5/core/controlplane/config/templates/cloud-config-controller#L1080-L1089). No problem, then!

omar-nahhas:
Thank you very much. And apologies for the slow response (I was on holidays)

PR #1427, opened 2018-07-30
[Feature] Add more etcd options

wwyiwzhang:
This Pull request is to address the option of passing arguments to `/usr/local/bin/etcd` such as setting `--quota-backend-bytes` accordingly to the size of the etcd nodes. 

Apologize for my rusty Golang and I am pretty sure there are more places I need to change. Really appreciate your feedback.

### DIFF
--- a/core/etcd/config/templates/stack-template.json
+++ b/core/etcd/config/templates/stack-template.json
@@ -497,6 +497,12 @@
                   "ETCD_VERSION='",
                     "{{$.Etcd.Version}}",
                   "'\n"
+                  {{if $.Etcd.FormatOpts}}
+                  ,
+                  "ETCD_OPTS='",
+                    "{{$.Etcd.FormatOpts}}",
+                  "'\n"
+                  {{end}}
                 ]]}
               }
             }
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -770,6 +770,12 @@ worker:
 #        Description=Example Custom Service
 #        [Service]
 #        ExecStart=/bin/rkt run --set-env TAGS=Controller ...
+#  # extra etcd options 
+#  userSuppliedArgs:
+#    # for example, setting the --quota-backend-bytes and --auto-compaction-retention options
+#    quotaBackendBytes:
+#    autoCompactionRetention:
+
 
 ## Networking config
 
--- a/model/etcd.go
+++ b/model/etcd.go
@@ -20,7 +20,7 @@ type Etcd struct {
 	DisasterRecovery   EtcdDisasterRecovery `yaml:"disasterRecovery,omitempty"`
 	VolumeMounts       []VolumeMount        `yaml:"volumeMounts,omitempty"`
 	EC2Instance        `yaml:",inline"`
-	UserSuppliedArgs   UserSuppliedArgs `yaml:"userSuppliedArgs,omitempty`
+	UserSuppliedArgs   UserSuppliedArgs `yaml:"userSuppliedArgs,omitempty"`
 	IAMConfig          IAMConfig        `yaml:"iam,omitempty"`
 	Nodes              []EtcdNode       `yaml:"nodes,omitempty"`
 	SecurityGroupIds   []string         `yaml:"securityGroupIds"`
--- a/model/etcd.go
+++ b/model/etcd.go
@@ -3,7 +3,13 @@ package model
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
+)
+
+const (
+	MaxQuotaBackendBytes     int = 8 * 1024 * 1024 * 1024
+	DefaultQuotaBackendBytes int = 2 * 1024 * 1024 * 1024
 )
 
 type Etcd struct {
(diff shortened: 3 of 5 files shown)


davidmccormick:
Hi @wwyiwzhang 

I am wondering what your usecase/rational is for wanting to set --quota-backend-bytes?  Are you experiencing issues with a cluster with a very large etcd which is running out of key space?  If you don't mind me asking, what sorts of objects are taking all your space?

wwyiwzhang:
Hi @davidmccormick, thanks for the feedback, yes we are experiencing with etcd storage running out of key space. We have an ad-hoc cluster where every engineer has his/her ASGs and they use them for ad-hoc related work by scaling their ASGs up/down on a daily basis. As the team grows, this cluster is showing slowness which mainly caused by the etcd. For example, we got errors like `mvcc: database space exceeded` in the past and had to do defragmentation. The etcd node for this cluster is on t2.medium and recently I set up a new cluster with better specs, for example, instance type is m4.large. However, we still cannot benefit from this large instance type if we don't tune the etcd options. Also, I checked with one of the contributors of kube-aws and he suggested and I agreed to add an option in cluster.yaml, instead of get inside the rkt container and manually do `etcd --quota-backend-bytes=<value>`.  So I made this PR to give the user s options to set the quota-backend-bytes to a larger value to avoid dealing with etcd space issue frequently. In the future, we could also add options like `--auto-compaction`. I hope this answers your question.

model/etcd.go
```diff
@@ -14,6 +15,7 @@ type Etcd struct {
 	DisasterRecovery   EtcdDisasterRecovery `yaml:"disasterRecovery,omitempty"`
 	VolumeMounts       []VolumeMount        `yaml:"volumeMounts,omitempty"`
 	EC2Instance        `yaml:",inline"`
+	OPTS               EtcdOpts     `yaml:"opts,omitempty`
```
davidmccormick: This is very generic - would something like UserSuppliedArgs be more descriptive?

model/etcd.go
```diff
@@ -29,6 +31,15 @@ type EtcdDisasterRecovery struct {
 	Automated bool `yaml:"automated,omitempty"`
 }
 
+type EtcdOpts struct {
+	QuotaBackendBytes int `yaml:"quotaBackendBytes,omitempty"`
+}
```
davidmccormick: How about adding another supported arg in order to test that your code is good with mutliple user provided args?

wwyiwzhang:
Hi @davidmccormick, I took a look at the number of keys by prefix, it seems that prefix `/registry/events/` has the most keys. Housekeeping is definitely an option but it means we need to keep an eye on this thing and do it manually. I am hoping to extend this PR to use flags such as, `--quota-backend-bytes`, `--auto-compaction-retention=3` so etcd take care of it in a more automatic way.

davidmccormick:
HIyah @wwyiwzhang yes `auto-compaction-retention` sounds interesting as does `auto-compaction-mode=revision` - these sound like a good way of reducing your key storage space.

Wow, you must have a *lot* of events! :) Events are normally removed after 1 hour but retention can be modified via the kube-apiserver `event-ttl` = duration Amount of time to retain events. (default 1h0m0s).  I must admit that I'm not entirely sure why the apiserver and not the controller is responsible for cleaning up events.

model/etcd.go
wwyiwzhang: will change.

mumoshu:
@wwyiwzhang Hey. Thanks for your efforts!

The first half of your commits seems to have wrong author. Perhaps missing/wrong in the git commit author?

image

I believe once you fixed the author info in your commits, the `cla/linuxfoundation` check starts passing:

image

mumoshu:
Regarding the build failure, this seems like the only issue remained failing:

```
# github.com/kubernetes-incubator/kube-aws/model
import cycle not allowed in test
package github.com/kubernetes-incubator/kube-aws/model (test)
	imports github.com/kubernetes-incubator/kube-aws/model
FAIL	github.com/kubernetes-incubator/kube-aws/model [setup failed]
```

core/etcd/config/templates/stack-template.json
```diff
@@ -496,7 +496,12 @@
                   "'\n",
                   "ETCD_VERSION='",
                     "{{$.Etcd.Version}}",
-                  "'\n"
+                  "'\n",
+                  {{if $.Etcd.FormatOpts}}
+                  "ETCD_OPTS='",
+                    "{{$.Etcd.FormatOpts}}",
+                  "'\n",
+                  {{end}}
```
mumoshu: Perhaps you have an extra `,` after the last element of the format opts, that is causing this error?

```
    	maincluster_test.go:3533: failed to create cluster driver : failed to initialize etcd stack: Error while rendering template: invalid character ']' looking for beginning of value:
    		json syntax error (offset=10035), in this region:
    		-------
    		,
    		                    "--quota-backend-bytes=2147483648",
    		                  "'\n",
    		                  
    		                ]]}
    		              }
    		            }
    		          }
    		
    		-------
```

model/etcd_test.go
```diff
@@ -0,0 +1,62 @@
+package model
+
+import (
+	"testing"
+)
+
+func TestEtcd(t *testing.T) {
```
mumoshu: Thank you so much for adding the test 🙏

core/root/config/templates/cluster.yaml
```diff
@@ -770,6 +770,12 @@ worker:
 #        Description=Example Custom Service
 #        [Service]
 #        ExecStart=/bin/rkt run --set-env TAGS=Controller ...
+#  # extra etcd options 
+#  userSuppliedArgs:
+#    # for example, setting the --quota-backend-bytes and --auto-compaction-retention options
+#    quotaBackendBytes:
+#    autoCompactionRetention:
```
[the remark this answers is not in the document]
wwyiwzhang: Hi @mumoshu, good suggestion! I can make a follow up PR or if anyone is interested, feel free to refactor as well.

PR #1428, opened 2018-08-02
feat: Ability to toggle node pool rolling strategy between Sequential and Parallel

Matei207:
### What

Ability to toggle in your `cluster.yaml`  file at a `nodePool:` or `worker:` level whether nodePools are going to be rolled out / rolled in sequentially or concurrently

### Why

After doing a fair share of rolls, it was noticed that at times, platforms are more resilient to a full `Availability Zone` going down rather than having 1 `nodePool` going down in each respective `Availability Zone`. Since sometimes it is needed to roll out nodePools in sequential order, it would be easier to have it as a toggleable option in your `cluster.yaml`.

This should allow for easier switching between rolling out workers sequentially or concurrently.

### How

* In the root stack-template, I made it so that when we loop through our currently declared `nodePools` we store their index. In the `DependsOn` field, I have added control flow to check if the current `nodePool`  has `sequentialRoll` enabled, and if it does, the previous nodePool is added as a dependency.

* To toggle `sequentialRoll` globally, I have added the sequentialRoll field in the `Worker` struct. When the config is read from a file, if the sequentialRoll field is enabled at a `Worker` level, then it overwrites all values of sequentialRoll false at a nodePool level, setting them to true. This ensures that if declared at a global level, then all of the nodePools are going to be rolled out sequentially. If that is not the desired behaviour, simply set the targeted nodePools to roll out sequentially on an individual basis.

*  The TemplateParams `NodePools` function now returns a slice of `nodePool` instead of a slice of `NestedStack` interfaces, this was done in order to be able to use the sequentialRoll flags in the stack-template.

#### Notes

* I tested this in cluster with 2 sets of nodePools (3 nodePools per set), one concurrent and one sequential, across 3 different AZs and it worked perfectly. There is a caveat, however, if there are nodePools being rolled out sequentially and concurrently, the nodePools declared sequentially will have to be declared before the concurrent ones, otherwise the first declared sequential nodePool will dependend on the last declared concurrent nodePool. 

* If nodePools are to be rolled sequentially, they will roll out in order of declaration in your `cluster.yaml` (unsure if I specified that anywhere) e.g if you have A, B, C declared in that order, thats the order they will also roll out in.

### DIFF
--- a/core/nodepool/cluster/cluster_test.go
+++ b/core/nodepool/cluster/cluster_test.go
@@ -106,8 +106,7 @@ func TestValidateKeyPair(t *testing.T) {
 s3URI: s3://mybucket/mydir
 apiEndpoints:
 - name: public
-  sequentialRoll:
-    enabled: false
+  nodePoolRollingStrategy: parallel
   dnsName: test-cluster.example.com
   loadBalancer:
     hostedZone:
--- a/core/nodepool/cluster/cluster_test.go
+++ b/core/nodepool/cluster/cluster_test.go
@@ -106,6 +106,8 @@ func TestValidateKeyPair(t *testing.T) {
 s3URI: s3://mybucket/mydir
 apiEndpoints:
 - name: public
+  sequentialRoll:
+    enabled: false
   dnsName: test-cluster.example.com
   loadBalancer:
     hostedZone:
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -269,12 +269,13 @@ worker:
 #  # Required if there are 2 or more API endpoints defined in `apiEndpoints`
 #  # Can be overriden per node pool by specifying `worker.nodePools[].apiEndpointName`
 #  apiEndpointName: versionedPublic
-#  # Used to globally toggle nodePool sequential rollout strategy
-#  # Default behaviour (false / disabled) will not overwrite anything
-#  # If enabled (set to true) it will overwrite all "sequentialRoll" false values at a nodePool level (see below)
-#  # For finer grain control, enable it at a lower (nodePool) level if only targeting specific nodePools.
-#  GlobalSequentialRoll:
-#    enabled: true
+#  # Used to globally specify nodePool rolling strategy across availability zones: 'sequential' or 'parallel'
+#  # Default behaviour ('parallel') will roll out nodes concurrently across availability zones
+#  # If set to 'sequential', the nodePools will roll out sequentially, in order of declaration, across availability zones
+#  # If set at a Worker level, it will only take effect if it is not set at a nodePool level
+#  # For finer grain control, declare it on a nodePool-by-nodePool basis
+#  nodePoolRollingStrategy: sequential
+
   nodePools:
     - # Name of this node pool. Must be unique among all the node pools in this cluster
       name: nodepool1
@@ -283,15 +284,13 @@ worker:
 #      # If omitted, public subnets are created by kube-aws and used for worker nodes
 #      subnets:
 #      - name: ManagedPublicSubnet1
-#      # Used to toggle how the worker nodepools will roll out.
-#      # Default behaviour (false / disabled) will create / upgrade each nodepool in parallel
-#      # If enabled, the worker nodepools will roll out one after the other
-#      # in sequence of how they are defined in the nodePool section
-#      # NOTICE: if there are multiple nodePool groups that are to be rolled sequentially and concurrently
-#      # the nodePools that have sequential roll out need to be declared before those with concurrent roll out
-#      # otherwise the first declared sequential nodePool will be dependent on the last declared concurrent nodePool
-#      sequentialRoll:
-#        enabled: false
+#      # Used to specify the strategy in which nodePools will roll out across availability zones: in sequence or parallel
+#      # By default (if not set at nodePool/worker level) the nodePools will roll out in parallel across all availability zones
+#      # If set to 'sequential', the nodePools will roll out one-by-one in order of declaration across all availability zones
+#      # NOTICE: if there are multiple nodePool groups that will use both rolling strategies ('sequential'/'parallel')
+#      # the nodePools rolling out in sequence should be declared before those that roll out in parallel
+#      # otherwise, the first declared nodePool to roll out in sequence will depend on the last concurrent nodePool
+#      nodePoolRollingStrategy: parallel
 #      # Existing "glue" security groups attached to worker nodes which are typically used to allow
 #      # access from worker nodes to services running on an existing infrastructure
 #      securityGroupIds:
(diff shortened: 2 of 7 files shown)


mumoshu:
@Matei207 Hey! Thanks a lot for your work. You seem to have interesting use-cases.

Just for clarification,

> platforms are more resilient to a full `Availability Zone` going down rather than having 1 `nodePool` going down in each respective `Availability Zone`

Regarding your use-case, I wondered if what you would like the most is something like `nodePoolRollingStrategy: BatchPerAvailabilityZone`. It is an imaginary rolling strategy, which will basically group up node pools per az, and adds corresponding `DependsOn` to ASGs.

davidmccormick:
I think that the essence of this change is rather simpler than it sounds - that is "sometimes a customer will prefer to roll out updates one node at a time irrespective of AZ's".  The default strategy of nodepools in parallel is good for most customers, but it is useful to be able to change the behaviour on clusters where less unavailable nodes are desired.

I don't think this sounds like 'BatchPerAvailabilityZone' but more like: -
nodePoolRollingStrategy: Parallel
nodePoolRollingStrategy: Sequential

I don't have an issue with this being a boolean, but it could changed to something like above.

An added complication is that this PR also allows for the sequential strategy to be applied to specific nodepools, so some pools can roll using the parallel strategy and others can be set for sequential rolling.  This seems like it could be useful for some users/teams.

Matei207:
I quite like the idea of changing it to `nodePoolRollingStrategy:` and have it default to "Parallel". I think that would be more self-explanatory and easier to use in your `cluster.yaml`.

Also, @davidmccormick  is right regarding the use-case. In this particular case, the issue we faced is with one of our internal teams and their usage of Kafka. Whenever the cluster is rolled out, it is more desirable for us to have a full AZ going down (by rolling the `nodePools` in sequential order). If the `nodePools` are rolled in parallel, then we risk data loss in our Kafka queue because we could have multiple replicas of the same queue that can go down at the same time across multiple AZs. Inside the AZ, the team in question told us that it is ok to increase the maxBatchSize so we can upgrade multiple workers at the same time in that AZ.

Matei207:
### Update:
***
Based on the previous comments, I have now updated the branch/PR to reflect the suggestions. 

* The `struct`s were all updated to now have a `nodePoolRollingStrategy` field of type string. My assumption is that at this stage it will either be `Parallel` or `Sequential`.

* In the `ConfigFromBytes` method I have changed the initialisation/control flow of the `nodePoolRollingStrategy` (previously `sequentialRoll`). As before, if there is no explicit value passed in your cluster.yaml file, the `nodePool` will inherit whatever value was declared in `Worker` (which should act as a global value). In case `Worker` does not have a value either, we default to `Parallel` to keep the default behaviour.

* Since the `sequentialRoll` field was changed to `nodePoolRollingStrategy`, at a `Worker` level, the `globalSequentialRoll` field was also removed. I have chosen not to prepend a "global" prefix to the worker-level field, I think that would make it a bit more confusing.

The functionality/behaviour should be the same as before. I think this looks a lot cleaner so thank you for the suggestions! Hopefully, this will make the PR better, waiting to hear some more feedback. 


`e.g cluster.yaml`

```
worker:
nodePoolRollingStrategy: Parallel
  nodePools:
     - &my-data-nodepool
       name: my-data-nodepool-a
       nodePoolRollingStrategy: Sequential
```

core/root/config/templates/cluster.yaml
```diff
@@ -269,7 +269,13 @@ worker:
 #  # Required if there are 2 or more API endpoints defined in `apiEndpoints`
 #  # Can be overriden per node pool by specifying `worker.nodePools[].apiEndpointName`
 #  apiEndpointName: versionedPublic
-#
+#  # Used to globally specify nodePool rolling strategy across availability zones: 'Sequential' or 'Parallel'
```
mumoshu: Can you safely remove the part `across availability zones` from the explanation?
I'm saying it because it seems like we don't actually take AZs of node pools into account, but rather the one of expected use-cases is to roll parallel per AZ.

core/root/config/templates/cluster.yaml
```diff
@@ -278,7 +284,13 @@ worker:
 #      # If omitted, public subnets are created by kube-aws and used for worker nodes
 #      subnets:
 #      - name: ManagedPublicSubnet1
-#
+#      # Used to specify the strategy in which nodePools will roll out across availability zones: in sequence or parallel
```
mumoshu: Same as https://github.com/kubernetes-incubator/kube-aws/pull/1428/files#r210823510

Matei207:
### Update
***

I've made some changes to the `cluster.yaml` template as requested, and have also added in tests.

##### For the `cluster.yaml` file:
* Removed the following from the documentation at both `Worker` and `nodePool` level:
> "across all availability zones"

Hopefully it is a bit more intuitive now and will avoid any confusions :) If there is anything else that I should change regarding the docs, do let me know.

##### Tests:

I was unsure of how to proceed with the tests, so I followed the format in the file.

* There are now three test functions: `hasDefaultNodePoolRollingStrategy`, `hasSpecificNodePoolRollingStrategy` and `hasNodePoolAndWorkerStrategy`.  The first function tests if a `nodePool` has the roll-out strategy set to parallel by default (with no explicit declarations in the cluster.yaml file whatsoever). The other two assert if the expected roll-out strategy equals the actual roll-out strategy specified in the cluster.yaml file.

* The three test functions form the basis for four test cases that I came up with (if there are any more cases that I haven't considered do let me know and I will add them in!): testing for default values at a nodePool level, testing if the strategy is set at a worker level and then inherited by a nodePool, testing for a specific roll-out strategy, and finally, a test case that has two nodePool groups with inherited and explicit declarations. The tests all passed and hopefully they will add value to the overall test class.

Again, I am quite unsure if the tests cover enough and if they follow the format, if there is anything that should be changed, do let me know.

---


@mumoshu  regarding the following question:
> @Matei207 Thanks for the clarification on your use-case! The kafka thing made sense to me. Out of curiosity, does it relate to how repartitioning and/or rebalancing works?

According to one of the guys that works with Kafka, "rebalance/repartition are admin tasks that should run when scaling the cluster up, but not when rolling the cluster".  I don't really know the details on how they do the rebalance/reparation right now, but I don't think it relates much to this particular use-case as I believe their kafka brokers are doing neither (rebalance/repartition) when they're evicted as part of the nodeDrain before a worker is killed and replaced.

To offer a bit more background on the issue, one of the teams we support run a kafka cluster of 66 brokers split over 3 AZ's in a single AWS region. Those brokers support 72 topics split over 812 partitions, with each partition replicated 3 times, so 2436 total replicas. I think because the kafka cluster is so big, rebalancing/repartioning takes a while so it wouldn't be done on every cluster roll/node upgrade.

core/nodepool/cluster/cluster_test.go
```diff
@@ -106,6 +106,7 @@ func TestValidateKeyPair(t *testing.T) {
 s3URI: s3://mybucket/mydir
 apiEndpoints:
 - name: public
+  nodePoolRollingStrategy: parallel
```
mumoshu: Nit, but this seems redundant because it is a config for an API endpoint, that does nothing to do with the feature

PR #1448, opened 2018-09-21
fix: Attach kube-aws controller policy to pre-existing IAM role

Matei207:
### Why

When I worked on the custom kiam role PR I overlooked the fact that kube-aws autogenerates the controller policy:
* By default, kube-aws will create an `IAMManagedPolicy` for the controller, after it creates an `IAMControllerRole` and attaches the ManagedPolicy to it.
* If you use a pre-existing role (through the `useStrict` feature implemented in the custom kiam role PR), the `IAMManagedPolicy` gets generated, but it does not get attached to any role. It would be up to the developer to manually attach it in the console or through the aws cli. Moreoever, since the controller policy will always have a randomly generated name it complicates things.

In my personal opinion this is a bug. When using a pre-existing IAM role everything should work just as before.

### What

To fix this, I added an `if` condition in the stack-template. When creating the `ManagedPolicy`:
* If we are using a pre-existing role, and if that role name is in our cluster.yaml, then we attach the policy to our role, on creation.
* If we are using the kube-aws generated role, then the behaviour is the same as before.

---

I'm open to suggestions if there is anything else that should be added in or changed. I personally regard this as a bugfix, apologies for overlooking this detail in the first place when I implemented the custom role feature.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -231,7 +231,12 @@
       "Type" : "AWS::IAM::ManagedPolicy",
       "Properties" : {
         "Description" : "Policy for managing kube-aws k8s controllers",
-        "Path" : "/",
+				"Path" : "/",
+				{{ if and (.Controller.IAMConfig.Role.UseStrict) (.Controller.IAMConfig.Role.Name) }}
+				"Roles" : [
+           "{{.Controller.IAMConfig.Role.Name}}"
+				],
+				{{ end }}
         "PolicyDocument" :   {
           "Version":"2012-10-17",
           "Statement": [


mumoshu:
@Matei207 Does this mean that, now, all you need to do before running kube-aws is, just manually creating an IAM role with no policy attached? Is the IAM policy managed by kube-aws automatically attached to the existing IAM role?

Matei207:
@mumoshu in short, yes!

We use a cluster provisioning tool that wraps around kube-aws. The main use case is to initialise our `cluster.yaml` file, in addition, it also creates our `IAM` roles (such as external-dns, S3 bucket access etc), so our roles are on the aws account before the cluster rolls out.

This is what I have in my `cluster.yaml` and what I used to test this fix:
```yaml
controller:
  iam:
    role:
      name: us-west-2-md-KiamTrust
      useStrict: true
```
---

When I rolled out my cluster with the above configuration, the controller policy created by kube-aws was attached to my (___existing___) `us-west-2-md-KiamTrust` role. The role also got attached to the EC2 instance through the `InstanceProfile` arn.

So, all that I had to do was create a role, pass that role in my cluster.yaml and let kube-aws generate the policy and attach it to the EC2 control-plane instance 👍

mumoshu:
@Matei207 Thanks a lot for the explanation and contribution! Good to know that cfn is able to attach iam policies to existing roles. This is AWESOME!

mumoshu:
@Fsero After seeing how this works, I wonder if we could rename today's `iam.role.name` to `iam.role.namePrefix`, and instead use `iam.role.name` for the behavior of `useStrict: true`. WDYT?

PR #1451, opened 2018-09-26
0.12 release: remove legacy networking (less is more)

davidmccormick:
The 0.11 release completed the migration from legacy flannel and calico networking to the newer Self Hosting 'canal' or 'flannel' networking daemonsets.

This change is clean up all of the **old code** for that supports the legacy flannel and calico installation and which enabled the migration from legacy to the network daemonsets and makes self-hosting always enabled.  The worker nodes no longer have to have any knowledge/access to the etcd servers as SelfHosting uses the kubernetes api for storage.

This simplifies kube-aws, which is always nice! :)

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -219,7 +219,6 @@ func NewDefaultCluster() *Cluster {
 				},
 				Networking: Networking{
 					SelfHosting: SelfHosting{
-						Enabled:         true,
 						Type:            "canal",
 						Typha:           false,
 						CalicoNodeImage: model.Image{Repo: "quay.io/calico/node", Tag: kubeNetworkingSelfHostingDefaultCalicoNodeImageTag, RktPullDocker: false},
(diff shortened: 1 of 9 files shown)


davidmccormick:
I've run a couple of manual tests - 1st upgrading my 0.10.x cluster to this release with canal networking and then an update where I change the networking to 'flannel' and both worked successfully.  This change requires that the legacy 'useCalico' setting be removed from cluster.yaml.

core/controlplane/config/config.go
```diff
@@ -1317,13 +1304,11 @@ func (c Cluster) validate() error {
...
 	}
 
-	if c.Kubernetes.Networking.SelfHosting.Enabled {
-		if (c.Kubernetes.Networking.SelfHosting.Type != "canal") && (c.Kubernetes.Networking.SelfHosting.Type != "flannel") {
-			return fmt.Errorf("networkingdaemonsets - style must be either 'canal' or 'flannel'")
-		}
-		if c.Kubernetes.Networking.SelfHosting.Typha && c.Kubernetes.Networking.SelfHosting.Type != "canal" {
-			return fmt.Errorf("networkingdaemonsets - you can only enable typha when deploying type 'canal'")
-		}
+	if (c.Kubernetes.Networking.SelfHosting.Type != "canal") && (c.Kubernetes.Networking.SelfHosting.Type != "flannel") {
```
kylehodgetts: Bit of a picky comment, but can we keep if statement style consistent.

`if (c.Kubernetes.Networking.SelfHosting.Type != "canal") && (c.Kubernetes.Networking.SelfHosting.Type != "flannel"){` can be replaced by
` if c.Kubernetes.Networking.SelfHosting.Type != "canal" && c.Kubernetes.Networking.SelfHosting.Type != "flannel"` right?

davidmccormick: Sure - that makes sense, have updated!

core/root/config/templates/cluster.yaml
```diff
@@ -1189,29 +1159,10 @@ kubernetes:
...
-# ISSUES2 - Existing pod ip addresses are not tracked and can cause clashing ips - please cordon all nodes whilst
-# performing the upgrade until all workers/nodes have been recycyled - then it will be ok.
-# ISSUES3 - Rolling back to old networking will require all nodes are cycled/rolled and the network is disrupted throughout.
-
+# increases substantially as the number of nodes is increased.
   networking:
     selfHosting:
-      enabled: true    # false will fall back to legacy coreos flannel/etcd2 installation
       type: canal      # either "canal" or "flannel"
       typha: false     # enable for type 'canal' for 50+ node clusters
```
mumoshu: Do we have any reasons not to enable typha by default btw?

davidmccormick: Typha is only needed on larger clusters with 50+ nodes.

mumoshu: Would the purpose of avoiding it when unnecessary is to avoid moving parts?
I just wondered if it would be a good idea to enable it by default if it doesn't have any bad side-effect, so that users won't suffer from degraded performance after reaching 50+ nodes

davidmccormick: I'm a little worried about trying to set it automatically because that would mean I would have to count the nodes in the nodepools and make allowances for possible cluster auto-scaling.  With it off by default, a cluster admin with a growing cluster and seeing high loads on their apiservers will hopefully have had some time to start getting acquainted with the cluster.yaml file and settings.

mumoshu: Makes sense. Thanks a lot for clarifying!

PR #1453, opened 2018-09-27
fix: `encryptionAtRest: true` should not break controller nodes

### LINKED ISSUE #1450: v0.11 fails at controller creation with encryptionAtRest enabled
When kubernetes.encryptionAtRest.enabled is true, the cluster fails at the controller creation phase.

After taking a look at the `cloud-config-controller`, looks like theres a missing equals sign at
```
--experimental-encryption-provider-config /etc/kubernetes/additional-configs/encryption-config.yaml
```

I thought putting in that equals sign might have fixed it but I was still unable to make it work. I am successfully able to spin up a new cluster with encryptionAtRest disabled.

mumoshu:
Fixes #1450


mumoshu:
This is a backport of #1452 to v0.11.x

Issue #1469, opened 2018-10-16
Notice for the upcoming major refactoring/improvements to kube-aws towards the EKS integration and for the future


mumoshu:
## What do we get from this?

- Better, documented kube-aws plugin system. Use builtin plugins or write your own with a single file called `plugin.yaml`, to customize every aspect of kube-aws clusters.
- aws-iam-authenticator for kubelet auth, as a kube-aws plugin
- (near future) EKS support #1434 powered by the plugin system

## Story

Take this as a continuation of https://github.com/kubernetes-incubator/kube-aws/issues/509#issuecomment-316805643

Towards the EKS integration(#1434), I have been working on the support for aws-iam-authenticator for kubelet authentication. And I was very annoyed by that I was forced to add small snippets of bash, cloud-config and go into several diverse locations of kube-aws's code-base, plus adding the feature to somehow obtain and transfer the `aws-iam-authenticator` "binary" onto kube-aws worker and controller nodes.

To summarize, things that I had to consider were:

- Creating a single keypair whose cert has two purposes - CA and Server Authn, encrypting it inside the `credentials/` dir, transferring to worker and controller nodes
- Somehow downloading and installing `aws-iam-authenticator` binary onto nodes
- Installing the aws-iam-authenticator deployment onto the k8s cluster
- Creating a configmap for aws-iam-authenticator that refers to IAM roles that are managed by kube-aws(Templating the configmap with the results of cloudformation functions? I'm tired of writing a long `{{ "Fn::Join": ["", ["long", "bash", "script", "separated and enclosed within double-quotes", "per", "line"]]}}` thing.

Being pretty annoyed, recalling #751 #791  #509, I decided to enhance the secret feature of kube-aws - "plugins" - to cover the use-case. It is resulting in a major refactoring of kube-aws.

It doesn't affect existing features of kube-aws at all. But it does affect public members exposed by several kube-aws go packages, and how golang packages are organized.

So to whom using kube-aws a golang library, this refactoring may affect you. Please feel free to send any feedback on that.

# Notables user-visible changes

## Improved plugin system

Create a `plugin.yaml` at `PROJECT_ROOT/plugins/NAME/plugin.yaml` and edit according to your needs. [The aws-iam-authenticator plugin](https://github.com/mumoshu/kube-aws/blob/aws-iam-node-auth/builtin/files/plugins/aws-iam-authenticator/plugin.yaml) would be the most informative source for learning how it write it.

It can be used for:

[...]

PR #1470, opened 2018-10-17
Allow plugins to add kube-controller-manager command line flags.

davidmccormick:
Plugins can set apiserver command line flags but not kube-controller-manager flags.  This enables these as well.

Intended as part of moving some core functionality into plugins https://github.com/kubernetes-incubator/kube-aws/issues/1469

### DIFF
--- a/plugin/clusterextension/extras.go
+++ b/plugin/clusterextension/extras.go
@@ -63,6 +63,7 @@ type worker struct {
 type controller struct {
 	APIServerFlags      pluginmodel.APIServerFlags
 	APIServerVolumes    pluginmodel.APIServerVolumes
+	ControllerFlags     pluginmodel.ControllerFlags
 	Files               []model.CustomFile
 	SystemdUnits        []model.CustomSystemdUnit
 	IAMPolicyStatements []model.IAMPolicyStatement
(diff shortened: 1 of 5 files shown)


plugin/pluginmodel/config.go
```diff
@@ -188,6 +189,12 @@ type APIServerVolume struct {
 	ReadOnly bool   `yaml:"readOnly,omitempty"`
 }
 
+type KubernetesControllerManager struct {
+	Flags ControllerFlags `yaml:"flags,omitempty"`
+}
+
+type ControllerFlags []CommandLineFlag
```
mumoshu: I'm fine with removing the dedicated type `APIServerFlags` and just use `[]CommandLineFlag` instead for simplicity!

plugin/clusterextension/extras.go
```diff
@@ -226,12 +228,23 @@ func (e ClusterExtension) Controller() (*controller, error) {
...
 						Value: v,
 					}
 					apiServerFlags = append(apiServerFlags, newFlag)
 				}
+				for _, f := range p.Spec.Kubernetes.ControllerManager.Flags {
+					v, err := render.StringFrom(f.Value)
+					if err != nil {
+						return nil, fmt.Errorf("failed to load controller-manager flags: %v", err)
+					}
+					newFlag := pluginmodel.CommandLineFlag{
```
mumoshu: Note: ALl the types defined in the `pluginmodel` pkg has been moved to `pkg/clusterapi` in #1473

PR #1471, opened 2018-10-18
renamed the --admission-control flag to --enable-admission-plugins in cloud-config-controller template

avinashpenmetsa:
I found kube-api-server config still using `--admission-control` flag which was deprecated in v1.10 and was replaced with `--enable-admission-plugins` flag.


avinashpenmetsa:
I signed it

PR #1472, opened 2018-10-18
WIP: Test preparation to allow legacy and plugins to co-exist using models

davidmccormick:
Re-implement control plane feature gates (now applying them to apiserver, controller and scheduler and kube-proxy as set).  The main difference is that they can now be set via plugins.  There has also been a refactor to show how logic can be removed from the cloud-config templating and moved elsewhere to provide backwards compatibility but still allow a migration to moving into plugins).

If model more of the state in the models package then the legacy config can be mapped to it as well as the plugins - then we simplify the render (see in this example how the complexity of cloud-config-controller is being reduced) from the model - removing as much logic as we can (this will eventually all live in core plugins).

@mumoshu - what do you think of these general ideas/direction?

### DIFF
--- a/core/controlplane/cluster/cluster.go
+++ b/core/controlplane/cluster/cluster.go
@@ -147,6 +147,9 @@ func NewCluster(cfgRef *config.Cluster, opts config.StackTemplateOptions, plugin
 		StackConfig: stackConfig,
 	}
 
+	// Map legacy core configurations to plugin compatible models
+	c.MapLegacySettings()
+
 	// Notes:
 	// * `c.StackConfig.CustomSystemdUnits` results in an `ambiguous selector ` error
 	// * `c.Controller.CustomSystemdUnits = controllerUnits` and `c.ClusterRef.Controller.CustomSystemdUnits = controllerUnits` results in modifying invisible/duplicate CustomSystemdSettings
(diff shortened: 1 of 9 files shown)


davidmccormick:
The definitely still needs quite a bit of work - it's just to highlight a possible way of migrating to plugins by abstracting more of the resultant configuration into models and allowing plugins to also interface to them.  There's a horrendously ugly legacy mapping function that is trying to map old style configs to a few test models... we could maybe find a way for individual core pluggins to each do their own mapping?

c-knowles:
I haven't gone into detail if this works but the general theme of it looks good. I like how `APIServerAggregator` has been combined, wondering if all API flags should be combined in the same way. It would hide some rendering details from the template but would centralise the logic alongside other app code, we have a mixture of both right now.

mumoshu:
@c-knowles Hey! Thanks for the comment.

> we have a mixture of both right now

Would you mind elaborating a bit more about this, perhaps with examples extracted from the current code-base?

I'm currently working on a HUGE refactoring in #1473. We'll (ideally) have `pkg/api` for 1:1 golang-representation of `cluster.yaml`, and `pkg/model` for the actual k8s cluster along with userdata and stack templating logics.

That's too huge to be reviewed, but I still want more info to align our direction!

davidmccormick:
Now closing this, it was meant to demonstrate an idea and has served its purpose.

PR #1474, opened 2018-10-22
Allow user control over controller.iam.role.name

davidmccormick:
We would like to control the name of the controller's IAM role without having to move its management outside of `kube-aws`.  Originally when we proposed the `useStrict` feature we were not aware of the powerful `kube-aws` plugins model - now we would like to add our `Kiam` policies through `kube-aws` plugins.  `useStrict` really means `manageExternally` and we would like a new `strictName` which can be `false` by default but if consciously set to `true` takes the name specified in `controller.iam.role.name` without pre-pending ClusterName-AWSRegion to the front of it.

e.g.
Default behavoir: -
```
controller:
  iam:
   role:
     name: myrole1
```
will result in a controller role something like `mycluster-us-west-2-myrole1`

```
controller:
  iam:
   role:
     name: myrole1
     strictName: true
```
will generate a controller role named `myrole1`

### DIFF
--- a/test/integration/maincluster_test.go
+++ b/test/integration/maincluster_test.go
@@ -1744,38 +1744,57 @@ worker:
 			},
 		},
 		{
-			context:    "WithControllerIAMDefaultUseStrict",
+			context:    "WithControllerIAMDefaultManageExternally",
 			configYaml: minimalValidConfigYaml,
 			assertConfig: []ConfigTester{
 				func(c *config.Config, t *testing.T) {
 					expectedValue := false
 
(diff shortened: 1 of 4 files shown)


test/integration/maincluster_test.go
```diff
@@ -1814,7 +1833,7 @@ waitSignal:
 			context: "WithNodePoolWithWaitSignalDisabled",
 			configYaml: minimalValidConfigYaml + `
 worker:
-  nodePools:
+  nodePools:F
```
mumoshu:
```suggestion
  nodePools:
```

davidmccormick:
Is this ok to merge?

Issue #1479, opened 2018-11-05
etcd fails to boot upon upgrading to v0.12.0


c-knowles:
I've just tried to upgrade a dev cluster from kube-aws master/713eeffd with self hosted canal enabled to v0.12.0. With a single etcd node, the cluster fails to get past the etcd stack changes with the logs below. I have not tried on a multi node etcd yet.

The biggest changes between these two versions:
* Removal of non-self hosted, but the cluster I'm upgrading already had self-hosted switched on
* Removal of etcd access from workers to etcd
* Upgrade to k8s 1.11.x from 1.10.x

I'm just digging into why this is occurring and filing here in case anyone else hits a similar problem. Note, there are some error logs from a datadog agent running on this host intermingled in there but suspect those are not related.

```
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal systemd[1]: Starting Exports Kubernetes Values from a remote Etcd cluster...
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_CACERT="/etc/ssl/certs/etcd-trusted-ca.pem"
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_CA_FILE="/etc/ssl/certs/etcd-trusted-ca.pem"
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_CERT="/etc/ssl/certs/etcd-client.pem"
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_CERT_FILE="/etc/ssl/certs/etcd-client.pem"
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_KEY="/etc/ssl/certs/etcd-client-key.pem"
Nov 05 03:11:56 ip-10-0-20-15.eu-west-1.compute.internal etcdadm[10496]: declare -x ETCDCTL_KEY_FILE="/etc/ssl/certs/etcd-client-key.pem"
[...]

PR #1480, opened 2018-11-05
Check stack existence via DescribeStacks

c-knowles:
Rather than ListStacks without any paging.

Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1479


c-knowles:
@mumoshu do you know why I cannot request a review from @davidmccormick? I cannot get his user to appear on the reviewers list.

mumoshu:
@c-knowles Thanks for the fix!

> do you know why I cannot request a review from @davidmccormick? I cannot get his user to appear on the reviewers list.

Just invited @davidmccormick as a collaborator w/ read/write privs to this repo, hoping it makes difference.

c-knowles:
@davidmccormick we could implement that although then it does not mirror the SDK so not sure if it's worth it.

PR #1484, opened 2018-11-07
feat: add CF stackNameOverride in cluster.yaml

koen92:
Kube-aws 0.11.x introduced a separate CloudFormation network stack. AWS VPCs are not movable between two CloudFormation stacks. This means that Kube-aws 0.11.x will create a new VPC when upgrading.

Because we added some extra LoadBalancers and RDS Databases that we cannot easily migrate to another VPC without downtime, creating a new VPC is not desired for us.

Therefore, we decided to keep the network stack in the original control plane stack and instead moved the control plane to a new separate stack. This means that we can stick with the original VPC.

The easiest way to do this with Kube-aws without adding legacy code, is by allowing the user to override the CF stacks names of the control plane, network and etcd stacks in cluster.yaml.

This PR adds an option in cluster.yaml to override the name of the CF controlPlane,
network and etcd stacks.

### DIFF
--- a/core/root/config/templates/cluster.yaml
+++ b/core/root/config/templates/cluster.yaml
@@ -19,6 +19,21 @@ amiId: "{{.AmiId}}"
 # Container Linux has automatic updates https://coreos.com/os/docs/latest/update-strategies.html. This can be a risk in certain situations and this is why is disabled by default and you can enable it by setting this param to false.
 disableContainerLinuxAutomaticUpdates: true
 
+# Override the CloudFormation logical sub-stack names of control plane, etcd and/or network.
+# Changing these names causes CF to delete the old and create a new stack and thus deleting and recreating all components in the stack.
+#
+# From kube-aws 0.11.x and onwards, separate networking and etcd CF stacks have been introduced.
+# Because CF does not support migration of resources between stacks (or renaming stacks), a new VPC will be created if you upgrade an existing 0.10.x cluster.
+# Creating a new VPC and deleting the old one can cause a lot of issues and might not be desired.
+# To make this migration more feasible it is posible to change the name of the network stack to the old controleplane stack to keep the vpc components in the same stack. This way CF will not create a new vpc.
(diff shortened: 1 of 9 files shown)


mumoshu:
Hey! Thanks for submitting this.

> we added some extra LoadBalancers and RDS Databases that we cannot easily migrate

Ahhh... this is something I strongly suggest NOT to do. Sorry If I wasn't very clear about that in the kube-aws docs or related GitHub issues.

Please give me a moment and I'll try to understand what you're going to do. Thanks.

mumoshu:
@koen92 Would you also mind submitting this to `v0.11.x` branch? So that we can cut v0.11.2 with this.

The current `master ` is for `v0.12.x` development.

core/root/config/templates/cluster.yaml
```diff
@@ -19,6 +19,21 @@ amiId: "{{.AmiId}}"
...
+# Changing these names causes CF to delete the old and create a new stack and thus deleting and recreating all components in the stack.
+#
+# From kube-aws 0.11.x and onwards, separate networking and etcd CF stacks have been introduced.
+# Because CF does not support migration of resources between stacks (or renaming stacks), a new VPC will be created if you upgrade an existing 0.10.x cluster.
+# Creating a new VPC and deleting the old one can cause a lot of issues and might not be desired.
+# To make this migration more feasible it is posible to change the name of the network stack to the old controleplane stack to keep the vpc components in the same stack. This way CF will not create a new vpc.
+# In this scenario the name of the networkstack becomes "ControlPlane". (which is the name of the stack which holds the vpc components in a kube-aws pre 0.11 cluster)
+# For clearity it is advised also rename the stack of the controlPlane components to, for example, "masters"
+# Note that this will keep some legacy naming around in your configs and CF stacks...
+#stackNameOverride:
```
mumoshu: Nit but:

`
cloudformation:
  stackNameOverrides:
    ...
```

sounds slightly nicer to me because it expresses you can one or more stack names customized, and it works in term of CloudFormation.

Issue #1491, opened 2018-11-09
Bumpy 10.2 -> 11.x migration


davidmccormick:
We're working on migrating our internal kube-aws clusters from 0.10.x to 0.11.x and looking at ways of reducing the "bumpiness" of it.  Whilst the etcd migration is working as expected we're finding issues between apiservers and nodes which are causing disruption until *all* nodes have finished being rolled during the migration.  On large clusters, this disruption is particularly painful.

The issue is that the new v0.11.x apiservers (a.k.a. the kube-controller-manager) can not connect to legacy nodes before they have been rolled/upgraded, and vice-versa legacy nodes can not connect to the new apiservers.  This is because of the Controller and Worker security groups migrating into the Network stack meaning that the new apiservers and nodes are effectively coming up in new groups.

I would like to propose a couple of small changes in order to relax the security group rules a little during the migration and allow the old and new to co-exist during the migration from 0.10.x to 0.11.x.

1) Put out a 0.10.3 release which relaxes the security group rules on the nodes so that the new 0.11.x apiservers/controller-managers can continue to manage the legacy nodes during the migration.  This pre-roll of 0.10.3 is necessary to put the right access on the legacy nodes before the new version and new groups is rolled in.  It is intended that people only deploy 0.10.3 as a step to migrating to 0.11.x.

2) Make a slight modification to the 0.11.x release to enable the user to set their own cidrs on the `SecurityGroupController` in the network stack using `APIEndpointConfigs.LoadBalancer.APIAccessAllowedSourceCIDRs` without requiring them to set APIEndpointLB.Type to "network".  This will allow users to relax the rules on the apiservers so that the legacy nodes can still talk to them.

PR #1492, opened 2018-11-09
v0.10.x Allow smoother migration to v0.11.x

davidmccormick:
Makes the migration from a v0.10.x cluster to v0.11.x smoother by allowing new kubernetes apiservers to connect to the existing nodes - despite them being in a different security group.  The relaxation of the security rules is intended only as a step to then migrating to the 0.11.x release.

### DIFF
--- a/core/controlplane/config/templates/stack-template.json
+++ b/core/controlplane/config/templates/stack-template.json
@@ -1278,6 +1278,15 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ $ports := splitList "," "5473,8472,10250,4194,10255,9100" -}}
+          {{ range $_, $port := $ports -}}
+          {
+            "CidrIp": "{{ $.VPCCIDR }}",
+            "FromPort": {{ $port }},
+            "IpProtocol": "tcp",
+            "ToPort": {{ $port }}
+          },
+          {{ end -}}
           {
             "CidrIp": "0.0.0.0/0",
             "FromPort": -1,


davidmccormick:
This is part 1 of the resolution for issue https://github.com/kubernetes-incubator/kube-aws/issues/1491

davidmccormick:
@mumoshu @c-knowles are you guys happy for us to merge this into the 0.10.x branch to smooth migrations to 0.11.x?

c-knowles:
Could we inject this only if running between those versions? I know we store the version but don’t think we have any code currently to check directly against that.

davidmccormick:
@c-knowles It has no way of knowing that the migration is going to happen the next run.  Because the 0.10.x branch is now quite old I think it makes sense to put this out a release specifically with helping migration to the 0.11.x branch in mind.

davidmccormick:
@c-knowles @mumoshu can we merge this into the v0.10.x branch?

core/controlplane/config/templates/stack-template.json
```diff
@@ -1278,6 +1278,15 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ $ports := splitList "," "5473,8472,10250,4194,10255,9100" -}}
+          {{ range $_, $port := $ports -}}
+          {
+            "CidrIp": "{{ $.VPCCIDR }}",
+            "FromPort": {{ $port }},
+            "IpProtocol": "tcp",
+            "ToPort": {{ $port }}
```
c-knowles: Can we start to add `Description` to these? It's now supported in CF.

PR #1493, opened 2018-11-09
v0.11.x - Smooth migration from v0.10.x nodes

davidmccormick:
Perform these actions in order to smooth the migration from v0.10.x: -

- Allow same CIDRs to access apiservers as are set on the api load-balancer
- Store and remove mutating/validating webhooks during install-kube-system and restore them again once it has run.
- Allow the new etcd's longer for creation to allow longer for the keys to be migrated from previous etcds to the new ones (3H)
- Roll-in upstream stack naming and stack paging fixes
- Disable go fmt diff check which breaks with newer versions of gofmt.
- Correct IAM max role name length checking when using StrictName


davidmccormick:
This is part 2 of a fix for https://github.com/kubernetes-incubator/kube-aws/issues/1491

mumoshu:
@davidmccormick Hey! Thanks for this. But I'm a bit confused.

This seems to open up access from the specified CIDRs on `controller nodes` by modifying `SecurityGroupController`. But assuming you're using CLB to serve the API endpoint, wouldn't opening up those ports on the CLB enough?

In that case, I think we need no changes to the template because we seems to support it already:

https://github.com/kubernetes-incubator/kube-aws/blob/b30dec67dd7e5b1bdbcf74be5b083d0c20b73d0f/core/network/config/templates/stack-template.json#L7-L38

davidmccormick:
Hi, yes, I wasn't expecting to need it either but found that whilst the kubelet and kube-proxy connect through to the controllers using the load-balancer which works, any pod based access of the api, such as calico, flannel etc. get routed via kube-proxy/iptables directly to the apiservers (and not via the load-balancer).

davidmccormick:
I think I need to do a couple more tests...

davidmccormick:
Hi, this is getting closer to being ready to merge.  I've found that a cluster with existing `mutatingwebhookconfigurations` or `validatingwebhookconfigurations` can cause a **kube-aws** cluster roll to fail.  The following conditions cause the roll to fail: -
  * during a migration where new apiservers are coming up in new groups.
  * you have an existing mutating or validating webhook that targets something that gets written by `install-kube-system`
  * you have modified the labels or annotations on kube-system.

We added an annotation to `kube-system` to enable *Kiam* and did so using a program that targets namespaces with a mutatingwebhook.  During the migration the webhook becomes unavailable and so kubectl apply can not patch the kube-system object and (rightly) refuses to delete and re-create it.  The failure causes `install-kube-system` to fail and go into a loop... until the timeout is elapsed and a roll-back then proceeds.

As well as loosening the rules for connecting to an apiserver, this change also saves and removes all mutating and validating webhooks when `install-kube-system` runs thus preventing failed rolls.

core/etcd/config/templates/stack-template.json
```diff
@@ -348,24 +348,24 @@
           {{$etcdInstance.SubnetRef}}
         ]
       },
-      {{if $.WaitSignal.Enabled}}
+      {{if $.WaitSignal.Enabled -}}
       "CreationPolicy" : {
         "ResourceSignal" : {
           "Count" : "1",
-          "Timeout" : "{{$.Controller.CreateTimeout}}"
+          "Timeout" : "{{if $.EtcdMigrationEnabled}}PT3H{{else}}{{$.Controller.CreateTimeout}}{{end}}"
```
mumoshu: Would you mind sharing me why it takes so much(3 hours)?!

core/controlplane/config/templates/cloud-config-controller
```diff
@@ -968,6 +968,40 @@ write_files:
...
+          if list_not_empty $file; then
+            echo "deleting $type webhooks..."
+            ensuredelete $file
+          fi
+        fi
+      }
+
+      # Disable all mutating and validating webhooks because they can interfere with the stack migration)
+      echo "Storing and removing all validating and mutating webhooks..."
+      save_webhooks validating /srv/kubernetes/validating_webhooks.yaml
```
mumoshu: So do we repeat these migration steps for each controller node being rolled, right?
Just asking for my understanding. Not thinking it is the cause of the slowness.

mumoshu:
Hey. Thanks for your awesome work!

> We added an annotation to kube-system to enable Kiam and did so using a program that targets namespaces with a mutatingwebhook. During the migration the webhook becomes unavailable and so kubectl apply can not patch the kube-system object and (rightly) refuses to delete and re-create it. 

Makes sense. Just for my understanding, I'm thinking that it wouldn't have affected install-kube-system if the mutating webhook was patching resources other than namespaces, right?

core/etcd/config/templates/stack-template.json
davidmccormick: Hi, I haven't had one take that long yet, but I have experienced a small cluster going above the default 15 minutes.  The issue, I think, is with the import (exporting the keys seems performant enough).  The import script is a single threaded shell script that executes within an etcdctl container and I think that writing larger items, such as big configmaps, is slower than smaller keys.  I noticed that a number of helm related config maps were importing quite slowly.  I haven't tried to migrate any of our large cluster yet, but I am concerned that I may need to refactor the import to improve performance.

core/controlplane/config/templates/cloud-config-controller
davidmccormick: Yes, I made it do this on each one in turn because I am worried about race conditions should all three controllers come up at the same time.

davidmccormick:
Still working on this, please do not merge yet!

davidmccormick:
@mumoshu @c-knowles can we squash and merge this?  We are happy that this is working as expected and have been using to migrate our clusters to the 0.11.x branch.  We experienced a number of issues early on and when we came to migrate our first 300+ node cluster (so it contains a number of different little fixes)

davidmccormick:
@mumoshu @c-knowles ok to merge this to the 0.11.x branch?

mumoshu:
@davidmccormick Yeah! Sorry for the delay, but I was able to catch this up just now 😄

PR #1495, opened 2018-11-10
Fix typos: additioanl->additional, creaes->creates

mooncak:
Fix typos: additioanl->additional, creaes->creates


mumoshu (approved):
LGTM. Thanks for the fix ☺️

PR #1497, opened 2018-11-20
IAM Role name length checks need to account for StrictName usage...

davidmccormick:
When you are using StrictName for IAMRole names then you should have the full 64 characters to play with, but `kube-aws` was still checking the length against the generated name.


davidmccormick:
@c-knowles would you be so kind as to take a look at this for me and let me know if it looks ok to you?

PR #1504, opened 2018-11-29
A couple of fixes for nodepool resources and cloudwatch logging

davidmccormick:
fix issue with nodepool names containing a - character.
Properly pass cloudwatch logging arn from root stack into the network stack

### DIFF
--- a/builtin/files/stack-templates/network.json.tmpl
+++ b/builtin/files/stack-templates/network.json.tmpl
@@ -1,8 +1,14 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
   "Description": "kube-aws network stack for {{.ClusterName}}",
+  {{ if .CloudWatchLogging.Enabled -}}
   "Parameters": {
+    "CloudWatchLogGroupARN": {
+      "Type": "String",
+      "Description": "CloudWatch LogGroup to send journald logs to"
+    }
   },
+  {{ end -}}
   "Resources": {
     {{range $i, $apiEndpoint := $.APIEndpoints -}}
     {{if .LoadBalancer.ManageELB -}}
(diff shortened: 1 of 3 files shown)


c-knowles:
@davidmccormick what problem did you have here? Asking since all our node pool names have hyphens in them and not spotted a problem with that so far but perhaps missed something.

davidmccormick:
@mumoshu @c-knowles are you happy to merge this?

c-knowles:
@davidmccormick probably just wondering what the problem was? Please see my comment.

davidmccormick:
The issue I was having is that our nodepool names contain hyphens - so am getting the following error when I try and build a cluster using the `master` branch: -

```
generating assets for control-plane, etcd, network, nodepool-a, nodepool-b, nodepool-c
Error: failed to validate network: invalid cloudformation stack template https://s3.amazonaws.com/034324643013-k8tool-us-west-2/cf/davem-lab-secure/kube-aws/clusters/davem-lab-secure/exported/stacks/network/stack.json: ValidationError: Template format error: Resource name nodepool-aIAMManagedPolicyWorker is non alphanumeric.
	status code: 400, request id: 14a5f3f8-fa39-11e8-80b2-7101c712114a
```

The change makes sure that the nodepool names get sanitized in much the same way we do for controllers and etcd stack names etc.

davidmccormick:
@c-knowles @mumoshu I'm not sure why Chris isn't seeing this but as I see it santizing the name can't hurt and this also fixes an issue with cloudwatch logging which I'm guessing you have disabled.  I'd like to get these PRs merged so we are not running a forked codebase.

c-knowles:
@davidmccormick we have CloudWatch logging turned on and I verified logs are coming through. I think this might only be a problem on latest large master refactor then, we are using v0.12.0 patched with the two PRs for stack existence checks with over 100 stacks in one account so a few commits behind.

As you say it seems like we'd want to sanitise anyway.

PR #1512, opened 2018-12-05
CA self-genereted cert: add the cn flag

### LINKED ISSUE #463: Create unique CN for generated CA certificates
When creating multiple test clusters, the generated self-signed CA certificates can confuse clients like Chrome because they all have the same CN. It also becomes difficult to trust these individual root CA certificates in various client software. This can be reproduced simply by creating two clusters with self-generated certificates and trying to log into the web dashboard of the second cluster will fail with HSTS errors because the CA is invalid (not just untrusted).

If the CNs of the generated CA certificates were unique, client software would have an easier time dealing with them. I would propose putting the given fqdn or some other unique information into the CN for the generated CA certificate.

sonant:
CA self-genereted cert: add the cn flag

all self-signed CA have the samee CN and this can confuse clients like
Chrome. Now we can set different CN

Fixes #463


davidmccormick:
This looks ok to me

davidmccormick:
Thanks very much for your effort on this!

PR #1514, opened 2018-12-14
'AvailabilityZone' nodepool rolling strategy

davidmccormick:
The default 'Parallel' nodepool rolling strategy is a bit dangerous because it can take down nodes across multiple availability zones and so, if you are unlucky and have an application with only a 1, 2 or 3 pods the parallel roll could take down all of your instances of at the same time - which is definitely not desirable!

The 'Sequential' nodepool rolling strategy is a lot safer because it will only roll a single nodepool at a time, but is slow as a result.  When you have multiple different node pools, or large clusters, you may find that your rolls, although safe, take a long time to complete.

The new 'AvailabilityZone' nodepool rolling strategy fits between the two.  It will roll all of nodepools that belong to the same AWS AvailabilityZone at the same time and roll the availability zones in order.  Example: -

```
worker:
  nodePoolRollingStrategy: AvailabilityZone
  nodePools:
    "TestA":
       subnet: a (in availability zone us-west-2a)
    "TestB":
      subnet: b (in availability zone us-west-2b)
    "TestC":
       subnet: a (in availability zone us-west-2a)
    "TestD":
      subnet: b (in availability zone us-west-2b)
```
In config above, "TestA" and "TestC" will roll in parallel with each other first (because they are both availability zone `us-west-2a`), then once they have finished rolling, "TestB" and "TestD" will roll in parallel.

Note: Only nodepools containing subnets in the same aws availability zone can be rolled using this strategy otherwise you will get an error when rendering the root cloud-formation stack.

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -302,12 +302,14 @@ worker:
 #  # Can be overriden per node pool by specifying `worker.nodePools[].apiEndpointName`
 #  apiEndpointName: versionedPublic
 #
-#  # Used to globally specify the strategy in which nodePools will roll out: in sequence or parallel
-#  # Default behaviour ('parallel') will roll out nodePools concurrently
-#  # If set to 'Sequential', the nodePools will roll out sequentially, in order of declaration
-#  # If set at a Worker level, it will only take effect if it is not set at a nodePool level
-#  # For finer grain control, declare it on a nodePool-by-nodePool basis
-#  nodePoolRollingStrategy: Sequential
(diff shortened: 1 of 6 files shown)


davidmccormick:
Rebased this on top of fix for workers not correctly starting when using the experimental TLS boostrap feature.  This issue was preventing me from performing real testing of this AZ rolling change.

davidmccormick:
This is now ready and tested.  I think that we should consider making this the default nodepool rolling strategy as it is less likely to cause outages than 'Parallel' because a it ensures that disruption is localised within in each well-defined failure domain rather than across all of the cluster.

davidmccormick:
Merge after https://github.com/kubernetes-incubator/kube-aws/pull/1522

PR #1520, opened 2018-12-21
The getting started guide should not recommend 2 controllers for HA

kevinhcross:
The doc recommends this:

> At least 3 etcd, 2 controller, 2 worker nodes are required to achieve high availability 

But the current (at the time of writing at least) [docs](https://kubernetes.io/docs/tasks/administer-cluster/highly-available-master/#best-practices-for-replicating-masters-for-ha-clusters) say this:

> Do not use a cluster with two master replicas. Consensus on a two-replica cluster requires both replicas running when changing persistent state. As a result, both replicas are needed and a failure of any replica turns cluster into majority failure state. A two-replica cluster is thus inferior, in terms of HA, to a single replica cluster.

My understanding is that this makes the "2 controller for HA" advice incorrect so it should be changed or removed.

My feeling is that HA Kubernetes controllers is probably out of the scope of a getting started guide so it should just be removed. If you would rather update the advice I think that would also work.

### DIFF
--- a/docs/getting-started/prerequisites.md
+++ b/docs/getting-started/prerequisites.md
@@ -3,7 +3,6 @@
 If you're deploying a cluster with kube-aws:
 
 * [EC2 instances whose types are larger than or equal to `t2.medium` should be chosen for the cluster to work reliably](https://github.com/kubernetes-incubator/kube-aws/issues/138)
-* [At least 3 etcd, 2 controller, 2 worker nodes are required to achieve high availability](https://github.com/kubernetes-incubator/kube-aws/issues/138#issuecomment-266432162)
 
 ## Deploying to an existing VPC
 


kevinhcross:
I did not realise I would need to sign the CLA for a PR. Apologies but I don't have time to look at this now. If someone wants to delete this PR that is fine but I don't want to close it as I think it is something that should be fixed.

mumoshu:
Hey! Thanks for this.

> My understanding is that this makes the "2 controller for HA" advice
incorrect so it should be changed or removed.

Nope. AFAIK, the upsteram doc mixes both control-plane and etcd into the
same set of machies - master, which is confusing.

By design, etcd should be at least 3. control-plane component like
controller-manager is HA for 2 or more nodes because their HA is based on
the master-slave architecture. apiserver can be HA for 2 or more nodes
because they are stateless.

kube-aws intentionally separates etcd into a dedicated set of machines, so
that we can scale etcd and control-plane independently.

2018年12月22日(土) 1:25 Codecov <notifications@github.com>:

kevinhcross:
Looking at this again I think you are correct. I was assuming the master referred to in upstream doc did not include etcd (as suggested in [this upstream doc](https://kubernetes.io/docs/concepts/)) but taking another looking at the HA doc I can see that their architecture does have etcd on the master node.

Apologies for the incorrect PR. Closing it now.

PR #1524, opened 2019-01-09
fix: Partial upgrades with stack name overrides

koen92:
https://github.com/kubernetes-incubator/kube-aws/pull/1487 introduced a
bug when doing a partial stack upgrade with a stack name override.

This is fixed by using the actual stack names in operation targets
instead of the hard coded constants.

### DIFF
--- a/core/root/operation_targets.go
+++ b/core/root/operation_targets.go
@@ -4,17 +4,8 @@ import "strings"
 
 // TODO: Add etcd
 const (
-	OperationTargetControlPlane = "control-plane"
-	OperationTargetEtcd         = "etcd"
-	OperationTargetNetwork      = "network"
-	OperationTargetAll          = "all"
+	OperationTargetAll = "all"
 )
-
-var OperationTargetNames = []string{
-	OperationTargetControlPlane,
-	OperationTargetEtcd,
-	OperationTargetNetwork,
-}
 
 type OperationTargets []string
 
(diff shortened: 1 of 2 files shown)


core/root/operation_targets.go
```diff
@@ -4,27 +4,18 @@ import "strings"
 
 // TODO: Add etcd
```
koen92: Don't know if this TODO comment is still relevant?

mumoshu: Seems irrelevant now. Good catch!

mumoshu:
@koen92 Thanks for your effort! Just to clarify, what was the exact issue? You had to say e.g. `apply --targets network` even what you wanted was updating only the network stack named `control-plane`?

koen92:
@mumoshu Thank you for your time!

When we try to do a partial validate and/or apply to only the Controlplane (or other target that has been renamed via the stackNameOverride option), we get the following error:
```
$ kube-aws-v0.12.0+dirty validate --targets control-plane
Validating UserData and stack template...
generating assets for control-plane
updating template url of control-plane
Error: failed to find assets for stack control-plane: [bug] failed to get the asset for the id "{control-plane stack.json}"
```
This happens because kube-aws checks for a hard coded Controlplane name (see https://github.com/kubernetes-incubator/kube-aws/blob/v0.12.x/core/root/operation_targets.go#L56 and https://github.com/kubernetes-incubator/kube-aws/blob/v0.12.x/core/root/operation_targets.go#L7 ).

With this PR I made this check dynamic and based on the actual Controlplane name (that is either 'control-plane' or the user provided name via cloudFormation.stackNameOverrides.controlPlane in cluster.yaml).

mumoshu:
Ah, makes sense. And I think we have no other way to fix the issue. Good job 👍

Issue #1500, opened 2018-11-23
New AWS autoscaling groups


sander-su:
AWS recently introduced the option of using autoscalinggroups and combine multiple instances.
https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

For us this would be a great feature. 

Is someone already working on this or should we create a PR?

sander-su:
Ok, we got this working in a 0.12 branch.
See: https://github.com/koen92/kube-aws/commit/073e95c4902fef2e6b943467b43295520312bbbe

Nodepool now spins an asg with lanchtemplate and mixedInstance options.

@mumoshu is this something for 0.12 of does it go to master only?
What is the timeline for a 0.13 release?

mumoshu:
@sander-su Hey!

Yes, this should go into the master because other versions are in the maintenance modeI and we usually accept only bug fixes and migration-related features only for the maintenance versions.

I think we have no specific blocker for the v0.13 release now

sander-su:
Ok we'll make a PR for the 0.13 branch.
Give us a few days as 0.13 is new for us and it looks like it has a lot of changes.

sander-su:
Pr merged

PR #1531, opened 2019-01-14
feat: support for AWS mixed instances

koen92:
AWS recently introduced the option of using autoscalinggroups and combine multiple instances.
https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/

This PR changes the CF LaunchConfiguration templates into CF [LaunchTemplates](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html), which will add the ability to combine multiple instance types within an AutoScalingGroup.

Please let me know what you think about this! (since it's quite a refactor in the CF node pool template).

Related issue: https://github.com/kubernetes-incubator/kube-aws/issues/1500

### DIFF
--- a/builtin/files/stack-templates/node-pool.json.tmpl
+++ b/builtin/files/stack-templates/node-pool.json.tmpl
@@ -135,8 +135,10 @@
             {{if .AutoScalingGroup.MixedInstances.SpotAllocationStrategy}}
             "SpotAllocationStrategy" : "{{.AutoScalingGroup.MixedInstances.SpotAllocationStrategy}}",
             {{end}}
-            "SpotInstancePools" : {{.AutoScalingGroup.MixedInstances.SpotInstancePools}},
-            "SpotMaxPrice" : "{{.AutoScalingGroup.MixedInstances.SpotMaxPrice}}"
+            {{if .AutoScalingGroup.MixedInstances.SpotMaxPrice}}
+            "SpotMaxPrice" : "{{.AutoScalingGroup.MixedInstances.SpotMaxPrice}}",
+            {{end}}
+            "SpotInstancePools" : {{.AutoScalingGroup.MixedInstances.SpotInstancePools}}
           },
           "LaunchTemplate" : {
             "LaunchTemplateSpecification" : {
--- a/builtin/files/stack-templates/node-pool.json.tmpl
+++ b/builtin/files/stack-templates/node-pool.json.tmpl
@@ -113,12 +113,10 @@
 {{end}}
 {{define "AutoScaling"}}
     "{{.LogicalName}}": {
+      "DependsOn": "{{.LaunchTemplateLogicalName}}",
       "Properties": {
         "HealthCheckGracePeriod": 600,
         "HealthCheckType": "EC2",
-        "LaunchConfigurationName": {
-          "Ref": "{{.LaunchConfigurationLogicalName}}"
-        },
         "MaxSize": "{{.MaxCount}}",
         "MetricsCollection": [
           {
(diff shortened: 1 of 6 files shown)


builtin/files/stack-templates/node-pool.json.tmpl
```diff
@@ -113,19 +113,52 @@
...
+            {{if .AutoScalingGroup.MixedInstances.OnDemandAllocationStrategy}}
+            "OnDemandAllocationStrategy" : "{{.AutoScalingGroup.MixedInstances.OnDemandAllocationStrategy}}",
+            {{end}}
+            "OnDemandBaseCapacity" : {{.AutoScalingGroup.MixedInstances.OnDemandBaseCapacity}},
+            "OnDemandPercentageAboveBaseCapacity" : {{.AutoScalingGroup.MixedInstances.OnDemandPercentageAboveBaseCapacity}},
+            {{if .AutoScalingGroup.MixedInstances.SpotAllocationStrategy}}
+            "SpotAllocationStrategy" : "{{.AutoScalingGroup.MixedInstances.SpotAllocationStrategy}}",
+            {{end}}
+            "SpotInstancePools" : {{.AutoScalingGroup.MixedInstances.SpotInstancePools}},
+            "SpotMaxPrice" : "{{.AutoScalingGroup.MixedInstances.SpotMaxPrice}}"
```
paalkr: SpotMaxPrice should be possible to leave out. If you are excluding SpotMaxPrice in the policy AWS will automatically set the on demand price as max price. This is especially important when having an override list of mixed instance types and sizes.

paalkr:
Please have a look at my comment
https://github.com/kubernetes-incubator/kube-aws/pull/1531/files#r252702059

davidmccormick:
What will happen when this change is rolled into existing clusters?

paalkr:
For what I understand a Launch Configuration will be replaced by a Launch template, and if you don't specify .AutoScalingGroup.MixedInstances.Enabled or  .AutoScalingGroup.MixedInstances at all the, AutoScaling group should just do a rolling update of all nodes.

The author @koen92 might have additional insight.

builtin/files/stack-templates/node-pool.json.tmpl
paalkr: I suggest that you add
            {{if .AutoScalingGroup.MixedInstances.SpotMaxPrice}}
            "SpotMaxPrice" : "{{.AutoScalingGroup.MixedInstances.SpotMaxPrice}}"
            {{end}}

sander-su:
> 
> 
> What will happen when this change is rolled into existing clusters?

As the co-author of this pr; 
The launchconfiguration for ALL worker autoscaling groups (only workers, not masters / etcd) will be replaced by a launchtemplate.
I thought it will do a rolling update of the nodes in the asg but I will retest this and post the result to be 100% sure. (could also be the replace of the asg)

@paalkr Good one, we will change this.

koen92:
@paalkr Fixed it!

paalkr:
Thx @koen92 

According to the CloudFormation AutoScalingGroup documentation, changing either the LaunchTemplate or the LaunchConfiguration requires no interruption.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-launchtemplate

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-launchconfigurationname

paalkr:
Btw, I have tested using a Mixed Instance Policy with 100% spot resources, and an instance distribution and instance type mix (overrides), that simulates my currently running worker pool powered by a regular spotfleet (40+ nodes). When changing something in the LaunchTemplate that will issue a rolling update (like AMI ID or user data), the AutoScaling group nicely performs a rolling update of the spot resources - one by one. This actually solves a huge nightmare of how AWS spot fleet currently are handled.  Where a change to a spot fleet request, replaces all spot instances in a one shot operation.

I can't see any situation where the old spot fleets are preferable over this new mixed instance policy with overrides.

sander-su:
Just confirmed that cloudformation is doing a nice rolling update moving from launchconfiguration to launchtemplate. Seeing is believing, always nice if the docs are right ;).

@davidmccormick can you also approve this pr? have not seen a reaction @mumoshu recently.

davidmccormick:
Thank you for your contribution - good work! 🎉

PR #1536, opened 2019-01-15
Master: Bump default Kubernetes Dashboard version and add AllowSkipLogin option

kylehodgetts:
We ran into some issues with the dashboard being slow at version v1.8.3 and these seem to have gone away in v1.10.1

However in v1.10.1 the skip login option is now disabled by default and requires the use of `--enable-skip-login` flag to bring it back. This has been added in as an option `AllowSkip` in the `KubernetesDashboard` struct and takes a default value of `false`


davidmccormick:
Thanks for you contribution!

PR #1541, opened 2019-01-17
Kube aws pr

omar-nahhas:
Adding the ability to add set arbitrary feature gates to controller components.


pkg/api/cluster.go
```diff
@@ -884,6 +884,24 @@ func (c *Cluster) AvailabilityZones() []string {
 	return result
 }
 
+func (c *Cluster) ControllerFeatureGates() FeatureGates {
```
davidmccormick: This is great work, thanks, but I wonder if it wouldn't be better to follow kubernetes here and also set these on by default too?  We could offer functionality to allow a user to choose to disable them.   Would you mind updating to do this?

builtin/files/userdata/cloud-config-controller
```diff
@@ -365,8 +365,8 @@ coreos:
         {{ else }}--cluster-dns={{.DNSServiceIP}} \
         {{ end }}--cluster-domain=cluster.local \
         --cloud-provider=aws \
-        {{if or (.Experimental.Admission.Priority.Enabled) (.Experimental.Admission.PersistentVolumeClaimResize.Enabled) -}}
-        --feature-gates=PodPriority={{.Experimental.Admission.Priority.Enabled}},ExpandPersistentVolumes={{.Experimental.Admission.PersistentVolumeClaimResize.Enabled}} \
+        {{if .ControllerFeatureGates.Enabled -}}
+        --feature-gates={{.ControllerFeatureGates.String}} \
```
davidmccormick: I love this removing complexity from the templating into the models - this simplifies this layer and makes plugin integration easier!

builtin/files/userdata/cloud-config-controller
```diff
@@ -3342,9 +3345,9 @@ write_files:
           - scheduler
           - --kubeconfig=/etc/kubernetes/kubeconfig/kube-scheduler.yaml
           - --leader-elect=true
-         {{if or (.Experimental.Admission.Priority.Enabled) (.Experimental.Admission.PersistentVolumeClaimResize.Enabled)}}
-          - --feature-gates=PodPriority={{.Experimental.Admission.Priority.Enabled}},ExpandPersistentVolumes={{.Experimental.Admission.PersistentVolumeClaimResize.Enabled}}
-         {{end}}
+          {{- if .ControllerFeatureGates.Enabled }}
+          - --feature-gates={{.ControllerFeatureGates.String}}
+          {{- end }}
```
davidmccormick: This practice of setting the same feature gates on all core components has been adopted by the kubernetes CI so it is good that we are doing the same.

pkg/model/node_pool_config.go
```diff
@@ -75,6 +75,19 @@ func (c NodePoolConfig) FeatureGates() api.FeatureGates {
 	if gates == nil {
 		gates = api.FeatureGates{}
 	}
+	//From kube 1.11 PodPriority and ExpandPersistentVolumes have become enabled by default,
```
davidmccormick: Again, can we instead follow kubernetes lead and also enable them but allow us to disable them if desired through our configs?

pkg/api/cluster.go
omar-nahhas: Done

builtin/files/userdata/cloud-config-controller
omar-nahhas: Done

davidmccormick:
Sorry @omar-nahhas, I noticed that the tests use `enabled` and `disabled` as values for configuring the feature gates.  Please could you change it to either `"true"` or `"false"` so that the value corresponds to the command line argument expected by kubernetes (which uses the boolean values)?  Otherwise, looks good and looking forward to this change! :)

PR #1544, opened 2019-01-18
Porting "Adding the ability to add feature gates to controller components" to 0.12.X

omar-nahhas:
Adding the ability to add feature gates to controller components.
//From Kube 1.11 PodPriority and ExpandPersistentVolumes have become enabled by default,
//so making sure it is not allowed if User has explicitly set them to false
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#changelog-since-v1110
This PR also ensure PodPriority is disabled if users specified so in the experimental features.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -353,6 +353,20 @@ func (c *Cluster) Load() error {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


omar-nahhas:
//assing @davidmccormick

davidmccormick:
Thanks for your contribution Omar!

PR #1548, opened 2019-01-25
feat: Option to provide existing private keys.

dominicgunn:
## Changes

Allow users to provide their own private keys when generating credentials through `kube-aws render credentials`.

This patch was paramount for my team to be able to upgrade from kube-aws 10.1 to kube-aws 12.2. It allowed us to use our existing clusters certificates and upgrade in a smooth fashion.

PR #1549, opened 2019-02-07
feat: Allow users to provide existing private keys

DominicGunn:
## Changes

Allow users to provide their own private keys when generating credentials through `kube-aws render credentials`.

This patch was paramount for my team to be able to upgrade from kube-aws 10.1 to kube-aws 12.2. It allowed us to use our existing clusters certificates and upgrade in a smooth fashion. 

Making a PR to ensure functionality is retained in future releases, previous PR [here](https://github.com/kubernetes-incubator/kube-aws/pull/1548)


davidmccormick:
Many thanks for your contribution! 🙏

PR #1551, opened 2019-02-08
Fixing adding cf resources to etcd and networking stack v.0.11.x

omar-nahhas:

This PR aims to fix both issues:


omar-nahhas:
Here are some results from some manual testing I did:

kube-aws-cluster

etcd
The entry at the button "Prometheus0" is some resource I added to the etcd stack through a plugin.

davidmccormick:
Thanks Omar!  It's always great to see some proper testing! :) 👍

PR #1552, opened 2019-02-08
Fixing adding cf resources to etcd and networking stack v.0.12.x

omar-nahhas:

This PR aims to fix both issues:


omar-nahhas:
Here some manual testing results with this branch:

In this test, I provision an external Prometheus server, using cloud formation inside a plugin. The plugin adds cloud formation to the network stack, which is run before controller and node pools stacks.

network_stack

network_stack_12 x

Proof of the cluster working:
kube_nodes

davidmccormick:
Thanks for your contribution and appreciate your testing! 🙏

PR #1553, opened 2019-02-08
Fixing adding cf resources to etcd and networking stack - master

omar-nahhas:


davidmccormick:
Thanks very much for your contribution! 🙏

omar-nahhas:
Manually tested, an upgrade from 0.12 to 0.13 (with plugins migration), including a plugin that put come cf inside the network stack.

test_13

test_2

PR #1554, opened 2019-02-11
Make cluster-autoscaler resource request/limit configurable

Luke-Humphreys:
## What
We've exposed the cluster-autoscaler deployment resources and made them configurable via cluster.yaml using the same approach taken regarding the kube-dashboard

## Why
We've started to experience OOM issues & CPU throttling in our bigger environments.


davidmccormick:
Any joy with those tests?

Luke-Humphreys:
> Any joy with those tests?

@davidmccormick  It's getting there, I've found the root cause, working on the solution now

davidmccormick:
Thank you for your contribution! 🙏

PR #1555, opened 2019-02-12
Revert etcd upgrade

omar-nahhas:
Etcd was upgraded from version 3.2.13 to version 3.3.10 on this PR https://github.com/kubernetes-incubator/kube-aws/commit/e8b7380d95c30239f252eb7ebad7586dd03b0716.

After testing an upgrade from a cluster created with kube-aws 0.12.X to kube-aws master, we have found that, if there is a problem during the upgrade, cloud formation is not able to rollback cleanly, and etcd cluster is left on an unavailable state due to being unable to downgrade etcd.

This is PR Is to revert the etcd upgrade version back to 3.2.13 until a two ways migration is implemented.

kube_aws_etcd_rollback_failed

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -734,6 +734,8 @@ worker:
 #      arn: "arn:aws:iam::YOURACCOUNTID:instance-profile/INSTANCEPROFILENAME"
 #
 #  # The version of etcd to be used. Set to e.g. "3.2.1" to use etcd3
+#  # CAUTION: Increasing etcd version will permanently change the state of etcd and a cloud formation rollback, will not succeed due to etcd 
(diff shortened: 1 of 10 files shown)


davidmccormick:
@omar-nahhas thanks for your testing efforts and work identifying this issue!  I agree, let's keep etcd on a stable version until we can implement a smoother path forward with upgrades.

davidmccormick:
Many thanks for your investigation and fix! 🙏

omar-nahhas:
Here are some results of manual testing I have done:
1 - create the cluster using kube 0.12
test_step_1_version_12

2 - try upgrade to kube 0.13, cause a rollback, and observe a successful rollback.
test_step_3_rollback_success

3 - cluster is healthy after rollback 
cluster_healthy

PR #1562, opened 2019-02-27
Init command refactor

kylehodgetts:


cmd/init.go
```diff
@@ -80,3 +111,29 @@ Next steps:
 	logger.Infof(successMsg, configPath, configPath)
 	return nil
 }
+
+func createClusterConfigFromTemplate(outputFilePath string, templateOpts interface{}, fileTemplate string) error {
```
mumoshu: My intention to keep this within the `filegen` package was to make some momentum to write unit tests for it. We don't usually write tests for anything under the `cmd` package in the golang world. Am I right?

kylehodgetts: I would disagree with this approach for 2 reasons.

One would be for scope. Being that this method is only used in a single place AND it's `cluster.yaml` specific, we can and should restrict its scope as much as possible.

The second reason is that there's nothing stopping us writing tests for the `cmd` package. We shouldn't be testing the commands themselves as that could result in us testing the library we use for commands, in our case Cobra. But nothing stops us from having a test file that tests the methods that make up the command if it's declared in the `cmd` package.

I guess the compromise would be to keep the `filegen` package but make the method not `cluster.yaml` specific and then call the method from the init command with `cluster.yaml` as a parameter.

LMK what you think @mumoshu

mumoshu: > 
this method is only used in a single place AND it's `cluster.yaml` specific, we can and should restrict its scope as much as possible.

That's a good point!

> 
But nothing stops us from having a test file that tests the methods that make up the command if it's declared in the `cmd` package.

True.

Also - I just noticed I had hard-coded `"cluster.yaml"` in the original `filegen/filegen.go`. What a non-sense!

> 
I guess the compromise would be to keep the `filegen` package but make the method not `cluster.yaml` specific and then call the method from the init command with `cluster.yaml` as a parameter

But it is still used only from `init.go`, right? Then I'm fine with your current work.

kylehodgetts: @mumoshu Let me know what you think of my latest change. I've made the templating more generic and seperating the concerns of creating a file and rendering the template into different layers

davidmccormick:
Hi @kylehodgetts - thanks very much for this PR.  Could you take a moment to give us some details in the PR that indicate your rationale for the change and your intended outcome?  It will help keep a record of the change without having to re-review the code.

kylehodgetts:
The rationale behind this PR is to start refactoring the code to be more readable and easier for new comers to the project to get their heads around. 

I started with the init command as a kind of PoC to show what I was going for.

The things I was trying to achieve were:
- removing methods that are not called anywhere - they just add confusion.
- For the structs / methods that are used in only one place, move them to the package where they are used to restrict scope.
- The CreateFileFromTemplate function was cluster.yaml specific. Made it more generic and changed name to `WriteTemplateWithOptions` and passed cluster.yaml information to it as arguments.

All of this has made the init command easier to follow, at least in my opinion. Let me know what you think

PR #1564, opened 2019-02-28
Allowing adding arbitrary flags to kubernetes core components through plugins

omar-nahhas:
This PR Is to allow adding specific configuration flags to the kube-proxy, kubelet and kube-scheduler through plugins, it is an extension of the currently existing functionality to add fags to the kubeapi and kube-controller-manager.

In the plugin.yaml you can add flags like follows:

kube-componets-flags

Then they will be injected on the configuration of the specific components:
kube-scheduler
kubelet
generated_kube_proxy_conf

### DIFF
--- a/pkg/api/kubernetes.go
+++ b/pkg/api/kubernetes.go
@@ -5,14 +5,22 @@ type Kubernetes struct {
 	EncryptionAtRest  EncryptionAtRest         `yaml:"encryptionAtRest"`
 	Networking        Networking               `yaml:"networking,omitempty"`
 	ControllerManager ControllerManager        `yaml:"controllerManager,omitempty"`
+	KubeScheduler     KubeScheduler            `yaml:"kubeScheduler,omitempty"`
+	KubeProxy         KubeProxy                `yaml:"kubeProxy,omitempty"`
+	Kubelet           Kubelet                  `yaml:"kubelet,omitempty"`
+	APIServer         KubernetesAPIServer      `yaml:"apiserver,omitempty"`
 
-	APIServer KubernetesAPIServer `yaml:"apiserver,omitempty"`
 	// Manifests is a list of manifests to be installed to the cluster.
 	// Note that the list is sorted by their names by kube-aws so that it won't result in unnecessarily node replacements.
 	Manifests KubernetesManifests `yaml:"manifests,omitempty"`
 }
 
 type ControllerManager struct {
+	ComputeResources ComputeResources `yaml:"resources,omitempty"`
+	Flags            CommandLineFlags `yaml:"flags,omitempty"`
+}
+
+type KubeScheduler struct {
 	ComputeResources ComputeResources `yaml:"resources,omitempty"`
 	Flags            CommandLineFlags `yaml:"flags,omitempty"`
 }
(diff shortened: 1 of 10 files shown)


omar-nahhas:
I have tested this by adding the flag metricsBindAddress 0.0.0.0 to the kube-proxy (through a plugin) and then spun the cluster. The cluster spun successfully and my external Prometheus could finally scrape all the kube-proxy(ies) 

version

cluster_spinned_up
kube_proxy

cluster_running

kube_proxy

pkg/api/kubernetes.go
```diff
@@ -5,8 +5,11 @@ type Kubernetes struct {
 	EncryptionAtRest  EncryptionAtRest         `yaml:"encryptionAtRest"`
 	Networking        Networking               `yaml:"networking,omitempty"`
 	ControllerManager ControllerManager        `yaml:"controllerManager,omitempty"`
+	KubeScheduler     KubeScheduler            `yaml:"kubeScheduler,omitempty"`
+	KubeProxy         KubeProxy                `yaml:"kubeProxy,omitempty"`
+	KubeLet           KubeLet                  `yaml:"kubeLet,omitempty"`
```
mumoshu: I think this should be `Kubelet           Kubelet                  `yaml:"kubelet,omitempty"`` given it isn't called `kube-let` in kebab-case.

pkg/model/config.go
```diff
@@ -28,9 +29,12 @@ type Config struct {
...
 
-	APIServerVolumes api.APIServerVolumes
-	APIServerFlags   api.CommandLineFlags
-	ControllerFlags  api.CommandLineFlags
+	APIServerVolumes   api.APIServerVolumes
+	APIServerFlags     api.CommandLineFlags
+	ControllerFlags    api.CommandLineFlags
+	KubeSchedulerFlags api.CommandLineFlags
+	KubeProxyFlags     api.CommandLineFlags
+	KubeLetFlags       api.CommandLineFlags
```
mumoshu: Nit but this should be `KubeletFlags`

test/integration/plugin_test.go
```diff
@@ -143,6 +143,22 @@ spec:
...
+          - name: "healthz-bind-address"
+            value: "0.0.0.0"
+      kubeScheduler:
+        flags:
+          - name: "secure-port"
+            value: "11259"
+      kubeProxy:
+        flags:
+          - name: "metricsBindAddress"
+            value: "0.0.0.0"
```
mumoshu: I believe this should be `config` rather than `flags` given it isn't actually command-line flags:

```yaml
kubeProxy:
  config:
    metricsBindAddress: "0.0.0.0"
```

The golang type of `config` would probably be `map[string]interface{}`

omar-nahhas: I agree it could be confusing.

omar-nahhas:
@mumoshu I have refactored it and test it again by spinning a cluster. 

rolled_ok2
rolled_ok

davidmccormick:
Thanks @omar-nahhas for your continued support! 🙏

paalkr:
I'm not sure if I understand how to include a plugin that will allow me to configure the kube-controller-manager flags. My goal is to add this flag
--horizontal-pod-autoscaler-downscale-stabilization-window=10m

Any guidance or examples would be highly appreciated.

PR #1566, opened 2019-02-28
Capture failures in decrypting assets

sgolightly:
"set -e" will capture when a line of code fails.  However, for a
pipeline such as "false | true", it will only handle the exit code
of the last command. In this case, the exit code will always be 0.

"set -o pipefail" will ensure that any failure in the pipeline results
in the whole line failing, which will then be handled correctly by
"set -e"

For example:
$ set +o pipefail ; false | true ; echo $?
0

$ set -o pipefail ; false | true ; echo $?
1

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -2657,6 +2657,7 @@ write_files:
           -ec \
           'echo decrypting assets
            shopt -s nullglob
+           set -o pipefail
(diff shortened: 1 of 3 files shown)


sgolightly:
Contributor License Agreement (CLA) signed.

mumoshu:
@sgolightly Hey! Thanks for the PR.

Just checked the pr status and saw that you might have missed a correct e-mail in the author info in your commit:
image

image

sgolightly:
I'm going to close this PR and submit a fresh one, which hopefully should fix the breaking CLA test (as well as ensuring the email address is correct).

PR #1569, opened 2019-03-04
Capture failures in decrypting assets

sgolightly:
"set -e" will capture when a line of code fails.  However, for a
pipeline such as "false | true", it will only handle the exit code
of the last command. In this case, the exit code will always be 0.

"set -o pipefail" will ensure that any failure in the pipeline results
in the whole line failing, which will then be handled correctly by
"set -e"

For example:
$ set +o pipefail ; false | true ; echo $?
0

$ set -o pipefail ; false | true ; echo $?
1

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -2657,6 +2657,7 @@ write_files:
           -ec \
           'echo decrypting assets
            shopt -s nullglob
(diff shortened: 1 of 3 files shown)


sgolightly:
I signed it

mumoshu:
@sgolightly Would you mind submitting this against `v0.12.x` and `master` as well?

PR #1573, opened 2019-03-07
BUGFIX: kubeletOptions not applied on controller nodes.

omar-nahhas:
kubeletOptions can be set at the cluster level, then they are injected on a file /etc/default/kubelet that is referenced on the kubelet systemd unit. Although the file gets created in the controller nodes, it is never referenced on the systemd unit. Hence kubeletOptions does not really work on Controllers.


davidmccormick:
Thank you very much for your continued support! 🙏

PR #1575, opened 2019-03-26
Allow plugins to add CFN tags and outputs

Luke-Humphreys:
## What
Adding the ability for plugins to inject cloud formation tags into control plane and node pool stacks.
Adding the ability for plugins to inject output cloud formation into control plane, node pool, etcd and network stacks.

## Why
In the process of developing a plugin for the cluster-autoscaler I noticed that you can only inject resources into cloud formation stacks from plugins, this meant I couldn't add the necessary ASG tags for the cluster-autoscaler. The functionality was already there, I just extended it a little bit.


dominicgunn (approved):

davidmccormick:
Thanks @Luke-Humphreys for this useful expansion of our plugin functionality! 🙏

PR #1576, opened 2019-03-26
Cluster-autoscaler plugin

Luke-Humphreys:
## What
A plugin for the cluster-autoscaler to allow more configuration.

## Why
We needed to change the configuration of the autoscaler and thought it would be good to contribute a plugin back to the community. 

## Depends on
https://github.com/kubernetes-incubator/kube-aws/pull/1575


dominicgunn (approved):
This feels useful, let's try and get it in!

davidmccormick:
Thanks @Luke-Humphreys!  It's great to see this functionality move into a plugin! 🎉 

Given that the legacy non-plugin code is still active could you introduce some logic so that the two implementations do not clash with each other.  We could add a warning on the old configuration that alerts the user that the legacy configuration is deprecated and will be removed in the next major release?  If that proves complicated then perhaps you can disable the legacy code path when the plugin is enabled?

PR #1585, opened 2019-05-01
Enable detailed monitoring

paalkr:
Launch Templates introduced in https://github.com/kubernetes-incubator/kube-aws/pull/1531 did not enable monitoring. This PR adds detailed monitoring, so that worker nodes will behave the same way they did prior to kube-aws 0.13


davidmccormick:
Many thanks for your contribution! 🙏

PR #1586, opened 2019-05-02
Custom APIServer SANs

DominicGunn:
## Changes

- Adds the ability to define custom SANs to the APIServer. Internally we found this incredibly useful when trying to roll changes to CIDRs inside of the cluster.


davidmccormick:
Thanks, this looks useful! 🙏

PR #1590, opened 2019-05-04
0.14 release prep

davidmccormick:
Take the release prep and update v1.14.1


davidmccormick:
> Just for you to know. I have been playing a bit with this and I've been able to provision a 1.14.1 cluster. The only thing that I tried that didn't work out-of-the-box has been metric-server.
> The problem is that it is deployed in workers but port 10250 is only open for a SG assigned to controllers only so I had to move it to controllers to be able to scrap every other node.

Hey @zonzamas, thanks for testing!  I just spun up a cluster and the server is connecting ok but it is not validating the kubelets self-signed certificate.  I'm going to try adding the flag `--kubelet-insecure-tls` and see if that resolves things.

zonzamas:
Ops, yes sorry I forgot I had to add that flag as well. After playing a little more I've seen other stuff:

- Missing CRD in thypa: **blockaffinities.crd.projectcalico.org** and **BGPPeers.crd.projectcalico.org**
- Missing labels in workers/controllers that used to be there in previous versions: **node-role.kubernetes.io/master** and **node-role.kubernetes.io/worker**. Im not sure if this is a real issue but for instance, without those labels 'kubectl get nodes' will not show the node role and sonobuoy test tool also have problems waiting for all workers to be ready.
-  **--kubelet-insecure-tls** worked for me, but, since I run metrics-server in a worker I algo need to manually open the 10250 port in worker SG

davidmccormick:
> * blockaffinities

Hi @zonzamas, thanks for these, I've been looking for the missing calico CRDs but not finding them on the canal install instructions for 3.7 https://github.com/projectcalico/calico/tree/master/v3.7 and https://docs.projectcalico.org/v3.7/manifests/canal.yaml.  I think that maybe the objects that you are relating to do not exist when using flannel as the backend?

davidmccormick:
> * Missing labels in workers/controllers that used to be there in previous versions: **node-role.kubernetes.io/master** and **node-role.kubernetes.io/worker**. Im not sure if this is a real issue but for instance, without those labels 'kubectl get nodes' will not show the node role and sonobuoy test tool also have problems waiting for all workers to be ready.

Yes, I had noticed this change - essentially the NodeRestriction access controller does not allow those labels - that's why I transitioned them over the 0.13 and 0.14 releases to node.kubernetes.io/role.  I appreciate that it means that the ROLE no longer appears when doing kubectl get nodes - which is an annoyance.  I could potentially ditch my attempt at enabling the NodeRestriction access controller and keep all the labels or is it liveable with?

davidmccormick:
> * **--kubelet-insecure-tls** worked for me, but, since I run metrics-server in a worker I algo need to manually open the 10250 port in worker SG

Hi @zonzamas, I just added a commit to make sure that 10250 is open worker->controller and worker->worker

zonzamas:
> > * blockaffinities
> 
> Hi @zonzamas, thanks for these, I've been looking for the missing calico CRDs but not finding them on the canal install instructions for 3.7 https://github.com/projectcalico/calico/tree/master/v3.7 and https://docs.projectcalico.org/v3.7/manifests/canal.yaml. I think that maybe the objects that you are relating to do not exist when using flannel as the backend?

To be honest Im no expert in calico/canal/typha. I just saw some annoying logs in typha pods complaining about missing api objects and googling a bit could relate it to those missing CRDs. Whether they are required or not for properly running typha I don't know.

davidmccormick:
Hi Oscar! @zonzamas
Thanks very much for your help.  I'm testing a commit to add those Calico/Typha objects to be safe (I wasn't convinced that Typha was happy).

I'll try to see if I can make the labels work and potentionally remove the noderestriction admission controller if they don't.  Then I think we should be in good shape to put out these releases as a beta.

PR #1593, opened 2019-05-16
Fixing broken apiserver feature gate logic for 1.9.3

Luke-Humphreys:
## What
A small fix in the logic for featuregate flags, if you are running kube < 1.11 these feature gates need to be explicitly set to true. The existing logic was to always set them to false.

## Why
When the flags are set in the cluster.yaml they were not being applied in the cluster.

### DIFF
--- a/core/controlplane/config/config.go
+++ b/core/controlplane/config/config.go
@@ -365,8 +365,14 @@ func (c *Cluster) ControllerFeatureGates() model.FeatureGates {
 	if !c.Experimental.Admission.Priority.Enabled {
 		gates["PodPriority"] = "false"
 	}
+	if c.Experimental.Admission.Priority.Enabled {
+		gates["PodPriority"] = "true"
+	}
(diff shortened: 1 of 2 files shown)


core/controlplane/config/config.go
```diff
@@ -365,9 +365,15 @@ func (c *Cluster) ControllerFeatureGates() model.FeatureGates {
 	if !c.Experimental.Admission.Priority.Enabled {
 		gates["PodPriority"] = "false"
 	}
+	if c.Experimental.Admission.Priority.Enabled {
```
kevtaylor: I think these can all be changed to this form:
`gates["PodPriority"] = c.Experimental.Admission.Priority.Enabled`

Luke-Humphreys: @kevtaylor Good spot, fixed now

davidmccormick:
Thanks for your continued contributions Luke! 🙏

Issue #1063, opened 2017-12-07
NAT gateway tagging


mumoshu:
The CloudFormation release of September 27, 2017 adds the support for tagging "AWS::EC2::NatGateway".
For consistency with other AWS resources managed by kube-aws, NAT gateways should also be tagged with `kubernetes.io/cluster/{{.ClusterName}}`.

Ref: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html

vsamidurai:
issue has been fixed in #1596

PR #1596, opened 2019-05-19
NAT Gateway tagging #1063

vsamidurai:
adding cloud formation template NAT Gateway tags #1063


davidmccormick:
Hi @vsamidurai! Many thanks for this contribution! 🙏

Issue #1595, opened 2019-05-17
Invalid cloudformation stack template in 0.14.0-rc1


cw-sakamoto:
I'm try to create cluster with 0.14.0-rc1, failed to validate control plane.

```
Unresolved resource dependencies [Test1aSubnet, Test1dSubnet, Test1cSubnet] in the Resources block of the template
```

I confirmed stack-template:
0.14.0-rc1
```
            "VPCZoneIdentifier":[
               {
                  "Ref":"Test1aSubnet"
               },
               {
                  "Ref":"Test1cSubnet"
               },
               {
                  "Ref":"Test1dSubnet"
               }
```
0.12.3
```
            "VPCZoneIdentifier":[
               {
                  "Fn::ImportValue":{
                     "Fn::Sub":"${NetworkStackName}-Test1aSubnet"
                  }
               },
               {
                  "Fn::ImportValue":{
                     "Fn::Sub":"${NetworkStackName}-Test1cSubnet"
                  }
               }
            ],
```

cluster.yaml:
```
subnets:
  - name: test-1a-subnet
    availabilityZone: ap-northeast-1a
    instanceCIDR: "10.0.5.0/25"

  - name: test-1c-subnet
    availabilityZone: ap-northeast-1c
    instanceCIDR: "10.0.5.128/25"

  - name: test-1d-subnet
    availabilityZone: ap-northeast-1d
    instanceCIDR: "10.0.6.0/25"

```

Have you changed the writing method for subnets in cluster.yaml?

davidmccormick:
ok, was wondering why I'm not seeing that issue and for me it be because I am deploying into existing subnets: -

```
subnets:
  - name: a
    availabilityZone: us-west-2a
    private: true
    id: subnet-xxx
  - name: b
    availabilityZone: us-west-2b
    private: true
    id: subnet-xxx
  - name: c
    availabilityZone: us-west-2c
    private: true
    id: subnet-xxx
```
So the issue here must be in the behavoir when creating the subnets...

davidmccormick:
Judging from your output it looks as though the subnet reference is being supplied by the logicalNameProvider function and not from the stackoutput as previously indicated.

```
func (i Identifier) Ref(logicalNameProvider func() string) string {
	if i.IDFromStackOutput != "" {
		return fmt.Sprintf(`{ "Fn::ImportValue" : %q }`, i.IDFromStackOutput)
	} else if i.ID != "" {
[...]

PR #1598, opened 2019-05-21
Issue when kube-aws manages/creates subnets

davidmccormick:
Subnets in the controller stack not being referenced against the network stack when being managed by kube-aws.  Map subnets back to the looked up set by name.  Various additional log messages to debug level.  Remove some helper functions on tempCtx types by calling the correct path in the stack templates.


davidmccormick:
Hopefully resolves https://github.com/kubernetes-incubator/kube-aws/issues/1595 which looks like it might be a long standing issue from the Netowork stack refactor.

PR #1599, opened 2019-05-21
Allow setting Kiam version <3.0

davidmccormick:
The Kiam server and agent command line flags changed in v3.0.
Although we install Kiam v3.2 by default - allow users to deploy older versions with the different command line options.

Tested deploying v2.7 and then upgrading to v3.2

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1467,14 +1467,14 @@ experimental:
   # This is intended to be used in combination with .controller.iam.role.name. See #297 for more information.
   kiamSupport:
     enabled: false
-    image:
-      repo: quay.io/uswitch/kiam
-      tag: v2.8
-      rktPullDocker: false
-    sessionDuration: 15m
-    serverAddresses:
-      serverAddress: localhost:443
-      agentAddress: kiam-server:443
+    # image:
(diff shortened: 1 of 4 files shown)


davidmccormick:
@paalkr does the PR aid your cluster migration strategy?

DominicGunn:
Slightly out of scope of this PR, but should we also update the image version of kiam in the [cluster.yaml.tmpl](https://github.com/kubernetes-incubator/kube-aws/blob/master/builtin/files/cluster.yaml.tmpl)? It's still set to a default of 2.8.

paalkr:
I think it's a good idea to let users decide which versions of kiam they would like to run, and it's definitely a good move to support backwards compatibility. When it comes to upgrade of kiam to 3.2, users have to update their kiam certs at some point. But this pr will probably make migration to 0.13 a little easier for many users, as there is just one thing less to focus on. The kiam upgrade can be carried out later, after the main cluster upgrade.

Personally I will probably make the move to kiam 3.2 during the initial cluster upgrade, as I have tested the cert update and documented in our systems how to execute the upgrade.

davidmccormick:
Updated cluster.yaml template as suggested and updated the default values so that it caches lookups by having a default session timeout of 30m and changing the server urls to remove the :port shenanigans.

PR #1603, opened 2019-05-24
[v0.13.x] Add support for PersistentVolumeClaim resize

DominicGunn:
## Changes

- Adds support for enabling the PersistentVolumeClaim admission-control.

### DIFF
--- a/builtin/files/userdata/cloud-config-controller
+++ b/builtin/files/userdata/cloud-config-controller
@@ -3353,7 +3353,7 @@ write_files:
           {{- else }}
           - --apiserver-count={{if .MinControllerCount}}{{ .MinControllerCount }}{{else}}{{ .Controller.Count }}{{end}}
           {{- end }}
-          - --enable-admission-plugins=EventRateLimit,ExtendedResourceToleration,PodSecurityPolicy{{if .Experimental.Admission.AlwaysPullImages.Enabled}},AlwaysPullImages{{ end }}{{if .Experimental.Admission.Initializers.Enabled}},Initializers{{end}}
+          - --enable-admission-plugins=EventRateLimit,ExtendedResourceToleration,PodSecurityPolicy{{if .Experimental.Admission.AlwaysPullImages.Enabled}},AlwaysPullImages{{ end }}{{if .Experimental.Admission.Initializers.Enabled}},Initializers{{end}}{{if .Experimental.Admission.PersistentVolumeClaimResize.Enabled}},PersistentVolumeClaimResize{{ end }}
           - --admission-control-config-file=/etc/kubernetes/auth/admission-control-config.yaml
           - --bind-address=0.0.0.0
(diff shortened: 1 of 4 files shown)


davidmccormick:
Isn’t this on by default in 1.13?

davidmccormick:
Ah, no it isn’t, we’ll spotted! 😉

davidmccormick:
Hi, went back to the docs to check.  I took this out as an experimental feature because I thought that it was now enabled by default.  The documentation here https://v1-13.docs.kubernetes.io/docs/reference/access-authn-authz/admission-controllers/ would seem to suggest that this is in fact the case and so this feature should always be on and no longer require explicit setting.

builtin/files/userdata/cloud-config-controller
```diff
@@ -3353,7 +3353,7 @@ write_files:
           {{- else }}
           - --apiserver-count={{if .MinControllerCount}}{{ .MinControllerCount }}{{else}}{{ .Controller.Count }}{{end}}
           {{- end }}
-          - --enable-admission-plugins=EventRateLimit,ExtendedResourceToleration,PodSecurityPolicy{{if .Experimental.Admission.AlwaysPullImages.Enabled}},AlwaysPullImages{{ end }}{{if .Experimental.Admission.Initializers.Enabled}},Initializers{{end}}
+          - --enable-admission-plugins=EventRateLimit,ExtendedResourceToleration,PodSecurityPolicy{{if .Experimental.Admission.AlwaysPullImages.Enabled}},AlwaysPullImages{{ end }}{{if .Experimental.Admission.Initializers.Enabled}},Initializers{{end}}{{if .Experimental.Admission.PersistentVolumeClaimResize.Enabled}},PersistentVolumeClaimResize{{ end }}
```
davidmccormick: PersistentVolumeClaimResize admission controller is enabled by default 1.13

builtin/files/userdata/cloud-config-controller
```diff
@@ -3408,6 +3408,9 @@ write_files:
           {{- if .ControllerFeatureGates.Enabled }}
           - --feature-gates={{.ControllerFeatureGates.String}}
           {{- end }}
+          {{- if .Experimental.Admission.PersistentVolumeClaimResize.Enabled }}
+          - --feature-gates=ExpandPersistentVolumes=true
```
davidmccormick: This will clash if .ControllerFeatureGates has been set above.  It looks like the expand persistent volume claim has been on by default since 1.11 https://v1-13.docs.kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

davidmccormick:
Maybe you could make this experimental feature for turning on the alpha feature "ExpandInUsePersistentVolumes"?

davidmccormick:
Closing as not needed (enabled by default)

PR #1607, opened 2019-05-28
Reserve resources for calico typha

zonzamas:
Calico-typha does not have any reserved resource so it works in a best effort mode. Since it is a 'system service' it should have some resources reserved. What do you think?

### DIFF
--- a/builtin/files/userdata/cloud-config-controller
+++ b/builtin/files/userdata/cloud-config-controller
@@ -1451,11 +1451,11 @@ write_files:
                 #  value: "9093"
               resources:
                 requests:
-                  cpu: "100m"
-                  memory: "100Mi"
+                  cpu: {{ if .Kubernetes.Networking.SelfHosting.TyphaResources.Requests.Cpu }}{{ .Kubernetes.Networking.SelfHosting.TyphaResources.Requests.Cpu }}{{ else }}"100m"{{ end }}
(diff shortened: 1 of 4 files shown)


davidmccormick:
Good idea!  In which case they should be configurable via the cluster.yaml so that users can increase them if the defaults prove too small.  Can you add them in as options?

DominicGunn:
Any chance you could sign the CLA and address David's comments @zonzamas?

zonzamas:
Hi, Im have it sing already with the other email I used to use to contribute to kube-aws. I will definitely revisit this when I have some free time or when I need it for my current company.

zonzamas:
I signed it

dominicgunn:
Thanks @zonzamas, I'd still like to see this be configurable through `cluster.yaml` before merging. Could you take a look at doing that?

zonzamas:
Sure, how about this?

builtin/files/userdata/cloud-config-controller
```diff
@@ -1449,6 +1449,13 @@ write_files:
                 #  value: "true"
                 #- name: TYPHA_PROMETHEUSMETRICSPORT
                 #  value: "9093"
+              resources:
```
davidmccormick: What you could do here is to put your default values into the function NewDefaultCluster() within pkg/api/cluster.go?  Then you don't need to do the conditional in the template... could you do that?

davidmccormick:
Hi, many thanks - would you be so kind as to cherry-pick this change and PR to our v0.13.x branch and v0.14.x branch?  That way your change will go out with those upcoming updates rather than when v0.15.x is released.  Thanks for your contribution! 🙏

PR #1615, opened 2019-06-04
Migrate to plugin based autoscaler and plugin system fixes

davidmccormick:
## Autoscaler related work
Some fixes to autoscaler plugin, now tested and ok to migrate over to it.
Remove all legacy autoscaler core code and references.
**WARNING - users must remove the experimental autoscaler values from their cluster.yaml as it no longer exists and migrate over to using the plugin - which is 100x newer and likely to work better anyway!**

## Kube AWS Plugin related work
This PR refactors the way that plugins load their values for Worker Nodepools.  Previously each nodepool would get its own plugin configs and would not inherit any settings from the root plugin section, ie.:-

```
<<ROOT PLUGIN SECTION>>
kubeAwsPlugins:
  awsIamAuthenticator:
    enabled: false
  clusterAutoscaler:
    enabled: true
    replicas: 2
    region: "us-west-2"
    image: "k8s.gcr.io/cluster-autoscaler:v1.13.4"
...
worker:
  apiEndpointName: Default
  nodePools:
  - &nodepoolNodePool
    <<NODEPOOL PLUGIN SECTION>>
    kubeAwsPlugins:
      clusterAutoscaler:
        enabled: true
        image: davexyz
```

This means that you can't define a plugin in the root plugin section which is enabled for ALL nodepools, and you can't have values in the root plugin section that can be inherited by the nodepool plugin sections which means that you have to template out all of your plugin settings in the Root section for the control plane and other stacks AND in each nodepool you want the plugin to be active - grrrr!

The new behaviour introduced in this PR *merges* each of the plugin configs in the nodepool plugin section with the values of the root plugin section when compiling each nodepool.  This means that nodePools values override the root ones and you only have to specify the values that you wish to override.  You can also disable a plugin for a particular nodepool.

Please note, another merge will also always take place - this is from the default values supplied in the plugin, so for nodepools the resultant plugin configs are the result of a merge of all three:

`plugin default values -> root values -> nodepool values`

and for everything else it is:

`plugin default values -> root values`

So taking the autoscaler plugin as an example, we can activate nodepool additions across the board from the root plugin config section and still disable the autoscaler on specific nodepools like this: -

```
<<ROOT PLUGIN SECTION>>
kubeAwsPlugins:
  awsIamAuthenticator:
    enabled: false
  clusterAutoscaler:
    enabled: true
    replicas: 2
    region: "us-west-2"
    image: "k8s.gcr.io/cluster-autoscaler:v1.13.4"
...
worker:
  apiEndpointName: Default
  nodePools:
  - <<: *nodepoolNodePool
    name: nodepool-c
    subnets:
      - name: c
    kubeAwsPlugins:
      clusterAutoscaler:
        enabled: false
```

and because we have a default image in the plugin itself then we can further reduce this to (enabled for all nodepools except nodepool-c):-

```
<<ROOT PLUGIN SECTION>>
kubeAwsPlugins:
  awsIamAuthenticator:
    enabled: false
  clusterAutoscaler:
    enabled: true
    region: "us-west-2"
...
worker:
  apiEndpointName: Default
  nodePools:
  - <<: *nodepoolNodePool
    name: nodepool-c
    subnets:
      - name: c
    kubeAwsPlugins:
      clusterAutoscaler:
        enabled: false
```


davidmccormick:
No reviews but gonna press on and merge. :(

PR #1632, opened 2019-06-19
Update OWNERS.

dominicgunn:
## Changes

- Adds @davidmccormick and @DominicGunn to owners so that @k8s-ci-robot is happy.


davidmccormick:
Possibly only need this in master

PR #1637, opened 2019-06-20
Expose CoreDNS to Prometheus

dominicgunn:
## Changes

- Updates the CoreDNS service deployment to allow scraping by Prometheus.
- This matches the Kubernetes upstream CoreDNS addon deployment, shown [here](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/coredns/coredns.yaml.in)


davidmccormick:
Thanks very much @dominicgunn 🎉

PR #1645, opened 2019-06-26
Fixed default etcd version

paalkr:


davidmccormick:
Hi - many thanks for this - it looks like I messed up the etcd upgrade commit.  If you don't mind I'm going to look whether we can keep the v3.2.x format of the version as this matches the image tags of the etcd releases.  I think what I need to do is maybe look at etcdadm and remove the adding of an additional v to it?

paalkr:
Sure, feel free to make that improvement. I guess the cluster-yaml template needs an update as well then. Does backwards compatibility need be handled in any way?

davidmccormick:
Thanks for the pointers of what needs to change, yes it looks like I changed the etcd-member service, i.e. from:-

```
        - name: 40-version.conf
          content: |
            [Service]
            Environment="ETCD_IMAGE_TAG=v{{.Etcd.Version}}"
```

to (in 0.13.0-rc.2): -

```
        - name: 40-version.conf
          content: |
            [Service]
            Environment="ETCD_IMAGE_TAG={{.Etcd.Version}}"
```

So this would explain why I've not seen errors with new clusters.  I'll look at the etcdadm changes and try and PR them tomorrow.  Thanks!

paalkr:
https://github.com/kubernetes-incubator/kube-aws/pull/1646

PR #1659, opened 2019-06-28
WIP: [Small] Enhance NLB support for api-endpoint

tyrannasaurusbanks:
**What**
Enhance NLB support for api-endpoint:
- add ability to use an existing NLB
- improve error handling around NLB configuration
- add cross-zone-balancing to created NLBs to mirror behavior of kube-aws-managed ELBs.

**Why**
We (Hotels.com) wish to front our clusters with an NLB managed outside of the kube-aws lifecycle, one which has already been configured with a VPC Endpoint to be shared across multiple AWS accounts.

**How**
I've amended the control-plane stack.json template to accommodate some extra logic if an NLB ID is provided.
The logic in the stack.json around endpoints is a bit ugly, but I favoured verbose checks in the hope it reads clearer. Shout if you're not happy and I'll try present and alternative.

I'll this feature is accepted then I'll make a corresponding PR for 0.12.x, 0.13.x & master.
Thanks!

### DIFF
--- a/model/api_endpoint_lb.go
+++ b/model/api_endpoint_lb.go
@@ -14,7 +14,7 @@ type APIEndpointLB struct {
 	APIAccessAllowedSourceCIDRs CIDRRanges `yaml:"apiAccessAllowedSourceCIDRs,omitempty"`
 	// Identifier specifies an existing load-balancer used for load-balancing controller nodes and serving this endpoint
 	Identifier Identifier `yaml:",inline"`
-	// Managed is set to true when want to create an ELB for this API endpoint. It is false by default i.e. considered to be false if nil
+	// Managed is set to true when we want to create an ELB for this API endpoint. It is false by default i.e. considered to be false if nil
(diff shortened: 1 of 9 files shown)


dominicgunn:
This is great @tyrannasaurusbanks! Could you look at fixing the build?

model/api_endpoint_lb.go
```diff
@@ -83,11 +83,20 @@ func (e APIEndpointLB) ManageSecurityGroup() bool {
 	return !e.NetworkLoadBalancer() && len(e.APIAccessAllowedSourceCIDRs) > 0
 }
 
+// ManageTargetGroup returns true if kube-aws should create a target group for this ELB
```
davidmccormick: As discussed, please take a look at this - it sounded like you might not need it! :+1:

tyrannasaurusbanks: done!

PR #1666, opened 2019-07-03
WIP: v0.13.x: Add handle-kubernetes-version-jump

davidmccormick:
When upgrading (or rolling back) you can run into issues when the two versions of kubernetes are far apart from each other (I tried upgrading a cluster from v1.9.3 to v1.13.7).  The difficulty appears to be from having a new apiserver and object schema installed from `install-kube-system` whilst the *kube-controller-manager* and *kube-scheduler* are most likely active on one of the older nodes.

This change makes a controller look for others when starting up for any that it finds which are either a major release or more than 1 release away from its version it will send a special *kube-aws-migration-disable-xyz* configmap to ask the other controller to shutdown it's scheduler and controller-manager so that this node's versions can guarantee to be the active versions of these.

It works both ways - so if a new controller is significantly older than the existing nodes it will also ask them to shutdown.

### DIFF
--- a/builtin/files/userdata/cloud-config-controller
+++ b/builtin/files/userdata/cloud-config-controller
@@ -5995,7 +5995,7 @@ write_files:
         namespace: kube-system
       data:
         kubernetesVersion: ${version}
-        disable: "kube-controller-manager kube-scheduler"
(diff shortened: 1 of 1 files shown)


davidmccormick:
Closing this will PR this functionality as a plugin which can be enabled if desired.

Issue #1332, opened 2018-05-26
Separate etcd stack does not work with managed subnet option


c-knowles:
Using version 9c735f9485445699d598df5c07fc73c811f682d2, the new separate etcd stack seems to not work with managed subnet options:

```
generating assets for control-plane, etcd, network, pool-A, pool-B, pool-C, pool-S
Error: Error validating cluster: failed to validate etcd plane: invalid cloudformation stack template https://s3.amazonaws.com/BUCKET/k8s-dev-v9/kube-aws/clusters/k8s-dev-v9/exported/stacks/etcd/stack.json: ValidationError: Template format error: Unresolved resource dependencies [ManagedPrivateSubnetWithExistingNGW1RouteToNatGateway] in the Resources block of the template
	status code: 400, request id: a5b59285-60b4-11e8-80b0-c11be6d8f9e1
make: *** [kube-aws-up] Error 2
```

In the autoscaling group of `etcd/stack.json` stored in S3:
```
"DependsOn": ["ManagedPrivateSubnetWithExistingNGW1RouteToNatGateway", "Etcd0EIP", "Etcd0EBS"]
```

Relevant configuration:
```yaml
etcd:
  subnets:
    - name: ManagedPrivateSubnetWithExistingNGW1
    - name: ManagedPrivateSubnetWithExistingNGW2
    - name: ManagedPrivateSubnetWithExistingNGW3
```

Relevant source seems to be this bit, but not sure what an appropriate fix would be:
https://github.com/kubernetes-incubator/kube-aws/blob/c430955dd249bb9629f414ec34f81ee713028848/core/etcd/config/templates/stack-template.json#L449

Ref https://github.com/kubernetes-incubator/kube-aws/pull/1233

mumoshu:
Thanks for the report! Probably we missed to pass those routes from the
network stack to the etcd stack?
Then you'd need to rewrite etcd cluster config in golang to refer the
routes by cross stack references.
2018年5月26日(土) 16:42 Chris Knowles <notifications@github.com>:

c-knowles:
My current workaround is to remove `{{if $etcdInstance.DependencyExists}}{{$etcdInstance.DependencyRef}},{{end}}` from `etcd.json.tmpl` before updating the cluster. Not sure if that affects anything else but the cluster appears to still function ok.

jcrugzz:
[...]

PR #1667, opened 2019-07-03
Etcd cannot depend on resources in another stack

koen92:
The dependency on the other stack gives errors. Starting the etcd node
before the natgateway is deployed is not a big issue as it keeps
restarting its proces to fetch its data. Wait signal should be long
enough for this to be resolved.

Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1332

PR #1670, opened 2019-07-03
Cannot depend on resources in another stack

koen92:
The dependency on the other stack gives errors. Starting the etcd node
before the natgateway is deployed is not a big issue as it keeps
restarting its proces to fetch its data. Wait signal should be long
enough for this to be resolved.

Fixes https://github.com/kubernetes-incubator/kube-aws/issues/1332

See also https://github.com/kubernetes-incubator/kube-aws/pull/1667


davidmccormick:
Thanks - that's the v0.14.x releases covered too!

PR #1671, opened 2019-07-08
Allow an AntiAffinty tolerant to help ensure even spread of DNS pods

dominicgunn:
## Changes

- Adds a preferred scheduling anti-affinity to try and target nodes in different AZs if enabled. 

Useful in situations where we have autoscaling enabled or are deploying DNS pods to nodes.

### DIFF
--- a/builtin/files/userdata/cloud-config-controller
+++ b/builtin/files/userdata/cloud-config-controller
@@ -4036,17 +4036,16 @@ write_files:
                         - "master"
                 {{- end }}
                 {{- if .KubeDns.AntiAffinityAvailabilityZone }}
-                affinity:
-                  podAntiAffinity:
-                    preferredDuringSchedulingIgnoredDuringExecution:
-                      podAffinityTerm:
-                        labelSelector:
-                        - matchExpressions:
-                           - key: k8s-app
-                             operator: In
-                             values:
-                             - "kube-dns"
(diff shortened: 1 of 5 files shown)


builtin/files/userdata/cloud-config-controller
```diff
@@ -4033,6 +4034,20 @@ write_files:
                         operator: In
                         values:
                         - "master"
+                {{- end }}
+                {{- if .KubeDns.AntiAffinityAvailabilityZone }}
+                affinity:
```
davidmccormick: This looks as though `affinity` is going to be repeated twice at two levels...

```
affinity:
  nodeAffinity:
  affinity:
    podAntiAffinity
```

davidmccormick: This issue does not seem to occur on your v0.13.x and v0.14.x branches! 👍

dominicgunn: Good eyes, fixed!

davidmccormick:
Many thanks @dominicgunn! 🎉

davidmccormick:
Be careful with this!  By using the topologyKey: "failure-domain.beta.kubernetes.io/zone" you are effectively saying I want one dns server per AvailabilityZone - which might make sense for small clusters but may not for large clusters where you desire more dns servers per AZ to serve the larger number of nodes.  If you are in a region with only 3 AZs - that is only 3 instances of dns!

PR #1672, opened 2019-07-08
[v0.13.x] - Allow an AntiAffinty tolerant to help ensure even spread of DNS pods

dominicgunn:
## Changes

- Adds a preferred scheduling anti-affinity to try and target nodes in different AZs if enabled. 

Useful in situations where we have autoscaling enabled or are deploying DNS pods to nodes.


davidmccormick (approved):
/lgtm

PR #1674, opened 2019-07-09
v0.13.x: fixes for rendering plugin systemd units

davidmccormick:
Changed Machine.SystemdUnits to inline rather than needing contents->content.

Fixed issue rendering systemd unit templates.
Fixed issue when RemoteFileSpec contains Template and not content.
Refactored extras.go to clean up repetitive code.


omar-nahhas:
Thank you very much @davidmccormick, I have also tested it and the rendering works fine,  would it be possible to extend the current tests on test/integration package to test the rendering of system units?

PR #1675, opened 2019-07-09
master: Fixes for rendering plugin systemd units

davidmccormick:
Systemd unit content is no longer: -

```
units:
- name: xyz
  contents:
     content: |
       abc
```
but is now...

```
units:
- name: xyz
  content: |
     abc
```

This is more in-line with how files and manifests are working.

Fixed issue rendering systemd unit templates.
Fixed issue when RemoteFileSpec contains Template and not content.
Refactored extras.go to clean up repetitive code.

### DIFF
--- a/pkg/api/plugin.go
+++ b/pkg/api/plugin.go
@@ -227,7 +227,7 @@ type SystemdUnits []SystemdUnit
 type SystemdUnit struct {
 	Name string `yaml:"name,omitempty"`
 	// Contents must be a valid go text template producing a valid systemd unit definition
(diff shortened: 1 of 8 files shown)


davidmccormick:
Hi @dominicgunn - do you use plugins?  Would you be able to cast your eyes over this?
Also @mumoshu if you have time then your input is always appreciated!

dominicgunn:
@davidmccormick, I don't use plugins, but the code itself looks pretty sound. If there any example plugins where this was previously failing, I'm happy to take a shot at verifying.

PR #1681, opened 2019-07-16
plugin: upgrade helper - more fixes and features

davidmccormick:


dominicgunn (commented):
I have mixed feelings about enabling this by default, but the code LGTM.

/lgtm

PR #1683, opened 2019-07-16
v0.14.x plugin: upgrade helper - more fixes and features

davidmccormick:


dominicgunn (commented):
I have mixed feelings about enabling this by default, but the code LGTM.

/lgtm

davidmccormick:
Yes, it should be disabled by default in cluster.yaml template (which it does appear to be)
Do not merge please, more testing required. ;)

PR #1695, opened 2019-08-05
Allow extra Coredns configuration

jorge07:
When touching the coredns configmap a cluster upgrade will overwrite the changes. This PR introduces a way to configure the changes inside the cluster config so no need to touch manually the config.

It will be ignored when other provider is specified


jorge07:
Is this is accepted I'll need to port this to +0.1X versions

davidmccormick:
Thanks very much for this contribution - could you also PR this for master, 13.x and 14.x branches?

jorge07:
Sure, will do

PR #1700, opened 2019-08-10
Remove duplicated kiam agent flags

paalkr:
The kiam agent flags `--whitelist-route-regexp=.*` and `--gateway-timeout-creation=1s` are not supported on kiam < 3.0. So removing them from the "base" set of flags, as they are added if kiam version > 3.0 is specified in cluster.yaml.


davidmccormick:
Good catch! Thank you for fixing this bug! 🙏

paalkr:
Should this fix also go into the 0.13 branch?

PR #1706, opened 2019-08-16
update ci golang version to 1.12.x

davidmccormick:
CI should use the same golang which I build the binaries with.


chenrui333:
Just saw this, I went ahead to bump the version to 1.13.x in https://github.com/kubernetes-incubator/kube-aws/pull/1820

davidmccormick:
This didn't work and is no longer needed

PR #1709, opened 2019-08-22
Update launch from up to apply

tedmiston:
Update the comment above to match the launch command being changed from `up` to `apply`.


davidmccormick (approved):
/lgtm

PR #1715, opened 2019-08-30
Toleration added to prevent node draininer scheduling on cordoned nodes

HarryStericker:
As title says. 

Its seems that if a node takes longer to drain than 300s [set here](https://github.com/kubernetes-incubator/kube-aws/blob/master/builtin/files/userdata/cloud-config-controller#L2851) the node drainer pod will try reschedule itself onto the node. This causes it to be in a restart loop until the node drains and goes offline, which makes our alerts pretty noisy! 

Since a cordoned node is given the label node.kubernetes.io/unschedulable, this gives the user the option to make the daemonset respect this/


HarryStericker:
I signed it

HarryStericker:
Closing and reopening to fix cla integration

PR #1716, opened 2019-09-02
Coredns prometheus metrics

HarryStericker:
The coredns annotation for prometheus metrics is tied to the wrong resource. This should re-point it to the actual deployment and hence pods, allowing metrics to be scraped.


HarryStericker:
Looks like the PR integrations got themselves in a tiff. Closing and reopening with a fresh PR

PR #1722, opened 2019-09-04
Toleration added to prevent node drainer scheduling on cordoned nodes

HarryStericker:
As title says.

Its seems that if a node takes longer to drain than 300s set here the node drainer pod will try reschedule itself onto the node. This causes it to be in a restart loop until the node drains and goes offline, which makes our alerts pretty noisy!

Since a cordoned node is given the label node.kubernetes.io/unschedulable, this gives the user the option to make the daemonset respect this

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1495,7 +1495,7 @@ experimental:
 
     # Prevents the nodeDrainer from scheduling on nodes that have been cordoned.
     # This happens if a nodes take more than the alloted time to drain
-    unschedulableWhenCordoned: true
+    # unschedulableWhenCordoned: true
     
   # Configure OpenID Connect token authenticator plugin in Kubernetes API server.
   # For using Dex as a custom OIDC provider, please check "contrib/dex/README.md".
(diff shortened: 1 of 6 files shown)


davidmccormick:
Hi thanks very much for your PR - I'm trying to understand your idea.  From the taint it looks like you are making the node drainer schedulable on nodes which have been tainted with `node.kubernetes.io/unschedulable:noschedule`. This seems contrary to your PR description and settings names?  I think I may have missed something important - could you possibly explain how your change functions?

HarryStericker:
Hey @davidmccormick - glad to contribute. Having thought about it - I believe you're right (unltimately rendering these PR's useless!). In essence since daemonsets don't respect unschedulable taints, I wanted to force nodeDrainer to respect that specific taint since I'm seeing it constantly try to re-deploy itself to a node if the 5 minute drain period is exceeded. However, the toleration here has inadvertently reinforced that the nodeDrainer _can_ tolerate the taint, and won't modify the behaviour.

HarryStericker:
Closing - re-evaluating and changing of strategy for this one.

PR #1724, opened 2019-09-04
[v0.14.x] Toleration to prevent node drainer scheduling on cordoned nodes

HarryStericker:
As title says.

Its seems that if a node takes longer to drain than 300s set here the node drainer pod will try reschedule itself onto the node. This causes it to be in a restart loop until the node drains and goes offline, which makes our alerts pretty noisy!

Since a cordoned node is given the label node.kubernetes.io/unschedulable, this gives the user the option to make the daemonset respect this


HarryStericker:
See #1722

PR #1726, opened 2019-09-09
move Kiam to a plugin

davidmccormick:
Create a new kiam plugin
Remove Experimental Kiam Support Code
Fix/Allow plugins to create ca's and use them to sign other certs.  Note, the CA must be created first.
Improve logging around plugin extensions.


go.mod
```diff
@@ -0,0 +1,56 @@
+module github.com/kubernetes-incubator/kube-aws
```
dominicgunn: We should be careful committing this until we shift the project to `GOMODULES`.

go.sum
```diff
@@ -0,0 +1,271 @@
+cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
dominicgunn: Same comment as `go.mod`.

builtin/files/plugins/kiam/plugin.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - source:
+          path: manifests/service.yaml
+      - source:
+          path: manifests/service-account.yaml
+      - source:
+          path: manifests/server-cluster-role.yaml
+      - source:
+          path: manifests/server-cluster-role-binding.yaml
+
+    pki:
```
dominicgunn: This is awesome!

go.mod
davidmccormick: Good point I'll remove both!

PR #1727, opened 2019-09-10
Allow CoreDNS resources to be configured

dominicgunn:
## Changes

- Allows CoreDNS resources to be configured through `cluster.yaml`.

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1328,7 +1328,7 @@ kubeDns:
   provider: coredns
   
   # Defines resources for the CoreDNS Deployment. Ignored if using kubedns.
-  # coreDnsResources:
+  # dnsDeploymentResources:
   #   requests:
   #     cpu: "100m"
   #     memory: "70Mi"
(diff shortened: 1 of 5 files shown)


builtin/files/cluster.yaml.tmpl
```diff
@@ -1326,6 +1326,15 @@ kubernetesDashboard:
 kubeDns:
   # Define which DNS provider to use (kube-dns or coredns), default coredns.
   provider: coredns
+
```
davidmccormick: Why not just make this resources and make it work for either kubedns or coredns?

dominicgunn: Done!

davidmccormick: Awsome! :)

davidmccormick:
One last question!  The dns resource sizing should be being governed on the fly by the dns-autoscaler pod - which is responsible for tweaking resource values and increasing the number of replicas in answer to the size of the cluster.  Can you check to see what it is doing in your cluster?  I would have hoped it would have been trying to scale it...

dominicgunn:
I can see it scaling the number of pods, but cannot see it scaling resources for the pods. Should it be?

```
05 21:10:48.636587       1 k8sclient.go:250] Cluster status: SchedulableNodes[33], SchedulableCores[280]
I0905 21:10:48.636607       1 k8sclient.go:251] Replicas are not as expected : updating replicas from 2 to 3
I0905 21:20:38.626543       1 k8sclient.go:250] Cluster status: SchedulableNodes[32], SchedulableCores[276]
I0905 21:20:38.626562       1 k8sclient.go:251] Replicas are not as expected : updating replicas from 3 to 2
```

PR #1730, opened 2019-09-10
Move kube2iam to a plugin

davidmccormick:
Move kube2iam out of the core codebase and into its own plugin


davidmccormick:
@dominicgunn could you take a quick look?  I've spun it up in a new cluster to test and looks ok.

PR #1751, opened 2019-10-08
[v0.13.x]Allow extra Coredns configuration

jorge07:
When touching the coredns configmap a cluster upgrade will overwrite the changes. This PR introduces a way to configure the changes inside the cluster config so no need to touch manually the config

Also realigned with gofmt


davidmccormick:
Looks great many thanks! 🙏

PR #1754, opened 2019-10-09
Migrate the kubernetes dashboard to a plugin and plugin templating improvements.

davidmccormick:
With optional ingress object and secrets creation.
With autoscaling based on cluster size/nodes.
Upgrade dashboard to kubernetesui/dashboard:v2.0.0-beta4

Templating system improvements: -
Allow use of templating in pki keypairs common names, orgs and dnsNames.
Allow use of templating in any string within a plugins `values` section.
Fix bugs scanning files for AWS CloudFormation Tags.


builtin/files/plugins/dashboard/manifests/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+  namespace: {{ .Values.namespace }}
+  {{ if or .Values.ingress.class .Values.tls.enabled -}}
+  annotations:
+    {{ if .Values.tls.enabled -}}
+    ingress.kubernetes.io/ssl-redirect: "true"
+    ingress.kubernetes.io/protocol: https
+    {{- end }}
+    {{ if .Values.ingress.class -}}
+    kubernetes.io/ingress.class: {{ .Values.ingress.class }}
+    {{- end }}
```
jorge07: It would be nice to allow custom annotations here so people can configure things like whitelisting ips or other custom rules

davidmccormick: Added, thanks!

PR #1767, opened 2019-10-24
Kube2iam resources adjustment

jorge07:
For cluster of +50 nodes kube2iam is throttling all the time with current limits. This one add some more margin and also remove the cpu limits what cause a performance issue

### DIFF
--- a/core/controlplane/config/templates/cloud-config-controller
+++ b/core/controlplane/config/templates/cloud-config-controller
@@ -4854,11 +4854,10 @@ write_files:
                     name: http
                 resources:
                   limits:
-                    cpu: 10m
-                    memory: 32Mi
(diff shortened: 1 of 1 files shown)


davidmccormick:
Interestingly, we ended up adjusting the limits for the v0.15.x kube2iam plugin.
            limits:
              memory: 128Mi
            requests:
              cpu: 100m
              memory: 64Mi
So this brings v0.12.x in line with the same settings.  One thing would you please consider also PRing to both v0.13.x and v0.14.x branches too please?  Otherwise, when you upgrade the resource setting is likely to be reset back to the old one...

davidmccormick:
Many thanks for contributing this change! 🙏

PR #1769, opened 2019-10-28
master: CoreDNS prometheus metric annotations exposed at pod level

### LINKED ISSUE #1768: Prometheus annotations are never added to CoreDNS pods
The CoreDNS prometheus metric annotations added in https://github.com/kubernetes-incubator/kube-aws/pull/1720 are never rendered because their `if` statement occurs inside the *kube-dns* block (rather than the *coredns* one above it).

kfr2:
These annotations would previously never be rendered because they
were added under the kube-dns section (rather than CoreDNS's).
Additionally, the annotations are now added to the Pods themselves
rather than the Deployment.

Fixes #1768

### DIFF
--- a/builtin/files/userdata/cloud-config-controller
+++ b/builtin/files/userdata/cloud-config-controller
@@ -4106,6 +4106,8 @@ write_files:
               labels:
                 k8s-app: kube-dns
               annotations:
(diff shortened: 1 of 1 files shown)


kfr2:
I signed it

PR #1782, opened 2019-11-26
Use nodePoolRollingStrategy of 'AvailabilityZone' by default.

davidmccormick:
The 'AvailabilityZone" nodePoolRollingStrategy was introduced a while ago and fits in between the existing 'Parallel' and 'Sequential' strategies - being safer for customer applications than 'Parallel' but not as slow as 'Sequential' (when there are multiple different nodepool types in play).  I think this strategy makes a better default. 

Changes: -
* Make 'AvailabilityZone' the default strategy.
* Set the MinInstancesAvailable to 0 by default.
* The maxBatchSize remains 1 by default (safety first).
* We invite users to set the maxbatchsize to the nodepool maxcount to test removing a whole AZ at a time (and when this happens we will remove the nodedrainer lifecycle hook).  This is so people can test what will happen in the event of an AZ failure and see if they have scaled and distributed their apps appropriately to handle this kind of event.
* Update cloudformation suspend automated processes which may interfere with the rolling update.
* Do not update autoscaler node levels unless they have changed in the template (i.e. don't roll back autoscaler or manual scaling changes)
* Remove the lifecycle hook from the controllers.  it is preferable the they shutdown quickly rather than hanging around for another 5 minutes and attempting to drain.

Sneaking in for clean up:
(Also clean up some unused refs to dashboard types - it's been migrated to a plugin)


builtin/files/stack-templates/control-plane.json.tmpl
```diff
@@ -96,15 +96,25 @@
       },
       {{end}}
       "UpdatePolicy" : {
+        "AutoScalingScheduledAction" : {
+          "IgnoreUnmodifiedGroupSizeProperties" : true
```
dominicgunn: Nice!

builtin/files/stack-templates/etcd.json.tmpl
```diff
@@ -367,15 +367,25 @@
...
         "AutoScalingRollingUpdate" : {
           "MinInstancesInService" : "0",
           "MaxBatchSize" : "1",
           {{if $.WaitSignal.Enabled}}
           "WaitOnResourceSignals" : "true",
-          "PauseTime": "{{$.Controller.CreateTimeout}}"
+          "PauseTime": "{{$.Controller.CreateTimeout}}",
           {{else}}
-          "PauseTime": "PT2M"
+          "PauseTime": "PT2M",
```
dominicgunn: This change feels somewhat significant, should be configurable or have you found 2m to be enough?

PR #1789, opened 2019-11-28
Add flag to cmds to use AWS profile

javipolo:
Add a new flag to command line utilities to allow to choose a AWS profile from the aws credentials file


davidmccormick:
Hi, I think we'll take this only as far back 0.13.x thanks.  Customers should upgrade from 0.12.x versions.

PR #1798, opened 2019-12-05
kube2iam metrics

jorge07:
Expose kube2iam metrics by default and allow, as feature flag, PodMonitor for Prometheus Operator.

### Motivation 

Expose kube2iam metrics for proper monitoring

### To discuss

- Enable metrics by default. When plugin is enabled will add an entry in worker and masters networking security group. The reason is that metrics, imo, should be exposed by default and this ones are the only ones available. 

- I set the prometheus pod monitor object base on the cluster autoscaler one but there's an inconsistent behaviour in `prometheusMetrics.selector` in cluster autoscaler, set as `.prometheusMetrics.labels` in this plugin. May need to be fixed in another PR to align plugins consistency.

## WIP

- [ ] Testing

### DIFF
--- a/builtin/files/plugins/kube2iam/manifests/daemonset.yaml
+++ b/builtin/files/plugins/kube2iam/manifests/daemonset.yaml
@@ -32,6 +32,7 @@ spec:
           name: kube2iam
           args:
             - "--app-port=8282"
+            - "--metrics-port=8181"
             - "--auto-discover-base-arn"
             - "--auto-discover-default-role"
             - "--iptables=true"
@@ -52,6 +53,9 @@ spec:
             - containerPort: 8282
               hostPort: 8282
               name: http
+            - containerPort: 8181
+              hostPort: 8181
+              name: metrics
           resources:
             limits:
               memory: 128Mi
--- a/builtin/files/plugins/kube2iam/manifests/podmonitor.yaml
+++ b/builtin/files/plugins/kube2iam/manifests/podmonitor.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.prometheusOperatorPodMonitor }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  labels:
+    app: kube2iam 
+    {{- range $key, $value := .Values.prometheusMetrics.labels }}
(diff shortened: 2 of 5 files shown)


davidmccormick:
Hi, many thanks for this!  Can you explain further about the "inconsistent behaviour"? (just want to understand better).

jorge07:
I'm talking about the cluster autoscaler plugin `values.yaml` and the naming of the prometheus metrics section:

https://github.com/kubernetes-incubator/kube-aws/blob/master/builtin/files/plugins/cluster-autoscaler/plugin.yaml#L21
VS
https://github.com/kubernetes-incubator/kube-aws/pull/1798/files#diff-32c9a649c9a3f62acd04b344048296a9R12

Using `selector` instead `labels`. I found it confusing so used `labels` here. I did not changed this in the autoscaler as is a breaking change.

builtin/files/stack-templates/network.json.tmpl
```diff
@@ -471,6 +471,36 @@
       "Type": "AWS::EC2::SecurityGroupIngress"
     },
     {{ end }}
+    {{if ((index .PluginConfigs "kube2iam").Enabled) }}
```
omar-nahhas: By assuming the port number as 8181, you could create conflict with other users who are using that same port for other applications, and other people may not need to open a hole in the worker's security groups in order to scrape that endpoint.

builtin/files/plugins/kube2iam/manifests/podmonitor.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.prometheusOperatorPodMonitor }}
```
omar-nahhas: I think that adding a reference to the Prometheus operator pod monitor here could add noise to the code, as it is unrelated to kube2ami.

builtin/files/plugins/kube2iam/manifests/daemonset.yaml
```diff
@@ -52,6 +53,9 @@ spec:
             - containerPort: 8282
               hostPort: 8282
               name: http
+            - containerPort: 8181
+              hostPort: 8181
```
omar-nahhas: making this pod configurable would help

PR #1813, opened 2020-01-03
Add proper indentation to extra CoreDNS config

javipolo:
When using more than one line of extra config, the indentation was not right and generated YAML was not valid

This should fix this issue


davidmccormick:
Many thanks for this fix! 🙏

PR #1819, opened 2020-01-04
Kubernetes 1.16 Support

davidmccormick:
ETCD_VERSION="v3.4.3"
KUBERNETES_VERSION="v1.16.4"
Decommission old v1beta1 apis
Add selectors to all DaemonSets and Deployments
Kubelet moves into a Docker container (issues with rkt fly container)
Calico bumped to version 3.11.1

### DIFF
--- a/builtin/files/plugins/dashboard/manifests/deployment.yaml
+++ b/builtin/files/plugins/dashboard/manifests/deployment.yaml
@@ -110,7 +110,7 @@ spec:
 
 {{ if .Values.ingress.enabled -}}
 ---
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: kubernetes-dashboard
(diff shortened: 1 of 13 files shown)


builtin/files/plugins/dashboard/manifests/deployment.yaml
```diff
@@ -110,7 +110,7 @@ spec:
 
 {{ if .Values.ingress.enabled -}}
 ---
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: extensions/v1beta1
```
Fsero: shouldn't be this the other way around? since extensions/v1beta1 is deprecated in 1.16 for ingress
```suggestion
apiVersion: networking.k8s.io/v1beta1
```

davidmccormick: Hi, you are quite right!  I'm not sure how I managed to do that!  Great spot - thanks!

davidmccormick:
I'm going to merge this as master should represent a jump to kubernete 1.16 working for a 0.16.0 release of kube-aws.  If there are breakages we need to fix them in master before the release.

PR #1821, opened 2020-01-05
Move to go modules

### LINKED ISSUE #1807: Move to go modules
https://github.com/golang/go/wiki/Modules

chenrui333:
Fixes #1807


chenrui333:
Circle back on the PR.

chenrui333:
Remove vendor folder in favor of go modules to manage the dependencies.

> The End of Vendoring
> Vendor directories serve two purposes. First, they specify by their contents the exact version of the dependencies to use during gobuild. Second, they ensure the availability of those dependencies, even if the original copies disappear. On the other hand, vendor directories are also difficult to manage and bloat the repositories in which they appear. With the go.mod file specifying the exact version of dependencies to use during vgo build, and with proxy servers for ensuring availability, vendor directories are now almost entirely redundant. They can, however, serve one final purpose: to enable a smooth transition to the new versioned world.

> When building a module, vgo (and later go) will completely ignore vendored dependencies; those dependencies will also not be included in the module's zip file. To make it possible for authors to move to vgo and go.mod while still supporting users who haven't converted, the new vgo vendorcommand populates a module's vendor directory with the packages users need to reproduce the vgo-based build.

davidmccormick:
Many thanks for this super useful contribution! 🙏

PR #1827, opened 2020-01-24
template: use of InitialConfig to build cluster.yaml

sei-nicolas:
This changes cluster.yaml initial skeleton to avoid the use of (not available) Config struct


davidmccormick:
Many thanks - looks great!

weisjohn:
@davidmccormick was this supposed to auto-create a release of something like v0.15.1 ?
I notice it's been approved and merged, but there's no additional release made

weisjohn:
perhaps that's because "tide" says it's not mergeable?

Screen Shot 2020-02-04 at 12 36 38 PM

PR #1832, opened 2020-02-11
v0.15.x branch: Make cloud-controller-manager an experimental feature

davidmccormick:
Fix issues caused by the v0.15.x release and the cloud-controller-manager breaking physical volume support.  This feature wasn't tested sufficiently and wasn't ready for release, apologies!  This PR makes it an experimental feature which is `disabled` by default.

I've added CSI provisioning for EBS volumes as another experimental feature so that the cloud-controller-manager and CSI can be tested together for the potential replacement of the in-tree drivers in the future (1.21 has been mooted for the removal of the in-tree drivers).  Again, this is disabled by default and should only be used by the brave.  My testing shows that it works but the migration of existing PVs is not working for me.

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1449,6 +1449,34 @@ experimental:
     usernameClaim: "email"
     groupsClaim: "groups"
 
+  # Enable cloudControllerManager to move AWS cloud-provider code out of the kube-controller-manager and into a separate cloud-controller-manager
+  # WARNING: enabling the cloud-controller-manager removes legacy PVC/Storage code - you will need to enable the alpha
+  # ContainerStorageInterface experimental feature to priovide EBS volume integtation and you risk losing any existing volumes.
+  cloudControllerManager:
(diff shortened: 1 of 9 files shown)


builtin/files/cluster.yaml.tmpl
```diff
@@ -1449,6 +1449,34 @@ experimental:
...
+  # Enable experimental Container Storage Interface (CSI)
+  # This subsystem is designed to replace legacy in-tree storage code with a standard interface (CSI) and external compliant drivers
+  # WARNING: this is for testing purposes only - their is no guarantee that your existing in-tree provisioned volumes will continue to
+  # work once CSI has been enabled (and they didn't work during my testing, although new volumes appeared to work)
+  containerStorageInterface:
+    enabled: false
+    # csiProvisioner:
+    #   repo: quay.io/k8scsi/csi-provisioner
+    #   tag: v1.3.1
+	  # csiAttacher:
```
Fsero: indentation 
```suggestion
     # csiAttacher:
```

PR #1836, opened 2020-02-18
Migrate to flatcar linux from coreos

davidmccormick:
Changes necessary to run flatcar linux instances instead of coreos.
This is intended as a shorter term fix to address supportability with the end-of-life of CoreOS.
Longer term I would like to move to a distribution with good support for both amd64 and arm architectures, such as Ubuntu or another.


jorge07:
Nice one David. You were faster than me. I was working on this this afternoon. 😂

Fsero:
Great job @davidmccormick thanks for putting work into this :smiley:

flatcar/amiregistry/amiregistry.go
```diff
@@ -3,31 +3,35 @@ package amiregistry
...
+				if hvm, ok := v["hvm"]; ok {
+					return hvm, nil
+				} else {
+					return "", fmt.Errorf("region %s does not have a flatcar hvm ami entry", region)
+				}
+			}
+		}
 	}
 
 	return "", errors.Errorf("could not find \"hvm\" image for region \"%s\" and channel \"%s\"", region, channel)
```
gopheryourself: Just a few things:

- There's already an error being returned from this function when a hvm image isn't found for a region and channel. That can probably just be reused.
- This file is the only file that uses `errors.Errorf`. They should probably all be replaced with `fmt.Errorf` for consistency (as mostly done with your changes).
- We can reduce the if nesting with early returns.

```go
	for _, v := range amis {
        	if v["name"] != region {
            		continue
        	}
        	if v["hvm"] != hvm {
            		break
        	}
       		return hvm, nil
	}

	return "", fmt.Errorf("could not find \"hvm\" image for region %q and channel %q", region, channel)
```

davidmccormick: @gopheryourself - thanks for your suggestions!  Take your point on the errors package, will remove.  I agree that the code is looking a little "nesty".  I can't use your code as above as it changes the behavoir of the code (namely we're not. checking if a hvm matcheds, we're looking it up if it exists) but I think this modification captures your suggestion?

```
	for _, v := range amis {
		if v["name"] != region {
			continue
		}
		if hvm, ok := v["hvm"]; ok {
			return hvm, nil
		} else {
			break
		}
	}
```

gopheryourself: @davidmccormick Opps, sorry! But yes, that's looking much better :)

A really picky alternative would be (but this is getting to be more personal preference now, rather than line-of-sight coding/early returns, so it's entirely up to you): 
> ```
> 	for _, v := range amis {
> 		if v["name"] != region {
> 			continue
> 		}
> 		if hvm, ok := v["hvm"]; ok {
> 			return hvm, nil
> 		}
> 		break
> 	}
> ```

gopheryourself: @davidmccormick Apologies for not being clearer, but my suggestion was to replace only `errors.Errorf` with `fmt.Errorf`, not `errors.Wrapf`. `errors.Wrapf` has been used elsewhere for wrapping errors, but `errors.Errorf` was only used in this file.

Maybe it'd best to handle such an inconsistency in another PR and have a project discussion on whether to use `github.com/pkg/errors` everywhere (for wrapping and formatting), or to use `fmt.Errorf` and Go 1.13's new error wrapping (https://blog.golang.org/go1.13-errors) functionality.

davidmccormick: Thanks for the advice! 👍

PR #1838, opened 2020-02-25
fixed managedELBImplied logic to check value of ELB RecordSetManaged is true

searsaw:
When reusing an existing ELB, I was receiving the following error when attempting to render credentials:

```
Error: failed loading cluster.yaml: invalid cluster: invalid api endpoint config at index 0: no appropriate subnets found for api load balancer with private=<nil>: invalid cluster: invalid api endpoint config at index 0: no appropriate subnets found for api load balancer with private=<nil>
```

Digging into the code a bit, it appears the logic for figuring out if the ELB was implicitly managed checked if the value of `recordSetManaged` was set but never checked that the value was actually `true`.

This PR simply checks that the `recordSetManaged` is set and that it is `true`. Otherwise, the implication is the ELB is not managed.

### DIFF
--- a/pkg/api/api_endpoint_lb.go
+++ b/pkg/api/api_endpoint_lb.go
@@ -152,7 +152,7 @@ func (e APIEndpointLB) managedELBImplied() bool {
 		e.explicitlyPublic() ||
 		e.HostedZone.HasIdentifier() ||
 		len(e.SecurityGroupIds) > 0 ||
-		e.RecordSetManaged != nil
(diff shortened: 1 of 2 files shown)


searsaw:
Closing due to I figured out my "issue." I had to remove `recordSetManaged: false` from `loadBalancer` when reusing an ELB. However, I don't think it's immediately obvious that this is how it should be. Just because I am reusing an ELB doesn't necessarily mean I don't want the record set to be managed. I guess the assumption is that if you are reusing an ELB, then you are managing the record set as well?

PR #1848, opened 2020-05-01
make flannel subnetLen configurable for big clusters

Fsero:
We had an issue in a big cluster (around 200 nodes) when rolling upgrade a change and replacing instances we ran out of subnet leases, that causes that new workers were unable to get a flannel lease and hence setting up the network  and were killed due to not sending the cfn-signal in time.

This patch aims to make subnetLen configurable for big clusters, using kube-aws default flannel config the maximum is 255 nodes which means around 127 nodes and a rolling upgrade that replaces all instances (like updating the base AMI).

In the case of being unset, flanneld will keep using the default which is /24 per node.


dominicgunn:
This is awesome, we have the same issue! Thank you.

PR #1851, opened 2020-05-12
Creating a switch to disable ICMP Ping instead of the default - allow 0.0.0.0/0

HarryStericker:
Currently kube-aws includes a rule in its security group to allow ping from all traffic (0.0.0.0/0). Unless this is necessary for a reason I have not foreseen, it is an inherent security risk, violates CIS compliance, and is probably superfluous if SG's are managed appropriately.

Therefore, I have created a switch in cluster.yaml to allow removal of this rule. I have left it enabled by default but am conscious that anyone upgrading needs to ensure this is in their cluster.yaml, so am happy to adjust the logic so that there's no impact unless you explicitly ask for it to be disabled. 

I don't think I've missed any other spots where this rule is used.

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1360,10 +1360,10 @@ kubeProxy:
 # It is enabled by default.
 #cloudFormationStreaming: true
 
-# When enabled, a security group rule is included on the generated SG to allow ICMP from all traffic (0.0.0.0/0).
-# This is applied to all nodes (worker & control plane) in the cluster
-#openICMP:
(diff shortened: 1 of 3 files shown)


dominicgunn:
Hey @HarryStericker, you need to sign the CLA before I can merge this :)

HarryStericker:
Yup. I signed it last time I made commits to this repo, but somethings fudged. Before i fixed it though, I wanted to get eyes on the change and make sure it's sane!

HarryStericker:
Clsoing this PR. Raised a clean one, since cla is in a funk.

PR #1853, opened 2020-05-18
bugfix: flannel SubnetLen expects an int not an string

Fsero:
I was creating a cluster from latest and it seems that flannel expects an int


dominicgunn:
Thanks for doing this, I haven't gotten around to testing the v0.16.x branch yet. Appreciate the due diligence.

PR #1854, opened 2020-05-18
[v0.16.0] Creating a switch to disable ICMP Ping instead of the default allow 0.0.0/0

HarryStericker:
Currently kube-aws includes a rule in its security group to allow ping from all traffic (0.0.0.0/0). Unless this is necessary for a reason I have not foreseen, it is an inherent security risk, violates CIS compliance, and is probably superfluous if SG's are managed appropriately.

Therefore, I have created a switch in cluster.yaml to allow removal of this rule. I have left it enabled by default but am conscious that anyone upgrading needs to ensure this is in their cluster.yaml, so am happy to adjust the logic so that there's no impact unless you explicitly ask for it to be disabled. 

I don't think I've missed any other spots where this rule is used.

### DIFF
--- a/builtin/files/stack-templates/control-plane.json.tmpl
+++ b/builtin/files/stack-templates/control-plane.json.tmpl
@@ -219,12 +219,14 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ if .openICMP.Enabled -}}
           {
             "CidrIp": "0.0.0.0/0",
             "FromPort": -1,
             "IpProtocol": "icmp",
             "ToPort": -1
           }
+          {{end -}}
         ],
         "Tags": [
           {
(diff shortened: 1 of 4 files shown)


HarryStericker:
Hey @dominicgunn, can i get a few pointers when you have time. Not sure what needs modifying to fix the build errors.

builtin/files/stack-templates/control-plane.json.tmpl
```diff
@@ -219,12 +219,14 @@
             "ToPort": 22
           },
           {{end -}}
+          {{ if .openICMP.Enabled -}}
           {
             "CidrIp": "0.0.0.0/0",
             "FromPort": -1,
             "IpProtocol": "icmp",
             "ToPort": -1
           }
```
dominicgunn, line 228: I think this might be your issue. Now that this is optional (`{{ if .... }}`) the block above is going to leave a trailing `,` if this is disabled, like this.

```
....
    "ToPort": 22
  },
]
```

We can't have that trailing slash if `openICMP` is disabled.

PR #1863, opened 2020-05-27
[v0.16.0] Switch CoreOS to Flatcar

dominicgunn:
## Changes

- Switches to use Flatcar in the v0.16.x branch, shamelessly stolen from @jorge07, thanks!


jorge07:
pre-stolen from @davidmccormick 😂

PR #1869, opened 2020-05-30
Bump golang to v1.14.3

chenrui333:
- designate go.mod with v1.14
- bump travis to use go 1.14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: bionic
 language: go
 go:
-  - 1.13.x
+  - 1.14.x
 
 script:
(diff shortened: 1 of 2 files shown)


dominicgunn:
Thanks for the PR! Master is now going to be for the v.1.17.x cut, could you also contribute the same patch for our v1.16.x branch?

chenrui333:
> Thanks for the PR! Master is now going to be for the v.1.17.x cut, could you also contribute the same patch for our v1.16.x branch?

sure thing

chenrui333:
Somehow I could not find the v1.16 branch.

PR #1872, opened 2020-06-08
[v0.16.x] Setup Calico Networking Option

dominicgunn:
## Changes

- Canal is EOL, and the only other option we offered was AWS CNI. This borrows (steals) the Calico configuration for kops, and dumps it into kube-aws so that we can use that as an option too.
- This PR also fixes node labelling issues that changed between v1.15.x and v1.16.x, as we can no longer use `kubelet` to self label `kubernetes.io` tags, making nodes difficult to differentiate.
- This PR also removes quay.io as the core provider for calico networking images because they keep falling over.
- This PR also opens ports 4789 to allow for vxlan networking between ports, and opens communication between workers & controllers for IPnIP tunnelling used by Calico's BGP (also requiring port 179) setup.

### DIFF
--- a/builtin/files/stack-templates/network.json.tmpl
+++ b/builtin/files/stack-templates/network.json.tmpl
@@ -211,6 +211,7 @@
         "GroupId": {
           "Ref": "SecurityGroupController"
         },
+        "Description": "controller to controller sg",
         "IpProtocol": "-1",
         "SourceSecurityGroupId": {
           "Ref": "SecurityGroupWorker"
@@ -224,6 +225,7 @@
         "GroupId": {
           "Ref": "SecurityGroupController"
         },
+        "Description": "controller to calico",
         "IpProtocol": "udp",
         "SourceSecurityGroupId": {
           "Ref": "SecurityGroupController"
@@ -238,6 +240,7 @@
         "GroupId": {
           "Ref": "SecurityGroupController"
         },
+        "Description": "worker to calico sg",
         "IpProtocol": "udp",
         "SourceSecurityGroupId": {
           "Ref": "SecurityGroupWorker"
--- a/builtin/files/stack-templates/network.json.tmpl
+++ b/builtin/files/stack-templates/network.json.tmpl
@@ -275,34 +275,6 @@
       "Type": "AWS::EC2::SecurityGroupIngress"
     },
     {{ end -}}
-    "SecurityGroupWorkerIngressFromControllerToCalicoBGP": {
-      "Properties": {
-        "FromPort": 179,
-        "GroupId": {
-          "Ref": "SecurityGroupWorker"
-        },
-        "IpProtocol": "tcp",
-        "SourceSecurityGroupId": {
-          "Ref": "SecurityGroupController"
-        },
-        "ToPort": 179
-      },
-      "Type": "AWS::EC2::SecurityGroupIngress"
-    },
-    "SecurityGroupWorkerIngressFromWorkerToCalicoBGP": {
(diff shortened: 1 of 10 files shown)


builtin/files/stack-templates/network.json.tmpl
```diff
@@ -194,6 +194,58 @@
...
+        }
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
+    "SecurityGroupControllerIngressFromWorkerToControllelr": {
+      "Properties": {
+        "GroupId": {
+          "Ref": "SecurityGroupController"
+        },
+        "IpProtocol": "-1",
```
Fsero, line 214: Add a description? 

```suggestion
        "Description": "controller to controller sg",
        "IpProtocol": "-1",
```

builtin/files/stack-templates/network.json.tmpl
```diff
@@ -194,6 +194,58 @@
...
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
+    "SecurityGroupControllerIngressFromControllerToCalico": {
+      "Properties": {
+        "FromPort": 4789,
+        "GroupId": {
+          "Ref": "SecurityGroupController"
+        },
+        "IpProtocol": "udp",
```
Fsero, line 227:
```suggestion
       "Description": "controller to calico",
        "IpProtocol": "udp",
```

builtin/files/stack-templates/network.json.tmpl
```diff
@@ -194,6 +194,58 @@
...
+      },
+      "Type": "AWS::EC2::SecurityGroupIngress"
+    },
+    "SecurityGroupControllerIngressFromWorkerToCalico": {
+      "Properties": {
+        "FromPort": 4789,
+        "GroupId": {
+          "Ref": "SecurityGroupController"
+        },
+        "IpProtocol": "udp",
```
Fsero, line 241:
```suggestion
        "Description": "worker to calico sg",
        "IpProtocol": "udp",
```

builtin/files/userdata/cloud-config-worker
```diff
@@ -387,7 +382,8 @@ coreos:
         {{- end }}
         $KUBELET_OPTS"
         Restart=always
-        RestartSec=10
+        RestartSec=10        
+
```
Fsero, line 386: nit: extra new line

Fsero:
Hi @dominicgunn thanks for your efforts and the contribution.

I was looking into Calico docs and didn't see any official information about canal deprecation. Can you share for better context?

Thanks!

builtin/files/userdata/cloud-config-worker
dominicgunn: I actually added this purposefully to more clearly separate the `[Service]` and `[Install]` block similar to other unit definitions here. 

We should probably unify on one in another pull request (extra line between unit blocks, or without extra line), what are your thoughts?

c-knowles:
Removed myself for review since unfortunately at the moment I don't have an appropriate stack available to check through this.

builtin/files/userdata/cloud-config-worker
Fsero: Sorry for the late response. I'm ok with the extra line but maybe would be much cleared if

```suggestion
        RestartSec=10        

{{/* END of kubelet [Service] */}}
```

dominicgunn: I'm happy with that, I pushed those changes up.

PR #1873, opened 2020-06-08
Migrate to new s3 URL method

tjamet:
Path based access has been deprecated and now returns a redirect error to bucketname.s3.amazonaws.com
as explained in this post:
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

This makes cloudformation to fail as it refers to the now redirected s3
bucket path, and raises the message:
S3 error: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint. For more information check http://docs.aws.amazon.com/AmazonS3/la

There is no strong documentation about what are the updates in China and
Gov regions, consider it follows the same logic.

### DIFF
--- a/pkg/api/assets.go
+++ b/pkg/api/assets.go
@@ -2,8 +2,9 @@ package api
 
 import (
 	"fmt"
-	"github.com/kubernetes-incubator/kube-aws/fingerprint"
 	"strings"
+
+	"github.com/kubernetes-incubator/kube-aws/fingerprint"
 )
 
 type AssetID struct {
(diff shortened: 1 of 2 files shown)


Fsero:
@tjamet could you please also patch 0.15.x and 0.14.x?

tjamet:
@Fsero yes sure, did this first one for the approach to focus discussion and planned to do so when convergence on the solution

dominicgunn:
Thanks for doing this @tjamet! The approach here seems good.

/lgtm

dominicgunn:
Thanks for doing this @tjamet! The approach here seems good.

/lgtm

PR #1880, opened 2020-07-01
[v0.12.x] [Unmaintained branch] Disable Flatcar auto-updates

jorge07:
This a patch we added in our side other may benefit from. Feel free to reject as I know well is unmaintained.


dominicgunn:
We'll accept, thanks for the contribution.

PR #1883, opened 2020-07-06
[v0.15.x] Allow injection of CoreDNS configuration for non-root zones

kfr2:
The existing `kubeDns.extraCoreDNSConfig`  allows the user to
inject additional values into the coredns configmap's root zone (".").
This commit allows the user to additionally specify a string to
`kubeDns.additionalZoneCoreDNSConfig` which will be injected
into the configmap after the root zone to allow the user to specify
configuration for additional zones. As an example, this might be used
to forward traffic for the .global zone to the istiocoredns service.

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1329,9 +1329,18 @@ kubeDns:
     nodesPerReplica: 16
     min: 2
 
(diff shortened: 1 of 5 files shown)


dominicgunn:
Thanks for the contribution @kfr2!

PR #1886, opened 2020-07-07
Keep newline, so apiserver starts

flah00:
* With WS stripping character in place configs would be malformed, ie
    * `--volume opt-bin,kind=host,source=/opt/bin \--mount volume=var-log,target=/var/log \`
* This code has been refactored in 0.16.x and master. As far as I'm aware, this is the only branch with this bug.
* Now there's an extra newline, but this function is only called once  and the the sh environment formatting is more forgiving, ie

```sh
--volume opt-bin,kind=host,source=/opt/bin \

--mount volume=var-log,target=/var/log \
```


dominicgunn:
Awesome! Thanks for the contribution @flah00.

PR #1891, opened 2020-07-30
[v0.14.x] Allow dnsmasq to be backed by a local copy of CoreDNS

kfr2:
This commit allows the user to specify that dnsmasq should be backed by a pod-local copy of CoreDNS rather than relying on the global CoreDNS service. If enabled, the dnsmasq-node DaemonSet will be configured to use a local copy of CoreDNS for its resolution while setting the global CoreDNS service as a fallback. This is handy in situations where the number of DNS requests within a cluster grows large and causes resolution issues as dnsmasq reaches out to the global CoreDNS service.

Additionally, several values passed to dnsmasq are now configurable, including its `--cache-size` and `--dns-forward-max`.

See [this postmortem](https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/docs/postmortems/jan-2019-dns-outage.md) for an investigation into this situation which was instrumental in understanding issues we were facing. Many thanks to @dominicgunn for providing the manifests which I codified into this commit.

---

These features can be enabled and tuned by setting the following values within cluster.yaml:

```yaml
kubeDns:
  dnsmasq:
    coreDNSLocal:
      # When enabled, this will run a copy of CoreDNS within each DNS-masq pod and
      # configure the utility to use it for resolution.
      enabled: true

      # Defines the resource requests/limits for the coredns-local container.
      # cpu and/or memory constraints can be removed by setting the appropriate value(s)
      # to an empty string.
      resources:
        requests:
          cpu: 50m
          memory: 100Mi
        limits:
          cpu: 50m
          memory: 100Mi

    # The size of dnsmasq's cache.
    cacheSize: 50000

    # The maximum number of concurrent DNS queries.
    dnsForwardMax: 500

    # This option gives a default value for time-to-live (in seconds) which dnsmasq
    # uses to cache negative replies even in the absence of an SOA record.
    negTTL: 60
```

### DIFF
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1359,6 +1359,9 @@ kubeDns:
   # nodeLocalResolverOptions:
   # - --neg-ttl=10
   # - --no-ping
+  # When enabled, this will run a copy of CoreDNS within each DNS-masq pod and
+  # configure the utility to use it for resolution.
+  # coreDNSLocal: false
 
   # When enabled, will modify the TTL of the coredns service.
   # ttl: 30
@@ -1381,6 +1384,8 @@ kubeDns:
   #
   # This configuration is injected into the CoreDNS config map after the root
   # zone (".") and can be used to add configuration for additional zones.
+  # If corednsLocal has been enabled, this configuration will additionally be injected
+  # into its ConfigMap.
   # additionalZoneCoreDNSConfig: |
   #  global:53 {
   #      errors
--- a/builtin/files/cluster.yaml.tmpl
+++ b/builtin/files/cluster.yaml.tmpl
@@ -1359,9 +1359,13 @@ kubeDns:
   # nodeLocalResolverOptions:
   # - --neg-ttl=10
   # - --no-ping
-  # When enabled, this will run a copy of CoreDNS within each DNS-masq pod and
-  # configure the utility to use it for resolution.
-  # coreDNSLocal: false
+
+  # Settings for the dnsmasq-node DaemonSet which must be enabled by setting
+  # `kubeDns.nodeLocalResolver` to true.
+  dnsmasq:
+    # When enabled, this will run a copy of CoreDNS within each DNS-masq pod and
+    # configure the utility to use it for resolution.
+    coreDNSLocal: false
 
   # When enabled, will modify the TTL of the coredns service.
   # ttl: 30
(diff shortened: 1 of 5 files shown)


builtin/files/cluster.yaml.tmpl
```diff
@@ -1405,6 +1410,13 @@ kubeProxy:
 # It is enabled by default.
 #cloudFormationStreaming: true
 
+<<<<<<< Updated upstream
```
dominicgunn, line 1413: Can we fix this merge conflict?

builtin/files/userdata/cloud-config-controller
```diff
@@ -3969,6 +3969,85 @@ write_files:
...
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: coredns-local
+          namespace: kube-system
+          labels:
+            application: coredns
+        data:
+          Corefile: |
+            cluster.local:9254 {{ .PodCIDR }}:9254 {{ .ServiceCIDR }}:9254 {
```
dominicgunn, line 4026: Nice.

builtin/files/userdata/cloud-config-controller
```diff
@@ -1058,6 +1058,9 @@ write_files:
         "${mfdir}/kube-dns-de.yaml"
       {{- end }}
       {{ if .KubeDns.NodeLocalResolver -}}
+      {{ if .KubeDns.dnsmasq.CoreDNSLocal.Enabled -}}
+      deploy "${mfdir}/dnsmasq-node-coredns-local.yaml"
```
dominicgunn, line 1062: I know that `kube-aws` doesn't do a fantastic job of clean up, but seems like this one would be easy. Could we throw in a 

```
{{- else }}
      remove "${mfdir}/canal.yaml"
{{- end }}}
```

pkg/api/types.go
```diff
@@ -209,6 +209,23 @@ type IPVSMode struct {
 	MinSyncPeriod string `yaml:"minSyncPeriod"`
 }
 
+type CoreDNSLocalLimits struct {
```
dominicgunn, line 212: Could we use `ComputeResources` instead?

kfr2: Yep -- great idea!

builtin/files/userdata/cloud-config-controller
dominicgunn: Sorry, `remove "${mfdir}/dnsmasq-node-coredns-local.yaml"`

kfr2: 👍 Will do.

PR #1897, opened 2020-08-10
[v0.17.x] Adds 'disable' flag for kubedns

brandong954:
Adds a flag for disabling the coredns/kubedns deployment. Useful when strictly using coredns within dnsmasq:

```
dnsmasq:
    coreDNSLocal:
      enabled: true
```


brandong954:
I signed it

dominicgunn:
Hey @brandong954,

Thanks for the contribution. So that we can effectively track all the changes into the project could we cut the `go.mod` & `go.sum` changes from this PR, and if we want to add them upstream open a new PR to address dependency updates?

Cheers!

dominicgunn (changes requested):
Same comment as the other PR regarding not removing already installed versions of CoreDNS after being enabled.

Additionally, tests are failing and my guess is it's because we haven't run `go fmt` on this, we tried to simplify some of this by dumping it in the make file, `make format` should sort it out. Could we also please remove the `go.sum` and `go.mod` changes from this PR? They're not necessary for what we're trying to in this PRs scope.

PR #1911, opened 2020-09-23
Fix trailing whitespaces in template

zonzamas:
Depending on the options combination you use, the controller
template may generate the kubelet.service unit with extra
white lines that break the format hence making it to fail.
   
I have hacked an example in playground comparing the previous
combination of - for trimming whitespaces and the new one
   
https://play.golang.org/p/OO5T7TZA1qR


dominicgunn:
Thank you! Much appreciated.

